PR #9, opened 2011-04-21
test wanted: client connection failure and check errno 

bigeasy:
try to connect to non-existent host


bigeasy:
Requirements for checking `errno` or `oio_error`.
- If a function fails immediately it returns -1.
- If a pending operation fails the callback is called with status -1.
- `oio_last_error()` can be used to retrieve the error codes.
- `oio_last_error()` returns a struct with one public member which is called `code`. That one stores a cross-platform error code.
- A list of all the errors is in `oio.h`.

bigeasy:
@piscisaureus Should I create a pull request, or is it enough to attach the sha as above?

bigeasy:
Oh, nice. I can see how it managed to attach the pull request to this thread.

Thanks for your patience while I got oriented. I'll work on it again tomorrow evening and see if I can't contribute whole test.

piscisaureus:
lgtm

PR #35, opened 2011-04-27
Use _NSGetExecutablePath on darwin

josh:
Use `_NSGetExecutablePath` on darwin as `/proc/self/exe` is linux only.

I wasn't sure if you wanted to fork off `runner-unix.c` to `runner-darwin.c`.

Tests at least run now on mac. 3 failing assertions.


piscisaureus:
lgtm. landed in 9b92e13

PR #56, opened 2011-06-13
uv_ip4_addr() should take a const char*, not a mutable char*.

bnoordhuis:


bnoordhuis:
That's so you can do `uv_ip4_addr("127.0.0.1", 8888)` without having gcc complain.

piscisaureus:
Landed in 1987728fa84f0869ab82d32e374d4be97889207c

PR #60, opened 2011-06-16
uvgetaddrinfo

HenryRawas:
Here are my getaddrinfo changes

### DIFF
--- a/msvs/libuv-benchmark.vcxproj
+++ b/msvs/libuv-benchmark.vcxproj
@@ -135,6 +135,7 @@
     </Link>
   </ItemDefinitionGroup>
   <ItemGroup>
+    <ClCompile Include="..\test\benchmark-getaddrinfo.c" />
     <ClCompile Include="..\test\benchmark-ping-pongs.c" />
     <ClCompile Include="..\test\benchmark-pump.c" />
     <ClCompile Include="..\test\benchmark-sizes.c" />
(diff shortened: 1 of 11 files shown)


piscisaureus:
These are the comments I sent by e-mail. Posting them here so we can practice pull request review a bit :-)

---

I reviewed Henry’s getaddrinfo patch and it looks good so far.

Question: should we force the caller to hang on the strings and structs that the addrinfo struct points to? Or would we rather copy them? I know libeio does the latter, I don’t really know what I prefer. Ryan, Igor?

These are just nits:
-   Declare internal stuff as static
-   Maybe initialize the WAKEUP req in uv_getaddrinfo instead of in the thread?
-   It may be worthwhile to use the Unicode version of getaddrinfo instead of the ascii one, and convert the input (which will be utf-8) to UT-16.

\- Bert

HenryRawas:
Using getaddrinfoW would mean parsing the returned data as well to convert any names from UNICODE back to UTF-8.

piscisaureus:
> Using getaddrinfoW would mean parsing the returned data as well to convert any names from UNICODE back to UTF-8.

That's true but the alternative is to not support any non-ascii characters. That's not good, unless there's not any conceivable case where people would need unicode support. (honestly I can't think of any case right now)

bnoordhuis:
@piscisaureus RFC 2181 allows octets > 127 but no browser or resolver I know of actually supports that. DNS in the real world doesn't talk Unicode, it's either ASCII or Punycode (which is ASCII too).

piscisaureus:
Landed in 295fa6f538e7975e44c2c495781c7b0fc308d6b1

PR #67, opened 2011-06-24
winsock extension call conventions

HenryRawas:
The stack pointer is not set properly on returning from these calls. Need PASCAL convention.


piscisaureus:
Landed in 9c00849c9bc5510284ab2e35a0270dbea7c7c94a

PR #69, opened 2011-06-27
Makefile

leetreveil:
uname_S wont work for me on Windows with minGW so I had to change the Makefile to look for $(SystemRoot).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,9 +18,7 @@
 # FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
 # IN THE SOFTWARE.
 
-uname_S := $(shell sh -c 'uname -s 2>/dev/null || echo not')
-
-ifneq (,$(findstring MINGW,$(uname_S)))
+ifneq ($(SystemRoot),)
 include config-mingw.mk
 else
 include config-unix.mk


piscisaureus:
So your mingw doesn't support `uname`?

leetreveil:
Error: 
process_begin: CreateProcess(NULL, sh -c "uname -s 2>/dev/null || echo not", ...
) failed.

Windows XP SP3
GNU Make 3.82

piscisaureus:
Looks like a broken setup rather than missing uname. Make is telling you that it can't spawn 'sh' for some reason, without that you're not gonna get far anyway.

leetreveil:
After the change I made make, make test etc work fine. I've tested this on 2
machines (XP+7) and they both display the same behaviour.

ry:
if uname can be avoided on windows that would be preferable. But what is `SystemRoot` is that guaranteed to be defined?

japj:
from http://en.wikipedia.org/wiki/Environment_variable#System_path_variables

The %SystemRoot% variable is a special system-wide environment variable found on Microsoft Windows NT and its derivatives. Its value is the location of the system folder, including the drive and path.

So my guess is that it is indeed guaranteed to be defined.

However uname works ok on my MingW setup, so I don't understand why it is not working for @leetreveil.

luislavena:
Hello, if I can chime in.

the failure around `uname -s` might be associated by the lack of MSYS tools, not MinGW

Please note that MSYS is a complementary toolset that provides a Minimal SYStem allowing Unix-like scripts and Makefiles works on Windows.

MinGW by itself might not work with current Makefile, but having MSYS plus MinGW will.

bnoordhuis:
Closing, should be fixed. If not, holler.

PR #70, opened 2011-06-27
IPv6 bind support

mattstevens:
Tested on OS X, Ubuntu, and Windows 7.

The only change for IPv4 should be using sockaddr_storage with accept() on Unix.


ry:
Looks good. Please agree to our CLA http://nodejs.org/cla.html

mattstevens:
Done

ry:
Thanks!

PR #94, opened 2011-07-11
test-loop-handles: test may finish before timeout expires.

bnoordhuis:
Remove 'has timeout callback run?' check, it's not reliable.

IOW, the test is timing dependant. Can someone review this?


bnoordhuis:
Note to self:

piscisaureus_> bnoordhuis_: check and prepare watchers only get called before and after libev polls the system for events. When a check watcher is invoked that means that either a FD event has occurrred, or there was a timeout. I want to know what fd or timeout is responsible. The idea of the loop handles test was that there would be no FD events occurring and only that one timer. So at the end of the test the timer callback must have been called.

IOW, what makes the check and prepare watchers go off if not the timer?

PR #103, opened 2011-07-12
fix uv_getsockname for listening socket

HenryRawas:
Fix to allow getsockname to be used with listening socket.


bnoordhuis:
Thanks, applied in 6a67b51.

PR #126, opened 2011-08-04
Support for unescaped arguments, suitable for use with cmd /c.

DrPizza:
Robust argument escaping that hopefully matches Windows' algorithm for unescaping.
Facility to disable argument escaping.


igorzi:
commited to libuv with d84b2496161c7a11ead6

PR #127, opened 2011-08-05
uv-unix: Fix missing symbols on FreeBSD7 systems.

ldm5180:
WEXITSTATUS and other symbols are missing when building on my FreeBSD7.3 machine. This patch fixes it simply by #include'ing sys/wait.h.


ry:
landed in 787ef5da8010933af4ad27f705ad976455c31a7c thanks

PR #132, opened 2011-08-09
Upstream master

DrPizza:
Resolve the problem wherein node can't spawn itself on Windows when using a custom environment.

The root cause was that some Windows components, e.g. winsock, must have certain environment variables defined, or else they will fail to initialize properly. Even when users specify custom environments, it's unlikely that they want to break their ability to spawn typical Windows programs.

This should fix node's test-init.js.

### DIFF
--- a/src/win/process.c
+++ b/src/win/process.c
@@ -494,6 +494,14 @@ error:
   return NULL;
 }
 
+typedef struct env_var {
+  const char* narrow;
+  const wchar_t* wide;
+  int len; /* including null or '=' */
+  int supplied;
+  int value_len;
+} env_var_t;
+
 /*
  * If we learn that people are passing in huge environment blocks
  * then we should probably qsort() the array and then bsearch()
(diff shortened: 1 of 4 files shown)


src/win/process.c
```diff
@@ -495,21 +495,68 @@ error:
...
+    "SYSTEMROOT",
+    "SYSTEMDRIVE",
+    "TEMP",
+  };
 
   for (env = env_block; *env; env++) {
     env_len += (uv_utf8_to_utf16(*env, NULL, 0) * sizeof(wchar_t));
   }
 
+  for (i = 0; i < sizeof(required_vars) / sizeof(*required_vars); ++i) {
```
igorzi, line 558: This loop could be optimized away by including the check for one of required_vars in the previous loop.

src/win/process.c
```diff
@@ -495,21 +495,68 @@ error:
...
+    if (!env_block_contains_var(env_block, narrow_required_vars[i])) {
+      env_len += wcslen(required_vars[i]) * sizeof(wchar_t);
```
igorzi: required_vars lenghts should all be known at compile time, so no reason to use wcslen.  Same for wcslen(L"=").

src/win/process.c
```diff
@@ -525,6 +572,24 @@ wchar_t* make_program_env(char** env_block) {
     }
   }
 
+  for (i = 0; i < sizeof(required_vars) / sizeof(*required_vars); ++i) {
+    if (!env_block_contains_var(env_block, narrow_required_vars[i])) {
```
igorzi: You could just remember from an earlier check that one (or more) ore required_vars need to be populated (instead of doing the expensive env_block_contains_var.

igorzi:
in master with 3409c9b & 54a6c6f.

Issue #155, opened 2011-08-25
uv_getaddrinfo provides no way for custom data


erickt:
The function `uv_getaddrinfo` function zeros out the `uv_getaddrinfo_t` structure, which clears out the user provided data field inside the `uv_getaddrinfo_t`. The simplest way to solve this is to just provide a `getaddrinfo_init` that zeros out the structure and gives the user the opportunity to set the data before calling `uv_getaddrinfo`, but I imagine implementing #75 would solve the problem as well.

PR #156, opened 2011-08-25
Fix uv_getaddrinfo to accept custom data.

erickt:
I noticed that the unix backend for `uv_getaddrinfo` is zeroing out the the `uv_getaddrinfo_t` handle, which prevents the user from providing custom data to the callback result. This patch adds a `uv_getaddrinfo_init` that zeros out the data before `uv_getaddrinfo_t` before calling the function.

Edit: I've updated the patches against HEAD, and pulled some of the patches that we need for compiling the [Rust language](http://github.com/graydon/rust).


erickt:
Added a couple fixes for mingw-w64.

japj:
@erickt it might help to split new features/fixes.

also from looking at#155 it is already included in libuv at 6793b53e626d3f39ad149f8bbbd3f5b8de1708c5 right?

erickt:
@japj, I split up the patch into just the bug fixes. I'll submit a pull request for the makefile changes later. Regarding #155 and commit 6793b53, I can't actually find that commit. Did it accidentally get dropped?

ry:
rather than adding another function, can we just add a `void* data` parameter to `uv_getaddrinfo` ?

ry:
i merged c28c780 063cf1e and 45e7f47

in 1a43429 fe6157b and f4e2d55

ry:
submit new pull for the others.

ry:
also landed 5b567b2

PR #170, opened 2011-09-01
Allow `make all` work under MinGW

luislavena:
This small change allows `make all` compile properly both run-tests.exe and run-benchmarks.exe


ry:
hi luis,

thanks for the fix but we require contributors to agree to http://nodejs.org/cla.html before accepting patches

luislavena:
@ry: I just signed electronically. Do I need to fax it also?

ry:
electronic is fine. thanks for the patchz

Issue #169, opened 2011-09-01
preprocessor typo in win/loop-watcher.c


luislavena:
Attempt to compile it result in error:

```
gcc -Iinclude -Iinclude/uv-private -g --std=gnu89 -D_WIN32_WINNT=0x0501 -Isrc/ares/config_win32 -o src/win/loop-watcher.o -c src/win/loop-watcher.c src/win/loop-watcher.c: In function 'uv_prepare_init':
src/win/loop-watcher.c:129:1: error: pasting "." and "prepare" does not give a valid preprocessing token
make: *** [src/win/loop-watcher.o] Error 1
```

Here is a small patch:

https://gist.github.com/b4b4eb85d345e74fe7aa

luislavena:
Closing this, submitted proper pull request #174

PR #174, opened 2011-09-02
Correct a preprocessor typo

luislavena:
GCC chokes on ##name## concatenation (since . and prepare are invalid
tokens)

The associated change solves Issue #169

Dunno if this breaks something on Visual Studio (/cc @igorzi and @piscisaureus)


bnoordhuis:
Thanks, Luis. Merged in 6422a14.

PR #178, opened 2011-09-05
alternative test callback checking: shorter, more informative

tshinnic:
For your consideration as a 'nicer' way to check for callbacks seen during
testing.  Especially for when the order of callbacks is known as it
additionally checks for that ordering, as well as the previous done "how many"
of each kind of callback.

Rather than a cumbersome counting of callbacks and then checking those counts,
a string accumulates each occurrence of each callback and then is checked at
the end of testing for the 'correct' match.

For example, if the order of callbacks should have been 'open_read_clos_unlk_'
but something else happened, the assert will trap with additional explanation:

> [%   0|+  50|-   1]: fs_file_async
> `fs_file_async` failed: exit code 134
> Output from process `fs_file_async`:
> callbacks seen were not as expected:
>   expect 'open_read_clos_unlk_'
>    found 'open_read_clos_buzz_'
> Assertion failed in test/test-fs.c on line 445: cb_check("open_read_clos_unlk_")

The list of callbacks can be accumulated over an entire test and checked at the end, accumulated and checked at various stages, or reset and accumulated for individual stages/subtests.  (see use of cb_seen_clear)

The string accumulator is guarded against overflow for the case of run-away code.

This change touches only test/test-fs.c and only one test there, as a
demonstration of the idea.  I think the "fs_file_async" test certainly reads better this way.  
Let me know what you think.

### DIFF
--- a/test/test-fs.c
+++ b/test/test-fs.c
@@ -89,6 +89,29 @@ static char buf[32];
 static char test_buf[] = "test-buffer\n";
 
 
+static char cbs_occurred[256] = "";
+
+void cb_seen(const char* cb_token) {
+  if ((sizeof(cbs_occurred) - strlen(cbs_occurred)) > (strlen(cb_token))) {
+    strcpy(strchr(cbs_occurred,'\0'), cb_token);
+  }
+}
+
+void cb_seen_clear() {
+  cbs_occurred[0] = '\0';
+}
+
+int cb_check(const char* expect_this) {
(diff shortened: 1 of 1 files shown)


bnoordhuis:
Sorry Thomas, I appreciate the effort but I don't see what value it adds. The test suite is there to check our assumptions. If  a test fails, I'll hook it up to gdb for a post-mortem and then a open_cb_called with a value of 0 or 2 tells me more than a string.

PR #182, opened 2011-09-09
Subclass uv_getaddrinfo_t from uv_req_t.

### LINKED ISSUE #155: uv_getaddrinfo provides no way for custom data

erickt:
This patch also fixes #155. Since we no longer
memset clear the uv_getaddrinfo_t, the user can
now set the `uv_getaddrinfo_t->data` field without
problems.


ry:
landed in efa1b54076349653267640d5e0a5fe27a9cd852d thanks!

PR #183, opened 2011-09-10
Fixed a typo that breaks windows build

erickt:


piscisaureus:
Thanks. Landed in af936a9.

PR #186, opened 2011-09-15
Fix a couple warnings, added some asserts, and compiling with gcc-4.5 and mingw-w64

erickt:


bnoordhuis:
Merged everything except 484d3ba. @piscisaureus: can you check that commit?

piscisaureus:
@bnoordhuis can't do right now, I have only mingw32 and I'm afraid installing mingw-w64 will break that. Tomorrow I will get mingw-w64 on my office machine and try.

nddrylliog:
@picsisaureus Sorry to disturb, but wouldn't it be a good idea to set up a 32-bit VM and a 64-bit VM ? I know it helps me a lot when developing. You can even install SSH servers on Windows (yeah, they exist) and set up port forwarding in VirtualBox or equivalent so that you can ssh into your virtual machines from anywhere, and automate the builds, it's damn convenient!

piscisaureus:
@erickt
Replacing ws2_32.lib b y -lws2_32 doesn't fly. It breaks the msvc build. Are you sure this is necessary?

luislavena:
If I can chime in, I don't use gyp for MinGW, AFAIK, is not even supported at this time. I use the makefiles directly.

@erickt that is what you tried to solve?

erickt:
I'm okay on punting on the mingw-w64 commit for now. I'll close.

PR #189, opened 2011-09-17
avoid freeing invalid data in uv_getaddrinfo_done, plus fixing gyp compilation

erickt:
I found that we uv_getaddrinfo on unix could try to free invalid addrinfos if the handle's data owns the handle. This fixes that bug by saving the addrinfo pointer and freeing that directly, rather than trying to free through the handle.

Also, I noticed that test-tcp-write-error.c wasn't added into the uv.gyp.


bnoordhuis:
Thanks, Erick. Ryan had already fixed the gyp build in 2f18cf5 but I merged the other patch in 70e1032.

bnoordhuis:
Good catch, by the way.

erickt:
Glad to help.

PR #196, opened 2011-09-22
Add uv_freeaddrinfo to deallocate the addrinfo list

erickt:


erickt:
Updated to fix a warning on darwin.

PR #199, opened 2011-09-24
Fix a bug in calling memcpy with the wrong size

jyasskin:
Clang warns about some other constructs, but I think this one's the only bug.


bnoordhuis:
Thanks. This was fixed in upstream c-ares at bagder/c-ares@22c527c and bagder/c-ares@1f8f14d. I thought I'd back-ported the fixes but apparently not. If you don't mind, I'll take the patches from upstream to ease future upgrades. Thanks for pointing it out though, I appreciate it.

jyasskin:
I don't mind at all. Thanks for fixing it.

PR #200, opened 2011-09-25
Pass sockaddrs around by address

erickt:
This lets us do some error checking in the uv_ip*_addr functions, and makes the api more like the traditional socket api.

I don't have a windows machine, so I'm not 100% sure it's all working there though.


erickt:
I added a couple other bug fixes, validations, and merged the windows and unix error reporting functions.

bnoordhuis:
I don't think this patch is going to be wildly popular. The idea behind the uv_ip?_addr functions is that they're convenient to use and this patch takes that away. It also touches nearly every file we have. A belated no from me.

PR #202, opened 2011-09-28
Fixed memcpy when copying hints on uv_getaddrinfo

saghul:
Hi,

Here is a one character fix for uv_getaddrinfo when hints are being passed. When memcpy-ing, handle->hints is already a pointer to addrinfo.

Hope I'm not mistaken, I tested it and seems to work :-)

Thanks,


bnoordhuis:
Good catch. I'll merge it but can you sign [the CLA](http://nodejs.org/cla.html)?

saghul:
I just signed (electronically) the CLA. Thanks for the quick response!

bnoordhuis:
Thanks, Saúl. Merged in b594dba.

PR #203, opened 2011-09-29
os: uptime

indutny:
Cross-platform implementation and test.


indutny:
Just tested on windows 7 - works fine and reports data close to GetTickCount().

src/unix/cygwin.c
```diff
@@ -53,6 +54,22 @@ int uv_exepath(char* buffer, size_t* size) {
...
+uint64_t uv_uptime(void) {
+  uint64_t amount;
+  char line[512];
+  FILE *fpUptime = fopen("/proc/uptime", "r");
+
+  amount = 0;
+
+  if (fpUptime) {
+    if (fgets(line, 511, fpUptime) != NULL) {
+      sscanf(line, "%d.%d %d.%d", &amount);
```
bnoordhuis: Memory corruption.

src/unix/cygwin.c
```diff
@@ -65,4 +82,4 @@ int uv_fs_event_init(uv_loop_t* loop,
 
 void uv__fs_event_destroy(uv_fs_event_t* handle) {
   assert(0 && "implement me");
-}
+*}
```
bnoordhuis: What's this?

src/unix/linux.c
```diff
@@ -64,6 +69,23 @@ int uv_exepath(char* buffer, size_t* size) {
   return 0;
 }
 
+uint64_t uv_uptime(void) {
+#ifdef UV_USE_MONOTONIC
+  struct timespec now;
+  if (0 == clock_gettime(CLOCK_MONOTONIC, &now)) {
+    double uptime = now.tv_sec;
+    uptime += (double)now.tv_nsec / 1000000000.0;
```
bnoordhuis: Don't use floating point here.

src/unix/sunos.c
```diff
@@ -62,6 +64,30 @@ int uv_exepath(char* buffer, size_t* size) {
...
+uint64_t uv_uptime(void) {
+  kstat_ctl_t   *kc;
+  kstat_t       *ksp;
+  kstat_named_t *knp;
+
+  long hz = sysconf(_SC_CLK_TCK);
+  ulong_t clk_intr;
+
+  if ((kc = kstat_open()) == NULL)
+    throw "could not open kstat";
```
bnoordhuis: Don't throw. This is C, not C++.

src/win/util.c
```diff
@@ -94,3 +95,123 @@ done:
...
+  DWORD size = 4096;
+  DWORD readSize = 0;
+  LONG status = ERROR_MORE_DATA;
+  LPBYTE data = NULL;
+
+  do {
+    if (readSize > size) {
+      size = readSize;
+    } else {
+      readSize = size += 4096;
```
bnoordhuis: Break this up in two statements.

src/win/util.c
```diff
@@ -94,3 +95,123 @@ done:
...
+
+    if (data == NULL) {
+      data = malloc(size);
+      // Handle malloc's fails
+      if (data == NULL) return NULL;
+    } else {
+      LPBYTE newData = realloc(data, size);
+      if (newData == NULL) break;
+      data = newData;
+    }
```
bnoordhuis: Don't need to branch on `data == NULL`

`realloc(NULL, size)` is equivalent to `malloc(size)`

src/win/util.c
```diff
@@ -94,3 +95,123 @@ done:
...
+  } else {
+    free(data);
+    return NULL;
+  }
+}
+
+DWORD uv__lookup_counter_title_index(char* table_, char* title) {
+  char* table = table_;
+
+  // Skip first record and it's length
```
bnoordhuis: Don't use C99 / C++ style comments.

src/win/util.c
```diff
@@ -94,3 +95,123 @@ done:
...
+
+uint64_t uv_uptime(void) {
+  // Receive performance counters
+  PPERF_DATA_BLOCK header =
+      (PPERF_DATA_BLOCK) uv__query_reg_data(HKEY_PERFORMANCE_DATA, TEXT("2"));
+
+  if (header == NULL) return -1;
+
+  // Receive performance counter's titles by ids
+  char* table = (char*) uv__query_reg_data(HKEY_PERFORMANCE_DATA,
```
bnoordhuis: Variable declarations should go at the top.

src/win/util.c
```diff
@@ -94,3 +95,123 @@ done:
...
+  }
+
+  // Find first object
+  PPERF_OBJECT_TYPE object = (PPERF_OBJECT_TYPE) ((LPBYTE) header +
+                                                  header->HeaderLength);
+
+  // If we don't have any objects or object has no counters - return error
+  if (header->NumObjectTypes == 0 || object->NumCounters == 0) {
+    free(header);
+    return -1;
```
bnoordhuis: I would like it better if this function had a single clean up and exit point (use `goto`), makes it easier to follow and check for resource leaks.

@piscisaureus: You should probably review it as well. It looks okay to me but I'm not a windows guru.

src/unix/cygwin.c
indutny: Secret C technique :)

indutny:
Fixed all errors and rebased on master

indutny:
Tested on mingw and linux.

src/win/util.c
piscisaureus: @bnoordhuis @indutny

I agree on @bnoordhuis' comment. Otherwise LGTM.

indutny: Please take a look at latest version, fixed here.

indutny:
@bnoordhuis fixed.

indutny:
@bnoordhuis please review, should be fixed now

bnoordhuis:
Sorry Fedor, @igorzi already implemented this.

PR #204, opened 2011-09-29
add uv_udp_set_membership for ipv4 multicast support

defunctzombie:
- test-udp-multicast-join tests that multicast packets can be received
- stub src/win/udp.c until support added

Notes: I can work on the other multicast methods if the general direction of this patch seems reasonable.


bnoordhuis:
Looks good otherwise. Can you sign [the CLA](http://nodejs.org/cla.html)?

defunctzombie:
updated code with given feedback

bnoordhuis:
Thanks. Merged in a3d1f6f.

PR #205, opened 2011-09-30
start unifying windows and unix bind and connect code

erickt:
Simple patch to start unifying the error checking code for bind and connect that's duplicated in the windows and unix backend. Plus a bug fix and a unit test.


ry:
LGTM. @piscisaureus or @igorzi should also review.

erickt:
I swapped to using EINVAL as recommended by Ben Noordhuis.

piscisaureus:
LGTM

PR #207, opened 2011-10-01
os: memory info functions

indutny:
`double uv_get_free_memory(void);`
`double uv_get_total_memory(void);`


bnoordhuis:
Thanks, Fedor. Squashed and merged in 33cb877.

PR #208, opened 2011-10-01
os: process title

indutny:
Tested on cygwin, mingw, linux, darwin.


bnoordhuis:
Things I don't like and would like to see changed:
1. The patch introduces static variables everywhere. That's sensitive to races in multi-threaded environments. It needs to be either clearly documented or wrapped in critical sections.
2. Most of those static variables are unnecessary, we can simply query the OS again.
3. The memory they point to is never freed (unless you call `uv_set_process_title()` again) and that generates valgrind noise.

src/unix/darwin.c
```diff
@@ -68,6 +73,119 @@ int uv_exepath(char* buffer, size_t* size) {
...
+static char* process_title;
+size_t uv_get_process_title(char* buffer, size_t size) {
+  pthread_mutex_lock(&process_title_mutex_);
+
+  size_t result = 0;
+  if (process_title != NULL) uv__strlcpy(buffer, process_title, size);
+
+  pthread_mutex_unlock(&process_title_mutex_);
+
+  return result;
```
bnoordhuis: `result` is always zero here.

bnoordhuis: Also, variable declarations should go at the top of the function.

test/test-process-title.c
```diff
@@ -0,0 +1,38 @@
...
+#include "uv.h"
+#include "task.h"
+#include <stdio.h>
+#include <string.h>
+
+TEST_IMPL(process_title) {
+  fprintf(stderr, "set!\n");
+  uv_set_process_title("hello world!");
+  fprintf(stderr, "get!\n");
+  char* title = (char*) malloc(32);
```
bnoordhuis: Variable declaration at the top. Also, make it a `char title[32]`.

test/test-process-title.c
```diff
@@ -0,0 +1,38 @@
...
+  uv_get_process_title(title, 32);
```
bnoordhuis: And use `sizeof title` here.

src/unix/sunos.c
```diff
@@ -62,6 +63,35 @@ int uv_exepath(char* buffer, size_t* size) {
...
+  pthread_mutex_unlock(&process_title_mutex_);
+
+  return result;
+}
+
+void uv_set_process_title(const char* title) {
+  pthread_mutex_lock(&process_title_mutex_);
+
+  if (process_title) free(process_title);
+  setproctitle(process_title = strdup(title));
```
bnoordhuis: Won't compile, there is no `setproctitle()` on SunOS.

src/unix/linux.c
```diff
@@ -83,6 +87,44 @@ double uv_get_total_memory(void) {
...
+
+  pthread_mutex_unlock(&process_title_mutex_);
+
+  return result;
+}
+
+void uv_set_process_title(const char* title) {
+  pthread_mutex_lock(&process_title_mutex_);
+
+  if (title != NULL) {
```
bnoordhuis: This is inconsistent with the other UNIX implementations. The caller is responsible for not passing in NULL.

src/unix/linux.c
```diff
@@ -83,6 +87,41 @@ double uv_get_total_memory(void) {
...
+
+  prctl(PR_SET_NAME, process_title);
+
+  pthread_mutex_unlock(&process_title_mutex_);
+}
+#else
+size_t uv_get_process_title(char* buffer, size_t size) {
+  return 0;
+}
+void uv_set_process_title(char* title) {
```
bnoordhuis: Prototype mismatch, should be `const char*`

src/unix/linux.c
```diff
@@ -83,6 +87,41 @@ double uv_get_total_memory(void) {
...
+
+  if (process_title) free(process_title);
+  process_title = strdup(title);
+
+  prctl(PR_SET_NAME, process_title);
+
+  pthread_mutex_unlock(&process_title_mutex_);
+}
+#else
+size_t uv_get_process_title(char* buffer, size_t size) {
```
bnoordhuis, line 116: Should zero terminate the buffer if `size > 0`

src/unix/linux.c
```diff
@@ -29,6 +29,10 @@
 
 #include <sys/inotify.h>
 #include <sys/sysinfo.h>
+#include <sys/prctl.h>
+#include <linux/prctl.h>
```
bnoordhuis: Do you need to include `linux/prctl.h`? `sys/prctl.h` should pull it in for you.

src/win/util.c
```diff
@@ -150,3 +149,76 @@ int uv_parent_pid() {
...
+  title = (char *) malloc(length);
+  if (!title) return 0;
+
+  /* Do utf16 -> utf8 conversion here */
+  if (!WideCharToMultiByte(CP_UTF8, 0, title_w, -1, title_, length, NULL, NULL)) {
+    free(title);
+    return 0;
+  }
+
+  return uv__strlcpy(buffer, title, size);
```
bnoordhuis: This cannot have compiled: `uv__strlcpy()` is only available on UNIX so far. :-/

indutny:
@bnoordhuis Everything fixed now ;)

bnoordhuis:
@indutny: the sunos code still uses `setproctitle()`

indutny:
@bnoordhuis should we turn off setproctitle for sunos for now?

indutny:
@bnoordhuis time to review!

bnoordhuis:
It seems @igorzi sneakily implemented all of this in 3d189de6.

PR #212, opened 2011-10-05
cmake: initial commit

okuoku:
Initial CMake support. Only builds static libuv.a (or uv.lib) but i think it's enough for node.

Win32 seems fine with MSVC10.
Doesn't compile on FreeBSD at src/unix/freebsd.c. I'll send another patch later..


ry:
Thanks for the thought but we've already got two build systems. I don't want to introduce another.

PR #213, opened 2011-10-06
FreeBSD: Fix FreeBSD build.

okuoku:
Fix FreeBSD 8.x/9.0 build.

Still fails 3 tests. I'll look into it later..
- Fatal error in test/test-tcp-close.c on line 151: never_cb should never be called
- Assertion failed in test/test-udp-ipv6.c on line 142: recv_cb_called == 1
- Assertion failed in test/test-pipe-bind-error.c on line 56: r == -1


bnoordhuis:
Right - that's Darwin code, not FreeBSD code.

I'll merge it but can you sign [the CLA](http://nodejs.org/cla.html)?

bnoordhuis:
PS: `test-pipe-bind-error` and `test-tcp-close` are known broken (not specific to FreeBSD).

`test-udp-ipv6` fails because the dual-stack UDP server doesn't receive datagrams from the IPv4 client.

okuoku:
I did.

okuoku:
Yup, tapped wrong place.. sorry.

bnoordhuis:
Thanks, much appreciated. Merged in d1016de.

PR #214, opened 2011-10-07
[uv.h] document memory functions and loadavg

indutny:
nt


include/uv.h
```diff
@@ -1063,7 +1063,7 @@ struct uv_fs_event_s {
 };
 
 
-/* Gets load avg */
+/* Gets load avg (in same format as in /proc/loadavg) */
```
ry: can you describe it in a cross platform way?

indutny: Is link to wkipedia article enough? :)

ry: sure

ry:
landed in 5ac040069eed526df7b280e3cdf39d06ca42444a. thanks

PR #217, opened 2011-10-11
set multicast loop and set multicast ttl

defunctzombie:
Add udp methods to set multicast loopback and ttl socket options.

Trying to reach feature parity with existing node multicast support.


bnoordhuis:
LGTM. Can you add tests?

bnoordhuis:
@shtylman: Can you add the tests? I'm fine with tests that simply set the TTL / loopback flag and check the return code (a loopback test helper gets you extra karma points, though :-).

I want to repurpose the old multicast test in Node, this pull request is one of the missing pieces.

defunctzombie:
I started working on the tests however ran into a bit of a snag. I am not sure how this deferred fd assignment is supposed to work. Apparently, the handle fd is not set on uv_udp_init but instead this happens sometime later. A bind call supposedly sets it but this is not needed for sending multicast packets. Any pointers on how to understand the fd assignment would be appreciated :)

bnoordhuis:
`uv__udp_maybe_deferred_bind()` will create the file descriptor for you (come to think of it, `uv_udp_set_membership()` should probably do that too).

If binding is not desirable, `uv__udp_maybe_deferred_bind()` should be split into two functions: one to create the socket, the other to do the binding. Ping me on IRC if you want to discuss that.

defunctzombie:
Thoughts on bringing this in? The new node.js version will release and one major hole still is the udp side of things.

isaacs:
@bnoordhuis Can you take a look at this and see why Github thinks it can't be merged cleanly?  Let's get this into 0.7 soon.

defunctzombie:
If I need to rejigger any patches here please let me know and I will be happy to take a look.

bnoordhuis:
@shtylman: Can you rebase it against master and squash it into one or two commits?

bnoordhuis:
Thanks, Roman. Multicast TTL and loopback support already landed but I merged your test in 9fa2cf2. Thanks again!

PR #219, opened 2011-10-13
Add support for OpenBSD

jedisct1:
This makes libuv compile and work on OpenBSD.

Tested on OpenBSD 5.0/amd64 and OpenBSD 5.0/i386.


bnoordhuis:
Thanks, Frank. Merged in 4ab1990.

PR #220, opened 2011-10-17
Fix pasto: uv_udp_recv_start() receives a UDP datagram, it doesn't send data.

jedisct1:
Just a copy/paste error in uv.h


bnoordhuis:
Thanks, Frank. Applied in e8a418e.

PR #222, opened 2011-10-20
win: MinGW32-w64 build fix

okuoku:
In Issue 221, several MinGW32-w64 installation lacks these definitions.
Definitions took from (original) MinGW project.

I don't know libuv's Win64 support status, but the tests seems good:

83 Success 3 Fail with MinGW64 (TDM-GCC on Windows, Ubuntu)
- fs_readdir_file
- udp_multicast_join
- tty

In addition, Tlhelp32.h => tlhelp32.h to fix build on case-sensitive file systems.


okuoku:
Added RWLOCK stuffs into patch. Hmm, same things in 2 place(uv-win.h and winapi.h)..

piscisaureus:
@piscisaureus Self-mention so I don't forget.

PR #224, opened 2011-10-21
fixed a memory error in fs__readdir on win32

mei-rune:


bnoordhuis:
@runner-mei: Thanks for pointing that out. We couldn't take your patch because you hadn't signed the CLA yet and we wanted this fixed in time for 0.5.10 but the attribution goes to you. Thanks again and please keep those patches coming.

PR #226, opened 2011-10-26
LLVM GCC support with OS X 10.7 and Xcode 4.2

CarterA:
Currently, the GYP system creates an Xcode project file with a hardcoded value for the GCC_VERSION setting. In newer versions of Xcode, plain GCC is not included. It has been replaced by LLVM-GCC, which uses GCC as a front-end for the LLVM compiler. Using a hardcoded value in the project file means that the initial build via the command line won't work:  you have to go into Xcode and switch the compiler setting to a compiler you have installed. This patch simply removes the hardcoded value, which Xcode will interpret as "use the default setting", which should work in all cases.


bnoordhuis:
Thanks, much appreciated. Merged in 1393ee7.

PR #229, opened 2011-10-28
fixed  write data overwrites the old data  when open a file with O_APPEND flag.

mei-rune:
fixed  write data overwrites the old data  when open a file with O_APPEND flag.


bnoordhuis:
@piscisaureus and @igorzi: Looks OK to me. Test passes on UNIX.

igorzi:
Looks good.

bnoordhuis:
@runner-mei: Thanks, I'll merge it. Can you sign [the CLA](http://nodejs.org/cla.html)?

mei-rune:
how to sign the CLA?

mmalecki:
Go to [this part](http://nodejs.org/cla.html#signelectronically), fill in your details and submit it (there are other options, like printing, signing and e-mailing it, but doing it electronically is the most convenient one).

mei-rune:
I sign it

mei-rune:
Note that this patch does not completely solve the problem, when the call fs__write with offset != -1, there will still be the problem.

piscisaureus:
I don't like the patch. We should specify the correct flags for CreateFile instead. From a user comment at the
[CreateFile documentation page](http://msdn.microsoft.com/en-us/library/windows/desktop/aa363858%28v=vs.85%29.aspx):

> Opening File for APPEND access
> 
> You can get atomic append on local files by opening a file with FILE_APPEND_DATA access and _without_ FILE_WRITE_DATA access. If you do this then all writes will ignore the the current file pointer and be done at the end-of file.
> 
> `HANDLE hFile = CreateFile(TEXT("c:\\file.txt"), FILE_APPEND_DATA , FILE_SHARE_WRITE, 0, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, 0);`
> 
> The append behavior is properly synchronized between multiple writes (with or without multiple handles), where the typical way I've seen this implemented (by seeking to EOF and then writing) has a race condition if multiple threads / processes are appending to the same file.
> 
> This behavior is documented in [Windows Driver Kit / Device and Driver Technologies / Installable File System / Reference / IO Manager Routines / IoCreateFileSpecifyDeviceObject](http://msdn.microsoft.com/en-us/library/ff548289.aspx)

mei-rune:
Oh, you have a better way, then I closed it. Please fix it as soon as possible.

igorzi:
This was landed in fe97c4dc63552ac830a82340f1c59c2a8bdfafe1.

Issue #228, opened 2011-10-27
TCP_KEEPALIVE is not defined under MinGW


luislavena:
Hello,

Attempting to compile latest master against MinGW GCC 4.5.2 got the following error:

```
gcc -Iinclude -Iinclude/uv-private -g --std=gnu89 -D_WIN32_WINNT=0x0501 -Isrc/ares/config_win32 -o src/win/tcp.o -c src/win/tcp.c
src/win/tcp.c: In function 'uv__tcp_keepalive':
src/win/tcp.c:74:28: error: 'TCP_KEEPALIVE' undeclared (first use in this function)
src/win/tcp.c:74:28: note: each undeclared identifier is reported only once for each function it appears in
make: *** [src/win/tcp.o] Error 1
```

Seems to be related to the usage of `TCP_KEEPALIVE` by `setsockopt`

I couldn't find any reference in MSDN to TCP_KEEPALIVE, only SO_KEEPALIVE:

http://msdn.microsoft.com/en-us/library/windows/desktop/dd877220(v=vs.85).aspx
http://msdn.microsoft.com/en-us/library/windows/desktop/ms740476(v=vs.85).aspx

I'm missing something?

Thank you.

PR #230, opened 2011-10-28
Seems setsockopt doesn't know TCP_KEEPALIVE (MinGW)

luislavena:
But instead clearly compiles using SO_KEEPALIVE, as documented in MSDN.

This corrects Issue #228


piscisaureus:
Thanks. Landed in 99b512e890bd424eaac8139a567733f03a7975a0.

Issue #160, opened 2011-08-30
Adding the way to set affinity


kuebk:
I would like to add a possibility of setting/getting affinity for each spawned process.

I want to:
- extend uv_process_options_t to contain an array(?) of cores to set process affinity to,
- add functions to set/get affinity on given process

What you think?

ry:
can you suggest some API?

kuebk:
Extend uv_process_options_t with unsigned long cpu_affinity?

int uv_set_affinity(uv_process_t\* process, unsigned long cpu_affinity);
int uv_get_affinity(uv_process_t\* process, unsigned long cpu_affinity*);

EDIT: Missclicked comment & close.

bnoordhuis:
Closing, we're not going to do this.

PR #232, opened 2011-11-03
set/get affinity

kuebk:
Lets you get/set affinity on the given pid.

Mentioned back in time here: https://github.com/joyent/libuv/issues/160

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -982,6 +982,15 @@ UV_EXTERN int uv_spawn(uv_loop_t*, uv_process_t*,
  */
 UV_EXTERN int uv_process_kill(uv_process_t*, int signum);
 
+/*
+ * Sets the cpu affinity on the given process pid
+ */
+UV_EXTERN int uv_set_affinity(int pid, char *core, int core_size);
+
+/*
+ * Gets the cpu affinity of the given process pid
+ */
+UV_EXTERN int uv_get_affinity(int pid, char **core, int *core_size);
 
 /*
  * uv_work_t is a subclass of uv_req_t
--- a/include/uv.h
+++ b/include/uv.h
@@ -946,7 +946,14 @@ typedef struct uv_process_options_s {
    * in. Stands for current working directory.
    */
   char* cwd;
-
+  /*
+   * Array of cores to run the process on
+   */
+  char* affinity;
+  /*
+   * Sizeof affinity
+   */
+  int affinity_size;
   /*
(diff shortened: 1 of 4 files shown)


include/uv.h
```diff
@@ -946,7 +946,14 @@ typedef struct uv_process_options_s {
...
   char* cwd;
-
+  /*
+   * Array of cores to run the process on
+   */
+  char* affinity;
+  /*
+   * Sizeof affinity
+   */
+  int affinity_size;
```
ry, line 956: Needs docs on how these should be set for default scheduling

include/uv.h
```diff
@@ -982,6 +989,15 @@ UV_EXTERN int uv_spawn(uv_loop_t*, uv_process_t*,
...
 
+/*
+ * Sets the cpu affinity on the given process pid
+ */
+UV_EXTERN int uv_set_affinity(int pid, char *core, int core_size);
+
+/*
+ * Gets the cpu affinity of the given process pid
+ */
+UV_EXTERN int uv_get_affinity(int pid, char **core, int *core_size);
```
ry, line 1000: Do we really need to add to `process_options_t` if we have these?

ry, line 1000: - should be with the type. `char**`. What should the `core` parameter be set to? It's unclear how this works. More docs?

kuebk, line 1000: If you want `uv_spawn` to do all the work then yes we need to modify `process_options_t`.

kuebk:
Test included.

ry:
@bnoordhuis @piscisaureus can you guys review this?

piscisaureus:
Should it not return uv_err_t so it can properly report errors?

bnoordhuis:
I suppose my first question is: why this feature? Why not let the OS figure it out?

kuebk:
We've experienced better results when we have set affinity on our own.

bnoordhuis:
Can you quantify that? Another thing: this patch is linux specific.

pietern:
For IPC on a NUMA system, I've seen ~2x performance when the processes were set to run on neighboring cores.

bnoordhuis:
After mulling it over, I've decided we won't take this patch.

@pietern: I suspect that you're looking at a badly tuned system or a bug in the OS scheduler if you see performance differences like that. (The stock linux 3.x scheduler for example has exactly that bug in power-saving mode.)

PR #235, opened 2011-11-05
Temporary fix to multiple initialisation of eio - pthread-specific

paddybyers:
Issue summary is here:

https://github.com/paddybyers/node/issues/15

### DIFF
--- a/src/unix/uv-eio.c
+++ b/src/unix/uv-eio.c
@@ -27,10 +27,6 @@
 #include <stdio.h>
 #include <pthread.h>
 
-/* TODO remove me! */
-static uv_loop_t* main_loop;
-
-
 static void uv_eio_do_poll(uv_idle_t* watcher, int status) {
   assert(watcher == &(watcher->loop->uv_eio_poller));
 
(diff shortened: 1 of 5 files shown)


src/unix/uv-eio.c
```diff
@@ -118,12 +127,7 @@ void uv_eio_init(uv_loop_t* loop) {
         uv_eio_done_poll_notifier_cb);
     uv_unref(loop);
 
-    eio_init(uv_eio_want_poll, uv_eio_done_poll);
-    /*
-     * Don't handle more than 10 reqs on each eio_poll(). This is to avoid
-     * race conditions. See Node's test/simple/test-eio-race.js
-     */
-    eio_set_max_poll_reqs(10);
+    pthread_once(&eio_initialised, eio_init_once);
```
bnoordhuis, line 124: `pthread_once()` is not fork-safe. If another thread calls `fork()` right when `pthread_once()` is doing its magic, `eio_initialised` will be in an undefined state.

It doesn't matter for Node because we always `execve()` immediately afterwards but it will be a problem in other applications that depend on libuv.

Then again, there is probably not much we can do about it. We can re-initialize libuv mutexes and guards in `pthread_atfork()` handlers but libeio uses mutexes internally that are out of our control.

Consider this the ramblings of a (not so) old man.

paddybyers, line 124: Not all platforms have pthread_atfork() either.

The other way to approach it is to pass the buck to the caller, just like eio does, and expose an initialiser that you say must only be called once.

paddybyers:
It's more complex than I thought and that second commit doesn't work. I commented here:
https://github.com/paddybyers/node/issues/14#issuecomment-2648139

The first commit is OK but the second one is not.

I will revert that .

bnoordhuis:
Closing, a newer version of this PR landed in abf9654.

PR #237, opened 2011-11-06
handle EPERM like EACCES

isaacs:
Granted, it's not quite 100% the same thing, but it seems pretty close.


bnoordhuis:
This is a potentially confusing change. Often EPERM == EACCES for all practical purposes but take `rename()`, for example: it returns EACCES when the source or the target aren't writeable but EPERM when the underlying file system doesn't support the operation. That's two pretty distinct errors.

isaacs:
Indeed.  Is EPERM meaningful on windows?  If so we should perhaps just define a UV_EPERM and do it that way.

Issue #206, opened 2011-10-01
src/uv-common.c:92: uv_err_name: Assertion `0' failed.


tommypowerz:
my node server.js is breaking from time to time with this error:

src/uv-common.c:92: uv_err_name: Assertion `0' failed.

it uses socket.io and get many connections...

is there a way to try/catch this somehow in my server?

i'm totally lost.. there are no more informations when it crashes....

thanks in advance for your help guys..

bnoordhuis:
Try this:

```
gdb --args node server.js
break uv-common.c:92
run
# wait until breakpoint is hit
backtrace
```

Post the backtrace here.

bnoordhuis:
Oh, and make sure you build Node with `./configure --debug` (and substitute node with node_g). The release build doesn't have debug symbols so the backtrace won't be very enlightening.

tommypowerz:
ok i will try that and post what i find out...

how do i (and substitute node with node_g). ?

node_g server.js ?

or gdb --args node server.js ?

tommypowerz:
Program received signal SIGPIPE, Broken pipe.
[Switching to Thread 0x7fabbdf6c6f0 (LWP 18634)]
0x00007fabbc7b275b in write () from /lib/libpthread.so.0

(gdb) backtrace
#0  0x00007fabbc7b275b in write () from /lib/libpthread.so.0
#1  0x0000000000803a5a in uv__write (stream=0x1d16208) at src/unix/stream.c:354
#2  0x00000000008048b3 in uv_write (req=0x1cd9d88, stream=0x1d16208, bufs=0x7fffa250af30, 
    bufcnt=1, cb=0x54a320 <node::StreamWrap::AfterWrite(uv_write_s*, int)>)
    at src/unix/stream.c:730
#3  0x000000000054a678 in node::StreamWrap::Write (args=<value optimized out>)
    at ../src/stream_wrap.cc:242
#4  0x00000817031d9b2a in ?? ()
#5  0x00007fffa250afd0 in ?? ()
#6  0x00007fffa250afd8 in ?? ()
#7  0x0000000000000001 in ?? ()
#8  0x0000000000000000 in ?? ()

then i start it over and get:

events.js:48
        throw arguments[1]; // Unhandled 'error' event
        ^
Error: accept ENOTCONN
    at errnoException (net_uv.js:561:11)
    at TCP.onconnection (net_uv.js:707:17)

Program exited with code 01.

backtrace gives no stack

[...]

PR #238, opened 2011-11-06
fix abort in linux due to connection reset

JerrySievert:
add linux connection reset code, should fix issue #206, or at least the abort


thejh:
I think that this only fixes symptoms, not the actual cause - libuv uses translated error codes, this shouldn't happen.

JerrySievert:
correct, this does not fix the low-lying issue with libuv, but the connection is being reset, and code 104 (connection reset) is being raised.  the connection reset code was missing from the enum and the error translation.

with this patch, the underlying error is handled more gracefully, connection is reset correctly on the client/server side, and node continues processing connections.

bnoordhuis:
I can't accept this patch but I'm aware of the issue. Expect a fix in the next couple of days.

PR #239, opened 2011-11-07
map ESRCH

tj:


ry:
Can we get a test for `uv_kill(SOME_RANDOM_NUMBER, 0)` returning `UV_ESRCH`?

ry:
We need Windows to map to `UV_ESRCH` too here: https://github.com/joyent/libuv/blob/51df5e3b00942a0b31a999e04b1b647544e7d2b6/src/win/process.c#L1089

PR #243, opened 2011-11-10
Added experimental iOS support

typester:
I've been using libev on iOS for long time, and hope libuv too.

Unfortunately libuv depends CoreServices.framework which is not included in iOS.
But libuv seems to depend CoreServices.framework for only AbsoluteToNanoseconds, and it's possible to replace.
This patch did this.

See also: http://developer.apple.com/library/mac/#qa/qa1398/_index.html

I also titled this experimental because I ran run-tests at my jailbroken phone but 8 tests were failed.
The output is here. https://gist.github.com/1354461

### DIFF
--- a/src/unix/darwin.c
+++ b/src/unix/darwin.c
@@ -25,7 +25,12 @@
 #include <stdint.h>
 #include <errno.h>
 
+#include <TargetConditionals.h>
+
+#if !TARGET_OS_IPHONE
 #include <CoreServices/CoreServices.h>
+#endif
+
 #include <mach/mach.h>
 #include <mach/mach_time.h>
 #include <mach-o/dyld.h> /* _NSGetExecutablePath */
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
Any reason to have this only for iOS and not just all Darwin targets?

typester:
I'm not sure which implementation is better on performance, so I kept darwin impl as-is.

But if there's no performance issues, I agree to remove CoreServices.framework dependency both Darwin and iOS platforms.

typester:
s/performance/performance or precision/

TooTallNate:
+1 (though the patch could be cleaned up a little bit)

ry:
looks good. @typester before we can merge you need to agree to the CLA http://nodejs.org/cla.html

typester:
@ry Thanks. I agreed and signed by electronically way

src/unix/darwin.c
```diff
@@ -33,7 +38,24 @@
 #include <sys/sysctl.h>
 #include <unistd.h>  /* sysconf */
 
+#if TARGET_OS_IPHONE
```
TooTallNate, line 41: Perhaps this should be changed to something like:

``` c
#if defined(Nanoseconds) && defined(AbsoluteToNanoseconds)
```

In case there's some other darwin platform someday that doesn't have this (you'd have to swap the order of the definitions).

ry:
Thanks. Please consider testing the GYP/Xcode build (see README for instructions) - you will probably need to remove https://github.com/joyent/libuv/blob/3cbe7c3f7952aa249450438b21cf4f346bb88e22/uv.gyp#L208-211

PR #245, opened 2011-11-10
Use (void) for a function taking no arguments

jedisct1:
This fixes gcc's -Wstrict-prototypes.


bnoordhuis:
Thanks, Frank. Applied in 9322dd1.

PR #249, opened 2011-11-13
Fix typo that broke libuv on NetBSD: psysconf() -> sysconf()

jedisct1:
The psysconf() function doesn't exist and never did AFAIK. And libuv just didn't compile on NetBSD-stable because of this typo.


bnoordhuis:
Thanks, Frank. Merged in 93faeb2.

PR #251, opened 2011-11-16
[tty] unreference readable tty streams

indutny:
See joyent/node#1934 for reference


bnoordhuis:
Not taking this, sorry Fedor. The whole reference counting scheme is about to be redone.

PR #252, opened 2011-11-19
Use .lib prefix only on MSVC builds by checking host_os (fix build on mingw)

adammw:
Assuming host_os is set either by options.gypi or node's configure script (by joyent/node#2148), we can use either the .lib version for windows visual studio builds, or without .lib (the .a version) for mingw or cross compile.
If host_os is not set, then we just default to target os, which would be "win" for msvc which would result in the same as what we already have...

Or is there a better way that doesn't modify the gyp files? (That would be good as it would mean I don't have to try and get v8 to do the same thing on their gyp files).


bnoordhuis:
Sorry @adammw, it appears this PR got overlooked. I think these issues have been resolved in the current master?

piscisaureus:
@adamw I think this has been resolved. If not, please file another issue.

PR #255, opened 2011-11-23
SO_REUSEADDR, 'uv_udp_set_broadcast' and 'uv_udp_set_multicast_ttl'

wankdanker:
In an attempt to move issue joyent/node#2163 forward I am submitting the libuv specific changes here.

### DIFF
--- a/src/unix/udp.c
+++ b/src/unix/udp.c
@@ -500,6 +500,24 @@ int uv_udp_set_membership(uv_udp_t* handle, const char* multicast_addr,
   return 0;
 }
 
+int uv_udp_set_multicast_ttl(uv_udp_t* handle, int ttl) {
+  if (setsockopt(handle->fd, IPPROTO_IP, IP_MULTICAST_TTL, &ttl, sizeof ttl) == -1) {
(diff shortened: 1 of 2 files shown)


src/unix/udp.c
```diff
@@ -494,6 +500,24 @@ int uv_udp_set_membership(uv_udp_t* handle, const char* multicast_addr,
   return 0;
 }
 
+int uv_udp_set_multicast_ttl(uv_udp_t* handle, int ttl) {
+  if (setsockopt(handle->fd, IPPROTO_IP, IP_MULTICAST_TTL, &ttl, sizeof ttl) == -1) {
```
bnoordhuis, line 504: The socket may not have been created (`handle->fd` will be -1).

@piscisaureus: I want to revisit lazy socket creation for 0.8. The benefits are questionable, the drawbacks numerous. It's not unreasonable to have the caller specify upfront what type of socket he wants. It doesn't break the Node API either.

bnoordhuis:
Thanks Dan, merged in b674187 and 497b1ec.

PR #257, opened 2011-11-23
processes should not wait for stdio before terminating

seebees:
If I understand https://github.com/joyent/node/issues/818, then here is a test for it.


bnoordhuis:
Thanks, Ryan. Merged in 59f9a73.

PR #264, opened 2011-11-29
adding a signal of support on win32.

mei-rune:
I am a developer on the win32 platform, I found libuv do not support signal, so adding a signal of support on win32, I hope it helps you.


mei-rune:
I accord with your request to modified

bnoordhuis:
@piscisaureus: What's the status? Is this PR still relevant?

piscisaureus:
Yes, it is. This patch was actually requested by me :-). I am just too scared to review it and I'm also not sure what to do with the crt signal support that's in there (I don't really think we need it. I don't want people to handle SIGFPE in node etc.)

migounette:
I am bit confuse by the current code, it works fine for internals signal emitted but how do you manage external signal. 
Eg Like under Linux if you want to send a signal from an external program such as kill with an equivalent of taskkill under windows or another process that wants to send SIGBREAK or SIGTERM ?

My 2 cents

piscisaureus:
For node, you can send the magic "start debugger" signal with `node debug -p «pid»` (from the command line) or  `process._debugProcess(«pid»)` (from js).

How it works is another story. In unix, this is indeed a wrapper for `kill(pid, SIGUSR1)`. On windows, node writes the pointer to the `EnableDebugThreadProc` function to a piece of named shared memory with a predictable name. When another node process wants to start the debugger, it reads this shared memory and injects a thread (with CreateRemoteThread) into the debuggee, with `EnableDebugThreadProc` as the thread entry point.

PR #269, opened 2011-12-01
Add interval and count to keep-alive

jameshartig:
We should be able to specify the probe interval and the failure count for keep-alive's on sockets.

Corresponding node pull request: https://github.com/joyent/node/pull/2239

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -483,9 +483,11 @@ UV_EXTERN int uv_tcp_nodelay(uv_tcp_t* handle, int enable);
 /* Enable/disable TCP keep-alive.
  *
  * `ms` is the initial delay in seconds, ignored when `enable` is zero.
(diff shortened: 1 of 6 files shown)


piscisaureus:
> We should be able to specify the probe interval and the failure count for keep-alive's on sockets.

Why is that needed?

jameshartig:
It is needed because the default probe interval and failure counts set by the OS are too extreme and if you have an application that keeps a lot of sockets open (say a server that hosts a lot of clients over socket connections) you'd want to fine-tune the values to be very low.

ibc:
I agree. Depending on OS default values is not (always) good.

jameshartig:
If I make a new pull request pointing at master and add the documentation for count/interval being 0, can it be merged?

jameshartig:
Updated with Windows support for interval, documentation added that @txdv suggested and targeted master.
https://github.com/joyent/libuv/pull/729

PR #272, opened 2011-12-09
test: make test runner return non-zero in case of failure

mmalecki:
Test runner always exited with 0.


bnoordhuis:
Thanks Maciej, merged in b06da4c.

PR #273, opened 2011-12-10
Travis CI

mmalecki:
As requested.

Workflow is the same as with joyent/node#2261.

Depends on #272 (`script` always returns `0` now which makes it pretty useless).


bnoordhuis:
Thanks Maciej, merged in 5989961 and b5762b2.

PR #276, opened 2011-12-12
Fix missing increments of loop->counters

shigeki:
I found this when I look up the result of process.uvCounters(). 
Obviously it's not a serious issue that some of init numbers of loop->counters are missing, however, I think that it would be nice to fix it for the future debug use and I would be glad if this is accepted.

Note that uv__req_init() in src/unix/core.c was changed into uv_req_init() so as to be consistent with the code of Win.
The comment and increments of counter was deleted in src/unix/udp.c as https://github.com/shigeki/libuv/commit/4ad5a9e7cdd8b7f669e9f850a26742c491c35299#diff-8 but I'm not sure the reason why the previous comment was written which seems to be appeared at the first commit.


bnoordhuis:
LGTM but can you undo the rename from `uv__req_init()` to `uv_req_init()`, it's inconsistent with how (most of) the other internal functions are named in src/unix/.

> The comment and increments of counter was deleted in src/unix/udp.c as https://github.com/shigeki/libuv/commit/4ad5a9e7cdd8b7f669e9f850a26742c491c35299#diff-8 but I'm not sure the reason why the previous comment was written which seems to be appeared at the first commit.

There was a bug at the time in `uv__req_init()` that zeroed the `.data` field. I couldn't fix it right away since other code depended on that behaviour.

shigeki:
Okay I will change my patch. Which one do you prefer?
1.  create new  uv_req_init() to wrap uv__req_init(). The uv__req_init() is left to have backward  compatibility. Every existing call of uv__req_init() is changed into that of uv_req_init(). 
   https://gist.github.com/1467739
   and the patch above as 4ad5a9e
2. increment the loop counter every time after the call of uv__req_init()
   For core.c case,
   https://gist.github.com/1467785
3. No change for req_init. Left as is.

I think  1 is better, but I will follow your preference.

bnoordhuis:
4ad5a9e is okay, just revert the rename from `uv__req_init()` to `uv_req_init()`. The `uv__` prefix is a convention we use to signal that the function is internal.

shigeki:
I got it.  Please review 7727385 .

bnoordhuis:
Thanks Shigeki, landed in v0.6 in ba52023 and merged into master in 4c6008f.

shigeki:
Ben, thanks for merging.
@ry Now, the process.uvCounters() in Node returns correct numbers. Can be it a public API or still private? If the former, I will ready to submit its description to docs.

Issue #262, opened 2011-11-28
Add asynchronous interface for libuv itself


rmtew:
Currently the control flow for the libuv framework is (as I understand it):

  ... set up application ...
  uv_run()
  ... return from blocking indefinitely ...

This  means that if an application is to use libuv, it needs to let libuv drive it.  This affects the way in which the application can be written.  If there was a minimal change to the libuv running mechanism to let it be pumped with a specified timeout, then an application could instead drive libuv more in the manner of a library and a main loop could be written in a straight-forward and synchronous manner.

I suggest that this could be done with the following changes:
1. The contents of the while loop in UV_RUN are moved to UV_PUMP.
2. A uv_pump() method is added that takes a optional timeout parameter.
3. The poll method observes the timeout parameter instead of the result of uv_get_poll_timeout(), or perhaps gets it from that function for this pending one-off iteration.

I can write and provide a patch, if there is an indication that one would be applied.

Cheers.

bnoordhuis:
Sure, I'd take a patch for that. I don't think libev has facilities for timed polls (without adding explicit timers to the event loop) but the diff below should be all you need for a simple immediate poll.

``` diff
diff --git a/src/unix/core.c b/src/unix/core.c
index cf3138f..2cf0ee5 100644
--- a/src/unix/core.c
+++ b/src/unix/core.c
@@ -194,6 +194,12 @@ int uv_run(uv_loop_t* loop) {
 }


+int uv_poll(uv_loop_t* loop) {
+  ev_run(loop->ev, EVRUN_NOWAIT);
+  return 0;
+}
+
+
 void uv__handle_init(uv_loop_t* loop, uv_handle_t* handle,
     uv_handle_type type) {
   loop->counters.handle_init++;
```

txdv:
looks like a run_once method to me, now the windows equivalent is the only thing missing.

mattn:
I hope `uv_run_once()`.

bnoordhuis:
I called it `uv_poll()` because it lets you poll for events at your convenience but the actual name is inconsequential. If someone wants to submit a patch that includes Windows support, please go ahead, else it will have to wait until someone on the team has some spare cycles.

mattn:
Do you mean that the `uv_poll()` do once of `uv_run()`'s loop?

bnoordhuis:
Yes, poll once without blocking if there are no events.

mattn:
[...]

PR #277, opened 2011-12-13
add uv_run_once()

mattn:
Added `uv_run_once()` for #262

### DIFF
--- a/src/win/core.c
+++ b/src/win/core.c
@@ -192,9 +192,7 @@ static void uv_poll_ex(uv_loop_t* loop, int block) {
   }
 }
 
-
-#define UV_LOOP(loop, poll)                                                   \
-  while ((loop)->refs > 0) {                                                  \
+#define UV_LOOP_ONCE(loop, poll)                                              \
     uv_update_time((loop));                                                   \
     uv_process_timers((loop));                                                \
                                                                               \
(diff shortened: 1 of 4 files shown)


src/win/core.c
```diff
@@ -208,7 +206,7 @@ static void uv_poll_ex(uv_loop_t* loop, int block) {
     uv_process_endgames((loop));                                              \
                                                                               \
     if ((loop)->refs <= 0) {                                                  \
-      break;                                                                  \
+      goto LOOP_BREAK;                                                        \
```
bnoordhuis: If you wrap the body of UV_LOOP_ONCE() in a `do { ... } while (0)`, you can `break` here instead of `goto`.

I would prefer that. A macro should be sanitary, this one has flow control leaking out.

mattn:
Updated. If you hope that this changes is included in one change set, I'll re-create pull request.

bnoordhuis:
Can you remove the LOOP_BREAK labels elsewhere in the code?

mattn:
Ah, sorry ,sorry... I'll fix soon

mattn:
done

bnoordhuis:
Thanks Yasuhiro, I landed your patch in the master branch in e53cecb.

mattn:
Thanks.
I'll update https://github.com/mattn/go-uv  too in few days. :)

PR #284, opened 2011-12-28
fix windows build.

mattn:
windows build does not refer `eio.h`.


bnoordhuis:
Thanks for the report. I didn't take your patch but I fixed the Windows build in 85f6b79.

PR #285, opened 2011-12-31
fix: libuv android(NDK) compilation in visual studio

llafuente:
Hi,

This will allow libuv to be compiled with "vs-android", http://code.google.com/p/vs-android/

some define(s) need to be tweaked.


bnoordhuis:
Don't take this the wrong way but I don't really want to take this patch. You can count the people that compile libuv for android with visual studio on one hand. You won't even need all fingers.

PR #290, opened 2012-01-11
EVRUN_NOWAIT is not run event once. This should be EVRUN_ONCE.

typester:
I'm not sure about history of `uv_run_once`, but its current implementation is not run once event at ev backend.

`EVRUN_NOWAIT` should be for different function name, not `uv_run_once`, IMO.


bnoordhuis:
Thanks Daisuke, merged in e8494dd with a slightly reworded commit.

PR #292, opened 2012-01-15
[unix] remove eio_set_max_poll_reqs

indutny:
See discussion here: bde908f565f9b90f6a18a6ce266cc0c144fd4acb and here: joyent/node#2508


bnoordhuis:
LGTM in general but the test is timing dependent. To wit:

```
$ out/Debug/run-tests eio_overflow eio_overflow
$ valgrind -q out/Debug/run-tests eio_overflow eio_overflow
Assertion failed in ../test/test-eio-overflow.c on line 59: opened - closed < max_delta
Killed
```

test/test-eio-overflow.c
```diff
@@ -0,0 +1,88 @@
...
+#include "uv.h"
+#include "task.h"
+#include <stdio.h>
+#include <stdlib.h>
+
+
+static uv_idle_t idle;
+
+static const int max_opened = 10000;
+static const int max_delta = 4000;
```
bnoordhuis, line 31: Tweaking a magic number until it works isn't a good idea... What does this test attempt to prove, Fedor? That libeio doesn't get too backlogged?

indutny, line 31: Yes, it's trying to approve that eio queue is growing. Delta is not changing in not monotonic way, that's why I'm doing this kind of check.

indutny:
This is not relevant now . /cc @bnoordhuis

PR #294, opened 2012-01-21
Fixes inability to turn on background-bright in Windows

ghost:
win/tty.c is has all the code to allow bright backgrounds but simply has no ansi escape mappings that set bg_bright on. This patch adds the following escape sequence mappings:

2 - foreground and background bright off
5 - background bright on (follows Microsoft's ansi.sys mapping of blink -> bgbright)
25 - background bright off (blink off)
90-97 - bright foreground color (aixterm)
100-107 - bright background color (aixterm)
@gissues:{"order":91.66666666666666,"status":"backlog"}


ghost:
Fixed a while ago

PR #295, opened 2012-01-22
Map `ENAMETOOLONG`

### LINKED ISSUE #247: ENAMETOOLONG not mapped
running the connect tests lead me to a UV_UNKNOWN for this one

mmalecki:
I haven't done the Windows part - Windows seems to call `ENAMETOOLONG` `ERROR_BUFFER_OVERFLOW`, which is way too vague to me (and I have no machine to test it on).

Test included.

Fixes #247.

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -115,7 +115,8 @@ typedef intptr_t ssize_t;
   XX( 45, EAISOCKTYPE, "") \
   XX( 46, ESHUTDOWN, "") \
   XX( 47, EEXIST, "file already exists") \
-  XX( 48, ESRCH, "no such process")
+  XX( 48, ESRCH, "no such process") \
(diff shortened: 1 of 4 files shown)


ghost:
The ERROR_BUFFER_OVERFLOW occurs when the parameter specifying the size of the provided buffer (regardless of the size of the buffer!) is too small, and will always be accompanied by a return value from the called function specifying the size of the required buffer. An example is this: http://msdn.microsoft.com/en-us/library/windows/desktop/aa365917(v=vs.85).aspx.

"The buffer to receive the adapter information is too small. This value is returned if the buffer size indicated by the pOutBufLen parameter is too small to hold the adapter information or the pAdapterInfo parameter was a NULL pointer. When this error code is returned, the pOutBufLen parameter points to the required buffer size."

PR #297, opened 2012-01-25
uv.h: add EPERM to errno map to fix regression

philips:
EPERM isn't mapped in so chown returns an unknown error. This is a
regression from 0.4.12.

philips:node/ (master*) $ cat chown.js
var fs = require('fs')
fs.chown("/tmp/foobar", 100, 100, function(er){ console.log(er);})

philips:node/ (master*) $ ls -la /tmp/foobar
total 0
drwxr-xr-x   2 root  wheel   68 Jan 24 17:21 .
## 0.4

philips:node/ (master*) $ /usr/local/Cellar/node/0.4.12/bin/node chown.js
{ stack: [Getter/Setter],
  arguments: undefined,
  type: undefined,
  message: 'EPERM, Operation not permitted \'/tmp/foobar\'',
  errno: 1,
  code: 'EPERM',
  path: '/tmp/foobar' }
## master

philips:node/ (master*) $ ./node chown.js
{ [Error: UNKNOWN, unknown error '/tmp/foobar'] errno: -1, code: 'UNKNOWN', path: '/tmp/foobar' }
## AFTER

philips:node/ (master*) $ ./node chown.js
{ [Error: EPERM, operation not permitted '/tmp/foobar'] errno: 49, code: 'EPERM', path: '/tmp/foobar' }

This is also needed to fix isaacs/npm#2080 under 0.6 and make npm work on nfs squashed filesystems.

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -115,7 +115,8 @@ typedef intptr_t ssize_t;
   XX( 45, EAISOCKTYPE, "") \
   XX( 46, ESHUTDOWN, "") \
   XX( 47, EEXIST, "file already exists") \
-  XX( 48, ESRCH, "no such process")
+  XX( 48, ESRCH, "no such process") \
+  XX( 49, EPERM, "operation not permitted")
 
 
 #define UV_ERRNO_GEN(val, name, s) UV_##name = val,
--- a/src/unix/error.c
+++ b/src/unix/error.c
@@ -59,6 +59,7 @@ void uv_fatal_error(const int errorno, const char* syscall) {
 uv_err_code uv_translate_sys_error(int sys_errno) {
   switch (sys_errno) {
     case 0: return UV_OK;
+    case EPERM: return UV_EPERM;
     case ENOSYS: return UV_ENOSYS;
     case ENOTSOCK: return UV_ENOTSOCK;
     case ENOENT: return UV_ENOENT;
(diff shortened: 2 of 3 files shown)


philips:
The test is probably completely bogus. But I don't know how to guarantee an EPERM under all circumstances.

isaacs:
I'm +1 on this.  UNKNOWN error codes are a pain, and this is a very common one.

@igorzi @piscisaureus Should something to into src/win/error.c as well?

isaacs:
@philips Can you please sign the CLA?  http://nodejs.org/cla.html

Thanks.

philips:
Signed..For both my personal and Rackspace email.

isaacs:
Great.  0.6.9 is just about wrapping up, but we'll try to get this in the next one.

Can you rebase onto the latest v0.6 branch?

philips:
Sorry it took a couple of pushes to get right. My v0.6 was out of date. Should be good on top of v0.6 now.

philips:
Here it is against master since ENAMETOOLONG was added too... https://github.com/philips/libuv/tree/add-eperm-to-errno-map-master

isaacs:
Landed on 4cfda74de4a3901828942e9418c43a434c0accf6.  Thanks!

PR #299, opened 2012-01-28
Fix uv__is_root typo causing incorrect fs.exists results

ghost:


bnoordhuis:
Thanks Brendon, merged in 74cff55.

PR #306, opened 2012-02-06
Add EXDEV to the errno map.

TooTallNate:
Fixes the "UNKNOWN" errno first reported in joyent/node#2703.

I tried adding a test case, but I couldn't figure out how to detect a remote network path, so I wasn't sure if it was possible.

Also, there might be an additional change needed for Windows, I wasn't quite sure.


bnoordhuis:
Thanks Nathan, merged (with some wedging) in d33ccb0.

TooTallNate:
@bnoordhuis Thanks! How come the `src/unix/error.c` change wasn't necessary?

bnoordhuis:
Damn it, I fail at wedging. Fixed in dbc046c.

PR #310, opened 2012-02-09
Stop reading from broken streams on error (unix)

creationix:
As discussed with Bert on irc, this patch calls `uv_read_stop` on a stream when there is a read error before calling the read callback.  This is for the unix version of stream.c.


bnoordhuis:
As discussed on said IRC channel, this patch is not necessary. It's the responsibility of the read callback to `uv_close()` the handle on error. That requirement needs to be documented a little better though.

bnoordhuis:
I've updated the API documentation in 62206c2.

PR #314, opened 2012-02-17
Fix libuv with uclibc < 0.9.32

jedisct1:
uclibc didn't provide ifaddrs.h before version 0.9.32

It explicitly didn't install it because (quoting) "the corresponding functionality is disabled".
So, fix libuv on uclibc < 0.9.32 by returning ENOSYS for uv_interface_addresses()

This makes libuv work on embedded operating systems like DD-WRT and TomatoUSB.

### DIFF
--- a/src/unix/linux.c
+++ b/src/unix/linux.c
@@ -28,7 +28,15 @@
 #include <assert.h>
 #include <errno.h>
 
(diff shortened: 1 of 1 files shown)


bnoordhuis:
Thanks Frank, merged in 75ab1ba.

Issue #288, opened 2012-01-10
read watcher for uv


txdv:
libev has a readwatcher - it just notifies if there is at all something to read so the user itself can decide how to read the data.
This would be useful for node-ncurses, since it uses the ev read_watcher to determine if there is something to read and then it uses getch (provided by the curses library) to actually get the data.
So it is impossible to compile node-ncurses on windows.

getch returns an integer of the key pressed including information like weather 'alt', 'ctrl', etc. was pressed or not.
Using the uv approach to read from the TTY object does not take this kind of information into consideration.

txdv:
https://github.com/txdv/libuv/tree/read_watcher

txdv:
I updated the branch, took boords comments into consideration.
But we need a windows implementation for this as well, someone out there with windows please?

txdv:
I talked to bnoordhuis on irc and he suggested to make the read_watcher a handle that you can instantiate from an arbitrary fd.
There is a draw back: kqueue needs all file descriptors to be in the same kqueue (loop).

So I coded it that one can instantiate a read_watcher handle from an arbitrary stream (so the read_watcher ends up in the same loop as the stream itself). The handle is an abstraction, all the action is done in the stream anyway, it makes the API just look nicer.

https://github.com/txdv/libuv/tree/read_watcher2

Any comments bnoordhuis?

piscisaureus:
Maybe we can somehow have full ev_io like support for some handle types? I can provide it for windows, although it will be much less efficient than overlapped io. I think it could be useful because it is sometimes needed for integration with other libraries (e.g. c-ares).

txdv:
what other ev_io functionality are you missing?

On a comment on the read_watcher2 (it has a uv_read_watcher_t which is a handle).
I don't know, the implementation is kind of heavy for a simple callback.
In my opinion this would be nicer:

``` c
UV_EXTERN int uv_read_start(uv_stream_t*);
UV_EXTERN int uv_read_stop(uv_stream_t*);

UV_EXTERN int uv_read_set_callbacks(uv_stream_t*, uv_alloc_cb, uv_read_cb,
    uv_read2_cb, uv_read_watcher_cb);
```

Any comments?

bnoordhuis:
Closing, continues in #316.

Issue #265, opened 2011-11-30
Foreign file descriptors


waywardmonkeys:
I have some libraries that I'd like to use in conjunction with libuv that expose their file descriptors for integration into a foreign event loop (in this case, I'd like that to be libuv).

libev seems to support this, so I suspect that it is a matter of adding a new handle type with readable / writable callbacks?

Examples of publicly available libraries that do this include libpq's asynchronous support and ZeroMQ.

bnoordhuis:
Sounds reasonable. @ry?

saghul:
I'd love too see this, and perhaps I could help (at least with the Unix part). Is anyone working on this atm?

And, how should the API look like? Should it be a thin wrapper around ev_io or perhaps make it inherit from uv_stream_t?

Thanks!

bnoordhuis:
> Is anyone working on this atm?

No.

> And, how should the API look like? Should it be a thin wrapper around ev_io or perhaps make it inherit from uv_stream_t?

I haven't thought deeply about it but it should probably be a direct descendant of uv_handle_t. The file descriptor could very well be fundamentally incompatible with the concept of streams (think datagram sockets).

saghul:
Thanks for the info, Ben. I'll see if I can work something out! :-)

bnoordhuis:
Continues in #316.

PR #316, opened 2012-02-20
Added io handle (unix)

saghul:
Added uv handle for monitoring arbitrary file descriptors, it's a thin wrapper around ev_io (sorry, unix only, I don't know how to do it for windows).

This should close issues #288 and #265.

I have a very dirty test which I'll clean and submit as well, but I'd like to receive some initial feedback on the API.

Thanks,

### DIFF
--- a/include/uv-private/uv-unix.h
+++ b/include/uv-private/uv-unix.h
@@ -233,4 +233,10 @@ typedef void* uv_lib_t;
 
 #endif
 
+/* UV_IO */
+#define UV_IO_PRIVATE_FIELDS \
+  ev_io io_watcher; \
+  uv_io_cb io_cb; \
+  int events; \
+
 #endif /* UV_UNIX_H */
--- a/include/uv.h
+++ b/include/uv.h
@@ -142,7 +142,8 @@ typedef enum {
   UV_ARES_TASK,
   UV_ARES_EVENT,
   UV_PROCESS,
-  UV_FS_EVENT
+  UV_FS_EVENT,
+  UV_IO
 } uv_handle_type;
 
 typedef enum {
@@ -192,6 +193,7 @@ typedef struct uv_fs_s uv_fs_t;
 /* uv_fs_event_t is a subclass of uv_handle_t. */
 typedef struct uv_fs_event_s uv_fs_event_t;
 typedef struct uv_work_s uv_work_t;
+typedef struct uv_io_s uv_io_t;
 
 
 /*
@@ -279,6 +281,7 @@ typedef void (*uv_after_work_cb)(uv_work_t* req);
 */
 typedef void (*uv_fs_event_cb)(uv_fs_event_t* handle, const char* filename,
     int events, int status);
+typedef void (*uv_io_cb)(uv_io_t* handle, int status, int events);
 
 typedef enum {
   UV_LEAVE_GROUP = 0,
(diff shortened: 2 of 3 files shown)


bnoordhuis:
Thanks, the patch looks good.

Bear with me for now, I can't merge it right away. @piscisaureus and I have been discussing a similar API but we haven't quite made up our minds on how to do the actual I/O: tie in with regular streams, provide read/write functions like your patch does, let the user handle it, etc.

saghul:
Hi,

Sure, I just wanted to get some feedback on the patch, feel free to suggest  changes :-) FWIW, when I was implementing it I was about to just let the suer handle it, but I thought that providing functions would help, at least with errno mapping.

ronkorving:
This is beautiful. I was trying dealing with the same problem, breaking my head on it for days now, and you've provided an answer, saghul. I hope the team will deal with this sooner rather than later. If it makes its way into Node 0.8, I might actually be able to use it :)

saghul:
Hi @ronkorving,

Thanks for the comment :-) Apparently it will be added, but my proposed API might not be the chosen one, though I guess it'll be somewhat similar. Another missing part is the Windows equivalent for this, but I think @piscisaureus said it can be done.

The reason why I need it is because I implemented Python wrappers for libuv (http://github.com/saghul/pyuv) and I'm now building a library on top of that using greenlets. So in order to make Python sockets cooperative I need to run an Io handle on their fileno().

ronkorving:
I'm working on a wrapper module for Couchbase, for which the library is based on an eventloop like libev(ent). So I really need a callback when select() is positive, but before the read(). No need for libuv to gather all read buffers together for me, since libcouchbase already implements that. I need libcouchbase to do the read, whenever the event system tells me there is data available.

piscisaureus:
Binding to external libraries is why we need to support this. However on windows it will not offer the best performance. So @saghul, I think you would be better off finding a different way to deal with it for pyuv.

saghul:
Thanks for the comment, @piscisaureus. My little project is still in diapers, so I can still try to find an alternative :-) Anyhow, other similar projects use select on Windows, do you think this 'Io' handle performance would be worse than select?

piscisaureus:
Well, select() on windows is very bad :-). Our performance might actually be a little better, but not as good as with IOCP.

saghul:
Thanks! That might be good enough to get started with :-)

ronkorving:
"Not as good as with IOCP" is still much better than "impossible to integrate your library, sorry" or circumventing libuv and having to hack into libev directly.

graydon:
If I understand correctly, this is the "emulate readiness using completion" side of the problem uv was designed to solve. If so I think this is probably undesirable. Since uv exists, if I understand correctly, in a large measure to provide the completion (IOCP-style) emulation on readiness interfaces of libev. In other words, if you want readiness and you want to paint yourself into a corner of only working on unix, you ... already have libev. There's no point using libuv. Doesn't buy you anything. Adding code to punch through to libev just breaks the things libuv was made to provide _beyond_ libev.

Maybe I misunderstand.

(Also, `select` is worse than just "slow" on windows. It doesn't actually work with most types of filehandle, and it has a fixed upper limit on the number of sockets it does handle. It's a very poor readiness emulation.)

piscisaureus:
@saghul Why did you add uv_io_read and uv_io_write ?

saghul:
@piscisaureus I thought they could be handy, specially for setting the error in the loop in case something bad  happens. That way the user doesn't need to translate errnos.

PR #321, opened 2012-02-23
Fix compilation on iOS

jedisct1:
iOS doesn't provide _NSGetEnviron(). So if we're cross-compiling for iPhone, just define environ as extern char **.


bnoordhuis:
Does this patch work for you? I feel it's slightly easier to read.

``` diff
diff --git a/src/unix/process.c b/src/unix/process.c
index becf60f..431f4d8 100644
--- a/src/unix/process.c
+++ b/src/unix/process.c
@@ -31,6 +31,10 @@
 #include <stdio.h>

 #ifdef __APPLE__
+# include <TargetConditionals.h>
+#endif
+
+#if defined(__APPLE__) && !defined(TARGET_OS_IPHONE)
 # include <crt_externs.h>
 # define environ (*_NSGetEnviron())
 #else
```

jedisct1:
Sure! Your version indeed feels easier to read.

Thanks.

TooTallNate:
Nice patch. I would always just copy over `crt_externs.h` manually (from the iPhoneSimulator directory). This is better :)

bnoordhuis:
Thanks guys, merged in 57c5fa1.

PR #322, opened 2012-02-23
Add 'detached' option for child processes

AvianFlu:
This introduces a new `detached` option in `uv_process_options_t`, which denotes that a given child process should be spawned in a 'detached' state.  Put simply, the child will be run in the background, without a connection to the parent, and without any controlling tty.
### Unix

On unix, the best way to achieve this was by adding a second `fork(2)` - in this implementation, there is a short-lived 'first child', which makes a call to `setsid()` before forking again.  
This ensures two things: 
- The second child will be a session/process group leader.
- The original parent is under no obligation to stay running after detaching the child.

To make sure that `process->pid` refers correctly to the detached child and not the intermediate child, `mmap()` is used to create a small amount of shared memory (`sizeof(pid_t)`) between the forks.  The first child saves the second child's pid to this map, and then the map is removed after `process->pid` is set in the parent.
### Windows

The Windows implementation was mostly a matter of finding the right combination of flags for `CreateProcessW()`.  `CREATE_NO_WINDOW`, `CREATE_NEW_PROCESS_GROUP`, and `DETACHED_PROCESS` were all necessary to produce the same functionality as unix.

It was also necessary to set the timeout value on `RegisterWaitForSingleObject()` to 0 when `options.detached` is set.  The timeout without `detached` set is `INFINITE` - i.e. keep waiting until the child exits - but a timeout of 0 simply checks for a successful start and then calls the exit callback.

On both platforms, the parent treats the child as if it exited successfully, and no stdio handles are preserved.


piscisaureus:
You should not use RegisterWaitForSingleObject at all - instead just pretend that the child exits immeditately as on unix. 
Make sure to close the process handle that CreateProcess returns.

It would be nice to have some sort of test for this.

AvianFlu:
@piscisaureus Thanks! I fixed it to skip `RegisterWaitForSingleObject` and added a test case.

src/unix/process.c
```diff
@@ -175,27 +184,45 @@ int uv_spawn(uv_loop_t* loop, uv_process_t* process,
...
-      uv__process_init_pipe(options.stdin_stream, stdin_pipe, 0)) {
-    goto error;
-  }
+  if (options.detached) {
+    pid_map = mmap(0, sizeof(pid_t), PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
+    if (pid_map == MAP_FAILED) {
+      goto error;
+    }
+    if (options.stdin_stream) {
+      uv_close((uv_handle_t*)options.stdin_stream, _detach_cb);
```
bnoordhuis: Might as well call `uv_close(handle, NULL)` here if `_detach_cb` is just a no-op.

AvianFlu:
@bnoordhuis Both implementations have now been updated to remove the no-op `_detach_cb`.  If I have time today, I'll make a one-fork implementation for unix as a test case for what we discussed in IRC.

bnoordhuis:
Thanks Charlie, merged.

> If I have time today, I'll make a one-fork implementation for unix as a test case for what we discussed in IRC.

If you have some cycles to spare, then yes please. I can live with a second fork() for now though.

PR #323, opened 2012-02-23
win: fs: handle EOF in read

philips:
in luvit after we upgraded libuv from 243cfc to d3efef readSync started
failing.  It seems that the code cleanup stopped handling EOF

Trivially reproduced with this

```
local fs = require('fs')
print(fs.readFileSync('foo.luvit'))
```


piscisaureus:
Thanks. landed in fca18c3.

PR #324, opened 2012-02-23
Fix memory leak in uv_exepath() on OSX.

jedisct1:
The uv_exepath() function leaks memory on OSX. We don't need to pre-malloc() a buffer for the result of realpath(). Just pass NULL for the resolved path.


bnoordhuis:
Thanks Frank, merged in f6c8e78.

PR #326, opened 2012-02-24
win: use performance counter instead of GetTickCount for uptime

mscdex:


mscdex:
Err... I somehow overlooked the fact that someone had already submitted a pull request a few months ago that does something similar. Oh well.

piscisaureus:
I agree that this needs to be done. However @indutny also did a pull request (#203) and by the look of it his patch is closer to being acceptable than yours.

piscisaureus:
@mscdex Ok I will take your patch if you fix the issues.

src/win/util.c
```diff
@@ -385,8 +385,70 @@ uv_err_t uv_resident_set_memory(size_t* rss) {
...
+    if (lError != ERROR_SUCCESS && getSize > 0) {
+      if (wcsncmp(pPerfData->Signature, "PERF", 4) == 0) {
+        break;
+      }
+    } else if (lError != ERROR_SUCCESS && lError != ERROR_MORE_DATA) {
+      err = uv__new_sys_error(GetLastError());
+      goto done;
+    }
+
+    dataSize += 1024;
```
piscisaureus, line 423: There is no need to spin to figure out the data size you need. From msdn:

> If the buffer specified by lpData parameter is not large enough to hold the data, the function returns ERROR_MORE_DATA and stores the required buffer size in the variable pointed to by lpcbData. In this case, the contents of the lpData buffer are undefined.
> 
> If lpData is NULL, and lpcbData is non-NULL, the function returns ERROR_SUCCESS and stores the size of the data, in bytes, in the variable pointed to by lpcbData. This enables an application to determine the best way to allocate a buffer for the value's data.

src/win/util.c
```diff
@@ -385,8 +385,70 @@ uv_err_t uv_resident_set_memory(size_t* rss) {
...
+    if (pPerfData) {
+      free(pPerfData);
+    }
+    pPerfData = (PERF_DATA_BLOCK*)malloc(dataSize);
+    if (!pPerfData) {
+      uv_fatal_error(ERROR_OUTOFMEMORY, "malloc");
+    }
+    getSize = dataSize;
+
+    lError = RegQueryValueEx(HKEY_PERFORMANCE_DATA, "2", NULL, NULL,
```
piscisaureus: Be explicit about the character type you want (use RegQueryValueExA or RegQueryValueExW)

src/win/util.c
```diff
@@ -385,8 +385,70 @@ uv_err_t uv_resident_set_memory(size_t* rss) {
...
+    dataSize += 1024;
+  }
+
+  RegCloseKey(HKEY_PERFORMANCE_DATA);
+
+  pObj = (PERF_OBJECT_TYPE*)((BYTE*)pPerfData + pPerfData->HeaderLength);
+  pDef = (PERF_COUNTER_DEFINITION*)((BYTE*)pObj + pObj->HeaderLength);
+
+  for (i = 0; i < pObj->NumCounters; ++i) {
+    if (pDef->CounterNameTitleIndex == 674) {
```
piscisaureus: Is `674` a fixed number associated with uptime?

src/win/util.c
```diff
@@ -385,8 +385,70 @@ uv_err_t uv_resident_set_memory(size_t* rss) {
 
 
 uv_err_t uv_uptime(double* uptime) {
-  *uptime = (double)GetTickCount()/1000.0;
-  return uv_ok_;
+  uv_err_t err;
+  PERF_DATA_BLOCK *pPerfData = NULL;
```
piscisaureus: Please don't use hungarian notation.

src/win/util.c
```diff
@@ -385,8 +385,70 @@ uv_err_t uv_resident_set_memory(size_t* rss) {
...
+  PERF_OBJECT_TYPE *pObj;
+  PERF_COUNTER_DEFINITION *pDef;
+  PERF_COUNTER_DEFINITION *pDefUptime = NULL;
+  DWORD dataSize = 1024;
+  DWORD getSize;
+  LONG lError = ERROR_MORE_DATA;
+  unsigned __int64 upsec;
```
piscisaureus: use uint64_t

src/win/util.c
```diff
@@ -385,8 +385,70 @@ uv_err_t uv_resident_set_memory(size_t* rss) {
...
+  UINT i;
```
piscisaureus: use windows types only for values passed to / return from win32 api calls.

src/win/util.c
```diff
@@ -385,8 +385,70 @@ uv_err_t uv_resident_set_memory(size_t* rss) {
...
+      pDefUptime = pDef;
+      break;
+    }
+    pDef = (PERF_COUNTER_DEFINITION*)((BYTE*)pDef + pDef->ByteLength);
+  }
+
+  pData = (BYTE*)pObj + pObj->DefinitionLength;
+  pData += pDefUptime->CounterOffset;
+
+  memcpy(&upsec, pData, 8);
```
piscisaureus: You really need to memcpy here? What about casting pData to int64_t*?

src/win/util.c
mscdex, line 423: Right after that it also says:

> If hKey specifies HKEY_PERFORMANCE_DATA and the lpData buffer is not large enough to contain all of the returned data, RegQueryValueEx returns ERROR_MORE_DATA and the value returned through the lpcbData parameter is undefined. This is because the size of the performance data can change from one call to the next. In this case, you must increase the buffer size and call RegQueryValueEx again passing the updated buffer size in the lpcbData parameter. Repeat this until the function succeeds. You need to maintain a separate variable to keep track of the buffer size, because the value returned by lpcbData is unpredictable.

src/win/util.c
mscdex: I just used RegQueryValueEx because that's what uv_cpu_info was already using for reading cpu speed and name info (cares uses RegQueryValueEx also). I'm not sure if the unicode version is preferred over the ansi version or vice versa for this case.

src/win/util.c
mscdex: Yep

src/win/util.c
piscisaureus, line 423: Right - ok, so I agree with your approach. Although I wouldn't mind starting off with a somewhat bigger buffer.

src/win/util.c
piscisaureus: So - uv_cpu_info is doing it wrong. The problem is that RegQueryValueEx maps to either RegQueryValueExA or RegQueryValueExW, depending on the setting of the _UNICODE directive. If you want to use RegQueryValueEx you should also encapsulate string constants in _TEXT. But don't. Just use the wide-character variant (RegQueryValueExW).

src/win/util.c
piscisaureus: Just curious, is this documented somewhere? Or is it documented somewhere that title indices are "stable" and never change?

mscdex: There isn't any official documentation of it that I could find. However, libraries like sigar, code samples from the msdn blog, and just about every other Windows uptime snippets on the internet that use performance counters instead of GetTickCount, all use 2 for the "System" counter type and 674 for the "System Up Time" counter id. Also, there was a post on microsoft.public.win32.programmer.kernel back in 1998 that mentions using these two constant values for retrieving the system up time :)

FWIW, the built-in counter types, ids, and names can be seen in: HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Perflib\<LangID>\Counter

src/win/util.c
mscdex, line 423: What size do you suggest? 4kb?

piscisaureus, line 423: Yeah, that would be nice.

piscisaureus:
Thanks, Brian. Landed in 9d3c00046f1e8d3800bcb161dd8e929399776d55 and 442aa1f4696d24a605790b8ad4317696758ee8de

piscisaureus:
@mscdex Why did we do this again? Was it only because GetTickCount could wrap, or where there other reasons?

mscdex:
Yeah, GetTickCount _does_ wrap after 49.7 days. Also one other reason for this specific solution was that GetTickCount64 is Vista+ only.

net147:
A downside of this is that the performance counter is affected by date/time changes on Windows XP (but not Windows Vista and later). See https://github.com/joyent/node/issues/5304.

PR #327, opened 2012-02-24
win: implement cpu times

mscdex:
On a related note (node), I wonder if it's better to use `Number::New` instead of `Integer::New` inside `GetCPUInfo()` in node_os.cc. That additional change will allow us to have more accurate (read: non-negative) values when cpu times get very very high, as double has higher precision than a 32-bit int.


mscdex:
Any comments/suggestions/questions/corrections? It'd be great to see support for this (and #326 or something like it) implemented in time for node 0.8 if at all possible (I do not know what the estimates are for a node 0.8 release these days).

bnoordhuis:
@piscisaureus: Review this PR. It LGTM at first glance.

src/win/util.c
```diff
@@ -45,6 +45,17 @@
  */
 #define MAX_TITLE_LENGTH 8192
 
+#define SystemProcessorPerformanceInformation  8
```
piscisaureus: Can you put SDK/DDK definitions in winapi.h?

src/win/util.c
```diff
@@ -45,6 +45,17 @@
...
+typedef struct {
+    LARGE_INTEGER IdleTime;
+    LARGE_INTEGER KernelTime;
+    LARGE_INTEGER UserTime;
+    LARGE_INTEGER DpcTime;
+    LARGE_INTEGER InterruptTime;
+    ULONG InterruptCount;
+} SYSTEM_PROCESSOR_PERFORMANCE_INFORMATION;
+typedef LONG (WINAPI *PROCNTQSI)(UINT,PVOID,ULONG,PULONG);
+PROCNTQSI NtQuerySystemInformation;
```
piscisaureus: Do it like the rest of libuv: https://github.com/joyent/libuv/blob/d07f2466d0a10ab02db588374dd97c2edd54c7fc/src/win/winapi.c#L87. 

You'll probably have to make sure that uv_init() is called, you could add a function uv__once_init for that. See https://github.com/joyent/libuv/blob/master/src/win/core.c#L98.

src/win/util.c
```diff
@@ -400,19 +411,50 @@ uv_err_t uv_cpu_info(uv_cpu_info_t** cpu_infos, int* count) {
...
   *count = 0;
 
-  for (i = 0; i < system_info.dwNumberOfProcessors; i++) {    
+  szSppi = sizeof(SYSTEM_PROCESSOR_PERFORMANCE_INFORMATION) * ncpus;
+  sppi = (SYSTEM_PROCESSOR_PERFORMANCE_INFORMATION*)malloc(szSppi);
+  if (!sppi) {
+    uv_fatal_error(ERROR_OUTOFMEMORY, "malloc");
+  }
+
+  SecureZeroMemory(sppi, szSppi);
```
piscisaureus: Why is memset not sufficient?

piscisaureus:
Good job! I'd like move the NtQuerySystemInformation acquisition to winapi.c and winapi.h where it belongs. After that this patch can land.

src/win/util.c
mscdex: Add a uv__once_init to core.c/uv.h and call that from util.c?

piscisaureus: yep

include/uv.h
```diff
@@ -212,6 +212,11 @@
 
 
 /*
+ * Ensures that uv is initialized.
+ */
+UV_EXTERN void uv__once_init(void);
```
piscisaureus: This is an internal api. It should go into src/win/internal.h and not be marked UV_EXPORT.

mscdex:
No luck with the Unicode version of the registry API, kept getting ERROR_FILE_NOT_FOUND. Switching to the ANSI version makes everything work as expected.

piscisaureus:
You probably have to use widechar strings. Compare

HANDLE h = CreateFileA("myfile", ...)
HANDLE h = CreateFileW(L"myfile", ...)

mscdex:
I tried that as well.

piscisaureus:
@mscdex I want to land this. Are you going to fix this up or should I do it?

mscdex:
Well, I got it to work, although it requires making uv_cpu_info_s's `model` a wchar_t*/uint16_t* instead of char*, which means we'd have to then change *nix's implementation to use 16-bit strings. Unless we want to convert the wchar_t* from the unicode registry call to char* (then we might as well use the ANSI version registry call in the first place)?

I've never seen unicode processor name strings.

PR #329, opened 2012-02-28
Implement detached child processes for unix and windows

AvianFlu:
This is a cleaned-up version of #322 with some important bug fixes for unix.  Now, on both unix and Windows, processes will detach properly, will not maintain connections to any prior `stdio` handles, and run as the head of a new process group.


klepikov:
PTAL at https://github.com/joyent/node/pull/3344, I would appreciate your comments, especially around killpg().

klepikov:
Is this pull request going to be superseded by https://github.com/AvianFlu/libuv/compare/detached2?

That one no longer has a force-close of std streams, or a second fork() -- both of which are IMHO good differences.

It still does not allow to call killpg() on UNIX.

AvianFlu:
Yes, this patch is defunct.  I'll close it.

PR #332, opened 2012-03-02
Windows: we need to zero the loop structure when we initialize it

jedisct1:
or else the list of ares handles is uninitialized and bad things can happen when we try to perform async lookups.

### DIFF
--- a/src/win/core.c
+++ b/src/win/core.c
@@ -59,6 +59,7 @@ static void uv_init(void) {
 
 
 static void uv_loop_init(uv_loop_t* loop) {
+  loop->uv_ares_handles_ = NULL;
   /* Create an I/O completion port */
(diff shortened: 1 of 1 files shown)


src/win/core.c
```diff
@@ -59,6 +59,7 @@ static void uv_init(void) {
 
 
 static void uv_loop_init(uv_loop_t* loop) {
+  memset(loop, 0, sizeof(*loop));
```
bnoordhuis: I'm functioning as a proxy for @piscisaureus here (who apparently has forgotten how the comment function works) but he mentioned that he would prefer it if the fields are zeroed explicitly.

jedisct1:
Sure! The pull request has been updated.

bnoordhuis:
Thanks Frank, merged in 1c8cf61.

PR #336, opened 2012-03-06
cygwin: we need to include uv-common.h for uv__set_sys_error()

jedisct1:


bnoordhuis:
Thanks Frank, merged in 4f1782a.

PR #345, opened 2012-03-13
Add uv_loop_fork()

jedisct1:
Add uv_loop_fork(), so that we can safely fork() after event loops have been initialized.

### DIFF
--- a/src/unix/core.c
+++ b/src/unix/core.c
@@ -215,6 +215,11 @@ int uv_run_once(uv_loop_t* loop) {
 }
 
 
+void uv_loop_fork(uv_loop_t* loop) {
+  ev_loop_fork(loop->ev);
+}
+
+
 void uv__handle_init(uv_loop_t* loop, uv_handle_t* handle,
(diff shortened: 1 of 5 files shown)


src/unix/core.c
```diff
@@ -215,6 +215,11 @@ int uv_run_once(uv_loop_t* loop) {
 }
 
 
+void uv_loop_fork(uv_loop_t* loop) {
+  ev_loop_fork(loop->ev);
```
bnoordhuis, line 219: Calling `ev_loop_fork()` is not enough. libeio for instance has a couple of mutexes that would need to be reinitialized (never mind the work queue and other state).

Fork safety has been discussed in the past. The consensus was that it's not worth the trouble.

piscisaureus, line 219: Plus the libeio threads would have to be recreated

bnoordhuis:
Sorry Frank, we can't take this patch. It won't work as advertised, there's simply too much that can go wrong if you fork but don't exec.

PR #352, opened 2012-03-20
map ENETUNREACH to UV_ENETUNREACH

AvianFlu:
UV_ENETUNREACH already exists, but was not mapped properly on unix.  This is the case on both `master` and in `v0.6`.


bnoordhuis:
Thanks Charlie, merged in 8e59042.

PR #357, opened 2012-03-25
Don't link with Carbon framework when using gyp (on Darwin)

saghul:
AFAIK it's not needed, the plain makefiles don't do it and test suite runs just fine without linking with it.


bnoordhuis:
Thanks Saúl, merged in e729a82.

PR #359, opened 2012-03-26
Ipv46agnostic

DylanLukes:
This set of commits modifies libuv's API to be IPv4/6 agnostic. This is accomplished by combining the 4/6 variants of functions, and replacing each `struct sockaddr_in(6)` parameter with a pair of `struct sockaddr *` and `socklen_t` parameters. 

This allows users to directly use the results from a `uv_getaddrinfo` call without regard for whether they are connecting to an IPv4 or IPv6 address.

In addition, the unit tests have all been edited to reflect these changes.

Lastly, internally there was a small bug wherein a UDP peer's address was written into a `sockaddr_storage` (correctly), but was then assumed to be and cast to `sockaddr_in`. Previously this would silently work for IPv4 and fail for IPv6. I was clued into this by an `EINVAL` from a POSIX sockets syscall, which resulted from the addrlen (128 for `sockaddr_storage`) not matching the actual structure (`sockaddr_in`).


bnoordhuis:
Closing, no longer applies. We may be moving to this model in the future.

PR #361, opened 2012-03-27
make clean made cleaner

dvv:


bnoordhuis:
Thanks Vladimir, merged in 2b3ba91.

PR #362, opened 2012-03-27
fix "for loop initial declaration used outside C99 mode" on openbsd

qbit:
This fix makes libuv compile on openbsd-current 

Note: this is not related to issue # 3016 ( now closed ) I opened for joyent/node


src/unix/openbsd.c
```diff
@@ -245,7 +245,8 @@ uv_err_t uv_cpu_info(uv_cpu_info_t** cpu_infos, int* count) {
   size = sizeof(info);
   which[0] = CTL_KERN;
   which[1] = KERN_CPTIME2;
-  for (int i = 0; i < numcpus; i++) {
+  int i;
```
bnoordhuis, line 248: Shouldn't this move to the function prolog?

qbit, line 248: yep! I will re-request.

bnoordhuis, line 248: @qbit: You can redo your commit (or commit on top and squash), then force-push it and post a "updated!" comment.

qbit:
closing for re-pull

PR #363, opened 2012-03-27
fix "for loop initial declaration used outside C99 mode" on openbsd

qbit:
Ewps - now with 100% more prolog'd var int!


bnoordhuis:
Looks good, Aaron. Can you sign [the CLA](http://nodejs.org/cla.html)?

qbit:
done.

bnoordhuis:
Thanks Aaron, merged in 40e6373.

PR #365, opened 2012-03-28
update conditional for OS X 10.6+ compatibility

avalanche123:
fixing conditional, OS X 10.6 and above defines TARGET_OS_IPHONE as 0


bnoordhuis:
Thanks Bulat, merged in 31ff986.

PR #370, opened 2012-04-03
add ability to lookup different handle sizes before allocation

avalanche123:
I'm working on Ruby FFI bindings for libuv and I find myself having to copy majority of libuv struct definitions in ruby just to get correct sizes for memory allocation. This change makes it possible to dynamically lookup memory necessary for libuv's handles, removing the need to duplicate structs.

Feedback is much appreciated!


txdv:
https://github.com/txdv/libuv/commit/ecea9882713d56b3cbfe19c489b2c3ef526115e3

this is my approach and I am for this, I have some C# bindings and they sizes of the objects differ on windows and linux, this means that this is the only sane approach to create platform independent bindings.

bnoordhuis:
LGTM in general but I noticed that the request type switch is not exhaustive.

@piscisaureus: What do you think?

piscisaureus:
@bnoordhuis @avalanche123 Fine with me. But yeah, include all the handle types in the switch.

avalanche123:
Ok, I'll work on adding more handle types

avalanche123:
Updated PR, this cannot be merged until private types are removed from UV_HANDLE_TYPE_MAP/UV_REQ_TYPE_MAP macros in uv.h

txdv:
was this merged?

piscisaureus:
@txdv https://github.com/joyent/libuv/commit/f09f7bc6a8b64585cd74d1fca712c6d0cab2f737

txdv:
you forgot ares_options, uv_rwlock_t, uv_once_t, uv_thread_t

avalanche123:
I'll submit another one once I get to those functions, or you could submit those

PR #373, opened 2012-04-08
Exposing the turn of the event loop

seebees:
I am proposing putting a counter on the event loop that will keep track of each "turn" and exposing this value so others can use it in magical ways.  Primarily I want to expose this value in node.  I will issue a pull and talk more there.

I also found that I needed to have a way to arbitrarily increment this counter to handle events like `process.nextTick`.  I would prefer if that happend on the node side, but I'm not sure how that works out cleanly... and this was simple.  Comments welcome.

I've written one test, which I'm not sure is sufficient, but I had trouble getting it to run automatically.

I have tested on OSX, and I will try and compile and test on windows tomorrow.

Oh, and suggestions on function names are welcome as well.

This is primarily implemented for node:[3074](https://github.com/joyent/node/pull/3074)

### DIFF
--- a/include/uv-private/ev.h
+++ b/include/uv-private/ev.h
@@ -640,8 +640,10 @@ void ev_unref (EV_P);
  */
 void ev_once (EV_P_ int fd, int events, ev_tstamp timeout, void (*cb)(int revents, void *arg), void *arg);
 
-# if EV_FEATURE_API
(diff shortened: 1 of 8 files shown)


bnoordhuis:
Closing, sorry. The patch no longer applies.

PR #378, opened 2012-04-14
unix: map `EROFS` to `UV_EROFS`

mmalecki:
No idea how to test it or implement on Windows.

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -123,7 +123,8 @@ typedef intptr_t ssize_t;
   XX( 52, EXDEV, "cross-device link not permitted") \
   XX( 53, ENOTEMPTY, "directory not empty") \
   XX( 54, ENOSPC, "no space left on device") \
-  XX( 55, EIO, "i/o error")
+  XX( 55, EIO, "i/o error") \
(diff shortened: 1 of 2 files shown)


bnoordhuis:
@piscisaureus: Is `ERROR_FILE_READ_ONLY` the corresponding Windows error code?

piscisaureus:
@bnoordhuis No - ERROR_FILE_READ_ONLY means that the file has the read-only bit set. It should report whatever error you get on unix when the user has no write permission on that file.

Windows has no concept of read-only mounts. Some mounts can be write protected in practice (like, the driver doesn't support it, or you're trying to write to a DVD). That would result in EACCESS on windows.

I think you should map ERROR_WRITE_PROTECT to EROFS though.

bnoordhuis:
Thanks Maciej, merged in 58733d1.

piscisaureus:
Hmm, so no ERROR_WRITE_PROTECT mapping uh?

bnoordhuis:
You haven't been paying attention, have you? :-) See here: d5e7786

Issue #383, opened 2012-04-19
windows: libuv does not compile with mingw-w64


bnoordhuis:
```
$ apt-cache show mingw-w64 | grep Version
Version: 2.0.1-1
$ make MSVC=1 AR=i686-w64-mingw32-ar CC=i686-w64-mingw32-gcc
i686-w64-mingw32-gcc -Iinclude -Iinclude/uv-private -g --std=gnu89 -D_WIN32_WINNT=0x0600 -Isrc/ares/config_win32 -o src/win/async.o -c src/win/async.c
In file included from src/win/internal.h:29:0,
                 from src/win/async.c:25:
src/win/winapi.h:4339:18: error: redefinition of ‘struct _OVERLAPPED_ENTRY’
/usr/lib/gcc/i686-w64-mingw32/4.6/../../../../i686-w64-mingw32/include/winbase.h:3259:16: note: originally defined here
src/win/winapi.h:4344:5: error: conflicting types for ‘OVERLAPPED_ENTRY’
/usr/lib/gcc/i686-w64-mingw32/4.6/../../../../i686-w64-mingw32/include/winbase.h:3264:3: note: previous declaration of ‘OVERLAPPED_ENTRY’ was here
src/win/winapi.h:4344:24: error: conflicting types for ‘LPOVERLAPPED_ENTRY’
/usr/lib/gcc/i686-w64-mingw32/4.6/../../../../i686-w64-mingw32/include/winbase.h:3264:22: note: previous declaration of ‘LPOVERLAPPED_ENTRY’ was here
In file included from src/win/internal.h:30:0,
                 from src/win/async.c:25:
src/win/winsock.h:117:16: error: redefinition of ‘struct _IP_ADAPTER_UNICAST_ADDRESS_XP’
/usr/lib/gcc/i686-w64-mingw32/4.6/../../../../i686-w64-mingw32/include/iptypes.h:70:18: note: originally defined here
src/win/winsock.h:134:3: error: conflicting types for ‘IP_ADAPTER_UNICAST_ADDRESS_XP’
/usr/lib/gcc/i686-w64-mingw32/4.6/../../../../i686-w64-mingw32/include/iptypes.h:86:5: note: previous declaration of ‘IP_ADAPTER_UNICAST_ADDRESS_XP’ was here
src/win/winsock.h:134:34: error: conflicting types for ‘PIP_ADAPTER_UNICAST_ADDRESS_XP’
[...]

PR #384, opened 2012-04-20
Actually detect mingw-w64 and not 64bits mode

luislavena:
`__MINGW64__` is only defined when using mingw-w64 in 64bits mode, not when
using the 32bits version of the compiler.

Instead, to detect usage of mingw-w64 instead of mingw version of GCC we
should look at `__MINGW64_VERSION_MAJOR` and `__MINGW64_VERSION_MINOR`. If one
of these is defined means we are running on mingw-w64 compiler and headers
and not mingw.org build.

Read more here: http://bit.ly/detect-mingw-w64-vs-mingw

This patch solves #383


bnoordhuis:
@piscisaureus: Review? LGTM.

PR #387, opened 2012-04-25
Make 'gmake test' work on OpenBSD

qbit:
gmake test was failing with kvm_\* errors - this adds -lkvm to LINKFLAGS, which resolves this issue.


bnoordhuis:
Thanks Aaron, merged in 109dcb2.

PR #389, opened 2012-04-25
Implement uid/gid argument to uv_spawn

isaacs:
This is necessary to replace the uid/gid arguments to child_process.spawn in Node, which apparently started silently not working as of 0.6.

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -1111,14 +1111,14 @@ typedef struct uv_process_options_s {
    * setgid to before executing.
    * Unix only.
    */
-  int gid;
+  gid_t gid;
 
   /*
    * If not -1, this represents the user ID that the subprocess should
    * setuid to before executing.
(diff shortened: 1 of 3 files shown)


tjfontaine:
I believe uid_t is generally uint32 on linux, it may not be wise to use int to describe uid/gid

piscisaureus:
gid_t is a type, right?

piscisaureus:
als, this could be backported to 0.6 as `uv_spawn2`, similar to what we did for 64 bit file offsets.

tjfontaine:
uid_t and gid_t are types yes

isaacs:
Updated to use uid_t and gid_t.  It's still using -1 as the sentinel value, so if you have a user with a uid of 0xffffffff or whatever, you won't be able to setuid to them, which seems rare and strange enough to not care about.

PR #397, opened 2012-04-29
Fix #382

### LINKED ISSUE #382: windows: _chsize_s not available on MinGW
After this change: https://github.com/joyent/libuv/commit/d5acfd0c0545acbef5ced2fbcbc6a91c8fc7c025#L3R580 libuv won't compile on MinGW because apparently _chsize_s is not available. Apparently _chsize is limited to 32bits offsets so we can't use that in case the _s version is not available.

Keno:
This should fix #382

### DIFF
--- a/src/win/fs.c
+++ b/src/win/fs.c
@@ -591,9 +591,11 @@ void fs__ftruncate(uv_fs_t* req, uv_file file, int64_t offset) {
                                     sizeof eof_info,
                                     FileEndOfFileInformation);
   
-  result = (nt_status==STATUS_SUCCESS)-1;
-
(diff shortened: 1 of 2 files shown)


piscisaureus:
Cool. Can you sign the node.js CLA?

http://nodejs.org/cla.html

Keno:
Done

piscisaureus:
Hmm. It doesn't compile, but I'll fix it myself

Keno:
whoops you're right I had a stale commit lying around that I didn't push onto the proper branch

piscisaureus:
Ok, landed with some modifications.

Issue #356, opened 2012-03-23
unix: handle (or at least report) EIO


bnoordhuis:
```
$ strace -e read out/Release/node -pe 'require("fs").readFileSync("/proc/self/mem")'
<snip>
read(5, 0x1410c40, 4048)                = -1 EIO (Input/output error)

fs.js:316
  var r = binding.read(fd, buffer, offset, length, position);
                  ^
Error: UNKNOWN, unknown error
    at Object.readSync (fs.js:316:19)
    at Object.readFileSync (fs.js:161:19)
    at Object.<anonymous> (eval at <anonymous> (eval:1:82))
    at Object.<anonymous> (eval:1:70)
    at Module._compile (module.js:444:26)
    at startup (node.js:77:27)
    at node.js:595:3
```

bnoordhuis:
@piscisaureus: What's the Windows equivalent? `man errno` has this to say about EIO:

```
EIO             Input/output error (POSIX.1)
```

Very helpful, very descriptive. :-/

piscisaureus:
@bnoordhuis There are many. to get the idea, look here: http://sources.redhat.com/cgi-bin/cvsweb.cgi/src/winsup/cygwin/errno.cc?rev=1.87&content-type=text/x-cvsweb-markup&cvsroot=src

indutny:
Proposed patch for unix version:

``` patch
diff --git a/src/unix/error.c b/src/unix/error.c
index aaeaf76..c2651a3 100644
--- a/src/unix/error.c
+++ b/src/unix/error.c
@@ -59,6 +59,7 @@ void uv_fatal_error(const int errorno, const char* syscall) {
 uv_err_code uv_translate_sys_error(int sys_errno) {
   switch (sys_errno) {
     case 0: return UV_OK;
+    case EIO: return UV_EIO;
     case EPERM: return UV_EPERM;
     case ENOSYS: return UV_ENOSYS;
     case ENOTSOCK: return UV_ENOTSOCK;
```

indutny:
Should this be closed now?

bnoordhuis:
This is fixed in 0efa3cf, right?

indutny:
indeed, yes

PR #399, opened 2012-05-01
err: handle EIO errors on win/unix

indutny:
see #356


bnoordhuis:
Merged in 0efa3cf, thanks.

PR #401, opened 2012-05-02
Poll

piscisaureus:
@igorzi Can you review 87b7e29a ?

@bnoordhuis Can you review e3875548 and 7e8ab521    ?

Both: please take a look at the test (c908bbd8). Or pretend you did and say LGTM.


txdv:
I just saw that it is in the commit tree.
Aweeesssooomeee! Thanks piscciaureus! That was exactly what I needed!

Issue #381, opened 2012-04-18
windows: cross-compiling tests with mingw32 fails


bnoordhuis:
Not sure if it matters but I'll post it anyway.

```
$ make test MSVC=1 AR=i586-mingw32msvc-ar CC=i586-mingw32msvc-cc
i586-mingw32msvc-cc -Iinclude -Iinclude/uv-private -Iinclude -Iinclude/uv-private -g --std=gnu89 -D_WIN32_WINNT=0x0600 -Isrc/ares/config_win32 -D_GNU_SOURCE  -o test/run-tests test/run-tests.c \
                test/runner.c test/runner-win.c test/blackhole-server.c test/echo-server.c test/test-*.c uv.a -lws2_32 -lpsapi -liphlpapi -lm
uv.a(fs.o): In function `fs__ftruncate':
/home/bnoordhuis/src/libuv-win/src/win/fs.c:580: undefined reference to `__chsize_s'
uv.a(getaddrinfo.o): In function `getaddrinfo_thread_proc':
/home/bnoordhuis/src/libuv-win/src/win/getaddrinfo.c:88: undefined reference to `__imp__GetAddrInfoW@16'
uv.a(getaddrinfo.o): In function `uv_process_getaddrinfo_req':
/home/bnoordhuis/src/libuv-win/src/win/getaddrinfo.c:215: undefined reference to `__imp__FreeAddrInfoW@4'
collect2: ld returned 1 exit status
make: *** [test/run-tests.exe] Error 1
```

Compiling master with MSVC works but I have 16 failing tests (almost all fs related but counters_init and pipe_connect_to_file are in there too).

luislavena:
@bnoordhuis that was master?

The undefined references I believe was merged from pull #333 in c344277450a2d2c35572183b5e25480a8966dc56 for `GetAdaptersAddresses` and `GetProcessMemoryInfo`

luislavena:
Hello @bnoordhuis and @piscisaureus I think this can be closed as the issue mentioned in #382 has been addressed.

bnoordhuis:
The issue with _chsize_s has been fixed but I'm still seeing the same link-time errors with mingw32 on ubuntu 10.04. It's not a terribly high-priority issue but I would like to get it fixed eventually, if only so I don't have to come out of my comfort zone (the UNIX shell).

luislavena:
Hmm, will double check tomorrow.
Can tell me which version of gcc it reports is using?

[...]

PR #403, opened 2012-05-04
windows: fix mingw32 cross-compilation

bnoordhuis:
@piscisaureus LGTY?


piscisaureus:
ntdef.h is not available for normal msvc users. Is there no other way?

luislavena:
@piscisaureus @bnoordhuis I think the issue is coming from `winapi.h` NTDEF checks and that it clashes with `uv-win.h` mentioned in #381:

https://github.com/joyent/libuv/blob/master/src/win/winapi.h#L28-34
https://github.com/joyent/libuv/blob/master/include/uv-private/uv-win.h#L131-134

Both are defining it, but none is defining `_NTDEF_` to avoid it being defined again.

Isn't the one in `winapi.h` redundant since `uv-win.h` will be included before?

bnoordhuis:
Sounds plausible. The core issue is that mingw is complaining that NTSTATUS and PNTSTATUS are being redefined (see below). A quick grep suggests that they're defined in ntdef.h, ntsecapi.h and subauth.h. Each file has #ifdef guards in place to prevent redefinition.

```
$ make test MSVC=1 AR=i586-mingw32msvc-ar CC=i586-mingw32msvc-gcc
i586-mingw32msvc-gcc -Iinclude -Iinclude/uv-private -g --std=gnu89 -D_WIN32_WINNT=0x0600 -Isrc/ares/config_win32 -o src/win/async.o -c src/win/async.c
In file included from src/win/internal.h:29,
                 from src/win/async.c:25:
src/win/winapi.h:32: error: redefinition of typedef ‘NTSTATUS’
include/uv-private/uv-win.h:132: note: previous declaration of ‘NTSTATUS’ was here
src/win/winapi.h:33: error: redefinition of typedef ‘PNTSTATUS’
include/uv-private/uv-win.h:133: note: previous declaration of ‘PNTSTATUS’ was here
make: *** [src/win/async.o] Error 1
```

luislavena:
> A quick grep suggests that they're defined in ntdef.h, ntsecapi.h and subauth.h.

Problem is that those headers might not be available on Windows native under some circumstances, mostly will require you install the Platform SDK with is an extra download.

I think that is why it was defined in `uv-win-h` first.

As I commented before, removing the definition from `winapi.h` seems to solve _that_ issue:

```
ubuntu@ubuntu64:~/projects/oss/joyent/libuv$ make test MSVC=1 PREFIX=i586-mingw32msvc-
i586-mingw32msvc-gcc -Iinclude -Iinclude/uv-private -g --std=gnu89 -D_WIN32_WINNT=0x0600 -Isrc/ares/config_win32 -o src/win/async.o -c src/win/async.c
i586-mingw32msvc-gcc -Iinclude -Iinclude/uv-private -g --std=gnu89 -D_WIN32_WINNT=0x0600 -Isrc/ares/config_win32 -o src/win/cares.o -c src/win/cares.c
i586-mingw32msvc-gcc -Iinclude -Iinclude/uv-private -g --std=gnu89 -D_WIN32_WINNT=0x0600 -Isrc/ares/config_win32 -o src/win/core.o -c src/win/core.c
...
```

``` diff
diff --git a/src/win/winapi.h b/src/win/winapi.h
index 4b1a249..ebc53b4 100644
--- a/src/win/winapi.h
+++ b/src/win/winapi.h
@@ -25,14 +25,6 @@
 #include <windows.h>


-/*
- * Ntdll headers
- */
-#ifndef _NTDEF_
-  typedef LONG NTSTATUS;
-  typedef NTSTATUS *PNTSTATUS;
-#endif
-
 #ifndef STATUS_SEVERITY_SUCCESS
 # define STATUS_SEVERITY_SUCCESS 0x0
 #endif
```

bnoordhuis:
@piscisaureus See @luislavena's comment, why are those typedefs in winapi.h?

luislavena:
@piscisaureus ping?

PR #405, opened 2012-05-05
make uv_handle_size return the size of UV_POLL.

txdv:


piscisaureus:
Thanks. landed in 985b961

PR #408, opened 2012-05-07
fix build on mingw32.

mattn:
on mingw32, NTSTATUS is defined in ntdef.h


bnoordhuis:
Related: #403

luislavena:
@mattn please see my comments on #403, `NTSTATUS` is already defined in `uv-win.h`

PR #413, opened 2012-05-09
unix: stub out new uv_spawn API

indutny:
Just a WIP branch, please put your comments on API or on any details here!

/cc @bnoordhuis @piscisaureus @isaacs


src/unix/process.c
```diff
@@ -313,35 +348,31 @@ int uv_spawn(uv_loop_t* loop, uv_process_t* process,
...
-    assert(options.stdin_stream);
-    assert(stdin_pipe[0] >= 0);
+    assert(options.stdio_count > 0);
     close(stdin_pipe[0]);
     uv__nonblock(stdin_pipe[1], 1);
-    flags = UV_STREAM_WRITABLE |
-            (options.stdin_stream->ipc ? UV_STREAM_READABLE : 0);
-    uv__stream_open((uv_stream_t*)options.stdin_stream, stdin_pipe[1],
+    flags = uv__process_stdio_flags(&options.stdio[0], 1);
+    uv__stream_open((uv_stream_t*)options.stdio[0].data.pipe, stdin_pipe[1],
```
bnoordhuis: Check the return value here.

src/unix/process.c
```diff
@@ -313,35 +348,31 @@ int uv_spawn(uv_loop_t* loop, uv_process_t* process,
...
-    assert(options.stdout_stream);
+    assert(options.stdio_count > 1);
     assert(stdout_pipe[1] >= 0);
     close(stdout_pipe[1]);
     uv__nonblock(stdout_pipe[0], 1);
-    flags = UV_STREAM_READABLE |
-            (options.stdout_stream->ipc ? UV_STREAM_WRITABLE : 0);
-    uv__stream_open((uv_stream_t*)options.stdout_stream, stdout_pipe[0],
+    flags = uv__process_stdio_flags(&options.stdio[1], 0);
+    uv__stream_open((uv_stream_t*)options.stdio[1].data.pipe, stdout_pipe[0],
```
bnoordhuis: Ditto.

src/unix/process.c
```diff
@@ -313,35 +348,31 @@ int uv_spawn(uv_loop_t* loop, uv_process_t* process,
...
-    assert(options.stderr_stream);
+    assert(options.stdio_count > 2);
     assert(stderr_pipe[1] >= 0);
     close(stderr_pipe[1]);
     uv__nonblock(stderr_pipe[0], 1);
-    flags = UV_STREAM_READABLE |
-            (options.stderr_stream->ipc ? UV_STREAM_WRITABLE : 0);
-    uv__stream_open((uv_stream_t*)options.stderr_stream, stderr_pipe[0],
+    flags = uv__process_stdio_flags(&options.stdio[2], 0);
+    uv__stream_open((uv_stream_t*)options.stdio[2].data.pipe, stderr_pipe[0],
```
bnoordhuis: Ditto.

indutny:
Updated. Going to work on support of more than 3 stdio pipes.

include/uv.h
```diff
@@ -1156,6 +1156,24 @@ UV_EXTERN int uv_getaddrinfo(uv_loop_t*, uv_getaddrinfo_t* handle,
...
+  UV_IPC_PIPE    = 0x04,
+
+  UV_RAW_FD      = 0x08
+} uv_stdio_flags;
+
+typedef struct uv_stdio_container_s {
+  uv_stdio_flags flags;
+
+  union {
+    uv_pipe_t* pipe;
```
Keno: Hmm, can't we make this uv_stream_t directly?

indutny: I'm absolutely ok with that. /cc @bnoordhuis @piscisaureus

bnoordhuis: I suppose so. It doesn't make sense for stream types like `uv_tty_t`, though.

indutny:
Added support for unlimited count of stdio, please review ;)

include/uv.h
Keno: Could we have uv_tty_t pass the associated stream through to the child (I.e. the equivalent of what setting the stream to 0 does right now)?

piscisaureus: Passing in an uv_tty_t makes the child inherit that stdio from the parent. So yeah, that would make sense.

src/unix/process.c
```diff
@@ -162,16 +221,21 @@ int uv_spawn(uv_loop_t* loop, uv_process_t* process,
    * by the child process.
    */
   char** save_our_env = environ;
-  int stdin_pipe[2] = { -1, -1 };
-  int stdout_pipe[2] = { -1, -1 };
-  int stderr_pipe[2] = { -1, -1 };
+
+  int* pipes = (int*) malloc(2 * options.stdio_count * sizeof(int));
```
bnoordhuis: Superfluous cast.

src/unix/process.c
```diff
@@ -184,19 +248,18 @@ int uv_spawn(uv_loop_t* loop, uv_process_t* process,
...
+    pipes[i * 2] = -1;
+    pipes[i * 2 + 1] = -1;
   }
 
-  if (options.stderr_stream &&
-      uv__process_init_pipe(options.stderr_stream, stderr_pipe, 0)) {
-    goto error;
+  /* Create socketpairs/pipes, or use raw fd */
+  for (i = 0; i < options.stdio_count; i++) {
+    if (uv__process_init_stdio(&options.stdio[i], (int*) pipes + i * 2,
```
bnoordhuis: Superfluous cast.

src/unix/process.c
```diff
@@ -236,31 +299,29 @@ int uv_spawn(uv_loop_t* loop, uv_process_t* process,
...
-    } else {
-      /* Reset flags that might be set by Node */
-      uv__cloexec(STDERR_FILENO, 0);
-      uv__nonblock(STDERR_FILENO, 0);
+    /* Child */
+
+    /* Dup fds */
+    for (i = 0; i < options.stdio_count; i++) {
+      int close_fd = i == 0 ? pipes[i * 2 + 1] : pipes[i * 2];
+      int use_fd = i == 0 ? pipes[i * 2] : pipes[i * 2 + 1];
```
bnoordhuis: This needs a comment explaining why it's swapped when i == 0.

src/unix/process.c
```diff
@@ -236,31 +298,29 @@ int uv_spawn(uv_loop_t* loop, uv_process_t* process,
...
+      int target_fd;
+
+      switch (i) {
+        case 0: target_fd = STDIN_FILENO; break;
+        case 1: target_fd = STDOUT_FILENO; break;
+        case 2: target_fd = STDERR_FILENO; break;
+        default: target_fd = i; break;
+      }
+
```
bnoordhuis: Superfluous switch.

src/unix/process.c
bnoordhuis: Also, I'd turn use_fd into a boolean. It's name and usage suggests it's a bool but it's not.

src/unix/process.c
```diff
@@ -312,49 +372,32 @@ int uv_spawn(uv_loop_t* loop, uv_process_t* process,
...
-    close(stdin_pipe[0]);
-    uv__nonblock(stdin_pipe[1], 1);
-    flags = UV_STREAM_WRITABLE |
-            (options.stdin_stream->ipc ? UV_STREAM_READABLE : 0);
-    uv__stream_open((uv_stream_t*)options.stdin_stream, stdin_pipe[1],
-        flags);
-  }
+  for (i = 0; i < options.stdio_count; i++) {
+    if (uv__process_open_stream(&options.stdio[i], pipes + i * 2, i == 0)) {
+      int j;
```
bnoordhuis, line 375: Superfluous cast.

src/unix/process.c
```diff
@@ -312,49 +372,32 @@ int uv_spawn(uv_loop_t* loop, uv_process_t* process,
...
-  close(stderr_pipe[0]);
-  close(stderr_pipe[1]);
+
+  free(pipes);
+
+  for (i = 0; i < options.stdio_count; i++) {
+    close(pipes[i * 2]);
+    close(pipes[i * 2 + 1]);
+  }
+
```
bnoordhuis, line 396: Use after free.

src/unix/process.c
indutny: No, it isn't boolean.

test/test-ipc.c
```diff
@@ -218,10 +218,17 @@ void spawn_helper(uv_pipe_t* channel,
   options.file = exepath;
   options.args = args;
   options.exit_cb = exit_cb;
-  options.stdin_stream = channel;
+
+  options.stdio = malloc(sizeof(uv_stdio_container_t) * 1);
```
bnoordhuis: Use an on-stack variable, easier to reason about.

src/unix/process.c
bnoordhuis: Right, missed the call to dup(). Still needs a comment though.

indutny:
All issues should be resolved now.

bnoordhuis:
The tests don't compile:

```
 CC(target) /home/bnoordhuis/src/nodejs/libuv/out/Debug/obj.target/run-benchmarks/test/benchmark-thread.o
../test/benchmark-spawn.c: In function ‘spawn’:
../test/benchmark-spawn.c:117: error: ‘uv_process_options_t’ has no member named ‘stdout_stream’
```

indutny:
Should be fixed now.

bnoordhuis:
Thanks, Fedor. LGTM, @piscisaureus?

include/uv.h
```diff
@@ -1156,6 +1156,24 @@ UV_EXTERN int uv_getaddrinfo(uv_loop_t*, uv_getaddrinfo_t* handle,
 UV_EXTERN void uv_freeaddrinfo(struct addrinfo* ai);
 
 /* uv_spawn() options */
+typedef enum {
+  UV_IGNORE      = 0x00,
+  UV_CREATE_PIPE = 0x01,
+  UV_DUPLEX_PIPE = 0x02,
+  UV_IPC_PIPE    = 0x04,
```
igorzi: Isn't this redundant with `uv_pipe_init` taking `ipc` argument?

/cc @piscisaureus & @bnoordhuis

bnoordhuis: Looks that way, yes.

@indutny: You can check that with `pipe->ipc != 0`.

Keno:
BTW, did we ever decide whether or not to replace `uv_pipe_t` by `uv_stream_t`

bnoordhuis:
It seems like a good change to me.

include/uv.h
indutny: It's not relevant on unix, since I'm creating `socketpair()` for any kind of pipe. And it'll be mandatory for windows if we'll move from `uv_pipe_t` to `uv_stream_t`, since the latter one doesn't contain this information.

indutny:
Cutted out UV_IPC_PIPE. /cc @igorzi

indutny:
landed

PR #416, opened 2012-05-16
Change #if FIONBIO to #ifdef FIONBIO.

jedisct1:
Fixes compilation on Debian/ppc 6.0.4.


bnoordhuis:
Thanks, merged in 07622e7. Curious that debian/ppc doesn't have FIONBIO.

PR #421, opened 2012-05-21
Unbreak compilation on Dragonfly BSD, that uses kqueue, too.

jedisct1:


bnoordhuis:
Thanks Frank, merged in d27a62d. Now I've met the other DragonFly BSD user, too.

PR #422, opened 2012-05-21
Fix compilation with MingW: NTSTATUS & PNTSTATUS were redefined.

jedisct1:
Once in uv-win.h, then in winapi.h

### DIFF
--- a/include/uv-private/uv-win.h
+++ b/include/uv-private/uv-win.h
@@ -134,6 +134,7 @@ typedef int (WSAAPI* LPFN_WSARECVFROM)
              LPWSAOVERLAPPED_COMPLETION_ROUTINE completion_routine);
 
 #ifndef _NTDEF_
+#define _NTDEF_ 1
   typedef LONG NTSTATUS;
   typedef NTSTATUS *PNTSTATUS;
 #endif


include/uv-private/uv-win.h
```diff
@@ -134,6 +134,7 @@ typedef int (WSAAPI* LPFN_WSARECVFROM)
              LPWSAOVERLAPPED_COMPLETION_ROUTINE completion_routine);
 
 #ifndef _NTDEF_
+#define _NTDEF_ 1
```
bnoordhuis, line 137: That doesn't look right to me. @piscisaureus?

luislavena:
@bnoordhuis @piscisaureus this is exactly the issue I described in #403

bnoordhuis:
It is. But changing a `#ifndef` to `#define` doesn't seem like a proper fix to me.

luislavena:
@bnoordhuis I see **adding** `#define` in the diff, to avoid the redefinition caused by winapi issue.

AFAIK is that inside `_NTDEF_` there are other things beyond `NTSTATUS`, and with this change we still have the duplicated code between the uv-internal and winapi headers.

bnoordhuis:
Sorry, that's what I mean. That's why it doesn't look like a proper fix to me but I'll let @piscisaureus decide.

piscisaureus:
Sorry for leaving this out there for so long, I was busy with other stuff.

luislavena:
@piscisaureus we understand, no problem. Bring chocolate next time and we will forgive you :wink:

Please also see my comments on #403, both pull requests attempt to workaround the real issue: duplicated code around `NTSTATUS` in the headers.

Let us know your thoughts.

PR #428, opened 2012-05-24
Shared library support

avalanche123:
Build will now generate `libuv.a` and `libuv.(so|dylib)` products


avalanche123:
made the diff easier to digest, feedback welcome

mrunge:
Any news here? I'd love to see libuv as a shared library and to have this patch included upstream. 
Being able to build a shared library also improves chances to get it included e.g. in Fedora, since static binaries are strictly forbidden by policy.

bnoordhuis:
I'm afraid this PR no longer applies but I have a patch pending that lets you compile libuv as a shared object. It's contingent on removal of libev though, see #485 for details. I'll try to work up the courage to finish and land it this week.

nathansobo:
:metal: :+1:  Sending moral support your way for landing it!

OCTAGRAM:
Where to get precompiled libuv.dll? node.js for Windows and luvit are linked statically against libuv. Rust has separate rustrt.dll or something like this that can be used instead of pure libuv.dll, but it weights 2.5Mb and depends on VC++ DLLs, and that means even more Mbs of dependencies.

bnoordhuis:
@OCTAGRAM: We don't provide one, you have to compile it from source.

txdv:
it weights 410K as a linux dynamic library

OCTAGRAM:
1. libuv is not friendly for Delphi & C++Builder development. Average developer is supposed to bother with never seen previously MinGW or Microsoft compilers and with GYP which immature website does not say anything about C++Builder support. Embarcadero went Mac OS X since XE2 and they promise to go Linux (after iOS and Android). And I think, bothering with GCC and GYP won't be desired for C++Builder for Linux developers
2. What's most important for me, shared libuv is a strong requirement for pluggable architecture. Plug-ins are supposed to share event loop and libuv handles, and so they are supposed to use shared libuv DLL. As opposed to other OSes, Windows has dynamic loader instead of dynamic linkers, so 2 DLL linking libuv statically would use 2 independent instances of libuv. This is not appropriate. node.js must use shared libuv.dll, luvit must do the same. Rust has libuv in a separate DLL, but there is redundant stuff in this DLL which is not required for Delphi programs. There should be a possibility to write a DLL in Delphi that is usable from node.js, luvit and probably any other host application using libuv. And vice versa: Delphi applications should be able to be host for another DLLs using libuv.

Making shared library libuv a mainstream will promote libuv.

Delphi and C++Builder are already using conveyor approach for its networking features. TTcpServer, TTcpClient and so on invoke handlers in a single GUI thread. There are several iocp client/server implementations for Delphi. I think, integrating libuv into Delphi would be best approach given the opportunity of interoperating with another libuv-enabled DLLs in another programming languages.

piscisaureus:
While pondering over the call for C++Builder support I got overwhelmed by nostalgia and memories of good old Borland C++ that taught me how to ride a bike. So I opted to try to add bcc support to gyp instead. Unfortunately on my quest for a modern version of c++builder I hit several roadblocks consisting of registration/billing forms and serious legalese warning me about how I would burn in hell if I'd dare to use it without a license.

The anticlimax end of this story is that I gave up.

OCTAGRAM:
Considering the importance of explicitly shared DLL, that doesn't change a lot. A Delphi user is supposed to download node.js distribution (or a separate libuv distribution), throw out node.exe and just use libuv.dll. When it comes to interoperating with another DLLs powered by libuv, everything is ready. As of XE2, Delphi is said to support COFF object files, so static libuv can also be helpful.

PR #429, opened 2012-05-24
Fix sunos build: undeclared variable 'handle'

isaacs:


piscisaureus:
Thanks. Landed in 604802a

PR #437, opened 2012-05-31
Support uv_interface_addresses on SunOS

mcavage:
Hello,

The Node.js `require('os').networkInterfaces()` is broken on Illumos (i.e., Solaris), and digging into the code, it's all there, but was #ifdef'd out, presumably because it was unclear if `getifaddrs` was available. `getifadds` is available on Illumos platforms (it seems it has been since late OpenSolaris days).  This PR simply #ifdef's out that code.  Running from node on smartos now spits out:

```
console.log(JSON.stringify(os.networkInterfaces(), null, 4))
{
   "lo0": [
       {
           "address": "127.0.0.1",
           "family": "IPv4",
           "internal": true
       },
       {
           "address": "::1",
           "family": "IPv6",
           "internal": true
       }
   ],
   "net0": [
       {
           "address": "10.88.88.130",
           "family": "IPv4",
           "internal": false
       }
   ]
}
```

### DIFF
--- a/src/unix/sunos.c
+++ b/src/unix/sunos.c
@@ -28,9 +28,7 @@
 #include <assert.h>
 #include <errno.h>
 
-#ifdef SUNOS_HAVE_IFADDRS
-# include <ifaddrs.h>
-#endif
+#include <ifaddrs.h>
 #include <net/if.h>
 
 #include <sys/loadavg.h>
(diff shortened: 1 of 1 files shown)


isaacs:
Landed on 9cb8bdc74edef8e3503c260a2f07150d211a48a9.  Thanks!

PR #441, opened 2012-06-02
process: /dev/null ignored stdio if index < 3

indutny:
nt


piscisaureus:
@bnoordhuis It looks good to me but I'd like you to take another look.

piscisaureus:
I landed the test (with slight modifications) in 1cd9642

src/unix/process.c
```diff
@@ -323,11 +325,25 @@ int uv_spawn(uv_loop_t* loop, uv_process_t* process,
 
       if (use_fd >= 0) {
         close(close_fd);
-        dup2(use_fd, i);
+      } else if (i < 3) {
+        /* `/dev/null` stdin, stdout, stderr even if they've flag UV_IGNORE */
+        if (i == 0) {
+          use_fd = open("/dev/null", O_RDONLY);
+        } else {
+          use_fd = open("/dev/null", O_RDWR);
```
bnoordhuis: I'm not wildly enthusiastic about the 5 levels of indent or the many if/then/else branches. Is there a way to flatten / simplify the code?

bnoordhuis:
Another thing, the commit log could be improved. Something like "unix: remap fds 0-2 to /dev/null if ignore flag set"

indutny:
@bnoordhuis moved init loop to separate function (for fixing indentation levels), and changed commit message. Please review!

bnoordhuis:
LGTM

indutny:
Can you please merge it then?

bnoordhuis:
Landed in a30e45f.

Issue #448, opened 2012-06-05
uv_close(TCP handle) does not immediately invoke uv_connect_cb(EINTR)


ibc:
(Linux libuv master branch). Pseudo-code:

<pre>
# A TCP socket that will be in connecting state for long...
uv_tcp_init(loop, tcp_handle);
uv_tcp_connect(tcp_handle, "1.2.3.4", 9999);
uv_read_start(tcp_handle);

# Run a timer that invokes uv_close(tcp_handle) after two seconds.
[...]

# We expect that uv_connect_cb(EINTR) should happen right now, but that
# does not happen.

# It will happen in next event of the loop (if there are...), so for example if we
# run another timer to fire in 5 seconds (and do nothing in the callback) then
# the uv_close_cb(tcp_handle) is called after 5 seconds and thus the
# on_connect_cb(EINTR).

uv_run(loop);
</pre>

Maybe related to #447?

ibc:
**NOTE:**  It's a regression, it worked fine in a copy of libuv from 2012-05-29 I had somewhere...

bnoordhuis:
Cannot reproduce, same for #447. Can you post a test case? I just wasted the better part of an hour trying to come up with one based on your description...

bnoordhuis:
That said, I'm pretty sure I fixed the issues you mention in 649ad50.

ibc:
Hi, I still reproduce this issue after the commit. Let me post a test case.

ibc:
Hi, working code that demonstrates the bug:  https://gist.github.com/2880865

bnoordhuis:
Fixed in dea4594.

ibc:
Just two lines of code???? XD

Thanks a lot, tested in my code and works.

PR #449, opened 2012-06-06
Test unit for #448

ibc:
This pull request includes a new test unit "tcp_close_while_connecting" which demonstrates the bug reported in #448, and could avoid a future regression. Comments within the text file should be (probably) modified once the bug is fixed.


bnoordhuis:
Thanks, I've landed a slightly modified version in b47af98. Don't worry, you still get the +1 commit. :-)

PR #450, opened 2012-06-07
linux: fix uv_hrtime() to return a right value

shigeki:
I struggled and annoyed against the timeout errors of the recent Node in master for these days and finally got it to be solved.
This tiny bug caused the stall of epoll_wait() in my 32bit Linux because it sometimes but not always generated an invalid timeout value.
I separated this PR into two commits, one is the fix and the other is the test because it needs many iterations and takes some time  to have the hr_time() error. If it need not to merge the revised test, please take only fix.


Sannis:
The same should goes [here](https://github.com/joyent/node/blob/master/src/node.cc#L1689)?

shigeki:
It may be but no harm at the moment because process.hrtime() is used only in benchmark.

Sannis:
`process.hrtime` is in public API. for example I use it for long-time (many hours) benchmarks.

bnoordhuis:
Thanks for pointing that out. I landed the test in e2aa39a and addressed the issue in c639f34.

shigeki:
@bnoordhuis Oh No, Ben. You missed to change the code for Linux. Added bb64032626c25bcdbdb9af437bf6eeb4b61614ec . But I'd like to confirm that the following warning message is shown with compile options of --std=c89 -pedantic as

```
src/unix/linux/core.c -o src/unix/linux/core.o
src/unix/linux/core.c: In function 'uv_hrtime':
src/unix/linux/core.c:74:23: warning: use of C99 long long integer constant [-Wlong-long]
```

Casting to uint64_t does not show the warning, that is why I took it.

bnoordhuis:
I guess you're right. Updated in 59cda86.

It's kind of silly that the compiler warns about unsigned long long but not uint64_t. They're both C99-isms.

shigeki:
@bnoordhuis  Thanks. I think this is odd, too. 
@Sannis Sorry, I was not aware that process.hrtime() has already been documented.
 I've just made a patch to fix process. hrtime() in https://github.com/shigeki/node/commit/72a85e7a7acdd59e00c282c5328c8a94120326ba . I will submit PR after libuv is upgraded in node.

Sannis:
> Sorry, I was not aware that process.hrtime() has already been documented.

I do not know what core team think about `process.hrtime()`, but i have no thoughts why this may be private. @bnoordhuis can resolve this.

TooTallNate:
@Sannis `process.hrtime()` is public API, hence being documented.

bnoordhuis:
`process.hrtime()` doesn't have to be fixed. It already uses uint64_t so NANOS_PER_SEC is automatically widened to 64 bits.

shigeki:
Oops, you are right, Ben. `t` and `seconds` have already been uint64_t.
As for the warning suppression, uint64_t of 32bit Linux is defined in /usr/include/stdint.h as

```
__extension__ typedef unsigned long long int  uint64_t;
```

and the gcc manual http://gcc.gnu.org/onlinedocs/gcc-4.6.3/gcc/Alternate-Keywords.html#Alternate-Keywords says that

```
-pedantic and other options cause warnings for many GNU C extensions. 
You can prevent such warnings within one expression by writing __extension__ before 
the expression.  __extension__ has no effect aside from this. 
```

PR #451, opened 2012-06-08
RFC: More flexible pipe API

Keno:
@piscisaureus As discussed on IRC and outlined in https://gist.github.com/2879135 and https://gist.github.com/2879181 this makes the pipe API more flexible by giving explicit control over both pipe ends. 
I did the windows side of this first as I figured that would be the more challenging of the two implementations and would thus reveal any issues.

### DIFF
--- a/include/uv-private/uv-win.h
+++ b/include/uv-private/uv-win.h
@@ -165,6 +165,8 @@ typedef struct uv_buf_t {
 
 typedef int uv_file;
 
+typedef HANDLE uv_os_handle_t;
+
 typedef SOCKET uv_os_sock_t;
 
 typedef HANDLE uv_thread_t;
@@ -351,7 +353,10 @@ RB_HEAD(uv_timer_tree_s, uv_timer_s);
 #define uv_pipe_connection_fields         \
   uv_timer_t* eof_timer;                  \
   uv_write_t ipc_header_write_req;        \
-  int ipc_pid;                            \
+  union {								  \
+    int pid;							  \
+    int *p_pid;							  \
+  } ipc_pid;							  \
   uint64_t remaining_ipc_rawdata_bytes;   \
   struct {                                \
     WSAPROTOCOL_INFOW* socket_info;       \
--- a/include/uv.h
+++ b/include/uv.h
@@ -904,20 +904,41 @@ UV_EXTERN uv_handle_type uv_guess_handle(uv_file file);
  * uv_pipe_t is a subclass of uv_stream_t
  *
  * Representing a pipe stream or pipe server. On Windows this is a Named
- * Pipe. On Unix this is a UNIX domain socket.
+ * Pipe. On Unix this is a UNIX domain sockets
+ *
+ * A single uv_pipe_t always represents one end of a pipe. You can use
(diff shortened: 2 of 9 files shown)


Keno:
Also, I'm getting 4 test failures on both this branch and master: https://gist.github.com/2893501. I investigated the first one (because of the "pipe" in the name I though I had broken it), but interestingly, the failure does not occur when a the debugger is attached which makes me think it might be something race-y.

piscisaureus:
@loladiro Did you figure out what to do about the issue that you had with ipc pipes the other day?

Keno:
Yes, I don't really like it, but I made it the other pipe end include a pointer to the pid field in the parent pipe:

```
+  union {                  \
+    int pid;                \
+    int *p_pid;                \
+  } ipc_pid;                \
```

If you have a better idea, I'm all ears.

piscisaureus:
Well, I think the that uv_read_start should just be modified to write the current process pid to the pipe, and (if that didn't happen yet), read the remote's pid from the pipe.

piscisaureus:
Bump @piscisaureus

vtjnash:
If it helps, I updated and rebased this pull request: https://github.com/JuliaLang/libuv/tree/upipes

indutny:
any update?

vtjnash:
JuliaLang has continued to maintain an up-to-date version of this patch. The latest is part of the https://github.com/JuliaLang/libuv/tree/julia-uv0.11.22 branch, where the relevant commits are generally the first few rebased to be first after the fork. Currently, those are:
https://github.com/JuliaLang/libuv/compare/cd0c19b1d3c56acf0ade7687006e12e75fbda36d...839ceedb7ae91cb9594aa8334dcb0352360e1fb5

saghul:
@trevnorris please hold reviewing this until we are done with 0.12. Also, we need to go over the design before style.

trevnorris:
@saghul okie dokie. deleted all my style comments to remove clutter. sorry about that :)

piscisaureus:
@Keno @vtjnash 

I believe that if we added UV_INHERIT_OS_HANDLE to [this list[(https://github.com/libuv/libuv/blob/master/include/uv.h#L788), julia could create it's own pipes with `pipe(2)` or `CreatePipe()`. You wouldn't need `uv_pipe_close_sync()`; instead you could close the pipe with the platform-specific `close()` or `CloseHandle()`.

I agree that it's not the nicest of solutions. Hopefully when we take another fresh look at the child process API we can structure things differently so this wouldn't be so much of a problem.

For now, would this work?

vtjnash:
that does seem fairly reasonable. if libuv were to also provide `uv_pipe_open_from_os_handle` and `uv_pipe2` functions, I think that would work fairly well.

tkelman:
@vtjnash do you think we should work on adding that as the first part of a rebase (more like partial do-over) against either master or the 1.x branch? Depending on whether we can get to it before libuv devs could.

vtjnash:
reviewing the current julia-libuv diff list, it seems quite reasonable to take this approach, it just needs someone to take the time to implement a prototype/demonstration

PR #452, opened 2012-06-08
Added missing include for limits.h

saghul:
I'm not sure if this is required when compiling with Visual Studio, but it is when using MinGW.


bnoordhuis:
Thanks, merged in 6a47e3a.

PR #455, opened 2012-06-09
Add support for condition variables

nikhilm:
This is tested on Mac and Linux, but I don't have a Windows box to test it on, so someone should test that.

Thanks!


piscisaureus:
This won't work on windows XP/3k. That was also the reason that @bnoordhuis's implementation didn't make it into master - he actually made an attempt to support it, but there were some flaws in it. (see https://github.com/bnoordhuis/libuv/compare/uv_cond)

bnoordhuis:
@nikhilm If you want to pick up from where I left off, please do. I added semaphore support in c76c206 so it should be possible now (with a bit of effort) to emulate condition variables on Windows XP and 2003.

meh:
Any update on this?

nikhilm:
https://github.com/joyent/libuv/pull/567

PR #459, opened 2012-06-09
win: map ERROR_PRIVILEGE_NOT_HELD to UV_EPERM

saghul:


saghul:
Can you check, @piscisaureus? Thanks!

piscisaureus:
I would say it should map to EPERM and not EACCESS. The reason that symlink fails is not that the user has no write access to this particular file; rather, the process is lacking privileges to create symlinks at all.

saghul:
I doubted between the two and picked EACCES, I can update the pr making it map to EPERM if that's more appropriate.

saghul:
Updated the commit, now it maps to EPERM. 

Btw, I force pushed to update previous commit, is that the right thing to do for pull requests? Thanks!

bnoordhuis:
Thanks, landed in 431d61a. Force-pushing is the right thing to do, yes. :-)

PR #460, opened 2012-06-11
stream: return error rather than do assertion

indutny:
uv__read_start_common should not throw assertion on uninitialized
streams (with fd < 0), it should gracefuly return error to let users
process it.


bnoordhuis:
This is about joyent/node#3402, right? I'll remove the assert someday but not just yet, I like that it catches bugs in node.

Issue #462, opened 2012-06-11
uv_dlopen fails even though it has a valid handle


joeferner:
When using node v0.7.9 on OSX built from source when I load my module it fails. When using node v0.7.8 on OSX built from source with the same javascript source it succeeds.

```
var java = require('java');
```

```
module.js:480
  process.dlopen(filename, module.exports);
          ^
Error: dlsym(RTLD_DEFAULT, JLI_MemAlloc): symbol not found
    at Object..node (module.js:480:11)
    at Module.load (module.js:356:32)
    at Function._load (module.js:312:12)
    at Module.require (module.js:362:17)
    at require (module.js:378:17)
    at Object.<anonymous> (/Users/admin/projects/rhesus-scrapp/node_modules/java/lib/nodeJavaBridge.js:5:16)
    at Module._compile (module.js:449:26)
    at Object..js (module.js:467:10)
    at Module.load (module.js:356:32)
    at Function._load (module.js:312:12)
```

After running git bisect I tracked it down to a commit made to node that upgraded libuv 039fac633eaf081a96bb1ed8bc7307d03dcbe9d9.

I then played with the code a bit and found that dlopen is returning a valid handle (i.e. lib->handle != NULL) yet dlerror is returning that an error occurred.

bnoordhuis:
Good catch, seems like it's a genuine error after all. Does this patch fix it for you?

``` diff
diff --git a/src/unix/dl.c b/src/unix/dl.c
index 9eb6600..01796e3 100644
--- a/src/unix/dl.c
+++ b/src/unix/dl.c
@@ -31,6 +31,7 @@ static int uv__dlerror(uv_lib_t* lib);


 int uv_dlopen(const char* filename, uv_lib_t* lib) {
+  dlerror(); /* Reset error status. */
   lib->errmsg = NULL;
   lib->handle = dlopen(filename, RTLD_LAZY);
   return uv__dlerror(lib);
```

joeferner:
At first that's what I thought it was, doesn't seem like resetting the error using dlerror is enough for me. I guess it couldn't hurt though. It could be that the library being loaded (java in my case) is calling dlopen itself and is doing something else with it internally, not really sure.

bnoordhuis:
Are you _sure_ it doesn't fix the problem? I was able to reproduce an issue very similar to the one you reported in joyent/node#3392 and 84f0d96 fixed it for me.

joeferner:
[...]

PR #463, opened 2012-06-11
Fix uv_dlopen to handle valid opens

joeferner:
This fixes issue #462


src/unix/dl.c
```diff
@@ -33,6 +33,14 @@
 int uv_dlopen(const char* filename, uv_lib_t* lib) {
   lib->errmsg = NULL;
   lib->handle = dlopen(filename, RTLD_LAZY);
+  if (lib->handle) {
+    if (lib->errmsg)
+      free(lib->errmsg);
+
+    lib->errmsg = NULL;
```
bnoordhuis, line 40: The reason I didn't take this patch is that `lib->errmsg` is always NULL here so most of this code is a no-op.

bnoordhuis:
Fixed in 1b68434.

PR #466, opened 2012-06-12
Implement uv_loop_get_backend_fd()

lpereira:
This allows obtaining the poller file descriptor for backends that support
it (mainly epoll under Linux and kqueue under *BSD); other backends (such as
the Windows one) will return -1.

This can be used to integrate with other main loops (such as EFL's Ecore or
Glib's).

### DIFF
--- a/include/uv-private/ev.h
+++ b/include/uv-private/ev.h
@@ -598,6 +598,8 @@ void ev_loop_fork (EV_P);
 
 unsigned int ev_backend (EV_P); /* backend in use by loop */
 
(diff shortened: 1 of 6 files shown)


bnoordhuis:
Thanks, but I won't take it for two reasons:
1. The poll fd itself is not enough. You also need to know the poll timeout, otherwise timers and idle handles won't work.
2. I'm _this_ close to dropping libev altogether. It's probably going to happen in the next week or two. When I do, I'll add an API to get at the poll fd (and timeout).

lpereira:
(2) is great news, bnoordhuis!

PR #467, opened 2012-06-13
Wrong compiler opiton in Makefile

verm666:
gcc / clang doesn't have --std option. Use -std instead.


bnoordhuis:
I know it's a oneliner but can you sign [the CLA](http://nodejs.org/cla.html)?

verm666:
Sorry, no.

PR #469, opened 2012-06-18
fs test fixes

AvianFlu:
These are some minor test improvements that will prevent unneeded test failures moving forward.

The file watching tests in `test-fs-events.c` all use the same directory for their fixtures - `watch_dir`.  If a test fails from an assertion error, it doesn't clean up, and this can cause other tests to fail, simply due to the other tests trying to rmdir("watch_dir") when there are still other files left in the directory.  I've replaced the repetitive and not always accurate manual cleanups with a helper method, `cleanup_watch_dir`, that reads the directory first, and cleans up whatever is actually there.

I added a second commit to skip one of the `uv_fs_chown` test cases if the tests are being run as root.  The test expects to fail - a normal user can't `chown` something to root - but if the current uid is 0, it succeeds, which makes the test fail.  This seemed spurious.

### DIFF
--- a/test/test-fs-event.c
+++ b/test/test-fs-event.c
@@ -38,6 +38,27 @@ static int close_cb_called = 0;
 static int fs_event_cb_called = 0;
 static int timer_cb_touch_called = 0;
 
+static void cleanup_watch_dir() {
+  uv_loop_t* loop = uv_default_loop();
+  uv_fs_t readdir_req;
+  uv_fs_t unlink_req;
+  int i, r;
+  char *buffer, *name;
+
+  r = uv_fs_readdir(loop, &readdir_req, "watch_dir", 0, NULL);
+  buffer = readdir_req.ptr;
+  uv_chdir("watch_dir");
+  for (i = 0; i < readdir_req.result; i++) {
+    name = buffer;
+    r = remove(name);
+    ASSERT(r == 0);
+    buffer += strlen(name) + 1;
+  }
+  uv_chdir("..");
+  r = remove("watch_dir");
+  ASSERT(r == 0 || uv_last_error(loop).code == UV_ENOENT);
+}
+
 static void create_dir(uv_loop_t* loop, const char* name) {
   int r;
   uv_fs_t req;
@@ -165,9 +186,7 @@ TEST_IMPL(fs_event_watch_dir) {
   int r;
 
   /* Setup */
-  uv_fs_unlink(loop, &fs_req, "watch_dir/file1", NULL);
-  uv_fs_unlink(loop, &fs_req, "watch_dir/file2", NULL);
-  uv_fs_rmdir(loop, &fs_req, "watch_dir", NULL);
+  cleanup_watch_dir();
   create_dir(loop, "watch_dir");
 
   r = uv_fs_event_init(loop, &fs_event, "watch_dir", fs_event_cb_dir, 0);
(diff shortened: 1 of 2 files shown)


AvianFlu:
I have force-pushed the `test-fixes` branch with some fixes for windows.  e0124ab and 04ade2e are now the two commits.

test/test-fs.c
```diff
@@ -191,9 +191,13 @@ static void chown_root_cb(uv_fs_t* req) {
   /* On windows, chown is a no-op and always succeeds. */
   ASSERT(req->result == 0);
 #else
-  /* On unix, chown'ing the root directory is not allowed. */
-  ASSERT(req->result == -1);
-  ASSERT(req->errorno == UV_EPERM);
+  /* On unix, chown'ing to the root user is not allowed, except by root. */
+  if (getuid() != 0) {
```
bnoordhuis: People should not be running the tests as root, it's dangerous.

I think it's a better idea to build an `if (is_root) die("don't do that, silly");` check into `test/runner-unix.c`

test/test-fs-event.c
```diff
@@ -38,6 +38,28 @@
...
 
+static void cleanup_watch_dir() {
+  uv_loop_t* loop = uv_default_loop();
+  uv_fs_t readdir_req;
+  uv_fs_t unlink_req;
+  int r = 0, i;
+  char *buffer, *name;
+
+  r = uv_fs_readdir(loop, &readdir_req, "watch_dir", 0, NULL);
+  buffer = (char*)readdir_req.ptr;
```
bnoordhuis: Superfluous cast.

test/test-fs-event.c
```diff
@@ -38,6 +38,28 @@
...
+  uv_chdir("watch_dir");
+  for (i = 0; i < readdir_req.result; i++) {
+    name = (char*)&buffer[i];
```
bnoordhuis: Ditto.

test/test-fs-event.c
```diff
@@ -38,6 +38,28 @@
 static int fs_event_cb_called = 0;
 static int timer_cb_touch_called = 0;
 
+static void cleanup_watch_dir() {
+  uv_loop_t* loop = uv_default_loop();
+  uv_fs_t readdir_req;
+  uv_fs_t unlink_req;
+  int r = 0, i;
```
bnoordhuis: Doesn't need to be initialized.

test/test-fs-event.c
```diff
@@ -38,6 +38,28 @@
...
+  char *buffer, *name;
+
+  r = uv_fs_readdir(loop, &readdir_req, "watch_dir", 0, NULL);
+  buffer = (char*)readdir_req.ptr;
+  uv_chdir("watch_dir");
+  for (i = 0; i < readdir_req.result; i++) {
+    name = (char*)&buffer[i];
+    r = uv_fs_unlink(loop, &unlink_req, name, NULL);
```
bnoordhuis: I'd just use `remove(name)` here. It's portable and less code.

AvianFlu:
@bnoordhuis - force-pushed again, commits now 9dd5295 and 244de4b.

bnoordhuis:
Thanks Charlie, landed in 7573f4a and f3fd8e3.

piscisaureus:
I reverted this in e9b17bc. It broke on windows, and the cleanup_watch_dir needs proper review. (e.g. don't remove files without checking if chdir succeeds etc.) Also I'd prefer to use `uv_unlink`/`uv_rmdir` instead of remove, because remove() actually doesn't really work well here.

bnoordhuis:
I'm going to close this.

@AvianFlu If you want to submit it again, please open a new PR.

AvianFlu:
Will do - been very busy lately.

bnoordhuis:
Aren't we all? :-) No rush.

PR #471, opened 2012-06-18
Allow suppression of ifaddrs.h on SunOS

isaacs:
Older Solaris versions do not have this feature.  This allows
them to set SUNOS_NO_IFADDRS to allow libuv to build.

Re: joyent/node#3465


bnoordhuis:
LGTM

bnoordhuis:
Merged in 382f2a2.

PR #472, opened 2012-06-18
Update c-ares to version 1.9.0

saghul:
Based on Ben's work for 1.8.0 this updates the bundled c-ares to version 1.9.0.

My Windows VM was a bit quirky at times, so if anyone else could run the gethostbyname tests would be great. Anyhow, the gethostbyname benchmark gives reasonable results now.

Also, I wasn't sure if this floating chunk of code needs to be included: https://github.com/joyent/libuv/blob/master/src/ares/ares_init.c#L701 if it is, then I'll apply it.

Sorry I couldn't get this done before the Node 0.8 release :-(

/cc @bnoordhuis, @piscisaureus


bnoordhuis:
LGTM. All tests pass.

@piscisaureus Can you review/test Windows?

piscisaureus:
@saghul 

> Also, I wasn't sure if this floating chunk of code needs to be included: https://github.com/joyent/libuv/blob/master/src/ares/ares_init.c#L701 if it is, then I'll apply it.

Yes, we need that too.

piscisaureus:
@saghul you should try to reapply cb474b24c1bd77fb2deec4be1830adeb17d435f6.

saghul:
Hi guys, thanks for looking into this :-)

@piscisaureus I applied a slightly modified version of that patch, since the code for reading DNS nameservers in Windows was completely overhauled recently. Also I realized that this fix only applies to the case when the nameservers are fetched with GetAdaptersAddresses(). Is this ok? c-ares will try to use GetAdaptersAddresses() then GetNetworkParams() and then it fallbacks to the registry: https://github.com/bagder/c-ares/blob/master/ares_init.c#L1056

piscisaureus:
@saghul GetAdaptersAddresses() always works on the windowses we support. The fallback methods are for Windows 2000 and Windows 95, so don't bother.

piscisaureus:
Thanks, around here: https://github.com/joyent/libuv/commits/5ee80f1a7a07eac257e497833c9db58ee5d8ba3a

saghul:
Great, thanks! :-)

bnoordhuis:
@piscisaureus Why is this closed?

piscisaureus:
@bnoordhuis Because it landed.

bnoordhuis:
No one ever tells me anything. :-(

piscisaureus:
Dude, it's all there: https://github.com/joyent/libuv/pull/472#issuecomment-6647460

bnoordhuis:
Mr. Belder, I expect TPS reports signed off in triplicate by our QA and acceptance testing teams, CC'd to all for each and every commit. Except mine, of course.

PR #476, opened 2012-06-25
darwin: get cpu model correctly on mac

upsuper:
The "hw.model" in OS X is something like "MacBookPro8,2", so I use CPU brand_string instead to let it get the CPU model more correctly.


bnoordhuis:
LGTM. Can you sign [the CLA](http://nodejs.org/cla.html)?

upsuper:
@bnoordhuis OK, I've signed it electronically.

bnoordhuis:
Thanks, merged in aec320b.

PR #478, opened 2012-06-26
uv.h: add members to any-union of handle and req

shigeki:
I would like to put my 2 cents  in the new libuv. 
Congratulations on the new release.  Recent progress of libuv surprised me a lot for its stability and performance. Thanks for your hard work.


bnoordhuis:
Thanks Shigeki, landed in c6f2ef2. And thanks for the kind words. :-)

PR #483, opened 2012-06-30
Stack Buffer Overflow/Segmentation Fault in uv_cpu_info

ghost:
Discovered this in NodeJS v0.8 when using os.cpus()

### DIFF
--- a/src/unix/freebsd.c
+++ b/src/unix/freebsd.c
@@ -261,7 +261,7 @@ uv_err_t uv_cpu_info(uv_cpu_info_t** cpu_infos, int* count) {
     return uv__new_sys_error(ENOMEM);
   }
 
-  if (sysctlbyname("kern.cp_times", &cp_times, &size, NULL, 0) < 0) {
+  if (sysctlbyname("kern.cp_times", cp_times, &size, NULL, 0) < 0) {
     free(cp_times);
     free(*cpu_infos);
     return uv__new_sys_error(errno);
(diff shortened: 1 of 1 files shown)


src/unix/freebsd.c
```diff
@@ -276,7 +276,7 @@ uv_err_t uv_cpu_info(uv_cpu_info_t** cpu_infos, int* count) {
     cpu_info->cpu_times.idle = (uint64_t)(cp_times[CP_IDLE+cur]) * multiplier;
     cpu_info->cpu_times.irq = (uint64_t)(cp_times[CP_INTR+cur]) * multiplier;
 
-    cpu_info->model = strdup(model);
+    cpu_info->model = strdup((const char *)&model);
```
bnoordhuis, line 279: Why is this cast necessary?

ghost, line 279: Guess its not, didn't realize that 'model' was a pointer to a stack buffer... It's on the stack, right? I rarely look into this kind of stuff (C,C++) and it's been a while...

I debugged with GDB, got a bit confused there when i saw that cpu_info->model referenced an invalid address before i realized that 'numcpus' had been overwritten from 1 to 0 after the call to sysctlbyname on line 264 and that loop was never entered...

In my mind i thought that since strdup takes (const char *) as an argument and didn't realise that 'model' was a pointer, so cast it..

ghost:
Just to touch upon a final point made briefly above: arrays don't require a reference operator (the ampersand) when you want to have a pointer to them. For example:

``` C
char *ptr;
char str[40];
ptr = str;  /* Gives the memory address without a reference operator(&) */
```

As opposed to

```
int *ptr;
int num;
ptr = &num; /* Requires & to give the memory address to the ptr */
```

The fact that arrays can act just like pointers can cause a great deal of confusion :P

bnoordhuis:
@oddmarthon-lende You haven't updated your PR and I more or less arrived independently at the same fix when debugging a node crash so I've landed that instead in be09be7. Thanks anyway for pointing out the issue.

PR #488, opened 2012-07-08
thread: use mach semaphores on osx

indutny:
nt


piscisaureus:
No real objections, but: why?

indutny:
That's a good question. AFAIK, because not every OSX version has support of posix semaphores (and I'm afraid that even on latest version support may be quite limited).

indutny:
Better ask @bnoordhuis anyway

include/uv-private/uv-unix.h
```diff
@@ -40,7 +40,14 @@
 #include <termios.h>
 #include <pwd.h>
 
+#if defined(__APPLE__) && defined(__MACH__)
+#include <mach/mach.h>
```
bnoordhuis, line 44: `# include <...>`

src/unix/thread.c
```diff
@@ -168,6 +168,45 @@ void uv_once(uv_once_t* guard, void (*callback)(void)) {
...
+    abort();
+}
+
+
+int uv_sem_trywait(uv_sem_t* sem) {
+  mach_timespec_t interval;
+  interval.tv_sec = 0;
+  interval.tv_nsec = 0;
+
+  /* A little bit hacky but should work */
```
bnoordhuis, line 200: You can remove the comment. I checked the xnu source - a zero timeout polls, it never blocks.

src/unix/thread.c
```diff
@@ -168,6 +168,45 @@ void uv_once(uv_once_t* guard, void (*callback)(void)) {
 }
 
 
+#if defined(__APPLE__) && defined(__MACH__)
```
bnoordhuis, line 171: Add a newline here.

EDIT: Ah, never mind. I'll touch it up when I merge it.

bnoordhuis:
Thanks Fedor, landed in 3a54805. I'm a bit unclear on how Mach semaphores interact with signals, i.e. if they're interruptible, but a quick scan of sync_sema.c suggests they're not.

indutny:
Landed in 3a548053b8d758f04e1d15c84daadc49fb9ac69e

PR #492, opened 2012-07-11
move uv_set_process_title to core

indutny:
Use hijacking argv array to change process' title. It seems to be
working fine on almost every platform (at least it should not break
anything as it's used in nginx in a similar way).


src/unix/core.c
```diff
@@ -653,3 +660,70 @@ void uv__io_feed(uv_loop_t* loop, uv__io_t* handle, int event) {
...
+  if (process_title.len)
+    strncpy(process_title.str, title, process_title.len - 1);
+
+  return uv_ok_;
+}
+
+
+uv_err_t uv_get_process_title(char* buffer, size_t size) {
+  if (process_title.str) {
+    strncpy(buffer, process_title.str, size);
```
bnoordhuis: Looking at this code again, I realize it should use uv_strlcpy() instead of strncpy() because it won't always zero terminate the output buffer now. No matter, we can fix that up in a separate commit.

bnoordhuis:
LGTM. Have you per chance tested it on freebsd?

indutny:
Unfortunatelly, no. I've only two osx machines now.

bnoordhuis:
Okay, no worries. I'll test it on my freebsd 9 machine later tonight (assuming I don't forget, of course. :-)

bnoordhuis:
Update: it doesn't work on freebsd.

indutny:
Well, it wasn't supported anyway. Does it crash on tests?

bnoordhuis:
No, what I mean is that the current approach - setproctitle() - works while hacking argv and the environ doesn't. You'll have to roll back the changes in freebsd.c.

indutny:
I made what Russians are calling "Ход конём" - creating title.c for every platform that supports argv pollution, and leaving freebsd.c as it is. Please verify that it's working correctly now, thank you!

bnoordhuis:
Landed with some fixups in dc97d44. Thanks, Fedor.

PR #498, opened 2012-07-21
tcp: uv_tcp_reset_timed(...)

indutny:
Usually close() syscall will send a FIN packet to the other side and
wait for it's acceptance. While it's waiting, socket will persist in
system since FIN packet may get lost in the way to other side and will
need to be sent again. But when dealing with TCP clients that are
unable to do so (i.e. mobile, or clients under firewall), hanging
TIME_WAIT or FIN_WAIT2 connections will be left.

Introduced function (when followed by close() call) will send RST
packet instead of FIN, and will immediately destroy socket and remove
it from system's list.

/cc @bnoordhuis


src/unix/tcp.c
```diff
@@ -293,6 +293,19 @@ int uv__tcp_keepalive(uv_tcp_t* handle, int enable, unsigned int delay) {
 }
 
 
+int uv__tcp_reset_timed(uv_tcp_t* handle, int enable) {
+  struct linger opt;
```
bnoordhuis, line 297: Newline.

indutny:
@bnoordhuis do you think this is still relevant?

bnoordhuis:
I'm closing this. I think we'll need to deal with linger timeouts eventually but in a more sophisticated way than simply disabling the timeout altogether.

PR #499, opened 2012-07-21
fs: fix warning on OSX

indutny:
OSX has no public API for fdatasync. And as pointed in `man fsync(2)`:

  For applications that require tighter guarantees about the integrity of
  their data, Mac OS X provides the F_FULLFSYNC fcntl. The F_FULLFSYNC
  fcntl asks the drive to flush all buffered data to permanent storage.
  Applications, such as databases, that require a strict ordering of writes
  should use F_FULLFSYNC to ensure that their data is written in the order
  they expect.  Please see fcntl(2) for more detail.


src/unix/fs.c
```diff
@@ -453,6 +453,13 @@ int uv_fs_fdatasync(uv_loop_t* loop, uv_fs_t* req, uv_file file, uv_fs_cb cb) {
    * do a full fsync instead.
    */
   WRAP_EIO(UV_FS_FDATASYNC, eio_fdatasync, fsync, ARGS1(file))
+#elif defined(F_FULLFSYNC)
+
+#define FCNTL_SYNC(fd) fcntl(fd, F_FULLFSYNC)
+  /* OSX >= 10.6 does have fdatasync, but better use fcntl anyway */
+  WRAP_EIO(UV_FS_FDATASYNC, eio_fdatasync, FCNTL_SYNC, ARGS1(file));
```
bnoordhuis: So why use eio_fdatasync then?

I suggest that you split off the fcntl(F_FULLSYNC) into a function that gets called by both code paths (synchronous and asynchronous).

indutny:
Done

bnoordhuis:
Landed with some nudges in 23dc564. Thanks, Fedor.

PR #500, opened 2012-07-21
unix: don't abort uv_mutex_trylock when it returns EDEADLK or EBUSY.

chobie:
`uv_mutex_trylock` only checks `EAGAIN`. this changes accepts `EDEADLK` and `EBUSY` errors.

Basically, `EDEADLK` only cause on same thread. I'm not sure I should add this error check here.
But this changes would be helpful in some cases (i. e single thread application). I would be glad if this is accepted.

Thanks,

### DIFF
--- a/src/unix/thread.c
+++ b/src/unix/thread.c
@@ -78,7 +78,7 @@ int uv_mutex_trylock(uv_mutex_t* mutex) {
 
   r = pthread_mutex_trylock(mutex);
 
-  if (r && (r != EDEADLK && r != EBUSY && r != EAGAIN))
+  if (r && r != EBUSY && r != EAGAIN)
     abort();
 
   if (r)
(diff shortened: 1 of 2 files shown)


bnoordhuis:
Not checking for EBUSY is a bug in libuv (`uv_rwlock_tryrdlock` and `uv_rwlock_trywrlock` should be updated as well). I can either fix it or take your patch if you amend it and sign [the CLA](http://nodejs.org/cla.html).

Not checking for EDEADLK is a feature because EDEADLK indicates an application bug. The mutex being locked by the same thread is the most common cause but not the only one.

chobie:
okay, I signed the CLA electronically. could you check my changes?

also removed needless parentheses regarding my changes.  As I looked around libuv sources and usually didn't use it in this situation.

Thanks,

bnoordhuis:
Looks great, thanks. Landed in 22f004d.

PR #501, opened 2012-07-24
Issue486 Autotool Support

JustinVenus:
Use the autotool build system for system packagers.

The old Makefile is renamed Makefile.old.  Please note this pull request updates .gitignore as well.

I have successfully tested the changes on OSX, Linux, and Solaris 11.

Enhancements over the previous build system:
- Solaris sunpro (solarisstudio) compiler 12.3 is supported.
- You may build with c99
- initial support in "configure.ac" for using the system provided libev, libeio, and libcares.

How to test the changes.

sh autogen.sh  #runs all of the utilities to generate `configure`
./configure       #generates an autotooled "Makefile"

make              #builds a shared and static library
make test        #runs the unittest test/runtests, note you may short circuit and skip "make"
make bench     #runs the unittest test/runbenchmarks, note you may short circuit and skip "make"
make dist        #creates a tarball gz file for distribution

make install    #installs the software on the system
make DESTDIR='/var/tmp/libuv' install    #installs the software at the specified prefix


configure.ac
```diff
@@ -0,0 +1,131 @@
...
+AM_CONDITIONAL([Cygwin], [false])
+
+AM_CONDITIONAL([UNIXLIKE], [true])
+#if the compiler is gcc add some extra options
+if test "x$GCC" = "xyes"; then
+   UV_CPPFLAGS+="-pedantic -Wextra -Wno-unused-parameter -D_GNU_SOURCE "
+fi
+
+#determine whether or not we need internal lib resources
+AC_CHECK_HEADERS([event.h],
```
bnoordhuis: Did you mean ev.h? event.h is from libevent.

bnoordhuis: It's not all that relevant, by the way - we're dropping the libev dependency.

autogen.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+#fix for solaris
```
bnoordhuis, line 4: Hah, really? It doesn't auto-create the m4 directory? Old autotools?

bnoordhuis, line 4: By the way, autogen.sh should have +x set.

bnoordhuis:
It works quite well, well done. Here are the loose ends I noticed:
1. `make dist` includes build artifacts in the tarball
2. `make install` doesn't install ev.h and eio.h
3. Extra karma if you get `make check` to work. (I noticed `make dist` runs the tests so it's probably a small issue.)

configure.ac
JustinVenus: Yes I did, thank you catching that.

Justin
On Jul 27, 2012 8:24 AM, "Ben Noordhuis" <
reply@reply.github.com>
wrote:

autogen.sh
JustinVenus, line 4: Yeah I was kind of irritated about the m4 directory.  I forgot about set.

JustinVenus:
I wonder if "make dist" is calling "make test" I will investigate that over the weekend.  "make check" should be pretty easy, that was an oversight on my part.  Thank you for the feedback.

JustinVenus:
I think I have resolved all of the requested fixes in my branch.  I am going to close this request and submit another pull request in a few moments.

PR #502, opened 2012-07-25
Add MAC address when enumerating network interfaces

mscdex:


include/uv.h
```diff
@@ -1346,6 +1346,7 @@ struct uv_cpu_info_s {
 
 struct uv_interface_address_s {
   char* name;
+  unsigned char* phys_addr;
```
bnoordhuis: Consistency: use `char` here, not `unsigned char`

src/unix/darwin.c
```diff
@@ -324,6 +329,36 @@ uv_err_t uv_interface_addresses(uv_interface_address_t** addresses,
...
+    address = *addresses;
+
+    for (i = 0; i < (*count); i++) {
+      if (strcmp(address->name, ent->ifa_name) == 0) {
+        sa_addr = (struct sockaddr_dl*)(ent->ifa_addr);
+        if (memcmp(LLADDR(sa_addr), empty_phys_addr,
+                   sizeof(empty_phys_addr)) == 0) {
+          phys_addr = NULL;
+        } else {
+          phys_addr = (unsigned char*)malloc(6);
```
bnoordhuis: Superfluous cast.

src/unix/darwin.c
```diff
@@ -324,6 +329,36 @@ uv_err_t uv_interface_addresses(uv_interface_address_t** addresses,
...
+          if (!phys_addr) {
+            return uv__new_artificial_error(UV_ENOMEM);
```
bnoordhuis: Leaks memory.

src/unix/linux/linux-core.c
```diff
@@ -397,12 +397,18 @@ uv_err_t uv_interface_addresses(uv_interface_address_t** addresses,
...
+  unsigned char *sa_data;
+  unsigned char *phys_addr;
+  unsigned char empty_phys_addr[6];
   uv_interface_address_t* address;
 
   if (getifaddrs(&addrs) != 0) {
     return uv__new_sys_error(errno);
   }
 
+  bzero(empty_phys_addr, sizeof(empty_phys_addr));
```
bnoordhuis: Use memset, bzero is deprecated.

src/unix/darwin.c
```diff
@@ -324,6 +329,36 @@ uv_err_t uv_interface_addresses(uv_interface_address_t** addresses,
     address++;
   }
 
+  /* Fill in physical addresses for each interface */
+  for (ent = addrs; ent != NULL; ent = ent->ifa_next) {
+    if (!(ent->ifa_flags & IFF_UP && ent->ifa_flags & IFF_RUNNING) ||
+        (ent->ifa_addr == NULL) ||
+        (ent->ifa_addr->sa_family != AF_LINK)) {
```
bnoordhuis: This should be `(sa_family != AF_INET && sa_family != AF_INET6)`. I realize that you're not the one introducing this, though. :-)

bnoordhuis: Belay that for now, we can address it in a later commit.

bnoordhuis:
Generally LGTM but the patch makes the platform_output test segfault on Linux.

```
(gdb) run platform_output platform_output
# output snipped
Program received signal SIGSEGV, Segmentation fault.
0x000000000041dfd5 in run_test_platform_output () at ../test/test-platform-output.c:84
84                   interfaces[i].phys_addr[5]);
(gdb) list
79                   interfaces[i].phys_addr[0],
80                   interfaces[i].phys_addr[1],
81                   interfaces[i].phys_addr[2],
82                   interfaces[i].phys_addr[3],
83                   interfaces[i].phys_addr[4],
84                   interfaces[i].phys_addr[5]);
85          }
86          printf("  internal: %d\n", interfaces[i].is_internal);
87
88          if (interfaces[i].address.address4.sin_family == AF_INET) {
(gdb) backtrace full
#0  0x000000000041dfd5 in run_test_platform_output () at ../test/test-platform-output.c:84
        buffer = "192.168.1.14\000uis/src/nodejs/libuv/out/Debug/run-tests", '\000' <repeats 458 times>
        rss = 786432
        uptime = 11029.724992138001
        cpus = 0x6e0470
        interfaces = 0x6e0e70
        count = 4
        i = 2
        err = {code = UV_OK, sys_errno_ = 0}
#1  0x0000000000406d62 in run_test_part (test=0x6df1d6 "platform_output", part=0x6df1e6 "platform_output") at ../test/runner.c:298
        task = 0x6896e0
#2  0x0000000000405c85 in main (argc=3, argv=0x6df010) at ../test/run-tests.c:57
No locals.
```

src/win/util.c
```diff
@@ -837,6 +839,16 @@ uv_err_t uv_interface_addresses(uv_interface_address_t** addresses,
       assert(name);
       address->name = name;
 
+      phys_addr = NULL;
+      if (adapter_address->PhysicalAddressLength > 0) {
```
piscisaureus: Some defensive coding would be nice. Maybe make it `== 6` ?

mscdex:
@bnoordhuis I tested it on my Linux systems and I did not experience segfaults related to these commits.

src/unix/darwin.c
```diff
@@ -324,9 +331,42 @@ uv_err_t uv_interface_addresses(uv_interface_address_t** addresses,
...
+    for (i = 0; i < (*count); i++) {
+      if (strcmp(address->name, ent->ifa_name) == 0) {
+        sa_addr = (struct sockaddr_dl*)(ent->ifa_addr);
+        if (memcmp(LLADDR(sa_addr), empty_phys_addr,
+                   sizeof(empty_phys_addr)) == 0) {
+          phys_addr = NULL;
+        } else {
+          phys_addr = malloc(6);
+          if (!phys_addr) goto error;
+          memcpy(phys_addr, LLADDR(sa_addr), sizeof(phys_addr));
```
bnoordhuis: Bad sizeof(). That's what was making it crash on my 64 bits Linux install: `sizeof(char*) == 8`

bnoordhuis:
I lie by omission. There's at least one more bug that stops address->phys_addr from being set in some code paths. It's the stray pointer that causes the SIGSEGV.

mscdex:
Ok, I think that should do it now.

wolfeidau:
Is there a reason why this was never merged? 

I would really like to get access to this information as it is handy for generating UUID numbers.

mscdex:
Updated for current master branch.

chilts:
This would be fantastic.

I'm currently using @wolfeidau's netif module to get the MAC address of an interface for my flake module. He's awesome, but having it in core would be awesomer. :)

include/uv.h
```diff
@@ -1475,6 +1475,7 @@ struct uv_cpu_info_s {
 
 struct uv_interface_address_s {
   char* name;
+  char* phys_addr;
```
bnoordhuis: I suggest replacing that with `char phys_addr[6]`; saves having to juggle malloc/free calls.

src/unix/darwin.c
```diff
@@ -355,12 +356,18 @@ uv_err_t uv_interface_addresses(uv_interface_address_t** addresses,
...
+  struct sockaddr_dl *sa_addr;
+  char *phys_addr;
+  char empty_phys_addr[6];
   uv_interface_address_t* address;
 
   if (getifaddrs(&addrs) != 0) {
     return uv__new_sys_error(errno);
   }
 
+  memset(empty_phys_addr, 0, 6);
```
bnoordhuis: Use `sizeof(empty_phys_addr)` for this (here and everywhere.)

include/uv.h
bnoordhuis: Oh, and zero the memory on error and in the sunos/freebsd/etc. uv_interface_addresses() implementations.

mscdex:
Ok, that ought to do it. I haven't tested on sunos because I don't have that available to me, but the implementation should be right.

src/unix/darwin.c
```diff
@@ -425,6 +431,31 @@ uv_err_t uv_interface_addresses(uv_interface_address_t** addresses,
...
+        (ent->ifa_addr->sa_family != AF_LINK)) {
+      continue;
+    }
+
+    address = *addresses;
+
+    for (i = 0; i < (*count); i++) {
+      if (strcmp(address->name, ent->ifa_name) == 0) {
+        sa_addr = (struct sockaddr_dl*)(ent->ifa_addr);
+        if (memcmp(LLADDR(sa_addr), empty_phys_addr,
```
bnoordhuis: Style: each argument on its own line.

src/unix/darwin.c
```diff
@@ -425,6 +431,31 @@ uv_err_t uv_interface_addresses(uv_interface_address_t** addresses,
...
+                   sizeof(empty_phys_addr)) != 0) {
+          memcpy(address->phys_addr, LLADDR(sa_addr),
+                 sizeof(address->phys_addr));
+        } else {
+          memset(address->phys_addr, 0, sizeof(address->phys_addr));
```
bnoordhuis: Logic bug. Hoist it out of the loop. (Just clear phys_addr before you enter the loop, it's only 6 bytes of memory.)

src/unix/freebsd.c
```diff
@@ -331,13 +335,115 @@ void uv_free_cpu_info(uv_cpu_info_t* cpu_infos, int count) {
...
+        (ent->ifa_addr == NULL) ||
+        (ent->ifa_addr->sa_family == AF_LINK)) {
+      continue;
+    }
+
+    (*count)++;
+  }
+
+  *addresses = (uv_interface_address_t*)
+    malloc(*count * sizeof(uv_interface_address_t));
```
bnoordhuis: Style: superfluous cast. Second, use `sizeof(**addresses)`.

bnoordhuis: Another thing is that it's somewhat of a best practice to not modify an out argument until function exit (usually successful exit.)

Libuv is not 100% consistent in that regard but a lot of the code that clobbers out arguments all over the place was carried over from node v0.4.

src/unix/freebsd.c
```diff
@@ -331,13 +335,115 @@ void uv_free_cpu_info(uv_cpu_info_t* cpu_infos, int count) {
...
+    return uv__new_artificial_error(UV_ENOMEM);
+  }
+
+  address = *addresses;
+
+  for (ent = addrs; ent != NULL; ent = ent->ifa_next) {
+    bzero(&ip, sizeof (ip));
+    if (!(ent->ifa_flags & IFF_UP && ent->ifa_flags & IFF_RUNNING)) {
+      continue;
+    }
```
bnoordhuis: Style (minor): it's allowed to drop the braces from single-statement if statements.

src/unix/freebsd.c
```diff
@@ -331,13 +335,115 @@ void uv_free_cpu_info(uv_cpu_info_t* cpu_infos, int count) {
...
+      address->address.address4 = *((struct sockaddr_in*) ent->ifa_addr);
+    }
+
+    if (ent->ifa_netmask->sa_family == AF_INET6) {
+      address->netmask.netmask6 = *((struct sockaddr_in6*) ent->ifa_netmask);
+    } else {
+      address->netmask.netmask4 = *((struct sockaddr_in*) ent->ifa_netmask);
+    }
+
+    address->is_internal = ent->ifa_flags & IFF_LOOPBACK ? 1 : 0;
```
bnoordhuis: Parentheses around `ent->ifa_flags & IFF_LOOPBACK`

src/unix/linux-core.c
```diff
@@ -681,6 +686,30 @@ uv_err_t uv_interface_addresses(uv_interface_address_t** addresses,
...
+        (ent->ifa_addr->sa_family != PF_PACKET)) {
+      continue;
+    }
+
+    address = *addresses;
+
+    for (i = 0; i < (*count); i++) {
+      if (strcmp(address->name, ent->ifa_name) == 0) {
+        sa_data = ((struct sockaddr*)(ent->ifa_addr))->sa_data;
+        sa_data += 10;
```
bnoordhuis: Magic value? What are you skipping over?

src/unix/openbsd.c
```diff
@@ -292,13 +296,108 @@ void uv_free_cpu_info(uv_cpu_info_t* cpu_infos, int count) {
...
+      address->address.address4 = *((struct sockaddr_in*) ent->ifa_addr);
+    }
+
+    if (ent->ifa_netmask->sa_family == AF_INET6) {
+      address->netmask.netmask6 = *((struct sockaddr_in6*) ent->ifa_netmask);
+    } else {
+      address->netmask.netmask4 = *((struct sockaddr_in*) ent->ifa_netmask);
+    }
+
+    address->is_internal = !!(ent->ifa_flags & IFF_LOOPBACK) ? 1 : 0;
```
bnoordhuis: Style: superfluous ternary operator. The result of !!expr is always 0 or 1.

bnoordhuis:
One more round of comments but most of them are style issues. It's shaping up nicely, Brian.

@piscisaureus You should take a look at the Windows code.

piscisaureus:
@mscdex I looked at the windows changes, but I can't comment on lines atm, something seems broken about github.

There's two changes in src/win/util/c:
- `uv_address_buf_size += 6;` -> this line is unnecessary
- https://github.com/mscdex/libuv/blob/2a318295513bca3b944f3f055178f71e8eca9ab0/src/win/util.c#L943-L944 -> adhere to the style guide, put all arguments on their own line and not just some.

src/unix/openbsd.c
mscdex: I had just left all of the non-mac address stuff as-is when copying from the darwin implementation. I can change it though.

src/unix/freebsd.c
mscdex: This was also something that was just copied as-is from the darwin implementation, I will change it as well.

src/unix/freebsd.c
mscdex: This was also copied from darwin. I will change it.

src/unix/darwin.c
mscdex: I thought it might be frowned upon if I zeroed it out first, only to write over it again with an actual address in case there is a physical address.

src/unix/linux-core.c
mscdex: I can't recall where now, but I had found that method on some page or project source code.

After some more research, it looks like there's a less hacky method available with sockaddr_ll.

bnoordhuis:
Landed the freebsd dtrace fix in c8ffee3.

src/unix/freebsd.c
```diff
@@ -331,13 +335,96 @@ void uv_free_cpu_info(uv_cpu_info_t* cpu_infos, int count) {
 
 uv_err_t uv_interface_addresses(uv_interface_address_t** addresses,
   int* count) {
-  /* TODO: implement */
-  *addresses = NULL;
+  struct ifaddrs *addrs, *ent;
+  char ip[INET6_ADDRSTRLEN];
```
bnoordhuis: Not used, only bzero'd.

src/unix/freebsd.c
```diff
@@ -331,13 +335,96 @@ void uv_free_cpu_info(uv_cpu_info_t* cpu_infos, int count) {
 
 uv_err_t uv_interface_addresses(uv_interface_address_t** addresses,
   int* count) {
-  /* TODO: implement */
-  *addresses = NULL;
+  struct ifaddrs *addrs, *ent;
```
bnoordhuis, line 329: Style issue: `type*` rather than `type *`.  (I know you copied this from elsewhere but still.)

src/unix/freebsd.c
```diff
@@ -331,13 +335,96 @@ void uv_free_cpu_info(uv_cpu_info_t* cpu_infos, int count) {
...
+        (ent->ifa_addr->sa_family != AF_LINK)) {
+      continue;
+    }
+
+    address = *addresses;
+
+    for (i = 0; i < (*count); i++) {
+      if (strcmp(address->name, ent->ifa_name) == 0) {
+        sa_addr = (struct sockaddr_dl*)(ent->ifa_addr);
+        memcpy(address->phys_addr, LLADDR(sa_addr), sizeof(address->phys_addr));
```
bnoordhuis, line 402: Correct me if I'm wrong but can't you fold this into the previous loop over addrs? The one with the bzero in it?

src/unix/openbsd.c
```diff
@@ -292,13 +296,94 @@ void uv_free_cpu_info(uv_cpu_info_t* cpu_infos, int count) {
...
+
+  *addresses = malloc(*count * sizeof(**addresses));
+
+  if (!(*addresses))
+    return uv__new_artificial_error(UV_ENOMEM);
+
+  address = *addresses;
+
+  for (ent = addrs; ent != NULL; ent = ent->ifa_next) {
+    if (!(ent->ifa_flags & IFF_UP && ent->ifa_flags & IFF_RUNNING))
```
bnoordhuis: Always put parentheses around `foo & bar` expressions when they're used inside more complex expressions.

src/unix/freebsd.c
mscdex, line 402: It's a similar thing with the Linux implementation. I don't think we can assume the order of the results from getifaddrs() (e.g. link-level information will always come before/after INET\* information).

piscisaureus:
The windows parts LGTM now.

mscdex:
Updated commits as per Ben's suggestions.

rvagg:
ping! would be great if we could get this in soon.

bnoordhuis:
This will need some rework after commit 3ee4d3f ("return error codes directly").

mscdex:
@bnoordhuis Updated.

bnoordhuis:
Thanks Brian, landed in e3a657c.

PR #504, opened 2012-07-25
Improved VS detection. Fix for GYP_MSVS_VERSION

gigi81:
Added visual studio 2010/2008 detection and set also the GYP_MSVS_VERSION bsed on the version detected


mscdex:
+100

I can't tell you how many times I've opened up a command prompt to use vcbuild, gotten errors of one kind or another (because GYP_MSVS_VERSION wasn't set), and then had to go and look up the environment variable name from one of the libuv github issues about the problem.

Yes, one could set a permanent environment variable, but it's better if the build tool creates the temporary variable since this is only a gyp-specific (and gyp version-specific too) issue.

piscisaureus:
I understand that the GYP_MSVS_VERSION fix is important, but what's the point of detecting vs2008 if we can't build with it?

gigi81:
The point is that is that it should be possible to build also with visual studio 2008 if someone hasn't 2010. There is no particular reason why it shouldn't build (at least that I'm aware of). The only problem from what I understood is with the inttypes http://code.google.com/p/msinttypes/. So this can be easily fixed. But it has nothing to do with the build script that it should just be smart enough to detect the latest available system compiler.

It's also a good example on how to add 2012 when it will be supported

piscisaureus:
LGTM.
@gigi81, can you sign the CLA? http://nodejs.org/cla.html

For the record, I'd take a patch that adds VS2008 support

piscisaureus:
ping @gigi81

gigi81:
I'm on it. Waiting for my project manager to approve it.

gigi81:
I signed and sent the cla. You can merge it now...

piscisaureus:
Thanks. I changed it a little bit to make it a little simpler. Also, please take care to not mix tabs and spaces next time.
I landed this patch together with #514 in 5bfb7c9.

Unfortunately the test runner doesn't link when a static library is being built, but this is not a regression.

PR #508, opened 2012-07-30
A small fix for solaris11

JustinVenus:
This allows me to remove a hack that I have in place for sunpro compiler support that is coming with the eventual resolution of issue486.


piscisaureus:
This should be fixed in 4eccb2ee525d5d5872ce185b59ba4f48f27af9a4

PR #514, opened 2012-07-31
Improved vcbuild with shared/static and x86/x64

gigi81:
Improved the vcbuild script adding support for the following flags:
- shared/static
- x86/x64

Now to build the uv.dll you can write: vcbuild release x86 shared

Note: the test program has some linking issues when building the DLL version of the library. Anyway the DLL itself builds fine.


common.gypi
```diff
@@ -115,6 +115,14 @@
           # POSIX names
           '_CRT_NONSTDC_NO_DEPRECATE',
         ],
+        'target_conditions': [
+          ['library=="shared_library"', {
+            'defines': ['BUILDING_UV_SHARED=1'],
+          }],
+		  ['target_arch=="x64"', {
+			'msvs_configuration_platform': 'x64'
```
bnoordhuis, line 123: Indentation.

PR #515, opened 2012-08-04
unix: Add fs truncate

isaacs:
This is lacking the windows portion.

Related: joyent/node#3805


bnoordhuis:
I don't really want to land this, `truncate()` is pretty unsafe in a concurrent environment. Sensible people use a combination of `open()` and `ftruncate()`.

@piscisaureus Thoughts?

isaacs:
Yeah, we should probably just simulate it in js, especially since
open/ftruncate will be required on windows anyway, effectively.

PR #519, opened 2012-08-07
Including <uv.h> shouldn't emit any warnings

trondn:
I'm trying to add a plugin that use libuv to my library and I'm normally compiling my modules with:

 gcc -DHAVE_CONFIG_H -I. -I./src -pipe -I./include -DHAVE_VISIBILITY=1 -fvisibility=hidden -Wall -pedantic -Wundef -Wshadow -fdiagnostics-show-option -Wformat -fno-strict-aliasing -Wno-strict-aliasing -Wextra -Werror -fprofile-arcs -ftest-coverage -pipe -std=gnu99 -Wstrict-prototypes -Wmissing-prototypes -Wredundant-decls -Wmissing-declarations -Wcast-align 

The constructs:
# if variable

will emit a warning if it's not defined, causing my build to break. The attached patch change these to #ifdef (this means that defining them to "0" instead of undef'ing them no longer works.. The alternative would be to extend the code check if they're defined _AND_ set to something else than 0...

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -53,6 +53,10 @@ extern "C" {
 #include <stdint.h> /* int64_t */
 #include <sys/types.h> /* size_t */
 
+#if defined(__SVR4) && !defined(__unix__)
+#define __unix__
+#endif
+
 #if defined(__unix__) || defined(__POSIX__) || defined(__APPLE__)
 # include "uv-private/uv-unix.h"
 #else
--- a/include/uv-private/eio.h
+++ b/include/uv-private/eio.h
@@ -251,7 +251,7 @@ struct eio_req
 
   eio_channel *channel; /* data used to direct poll callbacks arising from this req */
 
-#if __i386 || __amd64
+#if defined(__i386) || defined(__amd64)
   unsigned char cancelled;
 #else
   sig_atomic_t cancelled;
--- a/include/uv-private/ngx-queue.h
+++ b/include/uv-private/ngx-queue.h
@@ -62,7 +62,7 @@ struct ngx_queue_s {
     (q)->prev
 
 
-#if (NGX_DEBUG)
+#if defined(NGX_DEBUG)
 
 #define ngx_queue_remove(x)                                                   \
     (x)->next->prev = (x)->prev;                                              \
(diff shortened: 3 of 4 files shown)


include/uv.h
```diff
@@ -53,7 +53,7 @@
 #include <stdint.h> /* int64_t */
 #include <sys/types.h> /* size_t */
 
-#if defined(__unix__) || defined(__POSIX__) || defined(__APPLE__)
+#if defined(__unix__) || defined(__POSIX__) || defined(__APPLE__) || defined(__SVR4)
```
bnoordhuis: Wrap at 80 chars.

bnoordhuis:
I don't really want to pull in this patch. `#if FOO` has been used since the dawn of time.

piscisaureus:
@bnoordhuis I don't see much harm in it. However, I am surprised that it triggers a warning. It never did for me (and I tried many platforms).

bnoordhuis:
It only happens in gcc's overzealous mode, -Wundef, disabled by default because of that 'since the dawn of time' thing.

trondn:
Given that this is a "public" header for the library it would be nice if it didn't generate warnings causing building of clients of the library to fail.. I can fix the line wrapping, but if the patches won't be accepted I'd rather just skip adding support for libuv for libcouchbase given that it'll require me to keep me own branch of libuv in order to make it compile (making upgrading stuff etc harder).

ingenthr:
Since uv.h is something you want others to consume, and since others may want 'overzealous mode' for their projects, it seems like there's a good argument for including this change.  What's the downside?  It doesn't seem less readable to me.

bnoordhuis:
Trond, can you sign [the CLA](http://nodejs.org/cla.html)?

trondn:
I've just signed the CLA and submitted it on email. Please let me know if there is anything else I need to do.

bnoordhuis:
@isaacs Where do emailed CLAs end up? It's not in the list.

isaacs:
@trondn Where did you email the CLA?

isaacs:
@trondn Nevermind.  Yes.  Confirmed.

@bnoordhuis I'll add it to the spreadsheet.  @trondn is good.  Proceed as normal.

bnoordhuis:
Thanks Trond, landed in cbb9300 and 5143d54.

PR #520, opened 2012-08-09
implement uv_interface_addresses() & uv_free_interface_addresses() on freebsd.

vanillahsu:
darwin & freebsd use same codes to get interface information,

so I copy most code from darwin.c to implement uv_interface_addresses() & uv_free_interface_addresses()


bnoordhuis:
You mean it's a verbatim copy of the stuff in darwin.c? In that case it should be moved into a common file, I don't want it duplicated.

vanillahsu:
getifaddrs(3) available on darwin/freebsd/netbsd/openbsd,

I agreed to move to a common file.

bnoordhuis:
No reply in 4 months, closing.

PR #523, opened 2012-08-10
gai errno mapping

mnunberg:
This makes it easier to utilize the UV_ERRNO_MAP

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -68,7 +68,7 @@ extern "C" {
   XX( -1, UNKNOWN, "unknown error")                                           \
   XX(  0, OK, "success")                                                      \
   XX(  1, EOF, "end of file")                                                 \
-  XX(  2, EADDRINFO, "getaddrinfo error")                                     \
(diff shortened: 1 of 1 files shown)


bnoordhuis:
> This makes it easier to utilize the UV_ERRNO_MAP

This needs some further elaboration. Where/when does it make things easier?

mnunberg:
It's useful when trying to map back uv error codes to their system equivalents.. (there is the `sys_errno_` field which is marked as private...)

See for example https://github.com/mnunberg/couchnode/blob/master/io/common.c#L330 which needs to report the system errno equivalent...

bnoordhuis:
I'm not completely adverse against landing this patch but you should update `src/unix/core.c` and `src/win/getaddrinfo.c` to use the new error codes. Can you sign [the CLA](http://nodejs.org/cla.html) if you haven't yet?

bnoordhuis:
Stale, closing.

PR #528, opened 2012-08-16
wip

indutny:
wip


include/uv-private/uv-unix.h
```diff
@@ -186,6 +188,28 @@ struct uv__io_s {
...
+  uv__io_t read_watcher;                                                      \
+  uv__io_t write_watcher;                                                     \
+  ngx_queue_t write_queue;                                                    \
+  ngx_queue_t write_completed_queue;                                          \
+  uv_connection_cb connection_cb;                                             \
+  int delayed_error;                                                          \
+  int accepted_fd;                                                            \
+  int fd;                                                                     \
+
+#endif
```
bnoordhuis: Grep for UV_LOOP_PRIVATE_PLATFORM_FIELDS and use that approach instead. Duplicating the fields is bound to break later on.

src/unix/stream.c
```diff
@@ -34,6 +34,11 @@
 #include <sys/un.h>
 #include <unistd.h>
 
+#if defined(__APPLE__)
+#include <sys/event.h>
```
bnoordhuis: `# include <...>`

src/unix/stream.c
```diff
@@ -69,11 +74,139 @@ void uv__stream_init(uv_loop_t* loop,
   ngx_queue_init(&stream->write_completed_queue);
   stream->write_queue_size = 0;
 
+#if defined(__APPLE__)
+  stream->osx_hack = 0;
```
bnoordhuis: Give this field a better name.

src/unix/stream.c
```diff
@@ -69,11 +74,139 @@ void uv__stream_init(uv_loop_t* loop,
...
+
   uv__io_init(&stream->read_watcher, uv__stream_io, -1, 0);
   uv__io_init(&stream->write_watcher, uv__stream_io, -1, 0);
 }
 
 
+#if defined(__APPLE__)
+void uv__stream_osx_select(void* arg) {
+  uv_stream_t* stream = (uv_stream_t*) arg;
+  fd_set read, write, error;
```
bnoordhuis: Assign after declaration, put each fd_set on its own line.

src/unix/stream.c
```diff
@@ -69,11 +74,139 @@ void uv__stream_init(uv_loop_t* loop,
...
+    FD_ZERO(&read);
+    FD_ZERO(&write);
+    FD_ZERO(&error);
+    FD_SET(stream->fd, &read);
+    FD_SET(stream->fd, &write);
+    FD_SET(stream->fd, &error);
+
+    timeout.tv_sec = 0;
+    timeout.tv_usec = 250000; /* 250 ms timeout */
+    r = select(stream->fd + 1, &read, &write, &error, &timeout);
```
bnoordhuis: Handle EINTR here.

src/unix/stream.c
```diff
@@ -69,11 +74,139 @@ void uv__stream_init(uv_loop_t* loop,
...
+    if (FD_ISSET(stream->fd, &read)) {
+      stream->select_events |= UV__IO_READ;
+    }
+
+    if (FD_ISSET(stream->fd, &write)) {
+      stream->select_events |= UV__IO_WRITE;
+    }
+
+    if (FD_ISSET(stream->fd, &error)) {
+      stream->select_events |= UV__IO_ERROR;
```
bnoordhuis: This critical section is not really critical but... you'd normally aggregate the FD_ISSET checks outside of the critical section and then set them once with `stream->select_events |= events`.

src/unix/stream.c
```diff
@@ -69,11 +74,139 @@ void uv__stream_init(uv_loop_t* loop,
...
+    }
+
+    if (stream->select_events != 0) {
+      uv_async_send(stream->select_cb);
```
bnoordhuis: Which would let you move the `uv_async_send()` out of the critical section, it seems.

include/uv-private/uv-unix.h
```diff
@@ -186,6 +188,28 @@ struct uv__io_s {
   int delayed_error;                                                          \
   int accepted_fd;                                                            \
   int fd;                                                                     \
+  int osx_hack;                                                               \
+  uv_thread_t select_thread;                                                  \
+  uv_sem_t select_sem;                                                        \
+  uv_mutex_t select_mutex;                                                    \
+  uv_async_t* select_cb;                                                       \
```
bnoordhuis: Bad name, use something like `select_async`.

src/unix/stream.c
```diff
@@ -69,11 +74,139 @@ void uv__stream_init(uv_loop_t* loop,
...
+}
+
+
+int uv__stream_osx_hack(uv_stream_t* stream, int fd) {
+  /*
+   * kqueue doesn't work with some files from /dev mount on osx.
+   * select(2) in separate thread for those fds
+   */
+
+  int kq = kqueue();
```
bnoordhuis: Assignment after declaration. Also needs to handle a -1 return value (because of e.g. EMFILE).

src/unix/stream.c
```diff
@@ -69,11 +74,139 @@ void uv__stream_init(uv_loop_t* loop,
...
+  ret = kevent(kq, filter, 1, events, 1, &timeout);
+  close(kq);
+  if (ret < 1) return 0;
+  if ((events[0].flags & EV_ERROR) == 0 || events[0].data != EINVAL) {
+    return 0;
+  }
+
+  /* At this point we definitely know that this fd won't work with kqueue */
+  stream->osx_hack = 1;
+  stream->select_cb = malloc(sizeof(uv_async_t));
```
bnoordhuis: `malloc(sizeof(*stream->select_cb))` - which you will have renamed to select_async or some such :-)

src/unix/stream.c
```diff
@@ -69,11 +74,139 @@ void uv__stream_init(uv_loop_t* loop,
...
+  if (stream->select_cb == NULL) {
+    /* TODO: Return error */
+    abort();
+  }
+  uv_async_init(stream->loop, stream->select_cb, uv__stream_osx_select_cb);
+  stream->select_cb->data = stream;
```
bnoordhuis: Preferably embed it in another struct and use container_of so you don't have to go pointer chasing.

src/unix/stream.c
```diff
@@ -69,11 +74,139 @@ void uv__stream_init(uv_loop_t* loop,
...
+  uv_sem_init(&stream->select_sem, 0);
+  uv_mutex_init(&stream->select_mutex);
+  uv_thread_create(&stream->select_thread, uv__stream_osx_select, stream);
```
bnoordhuis: Check the return values.

src/unix/stream.c
```diff
@@ -69,11 +74,139 @@ void uv__stream_init(uv_loop_t* loop,
...
+  }
+}
+
+
+void uv__stream_osx_cb_close(uv_handle_t* async) {
+  free(async);
+}
+
+
+int uv__stream_osx_hack(uv_stream_t* stream, int fd) {
```
bnoordhuis: Return 0 on success, -1 on error.

bnoordhuis:
Lots of remarks but the concept seems sound enough to me. :-)

src/unix/stream.c
indutny: No way, uv_async_t is declared after uv_stream_t

include/uv-private/uv-unix.h
```diff
@@ -175,6 +175,24 @@ struct uv__io_s {
   uv_handle_t* next_closing;                                                  \
 
 
+#if defined(__APPLE__)
+
+# define UV_STREAM_PRIVATE_PLATFORM_FIELDS                                    \
```
piscisaureus: doesn't this waste a big chunk of memory per stream on os x?

indutny: you're right, I can allocate it on-demand

include/uv-private/uv-unix.h
```diff
@@ -175,6 +175,31 @@ struct uv__io_s {
...
+
+/* Forward declaration */
+struct uv_async_s;
+typedef struct uv__stream_select_s uv__stream_select_t;
+
+struct uv__stream_select_s {
+  uv_thread_t thread;
+  uv_sem_t sem;
+  uv_mutex_t mutex;
+  struct uv_async_s* async;
```
piscisaureus: Why not embed this?

indutny: It's unknown at the moment of declaration of this structure

piscisaureus: It's not possible to shuffle a little bit?

src/unix/stream.c
```diff
@@ -69,11 +74,182 @@ void uv__stream_init(uv_loop_t* loop,
...
+    /* TODO: Return error */
+    abort();
+  }
+
+  s->async = malloc(sizeof(*s->async));
+  if (s->async == NULL) {
+    /* TODO: Return error */
+    abort();
+  }
+  if (uv_async_init(stream->loop,
```
piscisaureus: You should unref() the async.

src/unix/stream.c
```diff
@@ -69,11 +74,182 @@ void uv__stream_init(uv_loop_t* loop,
...
+  if (uv_mutex_init(&s->mutex)) goto fatal2;
+
+  s->fake_fd = socket(PF_LOCAL, SOCK_STREAM, 0);
+  if (s->fake_fd == -1) goto fatal3;
+
+  if (uv_thread_create(&s->thread, uv__stream_osx_select, stream)) {
+    goto fatal4;
+  }
+
+  /* Replace stream's fd with fake (non-failing fd) */
```
piscisaureus: Why is this needed?

src/unix/stream.c
```diff
@@ -69,11 +74,182 @@ void uv__stream_init(uv_loop_t* loop,
...
+  EV_SET(&filter[0], fd, EVFILT_READ, EV_ADD | EV_ENABLE, 0, 0, 0);
+
+  /* Use small timeout, because we only want to capture EINVALs */
+  timeout.tv_sec = 0;
+  timeout.tv_nsec = 1;
+
+  ret = kevent(kq, filter, 1, events, 1, &timeout);
+  close(kq);
+  if (ret < 1) return 0;
+  if ((events[0].flags & 0x4000) == 0 || events[0].data != EINVAL) {
```
bnoordhuis: What is that hex value doing there?

src/unix/stream.c
bnoordhuis: And mark it internal.

include/uv-private/uv-unix.h
bnoordhuis: Fedor, you can turn the field in `UV_STREAM_PRIVATE_PLATFORM_FIELDS` into a `void*` and move the declaration of `uv__stream_select_s` into src/unix/stream.c.

src/unix/stream.c
indutny: `EV_ERROR` declared in <sys/events.h> is overwritten by ev.h, I'm using it's original value there. Probably better add constant or define for it?

src/unix/stream.c
indutny: Because io watcher will fail with error and stop watching when given original fd.

indutny: And comment is old and spooky there ;)

indutny:
Everything resolved.

src/unix/stream.c
```diff
@@ -34,6 +34,26 @@
...
+#define EV_ERROR_ORIG 0x4000
+
+/* Forward declaration */
+typedef struct uv__stream_select_s uv__stream_select_t;
+
+struct uv__stream_select_s {
+  uv_thread_t thread;
+  uv_sem_t sem;
+  uv_mutex_t mutex;
+  uv_async_t* async;
```
bnoordhuis: Why don't you embed the uv_async_t?

src/unix/stream.c
```diff
@@ -69,11 +89,188 @@ void uv__stream_init(uv_loop_t* loop,
...
+  fd_set read;
+  fd_set write;
+  fd_set error;
+  struct timeval timeout;
+  int events;
+  int fd;
+  int r;
+
+  stream = (uv_stream_t*) arg;
+  s = (uv__stream_select_t*)stream->select;
```
bnoordhuis: Superfluous casts.

src/unix/stream.c
```diff
@@ -69,11 +89,188 @@ void uv__stream_init(uv_loop_t* loop,
...
+    FD_ZERO(&write);
+    FD_ZERO(&error);
+    FD_SET(fd, &read);
+    FD_SET(fd, &write);
+    FD_SET(fd, &error);
+
+    timeout.tv_sec = 0;
+    timeout.tv_usec = 250000; /* 250 ms timeout */
+    r = select(fd + 1, &read, &write, &error, &timeout);
+    if (r < 0) {
```
bnoordhuis: Preferably use `r == -1` (minor style issue).

src/unix/stream.c
```diff
@@ -974,6 +1178,25 @@ int uv_is_writable(const uv_stream_t* stream) {
 
 
 void uv__stream_close(uv_stream_t* handle) {
+#if defined(__APPLE__)
+  /* Terminate select loop first */
+  if (handle->select != NULL) {
+    uv__stream_select_t* s;
+
+    s = (uv__stream_select_t*)handle->select;
```
bnoordhuis: Superfluous cast.

indutny:
Fixed everything.

src/unix/stream.c
```diff
@@ -69,11 +90,182 @@ void uv__stream_init(uv_loop_t* loop,
...
+}
+
+
+void uv__stream_osx_cb_close(uv_handle_t* async) {
+  /* Free container */
+  free(container_of(async, uv__stream_select_t, async));
+}
+
+
+int uv__stream_needs_select(uv_stream_t* stream, int fd) {
```
bnoordhuis: Convention: return -1 on error, 0 on success.

src/unix/stream.c
```diff
@@ -69,11 +90,182 @@ void uv__stream_init(uv_loop_t* loop,
...
+    return 0;
+  }
+  s->async.flags |= UV__HANDLE_INTERNAL;
+  uv__handle_unref((uv_handle_t*) &s->async);
+
+  if (uv_sem_init(&s->sem, 0)) goto fatal1;
+  if (uv_mutex_init(&s->mutex)) goto fatal2;
+
+  /* Create fake fd for io watcher */
+  s->fake_fd = socket(PF_LOCAL, SOCK_STREAM, 0);
```
bnoordhuis: s/PF_LOCAL/AF_UNIX/ - it's the same thing but we use AF_UNIX everywhere else.

src/unix/stream.c
bnoordhuis: Not fixed.

src/unix/stream.c
indutny: Well, I need to rename it then.

src/unix/stream.c
```diff
@@ -69,11 +90,182 @@ void uv__stream_init(uv_loop_t* loop,
...
+  struct timeval timeout;
+  int events;
+  int fd;
+  int r;
+
+  stream = (uv_stream_t*) arg;
+  s = (uv__stream_select_t*)stream->select;
+  fd = stream->fd;
+
+  while(1) {
```
bnoordhuis: `while (1) {`

src/unix/stream.c
```diff
@@ -69,11 +90,182 @@ void uv__stream_init(uv_loop_t* loop,
...
+    /* Terminate on semaphore */
+    if (uv_sem_trywait(&s->sem) == 0) break;
```
bnoordhuis: By the way, why trywait?

indutny: Because I'm ending poll loop once semaphore was fired. And if it wasn't - I'm continuing polling

indutny:
Does it looks better now?

indutny:
Oh, pushed into another repo :) Should be good now.

bnoordhuis:
Thanks Fedor, landed in 5da380a.

PR #532, opened 2012-08-21
Support compiling with musl

idunham:
These changes allow compilation with musl; most of them are needed because <features.h> is a no-op with musl.
If you merge this, you'll probably want to revert 5cdf31e (Turn off strict ANSI C89/define _GNU_SOURCE globally).
It is needed to build on musl, but config-unix.mk is easy enough to modify.

All changes are my own, and I consider them too trivial to be copyrightable.


config-unix.mk
```diff
@@ -19,12 +19,13 @@
 # IN THE SOFTWARE.
 
 E=
-CSTDFLAG=--std=c89 -pedantic -Wall -Wextra -Wno-unused-parameter
```
piscisaureus, line 22: I am not hapoy with this change. That `--std=c89` isn't used because it looks so nice.

idunham, line 22: > > @@ -19,12 +19,13 @@
> >  # IN THE SOFTWARE.
> > 
> >  E=
> > -CSTDFLAG=--std=c89 -pedantic -Wall -Wextra -Wno-unused-parameter
> 
> I am not happy with this change. That `--std=c89` isn't used because it
> looks so nice.
> Will revert.
> That's the one I said "You can revert".
> Note that with musl, reverting it makes for an FTBFS (the headers are
> written for C99).

jfhbrook, line 22: Is it possible to detect that one is using musl, and adjust the CFLAGS dynamically?

idunham, line 22: > > @@ -19,12 +19,13 @@
> >  # IN THE SOFTWARE.
> > 
> >  E=
> > -CSTDFLAG=--std=c89 -pedantic -Wall -Wextra -Wno-unused-parameter
> 
> Is it possible to detect that one is using musl, and adjust the CFLAGS
> dynamically?
> 
> No, but it is possible to use
> make CSTDFLAG="-pedantic -Wall -Wextra -Wno-unused-parameter"
>  (just tested).
> My last commit removes the modification.

idunham:
I tracked down the issue in musl (static inline in bits/syscall.h, which doesn't work in C89 mode) and submitted patches to fix it.
As soon as that gets merged, there will be be no problems.

idunham:
OK, musl upstream has merged the fix for <syscall.h>.
In other words, this patch should now be enough to compile libuv on musl.

config-unix.mk
```diff
@@ -25,6 +25,7 @@ CPPFLAGS += -Isrc -Isrc/unix/ev
 LINKFLAGS=-lm
 
 CPPFLAGS += -D_LARGEFILE_SOURCE
+CPPFLAGS += -D_GNU_SOURCE
```
bnoordhuis, line 28: This should be part of the `ifeq (Linux,$(uname_S))` block.

src/unix/eio/config_linux.h
```diff
@@ -4,6 +4,14 @@
 #include <linux/version.h>
 #include <features.h>
 
+/* 
+ * Handle non-gnu libc versions with nothing in features.h 
+ * We have no idea what they're compatible with, so always fail.
+ */
+#ifndef __GLIBC_PREREQ
+# define __GLIBC_PREREQ(x,y) 0
```
bnoordhuis, line 12: Style: space between arguments.

src/unix/ev/ev.c
```diff
@@ -235,7 +235,7 @@
 #endif
 
 #ifndef EV_USE_CLOCK_SYSCALL
-# if __linux && __GLIBC__ >= 2
+# if __linux && defined (__GLIBC__) && __GLIBC__ >= 2
```
bnoordhuis, line 238: I don't really care about the changes to ev.c because that file is going away soon but isn't something like the below less intrusive?

``` c
#ifndef __GLIBC__
# define __GLIBC__ 0
# define __GLIBC_MINOR__ 0
#endif
```

src/unix/linux/linux-core.c
```diff
@@ -35,12 +35,15 @@
 #include <fcntl.h>
 #include <time.h>
 
-#define HAVE_IFADDRS_H 1
-#ifdef __UCLIBC__
-# if __UCLIBC_MAJOR__ < 0 || __UCLIBC_MINOR__ < 9 || __UCLIBC_SUBLEVEL__ < 32
-#  undef HAVE_IFADDRS_H
+#ifdef __GLIBC__
+# define HAVE_IFADDRS_H 1
+# ifdef __UCLIBC__
```
bnoordhuis, line 40: I suspect that this breaks uclibc builds.

bnoordhuis:
Stale PR and no follow-up in 4 months time, closing.

PR #533, opened 2012-08-22
thread: handle return values of semaphores on mac

indutny:
nt


src/unix/thread.c
```diff
@@ -175,20 +175,24 @@ int uv_sem_init(uv_sem_t* sem, unsigned int value) {
...
+  if (semaphore_signal(*sem) != KERN_SUCCESS)
     abort();
 }
 
 
 void uv_sem_wait(uv_sem_t* sem) {
-  if (semaphore_wait(*sem))
-    abort();
+  int r;
+  do {
```
bnoordhuis: Newline here, no braces in do/while block.

indutny: Fixed in force-push

bnoordhuis:
Thanks, merged in 8969df6.

Issue #339, opened 2012-03-09
Proposal: add flags to uv_run, a la ev_run


saghul:
I noticed issue #291, and I think (please, correct me if I'm wrong), that the implementation of NOWAIT in Windows would just involve passing false to the call to poll here: https://github.com/joyent/libuv/blob/master/src/win/core.c#L237 (perhaps with a new parameter to UV_LOOP_ONCE, but I didn't hink too much of that yet).

So assuming that would be added there would be uv_run, uv_run_one and uv_run_nowait (or uv_run_once_nowait). What about having something like this:

```
typedef enum {
    UV_RUN_DEFAULT = 0,
    UV_RUN_ONCE = 1,
    UV_RUN_NOWAIT = 2,
} uv_run_flags;
```

And then just a single run function:

```
int uv_run(uv_loop_t *loop, int flags)
```

Thoughts? If I get an ok I could work on a patch and submit a pull request :-)

saghul:
Changing uv_run might be too much, so maybe having 2 functions would not hurt. What about this?

```
int uv_run(uv_loop_t *loop)
int uv_run2(uv_loop_t *loop, int flags)
```

uv_run would just call uv_run2(loop, UV_RUN_DEFAULT).

If it's not a good time for this (i see 0.8 is approaching) please let me know and I'll leave this open and tackle it later :-)

Thanks,

bnoordhuis:
I won't take patches right now but maybe I'll steal your idea and apply it in the upcoming refactor.

saghul:
Cool!

/Saúl

On 05/04/2012, at 18:49, Ben Noordhuis
reply@reply.github.com
wrote:

saghul:
Now that 0.8 is out can we reconsider this? /cc @bnoordhuis @piscisaureus

bnoordhuis:
Yes. I'll add it.

piscisaureus:
Fixed in 0820be7

PR #535, opened 2012-08-23
Implemented uv_run2 (adding flags to uv_run)

saghul:
Allows for running the event loop in 3 modes:
- default: loop runs until the refcount drops to zero
- once: poll for events only once and block until one is handled
- nowait: poll for events only once but don't block if there are
  no pending events

The patch contains the Unix implementation, I'll do the Windows one after some initial feedback.

I thought changing uv_run could be a bit too much, so the new functionality is implemented as uv_run2. I did remove uv_run_once, since it's very straightforward to mock with a define, for software which runs on 0.8 and 0.9.

Related issue: #339

/cc @bnoordhuis @piscisaureus

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -221,6 +221,13 @@ typedef struct uv_cpu_info_s uv_cpu_info_t;
 typedef struct uv_interface_address_s uv_interface_address_t;
 
 
+typedef enum {
+  UV_RUN_DEFAULT = 0,
+  UV_RUN_ONCE,
+  UV_RUN_NOWAIT
+} uv_run_mode;
+
+
 /*
  * This function must be called before any other functions in libuv.
  *
@@ -244,12 +251,18 @@ UV_EXTERN uv_loop_t* uv_default_loop(void);
 UV_EXTERN int uv_run(uv_loop_t*);
 
 /*
- * Poll for new events once. Note that this function blocks if there are no
- * pending events. Returns zero when done (no active handles or requests left),
- * or non-zero if more events are expected (meaning you should call
- * uv_run_once() again sometime in the future).
+ * This function runs the event loop. It will act differently depending on the
+ * specified mode:
+ *  - UV_RUN_DEFAULT: Runs the event loop until the reference count drops to
+ *    zero. Always returns zero.
+ *  - UV_RUN_ONCE: Poll for new events once. Note that this function blocks if
+ *    there are no pending events. Returns zero when done (no active handles
+ *    or requests left), or non-zero if more events are expected (meaning you
+ *    should run the event loop again sometime in the future).
+ *  - UV_RUN_NOWAIT: Poll for new events once but don't block if there are no
+ *    pending events.
  */
-UV_EXTERN int uv_run_once(uv_loop_t*);
+UV_EXTERN int uv_run2(uv_loop_t*, uv_run_mode mode);
 
 /*
  * Manually modify the event loop's reference count. Useful if the user wants
--- a/src/unix/core.c
+++ b/src/unix/core.c
@@ -267,27 +267,25 @@ int uv_backend_timeout(const uv_loop_t* loop) {
 }
 
 
-static int uv__run(uv_loop_t* loop) {
-  uv_update_time(loop);
-  uv__run_timers(loop);
-  uv__run_idle(loop);
-  uv__run_prepare(loop);
-  uv__run_pending(loop);
-  uv__io_poll(loop, uv_backend_timeout(loop));
-  uv__run_check(loop);
-  uv__run_closing_handles(loop);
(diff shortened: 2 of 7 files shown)


satoruhiga:
This pull request is still alive?
I want to use libuv in a single thread run loop.. (e.g. OpenGL application)

saghul:
This will be reworked soon, once libev is removed from libuv.

satoruhiga:
OK, Thanks! I'm really looking forward!

txdv:
libev is removed from libuv, go implement it!

saghul:
Ok, I rebased the code to work with latest master. Mr Travis says it's all good :-) This is just the unix side if of things, care to review it @bnoordhuis? I'll go for the windows side afterwards.

Btw, should we keep uv_run_once around? Perhaps with that new UV_DEPRECATED?

include/uv.h
```diff
@@ -244,12 +251,18 @@
...
+ *    zero. Always returns zero.
+ *  - UV_RUN_ONCE: Poll for new events once. Note that this function blocks if
+ *    there are no pending events. Returns zero when done (no active handles
+ *    or requests left), or non-zero if more events are expected (meaning you
+ *    should run the event loop again sometime in the future).
+ *  - UV_RUN_NOWAIT: Poll for new events once but don't block if there are no
+ *    pending events.
  */
-UV_EXTERN int uv_run_once(uv_loop_t*);
+UV_EXTERN int uv_run2(uv_loop_t*, uv_run_flags flags);
```
bnoordhuis: I suggest renaming `uv_run_flags flags` to `uv_run_mode mode`. 'flags' suggests it's possible to OR them together.

src/unix/core.c
```diff
@@ -267,27 +267,25 @@ int uv_backend_timeout(const uv_loop_t* loop) {
...
+    uv__run_timers(loop);
+    uv__run_idle(loop);
+    uv__run_prepare(loop);
+    uv__run_pending(loop);
+    uv__io_poll(loop, (flags & UV_RUN_NOWAIT ? 0 : uv_backend_timeout(loop)));
+    uv__run_check(loop);
+    uv__run_closing_handles(loop);
+    r = uv__has_active_handles(loop) || uv__has_active_reqs(loop);
+  } while (r && !(flags & (UV_RUN_ONCE | UV_RUN_NOWAIT)));
+  return (flags & UV_RUN_DEFAULT ? 0 : r);
```
bnoordhuis: Superfluous check? r should always be zero with UV_RUN_DEFAULT, right?

saghul: Actually, yes. All tests assert after uv_run returns, so we'll catch problems if something goes wrong anyway.

bnoordhuis:
Patch for uv.gyp:

``` diff
diff --git a/uv.gyp b/uv.gyp
index 9368329..2eda03f 100644
--- a/uv.gyp
+++ b/uv.gyp
@@ -275,6 +275,7 @@
         'test/test-poll-close.c',
         'test/test-process-title.c',
         'test/test-ref.c',
+        'test/test-run-nowait.c',
         'test/test-run-once.c',
         'test/test-semaphore.c',
         'test/test-shutdown-close.c',
```

saghul:
Updated addressing comments by @bnoordhuis, thanks!

bnoordhuis:
LGTM. Will you be adding the Windows side?

saghul:
Yes, I'm on it, though I'm a bit out of my confort zone ;-)

src/win/core.c
```diff
@@ -252,57 +252,47 @@ static void uv_poll_ex(uv_loop_t* loop, int block) {
...
+
+  if (pGetQueuedCompletionStatusEx)
+    poll = &uv_poll_ex;
+  else:
+    poll = &uv_poll;
+
+  r = UV_LOOP_ALIVE(loop);
+  while (r) {
+    uv_update_time((loop));
+    uv_process_timers((loop));
```
bnoordhuis: Extraneous parentheses.

saghul:
Added Windows counterpart now. @piscisaureus can you please review?

PD: @bnoordhuis I removed the extra parens.

piscisaureus:
Thanks @saghul, fixed and landed in 0820be7.

saghul:
Awesome, thanks!

On Tuesday, December 18, 2012, Bert Belder wrote:

> Thanks @saghul https://github.com/saghul, fixed and landed in 0820be7https://github.com/joyent/libuv/commit/0820be7
> .
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/joyent/libuv/pull/535#issuecomment-11490802.

## 

/Saúl
http://saghul.net | http://sipdoc.net

PR #537, opened 2012-08-24
windows: bug fixes of uv_signal_init/start and new support of signal handling in uv_kill()

shigeki:
This PR has the following three commits that have two bug fixes and one new feature. 
- bug fix: `uv__signal_control_handler_refs` needs to be incremented when multiple signal handlers are started.
- bug fix: `uv__handle_start()` was not needed in `uv_signal_init()`
- new feature: support signal handling in uv_kill(). This is nice to have consistent behaviors between Windows and Unix.


bnoordhuis:
Generally LGTM.

/cc @piscisaureus

piscisaureus:
@bnoordhuis @shigeki The fixes look good to me, but the `uv_kill` support won't work unfortunately - `uv__signal_dispatch` only triggers signal watchers, it doesn't send signals to another process.

shigeki:
@bnoordhuis Thanks for reviewing. I should have taken care of a coding style much more. I will fix it.
@piscisaureus Could you please explain it more?  `uv_kill` in this commit is only for hooking a callback of signal handlers of SIGINT, SIGHUP and SIGBREAK. I think that sending real signals with `uv_kill' is an another issue.

piscisaureus:
@shigeki Well, isn't this very unexpected:

``` js
var pid = 123; // Some process, not us
process.on('SIGINT', function() { console.log('sigint!') })
process.kill(pid, 'SIGINT');
 // "sigint!" printed
```

piscisaureus:
@shigeki I landed the "remove superfluous uv__handle_start" commit in 343be71.

bnoordhuis:
I think the relevant parts of this PR have landed? Tentatively closing, holler if I should reopen it.

PR #547, opened 2012-09-05
windows: support for NULL as uv_queue_work callback

lhecker:
This will make uv_queue_work consistent with other functions, where it's
possible to omit the callback. I'm pretty sure that this is already
implemented for unix systems, although I didn't test the unix part of
uv_queue_work.


piscisaureus:
I know it's a two-line fix, but can you sign the CLA? http://nodejs.org/cla.html

lhecker:
@piscisaureus I signed the CLA. Expect more to come :) (if I find the time)

piscisaureus:
Thanks. Landed in 15ba800.

PR #549, opened 2012-09-09
unix: support missing api on NetBSD

shigeki:
I need to run Node on NetBSD so I've just added several missing platform depended APIs with this commit.

Tests were made on only i386 with NetBSD-5.1.2. Several tests are still failed but it seems that they are not directly related to this commit. I will try to fix them later.
I have two points which should be noticed in this PR.
-  `uv_set_process_title` uses `setproctitle`
  On FreeBSD, 94355e4718a788c5f2d97f1d9da34cbcd2c74b03 uses sysctl instead of `setproctitle` because of libc fault. I tried to check it on NetBSD but vg4nbsd, valgrind for NetBSD, cannot be built on current NetBSD release. Is there any way to check it beside using valgrind? 
- `cpuspeed` is taken from `machdep.tsc_freq` . 
  It seems to be implemented only in i386 and amd64 form a kernel source. In case that this kernel parameter does not exist, it returns 0 in the current commit. I wonder it causes some side effects when someone would calculate its clock time. If a positive value is prefer, please let me know it.


src/unix/netbsd.c
```diff
@@ -115,3 +127,218 @@ uint64_t uv_get_total_memory(void) {
...
+  return uv_ok_;
+}
+
+
+uv_err_t uv_resident_set_memory(size_t* rss) {
+  kvm_t *kd = NULL;
+  struct kinfo_proc2 *kinfo = NULL;
+  pid_t pid;
+  int nprocs, max_size = sizeof(struct kinfo_proc);
+  int page_size = getpagesize();
```
bnoordhuis: Style: please split declarations and assignments (i.e. declarations first, then assignments). One declaration per line.

src/unix/netbsd.c
```diff
@@ -115,3 +127,218 @@ uint64_t uv_get_total_memory(void) {
...
+
+  pid = getpid();
+
+  kd = kvm_open(NULL, NULL, NULL, KVM_NO_FILES, "kvm_open");
+
+  if (kd == NULL) goto error;
+
+  kinfo = kvm_getproc2(kd, KERN_PROC_PID, pid, max_size, &nprocs);
```
bnoordhuis, line 181: max_size mismatch? It's set to `sizeof(struct kinfo_proc)` but `kvm_getproc2()` returns a `struct kinfo_proc2`.

src/unix/netbsd.c
```diff
@@ -115,3 +127,218 @@ uint64_t uv_get_total_memory(void) {
...
+
+  *rss = kinfo->p_vm_rssize * page_size;
+
+  kvm_close(kd);
+
+  return uv_ok_;
+
+error:
+  if (kd) kvm_close(kd);
+  return uv__new_sys_error(errno);
```
bnoordhuis: The kvm_\* functions don't set errno.

src/unix/netbsd.c
```diff
@@ -115,3 +127,218 @@ uint64_t uv_get_total_memory(void) {
...
+}
+
```
bnoordhuis: Style: two newlines between functions.

src/unix/netbsd.c
```diff
@@ -115,3 +127,218 @@ uint64_t uv_get_total_memory(void) {
...
+uv_err_t uv_uptime(double* uptime) {
+  time_t now;
+  struct timeval info;
+  size_t size = sizeof(info);
+  static int which[] = {CTL_KERN, KERN_BOOTTIME};
+
+  if (sysctl(which, 2, &info, &size, NULL, 0) < 0) {
```
bnoordhuis: I prefer `== -1`, that's what we use almost everywhere else in libuv.

src/unix/netbsd.c
```diff
@@ -115,3 +127,218 @@ uint64_t uv_get_total_memory(void) {
...
+  size = numcpus * CPUSTATES * sizeof(*cp_times);
+  cp_times = malloc(size);
+  if (cp_times == NULL) {
+    return uv__new_artificial_error(UV_ENOMEM);
+  }
+  if (sysctlbyname("kern.cp_time", cp_times, &size, NULL, 0) < 0) {
+    return uv__new_sys_error(errno);
+  }
+
+  *cpu_infos = (uv_cpu_info_t*)malloc(numcpus * sizeof(uv_cpu_info_t));
```
bnoordhuis: Superfluous cast. Also, prefer `sizeof(**cpu_infos)`.

src/unix/netbsd.c
```diff
@@ -115,3 +127,218 @@ uint64_t uv_get_total_memory(void) {
...
+  for (i = 0; i < numcpus; i++) {
+    cpu_info = &(*cpu_infos)[i];
+    cpu_info->cpu_times.user = (uint64_t)(cp_times[CP_USER+cur]) * multiplier;
+    cpu_info->cpu_times.nice = (uint64_t)(cp_times[CP_NICE+cur]) * multiplier;
+    cpu_info->cpu_times.sys = (uint64_t)(cp_times[CP_SYS+cur]) * multiplier;
+    cpu_info->cpu_times.idle = (uint64_t)(cp_times[CP_IDLE+cur]) * multiplier;
+    cpu_info->cpu_times.irq = (uint64_t)(cp_times[CP_INTR+cur]) * multiplier;
+    cpu_info->model = strdup(model);
+    cpu_info->speed = (int)(cpuspeed/(uint64_t) 1e6);
+    cur+=CPUSTATES;
```
bnoordhuis: Style: spaces between operator (in this whole block).

src/unix/netbsd.c
```diff
@@ -115,3 +127,218 @@ uint64_t uv_get_total_memory(void) {
...
+  for (i = 0; i < count; i++) {
+    free(cpu_infos[i].model);
+  }
+
+  free(cpu_infos);
+}
+
+
+uv_err_t uv_interface_addresses(uv_interface_address_t** addresses,
+  int* count) {
```
bnoordhuis: Style: should line up with the first argument.

src/unix/netbsd.c
```diff
@@ -115,3 +127,218 @@ uint64_t uv_get_total_memory(void) {
...
+    if (!(ent->ifa_flags & IFF_UP && ent->ifa_flags & IFF_RUNNING) ||
+        (ent->ifa_addr == NULL) ||
+        (ent->ifa_addr->sa_family != PF_INET)) {
+      continue;
+    }
+    (*count)++;
+  }
+  
+  *addresses = (uv_interface_address_t*)
+               malloc(*count * sizeof(uv_interface_address_t));
```
bnoordhuis: Superfluous cast and sizeof.

src/unix/netbsd.c
```diff
@@ -115,3 +127,218 @@ uint64_t uv_get_total_memory(void) {
...
+
+    address->name = strdup(ent->ifa_name);
+
+    if (ent->ifa_addr->sa_family == AF_INET6) {
+      address->address.address6 = *((struct sockaddr_in6 *)ent->ifa_addr);
+    } else {
+      address->address.address4 = *((struct sockaddr_in *)ent->ifa_addr);
+    }
+    
+    address->is_internal = ent->ifa_flags & IFF_LOOPBACK ? 1 : 0;
```
bnoordhuis: `!!(ent->ifa_flags & IFF_LOOPBACK)`?

src/unix/netbsd.c
shigeki, line 181: Sorry, I made it wrong. Fixed.

src/unix/netbsd.c
shigeki: They seems to set errno in internal system calls. For example,

``` C
#include <stdio.h>
#include <errno.h>
#include <fcntl.h>
#include <kvm.h>
#include <paths.h>

int main(int argc, char** argv) {
  kvm_t *kd;

  kd = kvm_open(NULL, _PATH_MEM, NULL, O_RDONLY, "kvm_open");
  if (kd == NULL ) {
    fprintf(stderr, "errno = %d\n",errno);
  }
}
```

``` sh
> ./kvm_error
kvm_open: /dev/mem: Permission denied
errno = 13
```

src/unix/netbsd.c
shigeki: Fixed all newline styles.

src/unix/netbsd.c
shigeki: Changed to follow your style.

src/unix/netbsd.c
shigeki: Thanks. Fixed.

src/unix/netbsd.c
shigeki: Sorry, I missed style. Fixed.

src/unix/netbsd.c
shigeki: Changed.

src/unix/netbsd.c
shigeki: Fixed

src/unix/netbsd.c
shigeki: Fixed

src/unix/netbsd.c
bnoordhuis: Sorry, I mean that they're not guaranteed to set errno. The kvm_\* functions clobber errno as a side effect of how they're implemented.

bnoordhuis: I guess you can work around it like this:

``` c
errno = 0;
kd = kvm_open(...);
if (kd == NULL && errno == 0) errno = EINVAL;
```

shigeki: @bnoordhuis Thanks, I got the point. You are right. This is more robust error handling. But I don't think UV_EINVAL is a proper error code but UV_UNKNOWN is in this case. 
I fixed it as below. How about it?

```
--- a/src/unix/netbsd.c
+++ b/src/unix/netbsd.c
@@ -173,6 +173,7 @@ uv_err_t uv_resident_set_memory(size_t* rss) {
   page_size = getpagesize();
   pid = getpid();

+  errno = 0;
   kd = kvm_open(NULL, NULL, NULL, KVM_NO_FILES, "kvm_open");

   if (kd == NULL) goto error;
@@ -188,7 +189,7 @@ uv_err_t uv_resident_set_memory(size_t* rss) {

 error:
   if (kd) kvm_close(kd);
-  return uv__new_sys_error(errno);
+  return uv__new_sys_error(errno ? errno : -1);
 }
```

bnoordhuis:
Thanks Shigeki, landed in 57e6113. Out of curiosity, what are you doing with (on) NetBSD?

shigeki:
@bnoordhuis Thanks! We have a router product based on NetBSD. http://www.netbsd.org/gallery/products.html#seil . I 'm attempting to run Node on it. But It is just beginning and a very early phase for evaluation of Node on the latest NetBSD release. Several tests are still failed and I will investigate them soon.

bnoordhuis:
Interesting. Guess I have a good reason to test node and libuv more on netbsd now. :-)

PR #550, opened 2012-09-11
Add uv_thread_self.

meh:
This adds a function to know the current thread's id.


txdv:
I don't think this is fits into the libuv API. libuv doesn't handle threads, you can run of course loops in different threads, but you need additional thread functionality and the functionality you are proposing should be included into this additional thread functionality.

meh:
You can create threads and run stuff in threads, you have locking functionalities, how doesn't knowing the current thread fit in there?

piscisaureus:
I don't think it is necessarily a misfit for libuv. The question is whether this API is what people need. @meh What do _you_ need it for?

meh:
OpenSSL multithread support requires a callback to know the current thread id.

bnoordhuis:
I'm not opposed either though I may rename it to `uv_thread_self()` or something.

@meh Can you sign [the CLA](http://nodejs.org/cla.html)?

meh:
> I'm not opposed either though I may rename it to `uv_thread_self()` or something.
> Yeah, I was unsure about the name, if it's easier I can change it when
> you have one and squash the commits.
> 
> @meh Can you sign [the CLA](http://nodejs.org/cla.html)?
> I sure can, but half of the required information is not applicable as
> I don't have a mailing address or telephone.

Is that ok or would it be better if I signoff the commit?

bnoordhuis:
> I sure can, but half of the required information is not applicable as I don't have a mailing address or telephone.

I guess you can put N/A in the fields that don't apply. We'll need your real name though.

meh:
Done. Any idea for a better name?

bnoordhuis:
`uv_thread_self()`? Principle of least disconnect.

meh:
Done.

bnoordhuis:
Thanks, landed in 974c856.

PR #553, opened 2012-09-13
aix: add initial platform support for aix using gcc/gxlc

apaprocki:
Adds initial libuv build/platform support for AIX. Builds work using gcc or the
IBM XL C compiler using its gxlc wrapper. Platform support is added for
uv_hrtime, uv_exepath, uv_get_free_memory, uv_get_total_memory, uv_loadavg,
uv_uptime, uv_cpu_info, uv_interface_addresses.


src/unix/eio/xthread.h
```diff
@@ -49,7 +49,7 @@
 #define X_THREAD_ATFORK(a,b,c)
 
 static int
-thread_create (xthread_t *tid, void *(*proc)(void *), void *arg)
+xthread_create (xthread_t *tid, void *(*proc)(void *), void *arg)
 {
```
apaprocki, line 53: These statics had to be renamed because thread_create is found in AIX system headers. Given that this is xthread.h, it only made sense that this should be xthread_create.

src/unix/error.c
```diff
@@ -95,7 +95,9 @@ uv_err_code uv_translate_sys_error(int sys_errno) {
     case ETIMEDOUT: return UV_ETIMEDOUT;
     case EXDEV: return UV_EXDEV;
     case EBUSY: return UV_EBUSY;
+#if ENOTEMPTY != EEXIST
```
apaprocki, line 98: This is done because these are equivalent on AIX.

src/unix/ev/ev.c
```diff
@@ -3912,8 +3912,8 @@ struct ev_once
           wl = wn;
         }
 #endif
-/* EV_STAT     0x00001000 /* stat data changed */
-/* EV_EMBED    0x00010000 /* embedded event loop needs sweep */
+/* EV_STAT     0x00001000 *//* stat data changed */
+/* EV_EMBED    0x00010000 *//* embedded event loop needs sweep */
```
apaprocki, line 3916: This was necessary because non-gcc compilers error when /\* _/ blocks contain nested /_.

test/test-thread.c
```diff
@@ -42,7 +42,7 @@ struct fs_req {
 };
 
 
-struct thread {
+struct test_thread {
```
apaprocki, line 45: Similar to the xthread.h change, 'struct thread' is found in system headers.

test/test-counters-init.c
```diff
@@ -18,12 +18,13 @@
  * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
  * IN THE SOFTWARE.
  */
-#define UNIX (defined(__unix__) || defined(__POSIX__) || defined(__APPLE__))
 #include "task.h"
 #include "uv.h"
 #include <fcntl.h>
 
-#if UNIX
+/* FIXME we shouldn't need to branch in this file */
```
apaprocki, line 25: Copied the FIXME already in place in another location so that if it is fixed, these locations are found as well.

piscisaureus:
@apaprocki Do you know what libev backend is used on AIX. You should know that libev is about to be removed, so if you really care you should probably start thinking about how to efficiently poll on AIX. pollset_poll ?

apaprocki:
@piscisaureus It uses select(). I'm aware libev is being removed, but I'm not quite sure what changes will need to happen once it is gone. Is it already removed on a branch somewhere? I can change whatever is needed to use pollset_poll().

piscisaureus:
@apaprocki I think it's in https://github.com/bnoordhuis/libuv/commits/rm-ev.

You probably want to talk to @bnoordhuis first. You can often catch him on IRC in #libuv (on freenode).

piscisaureus:
@apaprocki BTW do you have a garage full of old mainframes? First Sun Studio, now AIX... :-)

apaprocki:
@piscisaureus If by old you mean 128-way 4GHz POWER7... :) I'll take a look at the rm-ev branch. Since I already had this checked in, I wanted to get the request in before too much stuff changed in master. I'll definitely port whatever changes are needed, though.

bnoordhuis:
The patch LGTM but it doesn't apply to the current master. Can you rebase it? By the way, we split off c-ares into a [separate repository](https://github.com/piscisaureus/cares).

And yay for AIX support. Now the only hold-out is HP-UX.

piscisaureus:
Nevermind. Landed in 900ad30. Thanks.

PR #554, opened 2012-09-15
Struct layout

txdv:
Do not merge yet.

The idea of this pull request is to put all the private fields to the bottoms of the struct layouts.
This would make the struct layout offsets of the public fields indifferent on different platforms and therefore it would ease the process of creating bindings with ffi.

For example:

``` csharp
struct uv_signal_s {
  UV_HANDLE_FIELDS
  uv_signal_cb signal_cb;
  UV_SIGNAL_PRIVATE_FIELDS
  int signum;
};
```

This is what we have now, UV_HANDLE_FIELDS contains UV_HANDLE_PRIVATE_FIELDS, so the offset of uv_signal_cb differs from platform to platform, as well as the offset of signum.
This is an undesirable state, what I would like to have is:

``` csharp
struct uv_signal_s {
  UV_HANDLE_FIELDS
  uv_signal_cb signal_cb;
  int signum;
  UV_HANDLE_PRIVATE_FIELDS
  UV_SIGNAL_PRIVATE_FIELDS
};
```

This would make the offset of uv_signal_cb and signum consistent on all platforms.

I am facing a few issues right now to achieve this.
uv-common.h uses the field flags, which is defined in UV_HANDLE_PRIVATE_FIELDS. On unix it is defined as **int flags** on windows as **unsigned int flags**, whats up with the signed unsigned difference?
It is a common variable used on both platforms(and in uv-common) therefore it has to have a consistent offset for all struct layouts (and we can't put it to the end of struct layouts). We have to put it into UV_HANDLE_FIELDS , but it is considered to be a 'private' variable as well, private but not platform specific.

Comments and thoughts are welcome.


txdv:
Moving UV_STREAM_PRIVATE_FIELDS to the end of the struct adds a lot of complications, the only affected class is uv_pipe_t, the offset of the ipc field will be different on different platforms, which is easier to counter with a little extra code on the binding part.

piscisaureus:
The alternative approach (#555) was preferred.

PR #555, opened 2012-09-15
Fix some struct layouts.

txdv:
This is an alternative to https://github.com/joyent/libuv/pull/554 and this alternative more. Less intrusive.

The idea is just to use uv_req_size and uv_handle_size to get the first offsets.
For example, we would do something like this with ffi to retrieve the ipc field from uv_pipe_t:

``` c
struct uv_pipe_s {
  UV_HANDLE_FIELDS
  UV_STREAM_FIELDS
  int ipc; /* non-zero if this pipe is used for passing handles */
  UV_PIPE_PRIVATE_FIELDS
};
```

``` c
  struct pipe_t = {
    int rpc;
  }
  uv_pipe_t *ptr = /* we get somehow that pointer */;
  pipe_t *pipe = ptr + uv_handle_size(UV_NAMED_PIPE);
  pipe->rpc; // yeah
```

If you are writing bindings as a module in C, you do not need this, but if you are writing bindings using FFI (as I am doing with C#) then the struct layout has to be consistent.

In the patch I fixed the layouts a little bit so they would be more consistent and one could apply the logic I just described to retrieve all fields.


piscisaureus:
Thanks. Landed in 46bd5fb.

PR #556, opened 2012-09-15
Implemented uv_tcp_open and uv_udp_open

saghul:
Used to open the given file descriptor or SOCKET as a TCP or UDP handle respectively.

### DIFF
--- a/src/unix/pipe.c
+++ b/src/unix/pipe.c
@@ -156,10 +156,10 @@ void uv__pipe_close(uv_pipe_t* handle) {
 }
 
 
-void uv_pipe_open(uv_pipe_t* handle, uv_file fd) {
-  uv__stream_open((uv_stream_t*)handle,
-                  fd,
-                  UV_STREAM_READABLE | UV_STREAM_WRITABLE);
+int uv_pipe_open(uv_pipe_t* handle, uv_file fd) {
+  return uv__stream_open((uv_stream_t*)handle,
+                         fd,
+                         UV_STREAM_READABLE | UV_STREAM_WRITABLE);
 }
 
 
--- a/src/unix/udp.c
+++ b/src/unix/udp.c
@@ -316,17 +316,15 @@ static int uv__bind(uv_udp_t* handle,
     goto out;
   }
 
-  /* Check for already active socket. */
-  if (handle->fd != -1) {
-    uv__set_artificial_error(handle->loop, UV_EALREADY);
-    goto out;
+  if (handle->fd == -1) {
+    if ((fd = uv__socket(domain, SOCK_DGRAM, 0)) == -1) {
+      uv__set_sys_error(handle->loop, errno);
+      goto out;
+    }
+    handle->fd = fd;
   }
 
-  if ((fd = uv__socket(domain, SOCK_DGRAM, 0)) == -1) {
-    uv__set_sys_error(handle->loop, errno);
-    goto out;
-  }
-
+  fd = handle->fd;
   yes = 1;
   if (setsockopt(fd, SOL_SOCKET, SO_REUSEADDR, &yes, sizeof yes) == -1) {
     uv__set_sys_error(handle->loop, errno);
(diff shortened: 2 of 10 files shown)


piscisaureus:
This is nice and all, but is it also possible to open a socket that is already listening or already connected?

saghul:
Good question :-) My answer to that would be no. While it is possible to have it working in certain cases (at least in Unix) I think that could open the can of worms.

Is that something we want to support? If we do I can try to think of a solution for it, but for now I'd say that it is not supported.

The intent of this patch is to provide means to more easily manage sockets created by other applications / libraries, instead of using the poll handle, for example.

Thoughts?

test/test-udp-open.c
```diff
@@ -0,0 +1,161 @@
...
+
+  if (nread < 0) {
+    ASSERT(0 && "unexpected error");
+  }
+
+  if (nread == 0) {
+    /* Returning unused buffer */
+    /* Don't count towards sv_recv_cb_called */
+    ASSERT(addr == NULL);
+    return;
```
bnoordhuis, line 105: `free(buf.base)`? Maybe make the buffer static (if possible)?

src/unix/pipe.c
```diff
@@ -156,10 +156,10 @@ void uv__pipe_close(uv_pipe_t* handle) {
 }
 
 
-void uv_pipe_open(uv_pipe_t* handle, uv_file fd) {
-  uv__stream_open((uv_stream_t*)handle,
-                  fd,
-                  UV_STREAM_READABLE | UV_STREAM_WRITABLE);
+int uv_pipe_open(uv_pipe_t* handle, uv_file fd) {
+  return uv__stream_open((uv_stream_t*)handle,
+                        fd,
```
bnoordhuis: Style: needs to align with the first argument.

src/unix/udp.c
```diff
@@ -316,19 +316,16 @@ static int uv__bind(uv_udp_t* handle,
...
-    uv__set_artificial_error(handle->loop, UV_EALREADY);
-    goto out;
-  }
-
-  if ((fd = uv__socket(domain, SOCK_DGRAM, 0)) == -1) {
-    uv__set_sys_error(handle->loop, errno);
-    goto out;
+  if (handle->fd == -1) {
+    if ((fd = uv__socket(domain, SOCK_DGRAM, 0)) == -1) {
+        uv__set_sys_error(handle->loop, errno);
```
bnoordhuis: Style: indent error.

src/unix/udp.c
```diff
@@ -352,7 +349,7 @@ static int uv__bind(uv_udp_t* handle,
   if (flags & UV_UDP_IPV6ONLY) {
 #ifdef IPV6_V6ONLY
     yes = 1;
-    if (setsockopt(fd, IPPROTO_IPV6, IPV6_V6ONLY, &yes, sizeof yes) == -1) {
+    if (setsockopt(handle->fd, IPPROTO_IPV6, IPV6_V6ONLY, &yes, sizeof yes) == -1) {
```
bnoordhuis: Style: wrap at 80 columns. Probably applies to other lines too.

saghul:
I force pushed, all style issues should be fixed now.

bnoordhuis:
Thanks Saúl, landed in master.

txdv:
So I can't, or at least I shouldn't use it on already connected sockets, but I can use this on bound sockets?

saghul:
@txdv For the moment those functions are only guaranteed to work if the socket is not connected / listening.

bnoordhuis:
> For the moment those functions are only guaranteed to work if the socket is not connected / listening.

We'll probably add that functionality some day, though.

schlamar:
Hey @saghul pre-forking libuv/pyuv already works out of the box, see https://gist.github.com/schlamar/9691672. The only issue is that the fd is closed on handle close, 
you get a `(9, 'bad file descriptor')` on the parent if you uncomment the closing in the client process.

This should be handled with a flag to uv_tcp_open.

It should even already work on Windows if the user passes an inherited socket handle (and how he gets one is up to the user). 

Edit: if you can improve the script so both processes can be properly stopped with SIGINT it would be much appreciated =)

schlamar:
@saghul thanks for pointing me in the right direction, it works perfectly with `dup`. =) https://gist.github.com/schlamar/9691672 Just execute `telnet 127.0.0.1 8888` in two terminals to get a response from both processes.

saghul:
Hy @schlamar: yes, dup is the way to go IMHO. Basically, once a fd is passed to libuv with uv_*_open, libuv takes over, so if you don't want it to be closed you should pass a dupped one.

PR #559, opened 2012-09-19
unix: map errno EDQUOT

AvianFlu:
Encountered this as an 'unknown error' in node when the disk limit on a smartOS zone was hit.


piscisaureus:
I think I would prefer it to map to UV_ENOSPC.

AvianFlu:
Done.

bnoordhuis:
Thanks, landed in 2916a16.

PR #560, opened 2012-09-20
Feature fsevents more events

indutny:
nt


bnoordhuis:
Thanks. Landed in 73cf360.

PR #561, opened 2012-09-20
uv_wait() (everything except sunos, cygwin and windows)

indutny:
nt


txdv:
Is it possible to get an example (gist?) on how to embed the libuv loop into a qt loop?

src/unix/kqueue.c
```diff
@@ -143,3 +144,19 @@ void uv__fs_event_close(uv_fs_event_t* handle) {
   close(handle->fd);
   handle->fd = -1;
 }
+
+
+void uv_wait(uv_loop_t* loop, int64_t timeout) {
+  struct timespec ts;
+  int64_t poll_timeout;
+
+  poll_timeout = uv__poll_timeout(loop);
```
bnoordhuis: Overflow in unsigned to signed conversion is implementation defined.

bnoordhuis: To expand on that, this code works as long as `sizeof(unsigned int) < sizeof(int64_t)` - but think of the Cray port!

src/unix/kqueue.c
```diff
@@ -143,3 +144,19 @@ void uv__fs_event_close(uv_fs_event_t* handle) {
...
+  if (timeout == -1 || poll_timeout > timeout) {
+    poll_timeout = timeout;
+  }
+
+  ts.tv_sec = poll_timeout / 1000;
+  ts.tv_nsec = (poll_timeout - ts.tv_sec * 1000) * 1000000;
```
bnoordhuis: `(poll_timeout % 1000)` ?

src/unix/kqueue.c
```diff
@@ -143,3 +144,19 @@ void uv__fs_event_close(uv_fs_event_t* handle) {
...
   handle->fd = -1;
 }
+
+
+void uv_wait(uv_loop_t* loop, int64_t timeout) {
+  struct timespec ts;
+  int64_t poll_timeout;
+
+  poll_timeout = uv__poll_timeout(loop);
+  if (timeout == -1 || poll_timeout > timeout) {
```
bnoordhuis: Actually, now that I think about it... `uv__poll_timeout()` returns `(unsigned int) -1` (a.k.a. `UINT_MAX`) when we should block indefinitely. If an int is 32 bits, then the `timeout == -1` check will always be false because timeout won't be -1 but UINT_MAX.

test/test-embed.c
```diff
@@ -0,0 +1,90 @@
...
+#include "task.h"
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+static uv_thread_t embed_thread;
+static uv_sem_t embed_sem;
+static uv_timer_t embed_timer;
+static uv_async_t embed_async;
+static int embed_closed;
```
bnoordhuis: Needs to be volatile.

test/test-embed.c
```diff
@@ -0,0 +1,90 @@
...
+
+static int embed_timer_called;
+
+
+static void embed_thread_runner(void *arg) {
+  for (;;) {
+    if (embed_closed) break;
```
bnoordhuis: `while (embed_closed == 0) {`?

indutny:
@bnoordhuis I think I've fixed everything.

indutny:
@txdv I think test is a really good demonstration.

bnoordhuis:
Pet peeve: the commit log should tell (in detail) what changed and why.

src/unix/kqueue.c
```diff
@@ -143,3 +144,19 @@ void uv__fs_event_close(uv_fs_event_t* handle) {
...
+
+  max_timeout = uv__poll_timeout(loop);
+
+  if (max_timeout != UINT_MAX && (timeout == -1 || timeout > max_timeout))
+    timeout = max_timeout;
+
+  ts.tv_sec = timeout / 1000;
+  ts.tv_nsec = (timeout % 1000) * 1000000;
+
+  kevent(ev_backend_fd(loop->ev), NULL, 0, NULL, 0, timeout == -1 ? NULL : &ts);
```
bnoordhuis: Is ignoring the return value a good idea? At the very least you should handle EINTR here, shouldn't you, updating the timeout accordingly?

src/unix/linux/linux-core.c
```diff
@@ -571,3 +571,16 @@ void uv_free_interface_addresses(uv_interface_address_t* addresses,
 
   free(addresses);
 }
+
+
+void uv_wait(uv_loop_t* loop, int64_t timeout) {
+  struct timespec ts;
```
bnoordhuis: Unused.

src/unix/linux/linux-core.c
```diff
@@ -571,3 +571,16 @@ void uv_free_interface_addresses(uv_interface_address_t* addresses,
...
+  unsigned int max_timeout;
+
+  max_timeout = uv__poll_timeout(loop);
+
+  if (max_timeout != UINT_MAX && (timeout == -1 || timeout > max_timeout))
+    timeout = max_timeout;
+
+  epoll_wait(ev_backend_fd(loop->ev), NULL, 0, timeout);
```
bnoordhuis: Use uv__epoll_wait().

test/test-embed.c
```diff
@@ -0,0 +1,89 @@
...
+
+  uv_close((uv_handle_t*) &embed_async, NULL);
+}
+
+
+TEST_IMPL(embed) {
+  uv_loop_t* external;
+
+  external = uv_loop_new();
+  assert(external != NULL);
```
bnoordhuis: ASSERT

indutny:
@bnoordhuis you should be happy now :) fixed everything

indutny:
Closing in a favor of #631

PR #563, opened 2012-09-21
unix: map `EDQUOT` to `UV_EDQUOT`

mmalecki:
Seen in production on a SmartOS machine.


piscisaureus:
Already fixed in 2916a16.

PR #564, opened 2012-09-25
Fixed vcbuild.bat.

ghost:
gyp_uv was complaining because it never receives a value for which
type of library to build (static vs shared). Basically just set the
default to build a static library.

Maybe the default should be shared but either one fixes the build.


bnoordhuis:
LGTM. @piscisaureus?

piscisaureus:
I haven't noticed this problem, but it looks reasonable.

@shancat I know it's a one-liner but can you sign the CLA? http://nodejs.org/cla.html

ghost:
I've already signed the CLA for PRs to the node repo before.

piscisaureus:
@shancat Sorry, I must have overlooked it. Thanks, landed in 8afd71a.

PR #571, opened 2012-10-03
Moving queue stuff from fs.c to threadpool.c

arturadib:
I was going to ask why `uv_queue_*` was in `fs.c` for Unix whereas it's in `threadpool.c` on Windows (which seems to make more sense). Instead I'm asking via a pull request :)


bnoordhuis:
> I was going to ask why uv_queue_\* was in fs.c for Unix whereas it's in threadpool.c on Windows

Because that's where it lived before I removed libeio. :-)

Landed your patch in fed718c, thanks.

PR #577, opened 2012-10-05
Linux shared library support with autotools

sgallagher:
This patch is a first pass at making libuv compilable as a shared library that can be used by multiple projects. The configure.ac is minimalistic (providing checks only for the presence of libev at this time) and does not yet work on non-Linux environments (such as AIX).

Of particular note, this patch modifies the build so that it is linking against the system shared version of libev rather than the bundled copy. This is much safer, as libev has a stable ABI and so it's better to leave updates to the system library.


bnoordhuis:
I appreciate the effort but I'm in the process of ripping out libev. Expect that change to land in the next few days.

Next time, ping one of us on IRC. We're in #libuv on freenode.org.

PR #578, opened 2012-10-05
Struct layout

txdv:
Some struct layout.
The commit messages are very verbose, please read them.


piscisaureus:
Landed near 193432df.

PR #579, opened 2012-10-06
test: windows: "application crashed" popup still shows when using debug build

pH200:
[_CrtSetReportMode](http://msdn.microsoft.com/en-us/library/1y71x448.aspx) must be called to disable annoying debug popups.


piscisaureus:
Ok, fair enough. I know it's just a two line patch but could you please sign the [CLA](http://nodejs.org/cla.html)?

pH200:
@piscisaureus It's almost like these two lines were somewhat worthy.
Just kidding, I've signed the CLA :-P

piscisaureus:
Thanks! Landed in 8f1ec32

Issue #583, opened 2012-10-09
test/runner-win.c compile error on MinGW 


hnakamur:
with error message.
test/runner-win.c:27:20: fatal error: crtdbg.h: No such file or directory

https://github.com/joyent/libuv/commit/8f1ec328656c8fb792fce4a346c11141ab2cd9f9

bnoordhuis:
Does this patch fix it?

``` diff
diff --git a/test/runner-win.c b/test/runner-win.c
index 111a686..8ea459b 100644
--- a/test/runner-win.c
+++ b/test/runner-win.c
@@ -24,7 +24,10 @@
 #include <malloc.h>
 #include <stdio.h>
 #include <process.h>
-#include <crtdbg.h>
+
+#if !defined(__MINGW32__)
+# include <crtdbg.h>
+#endif

 #include "task.h"
 #include "runner.h"
```

hnakamur:
No. I sent a pull request. Please see https://github.com/joyent/libuv/pull/585

piscisaureus:
Fixed in 2684f876a.

pH200:
Oops, sorry for breaking the build :p

Although libuv is only supported build with either MSVC or MinGW for Windows. Given the fact that `crtdbg.h` is an MSVC specific header,  maybe it would be more appropriate to write:
`#if defined(_MSC_VER)`

PR #584, opened 2012-10-09
uv_barrier_t implementation

hnakamur:
This pull request implements
- uv_barrier_t
- uv_barrier_init()
- uv_barrier_destroy()
- uv_barrier_wait()
- On Unixes, it uses the native pthread_barrier_t implementation.
- On OSX and Windows, it uses the fallback implementation shown in Listing 3.10: Reusable barrier solution in [The Little Book of Semaphores](http://greenteapress.com/semaphores/)
  Ideally, if uv_sem_post would take the increment count as argument, we could use more efficient Listing 3.11: Reusable barrier solution.

Note 1: I tested this pull request on MinGW with the commit https://github.com/joyent/libuv/commit/8f1ec328656c8fb792fce4a346c11141ab2cd9f9 reversed. See https://github.com/joyent/libuv/issues/583
Note 2: Windows 8 has the native [Synchronization Barriers](http://msdn.microsoft.com/en-us/library/windows/desktop/hh706897(v=vs.85%29.aspx) implementation. However it is not included in MinGW header files. So I always use fallback implementation on Windows.


bnoordhuis:
I'm curious, what do you need barriers for? I don't mind landing the patch but, in my experience, barriers are one of the least used aspects of pthreads programming. They don't provide anything that you can't accomplish with condition variables or semaphores, do they?

hnakamur:
Well, I don't have particular needs at the moment. I read about barriers in the book "Multicore Application Programming" and I thought barriers are basic things because they are defined in pthread.

Also, as pointed in [The Little Book of Semaphores](http://greenteapress.com/semaphores/), there are many pitfalls for us to implement barriers, so I thought they are nice additions to libuv.

If you don't agree, feel free to reject this pull request. I don't mind.

bnoordhuis:
> If you don't agree, feel free to reject this pull request. I don't mind.

No, I'm fine with the patch, I was just wondering. If @piscisaureus doesn't object, I'll merge it. There are some style issues (functions should be separated by two newlines) but I'll fix that up.

hnakamur:
OK. I leave it up to you and @piscisaureus 
Thanks.

bnoordhuis:
Landed in acea3028. Thanks, Hiroaki!

PR #585, opened 2012-10-09
Disable crtdbg on MinGW. refs #583

hnakamur:


bnoordhuis:
LGTM. @piscisaureus?

piscisaureus:
@hnakamur Are you saying it's not necessary on mingw? Or is it just that crtdbg.h is not available?

hnakamur:
The latter. crtdbg.h is not just available on MinGW. I don't know it's not necessary on MinGW.

BTW, I found it in mingw-w64-v2.0.7, but _CrtSetReportMode is defined as no-op macro, not a real implementation.

piscisaureus:
Thanks, landed in 2684f876a

PR #590, opened 2012-10-15
Implement uv_set_process_title on darwin. Just copied linux implementati...

hnakamur:
...on and add 'extern char **environ'.

I found a library named libnostd impelemnts setproctitle on darwin.
http://www.25thandclement.com/~william/projects/libnostd.html
Its implemetaion is almost same between linux and darwin, so I copied libuv linux implementation of uv_set_process_title to src/unix/darwin.c

I confirmed all tests pass on darwin with this pull request applied.

I also tried a sample program using uv_set_process_title.

libuv_set_process_title_test.c

```
#include <stdio.h>
#include <unistd.h>
#include <uv.h>

int main(int argc, char **argv) {
  char buffer[256];

  argv = uv_setup_args(argc, argv);

  if (argc > 1)
    uv_set_process_title("hello uv_set_process_title");

  uv_get_process_title(buffer, sizeof(buffer));
  printf("title=%s\n", buffer);

  sleep(2);
  return 0;
}
```

The results:

```
c -o a.out libuv_set_process_title_test.c -I $(HOME)/src/libuv/include -L $(HOME)/src/libuv -luv -framework CoreServices
bash-3.2$ ./a.out & ps auxww | grep 'hello uv_set_process_title' | grep -v grep
[1] 24982
title=./a.out
bash-3.2$ 
[1]+  Done                    ./a.out
bash-3.2$ ./a.out 1 & ps auxww | grep 'hello uv_set_process_title' | grep -v grep
[1] 24986
title=hello uv_set_process_title
hnakamur       24986   0.0  0.0  2433000   1148 s006  R     7:55AM   0:00.01 hello uv_set_process_title 
```

### DIFF
--- a/src/unix/darwin.c
+++ b/src/unix/darwin.c
@@ -41,7 +41,12 @@
 #include <sys/sysctl.h>
 #include <unistd.h>  /* sysconf */
 
-static char *process_title;
+static void* args_mem;
+
+static struct {
+  char *str;
+  size_t len;
+} process_title;
 
 /* Forward declarations */
 void uv__cf_loop_runner(void* arg);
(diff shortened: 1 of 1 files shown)


bnoordhuis:
We used to have that (dc97d44) but I reverted it in 69a6afe, it's unsafe when you run as root.

hnakamur:
Oh. Sorry for the fuss.
Next time, I will check past pull requests.

bnoordhuis:
No worries, it happens.

PR #596, opened 2012-10-18
async: avoid full sweeps if possible

indutny:
/cc @bnoordhuis @piscisaureus


src/unix/async.c
```diff
@@ -81,11 +81,16 @@ int uv_async_send(uv_async_t* handle) {
     return 0; /* already pending */
 
   do
-    r = write(handle->loop->async_pipefd[1], "x", 1);
+    r = write(handle->loop->async_pipefd[1], (char*) &handle, sizeof(&handle));
```
bnoordhuis: You're taking the address of a local variable...

indutny: No, I'm writing data from this local address.

src/unix/async.c
```diff
@@ -81,11 +81,16 @@ int uv_async_send(uv_async_t* handle) {
...
   while (r == -1 && errno == EINTR);
 
-  if (r == -1 && errno != EAGAIN && errno != EWOULDBLOCK)
-    return uv__set_sys_error(handle->loop, errno);
+  if (r == -1) {
+    /* If pipe is full, event loop should sweep all handles */
+    if (errno == EAGAIN || errno == EWOULDBLOCK)
+      handle->loop->async_sweep_needed = 1;
+    else
+      return uv__set_sys_error(handle->loop, errno);
```
bnoordhuis: Why not abort()? Anything else is a fatal error anyway, right?

indutny: What else? :) There're no abort() calls in this function. Also, that's how it was before my patch.

bnoordhuis: Fair enough. We can change that some other time.

indutny:
Benchmarks are now working fine (w/o any assertions), additionaly this patch doesn't allow callbacks to be invoked by closed async handles.

src/unix/async.c
piscisaureus: Race condition: what if the following happens:
- Worker thread: write() EAGAIN
- Main thread: drain pipe
- Worker thread: set async_sweep_needed

I suppose you should retry the write after you set the async_sweep_needed bit.

indutny: You're right, fixed.

PR #597, opened 2012-10-18
Add function to convert a sockaddr to a host:port string

ghost:
Converts an AF_INET or AF_INET6 socket address to a host:port string.
Especially useful to output the addresses returned by getaddrinfo().


bnoordhuis:
I appreciate the effort but I don't really want to merge this. We try to keep the API fairly minimal and that means (with few exceptions) no convenience functions.

Thanks anyway. Maybe next time ping one of us first in #libuv on freenode.org before you start hacking.

Issue #595, opened 2012-10-17
Visual Studio 2008 support


saghul:
Yes, I know, it's old. But the python.org installer was compiled with it, so in order to properly compile pyuv I need to use this version.

I managed to compile ti already \o/ the only problem I ran into was the lack of stdint.h. We can use this one instead:

http://msinttypes.googlecode.com/svn/trunk/stdint.h

My quick and dirty was to do the following:

```
#ifdef _MSC_VER
    #include "uv-private/stdint-win.h"
#else
    #include <stdint.h>
#endif
```

wherever a file was included in the Windows build, just 6 files. (I added the file from that URL as stdint-win.h)

I'd like to submit a proper pull request, but I'd rather agree on how to do things from the start than doing things for nothing :-) so, does the suggested approach sound reasonable?

/cc @piscisaureus

piscisaureus:
@saghul The general approach seems fine with me. 

I would suggest you use this file only on VS 2008 and older; name the file to reflect this.

`#if defined(_MSC_VER) && _MSC_VER < 1600`

piscisaureus:
Also, update the LICENSE file.

saghul:
Thanks for your comments, Bert. Will do!

piscisaureus:
Fixed in 61ecb34.

PR #600, opened 2012-10-23
win: Support compilation with Visual Studio 2008

saghul:
As commented on issue #595


piscisaureus:
Thanks! Landed in 61ecb34.

PR #606, opened 2012-10-25
unix: Fixed compilation

saghul:
fd variable doesn't exist in that function.


bnoordhuis:
Thanks, Saúl. I didn't take your patch but I fixed the build error in 6150fed. The problem was that b5028c5b got dropped in a merge and `ninja -C out/Debug` somehow failed to pick up the changes.

saghul:
No problem, thanks Ben.

PR #608, opened 2012-10-25
Make it possible to compile for iOS

lhecker:
Relocated the include of TargetConditionals.h and fixed the use of
TARGET_OS_IPHONE. Furthermore uv__fsevents_init and uv__fsevents_close
are now empty functions for iOS, since the FSEvents-API is not
available there.


src/unix/internal.h
```diff
@@ -62,7 +62,7 @@
 # define HAVE_KQUEUE 1
 #endif
 
-#if defined(__APPLE__) && !defined(TARGET_OS_IPHONE)
+#if defined(__APPLE__) && !TARGET_OS_IPHONE
```
bnoordhuis, line 65: This should probably be `!defined(TARGET_OS_IPHONE) && !TARGET_OS_IPHONE` - gcc complains about dereferencing undefined macros in strict mode.

src/unix/fsevents.c
```diff
@@ -21,6 +21,14 @@
 #include "uv.h"
 #include "internal.h"
 
+#if TARGET_OS_IPHONE
+
+/* iOS (currently) doesn't provide the FSEvents-API (nor CoreServices) */
+int uv__fsevents_init(uv_fs_event_t* handle) { return 0; }
```
bnoordhuis: Split over multiple lines. Otherwise LGTM.

bnoordhuis:
Style issues but otherwise LGTM. Leonard, can you sign [the CLA](http://nodejs.org/cla.html)?

src/unix/internal.h
lhecker, line 65: As far as I know TARGET_OS_IPHONE should be used without defined()  (like http://stackoverflow.com/a/458388, or http://stackoverflow.com/a/3182873 and because it will be always defined since I relocated the include of TargetConditionals.h to uv-darwin.h).

lhecker:
I already signed it here: https://github.com/joyent/libuv/pull/547

src/unix/internal.h
bnoordhuis, line 65: You mean TargetConditionals.h always defines TARGET_OS_IPHONE? Looking at my copy of TargetConditionals.h that does indeed appear to be what's happening. In that case, okay, but can you fix up the style issues?

bnoordhuis:
Thanks Leonard, merged in 97c527a.

By the way, GH doesn't send notification emails for new commits - if you push a fix-up, post a "done" message in the issue. I noticed it because you mentioned it in #libuv. :-)

PR #612, opened 2012-11-02
If `options.env` is not provided, do not replace the child's environ

AvianFlu:
Currently, setting `options.env` to `NULL` for `uv_spawn` results in a child process for whom `environ == NULL`.  This can cause a wide variety of unexpected issues - this patch is a simple fix, and a test for the issue.


bnoordhuis:
Thanks Charlie, landed in 1d85815.

bnoordhuis:
And in fb64948 in v0.8.

piscisaureus:
You broke the windows build guys. Well done!

bnoordhuis:
Why? setenv?

PR #616, opened 2012-11-07
Support building libuv as a shared library

sgallagher:
The first patch in this branch modifies the libev headers so that there is a separate header for function prototypes. Source files that need internal access to these headers now include it directly.

The second patch is just a rebased version of 95b5331189f47ced9cf9d383a98a393140403873 written by Ben Noordhuis in his rm-ev branch.


bnoordhuis:
Thanks Stephen, landed with a minor fix in e318b06 and d56434a.

PR #619, opened 2012-11-09
Fix errors in linking libuv.so on Darwin (Mountain Lion)

hnakamur:


bnoordhuis:
I didn't take your PR but thanks anyway. It's addressed in 62332ea (which also fixes `make test` on SunOS).

hnakamur:
OK. thanks!

PR #622, opened 2012-11-12
unix: use select() for specific fds on OS X

indutny:
kqueue(2) on osx doesn't work (emits EINVAL error) with specific fds
(i.e. /dev/tty, /dev/null, etc). When given such descriptors - start
select(2) watcher thread that will emit io events.

@bnoordhuis "Select hack returns ;)"


src/unix/stream.c
```diff
@@ -34,6 +34,28 @@
 #include <sys/un.h>
 #include <unistd.h>
 
+#if defined(__APPLE__)
+# include <sys/event.h>
+# include <sys/time.h>
+# include <sys/select.h>
+
+/* ev.h is overwriting EV_ERROR from sys/event.h */
+#define EV_ERROR_ORIG 0x4000
```
bnoordhuis: Superfluous now that libev is removed.

src/unix/stream.c
```diff
@@ -97,11 +119,217 @@ void uv__stream_init(uv_loop_t* loop,
...
+  int events;
+  int fd;
+  int r;
+  int max_fd;
+
+  stream = arg;
+  s = stream->select;
+  fd = stream->fd;
+
+  max_fd = fd > s->int_fd ? fd : s->int_fd;
```
bnoordhuis: Use if/else here.

src/unix/stream.c
```diff
@@ -97,11 +119,217 @@ void uv__stream_init(uv_loop_t* loop,
...
+  if (s == NULL)
+    abort();
+
+  if (uv_async_init(stream->loop,
+                    &s->async,
+                    uv__stream_osx_select_cb)) {
+    return -1;
+  }
+  s->async.flags |= UV__HANDLE_INTERNAL;
+  uv__handle_unref((uv_handle_t*) &s->async);
```
bnoordhuis: Unnecessary cast.

src/unix/stream.c
```diff
@@ -97,11 +119,217 @@ void uv__stream_init(uv_loop_t* loop,
...
+fatal5:
+  close(s->int_fd);
+fatal4:
+  close(s->fake_fd);
+fatal3:
+  uv_mutex_destroy(&s->mutex);
+fatal2:
+  uv_sem_destroy(&s->sem);
+fatal1:
+  uv_close((uv_handle_t*) &s->async, uv__stream_osx_cb_close);
```
bnoordhuis, line 333: Use after free.

src/unix/stream.c
```diff
@@ -97,11 +119,217 @@ void uv__stream_init(uv_loop_t* loop,
...
+  /* At this point we definitely know that this fd won't work with kqueue */
+  s = malloc(sizeof(*s));
+  /* TODO: Return error */
+  if (s == NULL)
+    abort();
+
+  if (uv_async_init(stream->loop,
+                    &s->async,
+                    uv__stream_osx_select_cb)) {
+    return -1;
```
bnoordhuis: Memory leak.

src/unix/stream.c
```diff
@@ -97,11 +119,217 @@ void uv__stream_init(uv_loop_t* loop,
...
+  int ret;
+  struct kevent filter[1];
+  struct kevent events[1];
+  struct timespec timeout;
+  uv__stream_select_t* s;
+
+  kq = kqueue();
+  if (kq < 0) {
+    fprintf(stderr, "(libuv) Failed to create kqueue (%d)\n", errno);
+    abort();
```
bnoordhuis: Aborting on EMFILE is undesirable.

indutny:
@bnoordhuis rebased and force-pushed, fixed everything except EMFILE handling... any ideas on how it would be better to implement it?

indutny:
Unfortunatelly, it was required to introduce `uv__stream_fd()` routine, but I'm ok with replacing it with #define if it makes sense to you

bnoordhuis:
> fixed everything except EMFILE handling... any ideas on how it would be better to implement it?

Change the prototype of uv__stream_init() to return a status code and update the call sites to handle errors. That should probably be done in a separate commit, by the way.

bnoordhuis:
> Unfortunatelly, it was required to introduce uv__stream_fd() routine, but I'm ok with replacing it with #define if it makes sense to you

Yes, make that a simple #define when !defined(**APPLE**).

indutny:
@bnoordhuis fixed

src/unix/internal.h
```diff
@@ -192,6 +192,11 @@ void uv__work_submit(uv_loop_t* loop,
 void uv__prepare_close(uv_prepare_t* handle);
 void uv__process_close(uv_process_t* handle);
 void uv__stream_close(uv_stream_t* handle);
+#if defined(__APPLE__)
+int uv__stream_fd(uv_stream_t* handle);
+#else
+#define uv__stream_fd(handle) (handle)->io_watcher.fd
```
bnoordhuis: ((handle)->io_watcher.fd) - i.e. parentheses around the expression.

src/unix/internal.h
```diff
@@ -192,6 +192,11 @@ void uv__work_submit(uv_loop_t* loop,
 void uv__prepare_close(uv_prepare_t* handle);
 void uv__process_close(uv_process_t* handle);
 void uv__stream_close(uv_stream_t* handle);
+#if defined(__APPLE__)
+int uv__stream_fd(uv_stream_t* handle);
```
bnoordhuis: Maybe turn this into:

``` c
#define uv__stream_fd(handle) (uv_stream_fd((uv_stream_t*) handle))
```

So you don't have to cast to uv_stream_t everywhere.

indutny: No, it's potentially dangerous... I've found myself replacing the same calls in tcp/udp without realizing that tcp/udp handles ain't streams.

src/unix/stream.c
```diff
@@ -96,11 +116,230 @@ void uv__stream_init(uv_loop_t* loop,
...
+  /* Notify select() thread about state change */
+  int r;
+  uv__stream_select_t* s;
+
+  s = stream->select;
+
+  /* Interrupt select() loop */
+  do
+    r = write(s->int_fd, "x", 1);
+  while (r == -1 && errno == EINTR);
```
bnoordhuis: assert(r == 1)?

indutny:
@bnoordhuis Updated.

src/unix/internal.h
bnoordhuis: uv_tcp_t is a stream.

As to dangerous, people just need to be careful. It's not an uncommon pattern in libuv: uv__handle_init(), uv__req_init() and uv__is_active() do the same thing.

indutny: Ok, fixed

src/unix/stream.c
```diff
@@ -121,6 +360,16 @@ int uv__stream_open(uv_stream_t* stream, int fd, int flags) {
       return uv__set_sys_error(stream->loop, errno);
   }
 
+#if defined(__APPLE__)
+  r = uv__stream_try_select(stream, fd);
+  if (r == 0) {
+    /* Use fake fd */
+    fd = ((uv__stream_select_t*) stream->select)->fake_fd;
+  } else if (r > 0) {
```
bnoordhuis: Bug: uv__stream_try_select() returns either 0 or -1 and it's in fact returning -1 on my system.

bnoordhuis: In particular, it seems it's failing with ENOENT.

src/unix/stream.c
```diff
@@ -96,11 +116,230 @@ void uv__stream_init(uv_loop_t* loop,
...
+
+  EV_SET(&filter[0], fd, EVFILT_READ, EV_ADD | EV_ENABLE, 0, 0, 0);
+
+  /* Use small timeout, because we only want to capture EINVALs */
+  timeout.tv_sec = 0;
+  timeout.tv_nsec = 1;
+
+  ret = kevent(kq, filter, 1, events, 1, &timeout);
+  close(kq);
+  if (ret < 1)
```
bnoordhuis: ret seems to always be 0 for me...

bnoordhuis:
Landed in 731adac.

PR #626, opened 2012-11-15
Export uv_inet_* functions

sgallagher:
The uv_inet_ntop() and uv_inet_pton() functions need to be public in order for Node.JS to link against the libuv shared library.


bnoordhuis:
Thanks, landed in 05aac92.

PR #627, opened 2012-11-19
Add UV_PROCESS_WINDOWS_HIDE to avoid consoles on Windows

shanewholloway:
UV_PROCESS_WINDOWS_HIDE enables supression of process windows when launching subprocesses on Windows platforms from a executable linked to the WINDOWS subsystem (as opposed to the CONSOLE subsystem).

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -1314,7 +1314,12 @@ enum uv_process_flags {
    * parent's event loop alive unless the parent process calls uv_unref() on
    * the child's process handle.
    */
-  UV_PROCESS_DETACHED = (1 << 3)
+  UV_PROCESS_DETACHED = (1 << 3),
+  /*
+   * Hide the subprocess console window that would normally be created. This 
+   * option is only meaningful on Windows systems. On unix it is silently ignored.
+   */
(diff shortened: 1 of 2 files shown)


piscisaureus:
I'm fine with this. Can you sign the CLA? http://nodejs.org/cla.html

shanewholloway:
Fantastic!  CLA signed & submitted.

Do I need to port this to the 0.9 branch?

piscisaureus:
Landed (with some style modifications and fixing an assertion on unix) in 5d92ccc.

@shanewholloway PS. you may want to configure your git username, you show up as `"Shane Holloway ext:(%22) <shane.holloway@ieee.org>` here.

shanewholloway:
@piscisaureus thanks for letting me know — it is fixed for the future.

txdv:
Em, one question.
Why didn't you make HIDE the default setting and add a flag which would SHOW it?
Wouldn't it be more consistent behaviour since you can on unix create it in hidden mode?

shanewholloway:
When I was writing it, I was more concerned about not changing any existing behavior — SW_SHOWDEFAULT is the default behavior for Win32's CreateProcess. Making HIDE the default makes great sense based on your argument.

PR #629, opened 2012-11-22
windows: support for NULL as uv_work_cb of uv_queue_work callback

taichi:
This will be able to omit uv_work_cb on Windows.
This behavior is same as Unix.

I already signed the CLA.

### DIFF
--- a/src/win/threadpool.c
+++ b/src/win/threadpool.c
@@ -43,9 +43,9 @@ static DWORD WINAPI uv_work_thread_proc(void* parameter) {
 
   assert(req != NULL);
   assert(req->type == UV_WORK);
-  assert(req->work_cb);
 
-  req->work_cb(req);
+  if(req->work_cb)
+    req->work_cb(req);
 
   POST_COMPLETION_FOR_REQ(loop, req);
 


bnoordhuis:
I know uv-unix allows you to omit work_cb but IMO it's a silly feature (and I say that as the one who wrote it). Can you explain why you (or rather, node-memwatch) need that?

denik:
@bnoordhuis for example, to cancel getaddrinfo. Currently there does not seem to be a way to do it. If I know I no longer need the result of particular getaddrinfo call I'd like at least to set `work_cb` to NULL.

(If there _is_ actually another way to cancel async operation, please let me know and sorry for the noise).

bnoordhuis:
> for example, to cancel getaddrinfo. Currently there does not seem to be a way to do it. If I know I no longer need the result of particular getaddrinfo call I'd like at least to set work_cb to NULL.

That's not safe (at all). If thread A sets `req->work_cb = NULL` right between thread B's `if (req->work_cb)` check and `req->work_cb(req)` call, bad things will happen.

> (If there is actually another way to cancel async operation, please let me know and sorry for the noise).

There is no way to do that right way but this PR is not the solution. If you open an issue, we'll look into it.

lloyd:
In memwatch I'm using uv_queue_work as a process.nextTick() analogue in native code.  I care about running a function on the main loop, but not one on a thread.  I need this because I trigger work after gc is complete and cannot safely call into JavaScript in my post-gc callback.

So for me this usage was incredibly useful, and something that works on unix shouldn't _crash_ on doze.

I guess I should provide a noop function to workaround the flaw in libuv?  Is there a different function in libuv that would be better for my usage?

taichi:
```
That's not safe (at all). If thread A sets req->work_cb = NULL right between thread B's if (req->work_cb) check and req->work_cb(req) call, bad things will happen.
```

you are right @bnoordhuis. I was not aware of the problem.

But this PR is better code. because...
- current code contains multi threading ploblem.
- @lloyd code chrashes current code. but this PR doesn't crash.
- This behavior is same as Unix. I think that these differences in behavior and increase the problem.

i'm sorry that I can not seem to solve the problem pointed out by @bnoordhuis.

@lloyd please write workaround code because I want to use memwatch on Windows as soon as possible.

lloyd:
Yeah, I'll add a noop work function or merge a pr that does this so memwatch works on doze.  Thanks for the info and diagnosis.

-- lloyd (thumb-typing)

lloyd:
@taichi try tip of master, see https://github.com/lloyd/node-memwatch/commit/a2547b444adb7abff2d093cd1bc7d26d137846dc

If this resolves your crash then it confirms the diagnosis I guess.  cheers!

taichi:
@lloyd awesome!! 
I've tried three samples of your code and mine. all code did not crash. 
thanks for your quick response.

shigeki:
@lloyd a noop or null worker leads a thread go to waste because `after_work_cb` runs as a pending queue in the main loop not in the thread. If you want to execute  a native function as a `process.nextTick()`, which runs once in each one event loop, `uv_check()` or `uv_prepare()` is the right place to do.

bnoordhuis:
Okay, it looks like we have two issues here:
1. As @shigeki points out, using uv_queue_work() here is the wrong solution to the right problem. A uv_prepare or a uv_check is more appropriate.
2. There are inconsistencies between uv-win and uv-unix. I'll rectify that but I'm more inclined to disallow NULL callbacks altogether.

And a semi-related issue is the ability to cancel work requests. That feature request has come up once or twice before, it's something I'll look into.

shigeki:
> There are inconsistencies between uv-win and uv-unix. I'll rectify that but I'm more inclined to disallow NULL callbacks altogether.

I agree with you, Ben. I made a patch and a test in 254a289d7afea6747ecd58ca17e91a3e806244fa . Please review it.

shigeki:
@bnoordhuis Thanks for your review and patience. Fixed in 0e87f83ababbcfe27ed7b022b0565e088425f84f . I also check make test on my Windows.

bnoordhuis:
Very nice, thanks @shigeki. Landed in 0cca539.

PR #632, opened 2012-11-27
build: link static library on run-tests/benchmark

shigeki:
`make test` and `make bench` was failed in linking a shared library as below. 
It is better to link static library to build  run-tests and run-benchmark as gyp_uv script does. 
I tested this patch on Linux, SmartOS and NetBSD but not on MacOS. Please add RUNNER_LINKFLAGS if it is necessary on MacOS

```
> make test
cc -shared -o libuv.so src/unix/async.o src/unix/core.o src/unix/dl.o src/unix/error.o src/unix/fs.o src/unix/getaddrinfo.o src/unix/loop.o src/unix/loop-watcher.o src/unix/pipe.o src/unix/poll.o src/unix/process.o src/unix/signal.o src/unix/stream.o src/unix/tcp.o src/unix/thread.o src/unix/threadpool.o src/unix/timer.o src/unix/tty.o src/unix/udp.o src/fs-poll.o src/uv-common.o src/inet.o src/unix/linux/linux-core.o src/unix/linux/inotify.o src/unix/linux/syscalls.o -lm -ldl -lrt
/usr/bin/ld: src/unix/async.o: relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC
src/unix/async.o: could not read symbols: Bad value
collect2: ld returned 1 exit status
make: *** [libuv.so] Error 1
```


bnoordhuis:
Did you run `make all` before `make test`? There's no good story for that yet, the former builds libuv.a while the latter expects libuv.so.

A fix will probably involve building static and shared object files in separate build directories, e.g. out/static and out/shared.

shigeki:
@bnoordhuis Thanks. I understand you need to run testing with a shared library.  I used to execute `make; make test` unconsciously but I will not do it with careful attention. I close this ticket.

PR #633, opened 2012-11-27
test: fix finished consumer number check

shigeki:
I made a bug in f7c991f1cef1e0691b384b900f551a5c30745c36 that an error sometimes occurs in checking the finished consumer number in the test because of racing when a producer thread finishes before the last consumer does. 
Checking should be put in the place where joining all threads is finished.


bnoordhuis:
Thanks, landed in fa11db9. That was probably what was making the test fail intermittently on travis-ci.

shigeki:
Thanks, Ben. Sorry for the travis error. I should have fixed it more earlier.

bnoordhuis:
No problem. Glad it got fixed.

PR #639, opened 2012-12-02
Use putenv instead of setenv on MinGW

hnakamur:
It seems MinGW does not have setenv(). So we must use putenv() instead.
http://octave.1599824.n4.nabble.com/MinGW-build-error-setenv-was-not-declared-in-this-scope-td4644240.html


piscisaureus:
It seems that this is already fixed in master?

hnakamur:
Oh, you are right. I might have seen an old commit. I am sorry for making a wrong pull request.
I close this pull request.

PR #640, opened 2012-12-02
uv_thread_self() should return uv_thread_t instead of unsigned long

hnakamur:
Also, GetCurrentThread() should be used instead of GetCurrentThreadId() on Windows.

### DIFF
--- a/src/uv-common.c
+++ b/src/uv-common.c
@@ -316,7 +316,7 @@ int uv_thread_create(uv_thread_t *tid, void (*entry)(void *arg), void *arg) {
 
 uv_thread_t uv_thread_self(void) {
 #ifdef _WIN32
-  return GetCurrentThreadId();
+  return GetCurrentThread();
 #else
   return pthread_self();
 #endif
--- a/src/uv-common.c
+++ b/src/uv-common.c
@@ -314,11 +314,11 @@ int uv_thread_create(uv_thread_t *tid, void (*entry)(void *arg), void *arg) {
 }
 
 
-unsigned long uv_thread_self(void) {
+uv_thread_t uv_thread_self(void) {
 #ifdef _WIN32
-  return (unsigned long) GetCurrentThreadId();
+  return GetCurrentThreadId();
 #else
-  return (unsigned long) pthread_self();
+  return pthread_self();
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


src/uv-common.c
```diff
@@ -314,11 +314,11 @@ int uv_thread_create(uv_thread_t *tid, void (*entry)(void *arg), void *arg) {
 }
 
 
-unsigned long uv_thread_self(void) {
+uv_thread_t uv_thread_self(void) {
 #ifdef _WIN32
-  return (unsigned long) GetCurrentThreadId();
+  return GetCurrentThread();
```
bnoordhuis, line 319: Using GetCurrentThreadId() here is deliberate. GetCurrentThread() returns a pseudo handle that's not valid outside the calling thread.

hnakamur:
I see. Thanks for explanation. I close this pull request.

piscisaureus:
> Using GetCurrentThreadId() here is deliberate. GetCurrentThread() returns a pseudo handle that's not valid outside the calling thread.

That's not really _the_ issue, because a globally valid handle can be obtained by duplicating the pseudo handle.
The real issue hear was that many (real) handles can refer to the same thread, so it is not safe to compare thread handles.

hnakamur:
How can we get the id from a thread to compare with the result of GetCurrentThreadId()?
I found DWORD GetThreadId(HANDLE hThread)
http://msdn.microsoft.com/library/windows/desktop/ms683233(v=vs.85).aspx
but it is supported only on Vista+.

hnakamur:
Ah, we can get the thread id in _beginthreadex()
http://msdn.microsoft.com/en-us/library/kdzttdcb.aspx
Then we should define uv_thread_t in Windows as below, shouldn't we?
typedef struct {
  HANDLE thread;
  unsigned thrdaddr;
} uv_thread_t;

And we should add a new function uv_thread_id().

unsigned long uv_thread_id(uv_thread_t\* tid) {
#ifdef _WIN32
  return (unsigned long) tid->thrdaddr;
#else
  return (unsigned long) *tid;
#endif
}

hnakamur:
I made another pull request.
https://github.com/joyent/libuv/pull/642
Could you review that?

PR #642, opened 2012-12-03
Add uv_thread_id(). Modify uv_thread_t on Windows to have the thread id

hnakamur:
as well as the thread handle.

### DIFF
--- a/src/uv-common.c
+++ b/src/uv-common.c
@@ -301,8 +301,9 @@ int uv_thread_create(uv_thread_t *tid, void (*entry)(void *arg), void *arg) {
   ctx->arg = arg;
 
 #ifdef _WIN32
-  *tid = (HANDLE) _beginthreadex(NULL, 0, uv__thread_start, ctx, 0, NULL);
-  if (*tid == 0) {
+  tid->thread = (HANDLE) _beginthreadex(NULL, 0, uv__thread_start, ctx, 0,
+    &tid->thrdaddr);
+  if (tid->thread == 0) {
 #else
   if (pthread_create(tid, NULL, uv__thread_start, ctx)) {
 #endif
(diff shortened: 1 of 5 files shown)


src/uv-common.c
```diff
@@ -301,8 +301,9 @@ int uv_thread_create(uv_thread_t *tid, void (*entry)(void *arg), void *arg) {
   ctx->arg = arg;
 
 #ifdef _WIN32
-  *tid = (HANDLE) _beginthreadex(NULL, 0, uv__thread_start, ctx, 0, NULL);
-  if (*tid == 0) {
+  tid->thread = (HANDLE) _beginthreadex(NULL, 0, uv__thread_start, ctx, 0,
+    &tid->thrdaddr);
```
bnoordhuis, line 305: Style: arguments need to line up.

bnoordhuis:
I'm not against this change per se but can you explain why you want it?

hnakamur:
I thought we can use uv_thread_id() to check if the thread is self, like:
uv_thread_id(tid) == uv_thread_self().

However, I found we need to use pthread_equal() to check if two threads are equal.
http://pubs.opengroup.org/onlinepubs/007904975/functions/pthread_equal.html

Now I think it may be better to add uv_thread_equal(uv_thread_t\* t1, uv_thread_t\* t2) instead of adding uv_thread_id().
But then I don't understand the supposed usage of uv_thread_self(). Could you explain why it is added?

bnoordhuis:
> But then I don't understand the supposed usage of uv_thread_self(). Could you explain why it is added?

node.js uses it for its openssl bindings: https://github.com/joyent/node/blob/1f76a2e/src/node_crypto.cc#L88

hnakamur:
Thanks for explanation. I cancel this pull request.
Sorry to bother you.

bnoordhuis:
No problem. Yours is the more correct approach actually but uv_thread_self() is somewhat easier in everyday use.

PR #644, opened 2012-12-04
Fix Makefile build system for 'make test' to run ok on Linux.

hnakamur:


shigeki:
Duplicated in #632 . Need kvm link flags on *BSD but not sure on MacOS. 
@bnoordhuis Any thoughts?

bnoordhuis:
So the reason I switched it to dynamic linking is that I want to iron out the bugs in the .so build. If you run into issues, please report them and I'll (help you) fix them.

shigeki:
@bnoordhuis I understand you need a check in a shared library. An link error of src/async.o as is shown in #632 . It occurs after `make` then `make test` where objects are not compiled with -fPIC.

hnakamur:
I close this pull request as this is duplicate of #632.
Please comment on #632 for further discussion.

PR #650, opened 2012-12-12
unix: rework uv_cancel()

bnoordhuis:
Bert Belder informs me the current approach, where a request is immediately
cancelled, is impossible to implement on Redmond OS.

Rework the API to always invoke the "done" callback with an appropriate status
code.

Reviewers: @indutny @piscisaureus

TBD: Change the prototype of uv_cancel to `void uv_cancel(uv_req_t*)`, i.e. don't return a status code?


indutny:
lgtm

piscisaureus:
lgtm

bnoordhuis:
Landed in 92fb84b, thanks for reviewing.

PR #652, opened 2012-12-12
WIP Make IPV6ONLY the default behaviour on systems supporting dual stack.

txdv:
Instead of UV_UDP_IPV6ONLY we have now UV_UDP_DUALSTACK, which will try to turn
on the dual stack behaviour.


bnoordhuis:
What is the motivation for this change? Most platforms are dual stack by default.

txdv:
Except the platforms which don't support dualstack.
So I have a different behaviour of uv_udp_bind6(handle, addr, 0) on platforms which support it and on platforms which don't support it (on one it will be dualstack, on the other it will be only ipv6).

bnoordhuis:
@piscisaureus Opinions?

PR #653, opened 2012-12-12
unix: fix small write immediately followed by uv_read_stop never completes

piscisaureus:
uv__io_feed() would not update the pending events masked to include
UV__POLLOUT, so a subsequent uv__io_stop() call to remove the
UV__POLLIN event would cause the pevents mask to become zero. When
the pevents mask would be zeroed, the io watcher would be removed from
the pending queue.

@bnoordhuis review?
/cc @isaacs This should fix your problem.


isaacs:
Confirmed that this fixes the problem I was seeing.

A test in libuv would be nice, but I can of course put my big-write test cases into node's tests.

PR #654, opened 2012-12-13
Solaris directory watching resource leak / segfault bug fix

ashaffer:
Libuv doesn't properly clean up directory watches on solaris if they are in the PORT_LOADED state.  I encountered this problem in node.js.  To reproduce it, execute the following in a node console on solaris:

```
fs = require('fs');
h = fs.watch(<directory>);
h.close();
```

Then in a separate shell (without closing the node shell, of course) touch the folder you watched and you should get a seg fault.


bnoordhuis:
The change itself LGTM but would it be possible to add a regression test?

ashaffer:
Added.

bnoordhuis:
Thanks. Can you sign [the CLA](http://nodejs.org/cla.html)? I believe you're not on the list yet.

ashaffer:
Ok, signed.

bnoordhuis:
Thanks Andrew, landed in 4997738. It'll be part of the next v0.8 node release.

PR #656, opened 2012-12-17
kqueue: ignore ENOENT errors when EV_EOF

indutny:
Some sockets are automatically removed from kqueue upon their
termination, and doing EV_DELETE results in ENOENT. It could be safely
ignored as it isn't a big deal anyway.

This blocks me from using tlsnappy with node 0.9.x, as it `abort()`s suddenly on a benchmark.

### DIFF
--- a/src/unix/kqueue.c
+++ b/src/unix/kqueue.c
@@ -197,7 +197,9 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
           /* TODO batch up */
           struct kevent events[1];
           EV_SET(events + 0, fd, ev->filter, EV_DELETE, 0, 0, 0);
-          if (kevent(loop->backend_fd, events, 1, NULL, 0, NULL)) abort();
+          if (kevent(loop->backend_fd, events, 1, NULL, 0, NULL))
+            if (errno != ENOENT)
+              abort();
         }
       }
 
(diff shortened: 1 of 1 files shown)


src/unix/kqueue.c
```diff
@@ -197,7 +197,10 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
           /* TODO batch up */
           struct kevent events[1];
           EV_SET(events + 0, fd, ev->filter, EV_DELETE, 0, 0, 0);
-          if (kevent(loop->backend_fd, events, 1, NULL, 0, NULL)) abort();
+          if (kevent(loop->backend_fd, events, 1, NULL, 0, NULL) &&
+              !(ev->flags & EV_EOF)) {
```
bnoordhuis: Shouldn't you check for ENOENT here?

indutny: Yes, I can do this too...

indutny: Force pushed.

indutny:
Ok, I nailed it down... Please refer commit message. /cc @bnoordhuis

src/unix/kqueue.c
```diff
@@ -167,6 +167,15 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
       fd = ev->ident;
       w = loop->watchers[fd];
 
+      /* File descriptor might be closed during callback,
+       * all events that was reported before callback are not valid in such
+       * cases and can be safely ignored
+       */
+      if (w != ev->udata) {
```
bnoordhuis: What happens if the fd is closed, then reopened and added with the same watcher?

Also, this effectively turns the NULL check below into dead code. Probably not what you want.

indutny: Actually, after sleeping well, I think that ignoring ENOENT is enough.

indutny:
Force pushed. @bnoordhuis

PR #660, opened 2012-12-18
 build libuv.a and libuv.so in different directories fix #659

### LINKED ISSUE #659: build: build libuv.a and libuv.so in different directories
See e.g. #632.

Running `make all && make test` fails to build the test suite.

`make all` first builds libuv.a, `make test` then tries to build libuv.so but dies with relocation errors.

Possible solutions:
1. Always build with -fPIC. Bad but easy.
2. Build libuv.a and libuv.so in different directories. Good but hard(er).

vtudose:


bnoordhuis:
Build output should go in out/ (build/ is used by gyp) but apart from that LGTM, thanks! Vlad, can you sign [the CLA](http://nodejs.org/cla.html)?

vtudose:
I've signed the CLA.
I changed the folder to out/
I also modified the targets to take into account the .h files

shigeki:
@vtudose Awesome!  #632 is also resolved. I've never think of MAKECMDGOALS . Thanks.

bnoordhuis:
Thanks Vlad, landed in 4b115f8.

PR #662, opened 2012-12-21
linux-core.c: return time values in ticks, not seconds

tfeb:
(still less the bogus values that were being returned previously)

CPU time fields were being spuriously multiplied by the ticks/second value resulting in insane values on Linux.  This patch changes that so they're just returned as raw ticks.

Since node.js offers no way to get the ticks/second value this is still not very useful but it does mean the values actually correspond to what the OS thinks.

In an ideal world I think it would be better to either provide an interface to getting the ticks/second value from the OS, or to return CPU times in seconds.  Both of those would be fairly easy changes I think: probably the former is better as it does not change things incompatibly (though I assume no one has being relying on the existing values!)


bnoordhuis:
I can't take this patch, it's inconsistent with what libuv does on other platforms.

> or to return CPU times in seconds

I think that was the intended goal. (The original code wasn't written by me; it's bug-for-bug adapted from node v0.4.)

Does this patch look acceptable to you?

``` diff
diff --git a/src/unix/linux/linux-core.c b/src/unix/linux/linux-core.c
index 95df667..102be43 100644
--- a/src/unix/linux/linux-core.c
+++ b/src/unix/linux/linux-core.c
@@ -640,11 +640,11 @@ static void read_times(unsigned int numcpus, uv_cpu_info_t* ci) {
                     &irq))
       abort();

-    ts.user = clock_ticks * user;
-    ts.nice = clock_ticks * nice;
-    ts.sys  = clock_ticks * sys;
-    ts.idle = clock_ticks * idle;
-    ts.irq  = clock_ticks * irq;
+    ts.user = user / clock_ticks;
+    ts.nice = nice / clock_ticks;
+    ts.sys  = sys / clock_ticks;
+    ts.idle = idle / clock_ticks;
+    ts.irq  = irq / clock_ticks;
     ci[num++].cpu_times = ts;
   }
   fclose(fp);
```

tfeb:
I'm slightly confused by this, I think.

What I think is the _right_ thing to do is return times in seconds, as node provides no way of knowing how long a tick is, so times in ticks are not really useful other than for comparison).

However what I looked at to work out what other platforms do was two things: firstly the documentation (http://nodejs.org/api/os.html#os_os_cpus) talks about ticks, not seconds,  and secondly the Solaris code (as I know Solaris somewhat) does things like:

```
  knp = (kstat_named_t *) kstat_data_lookup(ksp, (char *)"cpu_ticks_user");
  assert(knp->data_type == KSTAT_DATA_UINT64);
  cpu_info->cpu_times.user = knp->value.ui64;
```

So that looks as if Solaris is returning times in ticks, as well.  I confess to not having checked other platforms however.

So given that, I decided to make my changes return ticks.  I would be very happy for it to return seconds instead and I have fixes for that which are pretty much the same as yours but a little more careful about rounding.  I can submit pull requests for those as well if you like.

(Sorry if I am botching the process here, I've not done things via Gitub previously so I don't really know the protocol!)

--tim

tfeb:
On 29 Dec 2012, at 21:19, Tim Bradshaw wrote:

> So given that, I decided to make my changes return ticks.  I would be very happy for it to return seconds instead and I have fixes for that which are pretty much the same as yours but a little more careful about rounding.  I can submit pull requests for those as well if you like.

FWIW those changes are now in a branch called tfeb.cpu-fix-seconds, which I can send a pull request for if that seems like right.

bnoordhuis:
> What I think is the _right_ thing to do is return times in seconds, as node provides no way of knowing how long a tick is, so times in ticks are not really useful other than for comparison).

Yes, seconds or possibly milliseconds.

It's not unlikely that the actual code does something different; most of the code comes straight from node v0.4 and was patched together by a small army of volunteers.

A quick look at the Windows implementation suggests in returns the times in milliseconds.

tfeb:
On 31 Dec 2012, at 01:15, Ben Noordhuis wrote:

> Yes, seconds or possibly milliseconds.
> 
> It's not unlikely that the actual code does something different; most of the code comes straight from node v0.4 and was patched together by a small army of volunteers.
> 
> A quick look at the Windows implementation suggests in returns the times in milliseconds.

OK, I vote that: you ignore this pull request (or I'll withdraw it); I'll submit another one with two commits: one for Linux which makes things work in ms (and which I can test before sending) and one for Solaris which does the same thing but which I can't currently test.  Then if you have a Solaris test platform the second commit can be tested, but if not you can just apply the Linux one.

There needs to be a corresponding change to the node.js documentation to say things are in ms, not ticks I think.

Does that sound right?

bnoordhuis:
Yes, sounds good. The BSD + Darwin implementations will need to be reviewed as well though I believe they already do the right thing. I'll close this PR.

PR #664, opened 2012-12-23
uv_fs_skip() (or perhaps uv_fs_seek() ?)

rvagg:
What are the chances of getting an `lseek()` / `SetFilePointerEx()` in? Not sure about the naming, I went with _skip_ but perhaps _seek_ would be a better choice.


bnoordhuis:
What do you need lseek() for? Phrased differently: why would you use lseek() instead of a positional read or write?

Keep in mind that file operations run in a thread pool in undefined order. In such an environment, lseek() is at best useless and possibly harmful.

rvagg:
Harmful, in the same way that other async I/O ops in libuv are harmful when order isn't managed properly? I'm not suggesting this be added to Node, just fleshing out libuv's fs ops, for use by people who need such functionality and have a clue. Caveat emptor.

I need this for a port of LevelDB to libuv, it requires a separate skip/seek call and I'm actually only interested in the sync version but it's not totally unreasonable that someone may want an asynchronous seek.

bnoordhuis:
> for use by people who need such functionality and have a clue. Caveat emptor.

That probably means I'll be explaining the caveats on IRC for years to come. :-/

I'm generally -1 on this PR but I'll let @piscisaureus weigh in as well.

> I need this for a port of LevelDB to libuv, it requires a separate skip/seek call and I'm actually only interested in the sync version

"Requires" as in "that's what it uses now"? I would recommend using positional reads / writes, even when you're doing just synchronous I/O. It saves a syscall and it'll let you implement concurrent readers without undue hardship later on.

It'll probably be a little faster as well. Most all Unices implement read() like this, basically:

``` c
int sys_read(int fd, void *buf, size_t n)
{
  struct file *file = fdget(fd);
  if (file == NULL) return -EBADF;
  return do_pread(file, buf, n, atomic_read(&file->pos));
}
```

lseek() + read() is essentially a slightly less efficient way of doing pread().

rvagg:
"Requires" as in it's an extension point for custom _environments_: https://github.com/rvagg/node-levelup/blob/master/deps/leveldb/leveldb-1.7.0/include/leveldb/env.h#L177

The current posix implementation uses a simple [`fseek`](https://github.com/rvagg/node-levelup/blob/master/deps/leveldb/leveldb-1.7.0/util/env_posix.cc#L63) but there's currently no satisfactory Windows port, which is what I'm trying to use libuv for.

bnoordhuis:
You can maintain a `position_` field in your implementation, right? I.e. when Skip() is called, you set `position_ += skip`, same after a successful Read().

rvagg:
yes, that's an alternative but I'd also need to know the current length of the file or things could get crazy, consider a _skip_ that goes beyond the end of the file, followed by a _skip_ that goes backwards. A proper OS implementation will stop at the end of the file for the first so the second would go back the correct amount from the end of the file, while a naive solution would go beyond the end of the file and come back the wrong amount.

If I can't get this into libuv then I'll just have to implement my own. If you **_really**_ don't want it then it's your choice and that's fine by me. However, `seek` is a standard fs operation that's missing from libuv.

rvagg:
fwiw, the Node docs on `fs.read()` have this to say about `position`:

> `position` is an integer specifying where to begin reading from in the file. If `position` is `null`, data will be read from the current file position.

Obviously the current position can be affected by a `read()` but surely there's also a case for being able to move it around without a `read()`?

ghost:
Is there a way forward here with this. Having leveldb work on windows would be pretty useful for lots of use cases guys !!

piscisaureus:
I am with Ben here. This is not what people should be doing.

@rvagg 

> A proper OS implementation will stop at the end of the file for the first so the second would go back the correct amount from the end of the file, while a naive solution would go beyond the end of the file and come back the wrong amount.

That's not true. Posix operating systems won't let you seek past the end of file, the lseek() call will fail and errno is ENXIO. On windows, SetFilePointerEx will just seek beyond the end of file.

I don't think that's true.

Issue #625, opened 2012-11-15
Addition of shared libraries broke MinGW build


luislavena:
Hello,

After cc36fd036606a3dec452b6e40ded41db6e5bf322 it is no longer possible to compile tests or benchmarks under Windows with GCC:

```
C:\Users\Luis\Code\joyent\libuv>make test
make: *** No rule to make target `libuv.so', needed by `test/run-tests.exe'.  Stop.
```

Even if we add the necessary changes, all the C files needs to be compiled with `BUILDING_UV_SHARED` and then tests with `USING_UV_SHARED`

I wonder when gyp will support MinGW so we can stop worrying about the makefiles?

bnoordhuis:
> I wonder when gyp will support MinGW so we can stop worrying about the makefiles?

I wouldn't put your hopes up. By and large, the gyp people only care about the platforms they use themselves. MinGW isn't one of them.

I'll take patches for the build issue (or maybe @piscisaureus can fix it). It should be pretty straightforward:
1. Add a libuv.so target to config-mingw.mk that compiles with`CPPFLAGS+=-DBUILDING_UV_SHARED`
2. Add `-DUSING_UV_SHARED` to RUNNER_CFLAGS.

You can look at config-unix.mk for inspiration.

vtudose:
Sorry for the mess, I accidentally referenced  the commits twice.

saghul:
We do have a MinGW makefile now, closing.

luislavena:
@saghul there was a MinGW makefile in the past, but it failed to build tests (that is why this was opened)

Do not have time right now to verify if that issue was sorted out after, but will during the weekend.

Can you confirm the issue mentioned above was solved?

saghul:
That's right, it doesn't build tests, sorry about that. Last update was a year ago, so feel free to open a PR with a patch when you have the time :-)

PR #665, opened 2012-12-23
fix MinGW build #625

vtudose:


src/win/process.c
```diff
@@ -379,7 +379,7 @@ static WCHAR* search_path(const WCHAR *file,
  * Quotes command line arguments
  * Returns a pointer to the end (next char to be written) of the buffer
  */
-WCHAR* quote_cmd_arg(const WCHAR *source, WCHAR *target) {
+UV_EXTERN WCHAR* quote_cmd_arg(const WCHAR *source, WCHAR *target) {
```
bnoordhuis: Shouldn't that be `static` instead of `UV_EXTERN`? Those functions aren't used outside of process.c.

src/win/util.c
```diff
@@ -87,7 +87,7 @@ int uv_utf16_to_utf8(const WCHAR* utf16Buffer, size_t utf16Size,
 }
 
 
-int uv_utf8_to_utf16(const char* utf8Buffer, WCHAR* utf16Buffer,
+UV_EXTERN int uv_utf8_to_utf16(const char* utf8Buffer, WCHAR* utf16Buffer,
```
bnoordhuis: This function isn't exported, is it?

src/win/process.c
vtudose: Yes they are.
The function is used in test\test-spawn.c:683.

src/win/util.c
vtudose: This one is used in test\test-spawn.c:745.

bnoordhuis: Ah, indeed. I see that it's defined in include/uv-private/uv-win.h. How nasty.

piscisaureus:
These functions aren't actually for public use. I'm okay with exporting `quote_cmd_arg` but you should probably rename it to `uv__quote_cmd_arg` then.

As for `uv_utf8_to_utf16`, I am slowly removing all the uses of it. I thing you'd better replace it's usage in test-spawn.c with a MultiByteToWideChar call.

vtudose:
@piscisaureus, @bnoordhuis  all fixed.

bnoordhuis:
Closing, no longer applies. Thanks though.

PR #666, opened 2012-12-24
unix: fix running loop when all handles are unref'd

saghul:
Hi!

This pull request contains 2 fixes: the first commit disallows NULL to be passed as a callback to prepare/idle/check handles and modifies the tests accordingly. 

After the first commit, the build fails on unix (test-ref) because the event loop will run once even if all handles are unref'd. This wasn's apparent because the tests in test-ref used NULL as a callback. The second commit fixes this.

/cc @bnoordhuis @piscisaureus

### DIFF
--- a/src/unix/core.c
+++ b/src/unix/core.c
@@ -267,8 +267,18 @@ int uv_backend_timeout(const uv_loop_t* loop) {
 }
 
 
+static int uv__loop_alive(uv_loop_t* loop) {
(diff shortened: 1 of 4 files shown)


saghul:
Thanks for the lighning fast review @bnoordhuis ;-) I just force pushed with the suggested changes.

bnoordhuis:
Thanks Saúl, landed in 33d5c49 and 4650a59.

saghul:
Awesome, thanks!

PR #669, opened 2012-12-26
unix: Fix canceling request if it's the only one

saghul:
This is a fix for the following case (which I surprisingly run into): there is a single request in the thread pool (say a getaddrinfo) and it's cancelled before it gets a chance to run. If no other request is put in the thread pool, the work_async will not be sent and thus the done callback will never be called. The fix consists in sending the async in case the cancelled request is the last request.

I don't know if a reliable test case can be written for this, though :-S

/cc @bnoordhuis

### DIFF
--- a/src/unix/threadpool.c
+++ b/src/unix/threadpool.c
@@ -141,6 +141,7 @@ void uv__work_submit(uv_loop_t* loop,
 
 int uv__work_cancel(uv_loop_t* loop, uv_req_t* req, struct uv__work* w) {
   int cancelled;
+  int wq_empty;
 
   uv_mutex_lock(&mutex);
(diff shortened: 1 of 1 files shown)


bnoordhuis:
@saghul Can you test 5ae28c2? Quoting the commit log:

```
Wake up the event loop with uv_async_send() when a request is cancelled.
Ensures the done_cb is run on the next tick of the event loop.

Not sending a wakeup signal results in the done_cb not getting called until
another request completes, which may be a long time coming when it's the only
request in the queue or when other requests are executing long-running jobs.
```

saghul:
@bnoordhuis that works :-)

bnoordhuis:
Thanks for testing, landed in 92a19a1.

Issue #671, opened 2012-12-30
unix: Loop doesn't iterate when a signal is received


saghul:
Before libev was removed, if a signal was received when the loop was blocked for io, the signal would "wakeup" the loop (the poll would be interrupted with EINTR as far as I see) and another loop iteration would begin.

This is no longer the case and I kind of relied on it for interfacing with Python signal handlers in the following way:
- Have a internal prepare handle
- Call a specific Python API function which checks is there are pending signals and executes their respective callbacks

I know we have uv_signal_t now, but that means that the standard Python signal module can't be used together with pyuv. I'm ok with that but I'd like to know if this is a bug or a feature :-)

I played around a bit with the code by just doing `return` here: https://github.com/joyent/libuv/blob/master/src/unix/kqueue.c#L152 which is apparently what libev did, but I'm not sure if that would open a can of worms.

/cc @bnoordhuis

saghul:
Hmm, after some more testing things look worse than I expected :-S In Python, when Ctrl-C is pressed KeyboardIterrupt is raised unless a handler for SIGINT is registered with the builtin signal module. I'm unable to get this behavior working with uv_signal_t handles, becuase libuv is handling the signal. If possible it would be nice to have the old behavior, where the loop would iterate after being interrupted by a signal back :-)

bnoordhuis:
The current behavior (retry on EINTR) is intentional; I've always considered the libev behavior borderline buggy.

The reason is that prepare and idle handles are only supposed to run when the event loop is making some kind of forward progress (timeout, I/O pending, etc.). Not retrying on EINTR means they also run when nothing actually happened.

What solutions are possible that don't involve reverting to the old behavior?

saghul:
I see. I supposed there was some reason behind it since the code didn't look arbitrary.

I'm not sure what can be done to achieve the same behavior (in pyuv) that doesn't involve reverting to the old behavior. If timeout is 0 then the poll will return on EINTR, but there are many cases where timeout will be -1 or some other value, so it's very easy to run into the problem. Basically I'd need to be somehow notified when a signal is received so that I can call the Python API function to process it. If you have some other idea I'm all ears :-)

I sent a pull request reverting libuv to the old behavior and tests seem to pass. Do you have practical cases where this would be a problem?

saghul:
I'm not sure if anybody would use it (other than me), but maybe it could be made configurable? By using a loop flag:

```
if (loop->flags & UV_ITERATE_ON_EINTR)
    return;
[...]

PR #672, opened 2012-12-30
unix: Make a loop iteration if polling is interrupted by a signal

saghul:
Possible fix for #671. Take with a grain of salt ;-)


saghul:
Looks like this may not be needed after all, see comments on #671.

PR #673, opened 2012-12-31
CPU times should be reported in ms

tfeb:
This changes CPU times to be reported in milliseconds, not ticks:
- for Linux this is done by 38032ec: this commit has been tested reasonably well;
- for Solaris this is done by e79d4e7: I haven't tested this as I don't currently have a Solaris VM to try it on although I'm reasonably sure it is OK.

In both cases the division is done slightly carefully to round rather than truncate (ie `9/10` should be `1` not `0` which is achieved by `(9 + 10/2)/10`): this probably does not matter much now but it did in an earlier version of things where I had things returning seconds.

I have also checked (by reading, not building) the Darwin and *BSD equivalents of this code and I think they all return things in ms as well.

I _haven't_ checked the AIX code: I don't know AIX and don't even potentially have a machine to test it on, so I'm not sure what it does.


src/unix/sunos.c
```diff
@@ -528,21 +533,28 @@ uv_err_t uv_cpu_info(uv_cpu_info_t** cpu_infos, int* count) {
       cpu_info->cpu_times.idle = 0;
       cpu_info->cpu_times.irq = 0;
     } else {
+      /* conversions to ms below have good rounding but potential
+       * overflow */
       knp = (kstat_named_t *) kstat_data_lookup(ksp, (char *)"cpu_ticks_user");
       assert(knp->data_type == KSTAT_DATA_UINT64);
-      cpu_info->cpu_times.user = knp->value.ui64;
+      cpu_info->cpu_times.user = (1000 * knp->value.ui64 + clock_ticks/2)
+	/ clock_ticks;
```
bnoordhuis, line 541: Style error.

src/unix/linux/linux-core.c
```diff
@@ -640,11 +640,17 @@ static void read_times(unsigned int numcpus, uv_cpu_info_t* ci) {
...
+     * most accurate way of doing this but is more liable to overflow.
+     * I don't think this matters as anything using these values needs
+     * to cope with potential overflows.
+     */
+
+    ts.user = (1000 * user + clock_ticks/2) / clock_ticks;
+    ts.nice = (1000 * nice + clock_ticks/2) / clock_ticks;
+    ts.sys  = (1000 * sys + clock_ticks/2) / clock_ticks;
+    ts.idle = (1000 * idle + clock_ticks/2) / clock_ticks;
+    ts.irq  = (1000 * irq + clock_ticks/2) / clock_ticks;
```
bnoordhuis, line 653: I suggest you do away with banker's rounding here. As you say, it's not that relevant when counting in milliseconds; if someone else reads the source, he'll have to do a double-take to figure out _what_ it's doing, let alone _why_.

tfeb, line 653: On 8 Jan 2013, at 04:36, Ben Noordhuis wrote:

> I suggest you do away with banker's rounding here. As you say, it's not that relevant when counting in milliseconds; if someone else reads the source, he'll have to do a double-take to figure out what it's doing, let alone why.

I don't mind if you do, but I won't. The designers of C might have thought it was OK to save some tiny amount of time by doing arithmetic wrong (and this might even have been justified on a PDP11, though it was clearly short-sighted even then) but I'm having none of it: I don't want anything with my name on it which is just that sloppy.

bnoordhuis:
Closing, no longer applies. Thanks though.

PR #674, opened 2013-01-02
stream: fix infinite select() polling on osx

indutny:
Interruption FD buffer should be emptied after write, otherwise select()
loop will poll indefinitely.

/cc @bnoordhuis


saghul:
Sorry if it's a dumb question, but, can `read` be interrupted after it has read some data? That is, is it possible that r is > 0 and errno is EINTR because there is data to be read?

indutny:
@saghul no, it can't:

```
If successful, the number of bytes actually read is returned.
Upon reading end-of-file, zero is returned.
Otherwise, a -1 is returned and the global variable errno is set to indicate the error.
```

saghul:
Thanks for clarifying it, Fedor.

src/unix/stream.c
```diff
@@ -128,10 +128,10 @@ void uv__stream_init(uv_loop_t* loop,
...
   uv_stream_t* stream;
   uv__stream_select_t* s;
-  fd_set read;
-  fd_set write;
-  fd_set error;
-  struct timeval timeout;
+  char buf[1024];
+  fd_set sread;
+  fd_set swrite;
+  fd_set serror;
```
bnoordhuis, line 134: Not really related to this pull request but do you expect the error set to ever have anything in it? I'm not sure about OS X, but on Linux it's only useful for out-of-band data on TCP sockets and monitoring sysfs changes.

src/unix/stream.c
```diff
@@ -178,13 +177,33 @@ void uv__stream_osx_select(void* arg) {
     if (r == 0)
       continue;
 
+    /* Empty socketpair's buffer in case of interruption */
+    if (FD_ISSET(s->int_fd, &sread))
+      while (1) {
+        r = read(s->int_fd, buf, sizeof(buf));
+
+        if (r == sizeof(buf))
```
bnoordhuis, line 185: What if you read less than 1024 bytes?

bnoordhuis, line 185: Oh wait, that's caught a few lines below (r != -1). Seems like you could coalesce those.

indutny, line 185: I've copied this from async.c

bnoordhuis, line 185: Oh, okay. Still - don't trust the guy who wrote that.

bnoordhuis:
LGTM

indutny:
Thanks, landed in ba83510

PR #676, opened 2013-01-04
Adding uv_break, because sometimes the world is not perfect :-)

saghul:
I commented this a while ago on the mailing list and today I had some hours to kill at an airport so here it is. I couldn't comment on the implementation since I had no Internet when I was writing it, so all comments are welcome :-)

/cc @bnoordhuis @piscisaureus


bnoordhuis:
A libuv user can accomplish the same thing by combining UV_RUN_ONCE or UV_RUN_NOWAIT and a global flag, right? I don't particularly like adding API functions that are easily emulated in a user's code.

saghul:
It can be kind of emulated, but it can be expensive. In pyuv I don't like the idea of inventing APIs libuv doesn't have so I'd have to do the loop in Python, and going back and forth between C and Python has an overhead. This is of course a particularity of how pyuv works, but IMHO uv_break fits in, and I (for one) would take advantage of it :-)

saghul:
Is there anything I can do to convince you this is a nice to have thing? :-)

txdv:
What? I love inventing APIs that libuv doesn't have.
I have coded RunOnce before your run2 patch.

In what scenarios would you use it?

saghul:
pyuv is a shallow wrapper around libuv, not a framework built on top of it. Of course it's adapted to the Python universe/syntax, but right now I don't have any "extra" APIs, and I'd rather stay that way.

I've used pyuv to monkeypatch different frameworks like Twisted or Tornado, and some of these provide 3 functions for a given loop: run, stop and close. Run does the obvious, that is, run the loop until there is nothing left to be done or stop is called, stop just stops the loop and close frees all resources and fds associated with the loop.

So, that would be very easy to implement as follows: run -> uv_run(default), stop -> uv_break, close -> uv_walk + uv_close all handles + uv_run(nowait).

Yes, I can do it directly in Python, but it's a lot slower :-S

Also, since we can run a loop it makes sense to me that we should be able to stop it at will.

Maybe s/uv_break/uv_stop/ ?

txdv:
Yeah, stop would be nice, i'm for this.

saghul:
Hi, it's stubborn @saghul again, last attempt :-)

I wrote a silly example which involves a loop that runs until a idle handle has made a million iterations. In one case it's stopped with loop.stop (I'm thinking that uv_stop looks better that uv_break) and the other has a global variable to do the same trick. Here are the results: https://gist.github.com/4497505

As it can be seen there is a performance penalty in doing the custom loop thing (due to Python being used here), but I don't want to focus on that. IMHO, the version with the run once loop looks a bit weird while the other one is more straightforward and just Looks Better (TM).

Not that it matters much, but similar libraries such as libevent and libev do provide a similar mechanism. The semantics of uv_break are, however, simpler: end the event loop as soon as possible, actually in the next loop iteration. It's up to the user to decide if he wants to let the loop block for io or he wants to start an idle handle to prevent it from blocking.

IMHO this is a convenient API function with a clear meaning and a really simple implementation, so it's not like I'm asking to merge a bunch of crap anyway ;-)

PS: If you guys are willing to merge it, uv_break or uv_stop?

saghul:
Ok, after some more thoughts I simplified the implementation and renamed uv_break to uv_stop. I made an important change in the behavior: if uv_stop is called _before_ blocking for i/o, it will prevent the blocking. The reason is that if you called uv_stop you _do want_ to stop, not wait in case a repeating timer or something would make the loop block for i/o.

Now, with the aforementioned change in semantics implementing it manually becomes trickier because UV_RUN_ONCE would block and UV_RUN_WAIT would prevent blocking for i/o always, so UV_RUN_ONCE would need to be used in combination with an idle handle for example. Still doable, but uv_stop doesn't require an extra handle :-)

bnoordhuis:
@saghul You mention that going back and forth between Python and C land is slow. Can you quantify that?

And when is that an actual issue? I mean, how often do you need to break out of the event loop?

saghul:
@bnoordhuis Thanks for coming back to this, Ben. I was hoping that we could reach some common ground before 0.10.0.

I did a performance benchmark a while ago (https://gist.github.com/4497505), but please just don't look at it. The performance argument is not the most important reason why I want uv_stop.

As for how often do I need to break out of the loop, actually almost always. Twisted and Tornado are the most used event loops in Python and they both have the concept of "running forever" until the user calls stop o the event loop. I wrote pyuv reactors for both frameworks and had to use some sort of trick to break out of run while still leaving the loop in a usable state, because it can be run later again.

The upcoming event loop implementation for the standard library (Tulip) also has this concept.

Now, I could understand why uv_stop wasn't that useful in its first implemented form, because it was nothing more than a flag. The current code up for review, however, takes an extra step and prevents the loop from blocking for i/o on that iteration, which IMHO is a good thing.

I know there is no use for this in Node, but I think it's a useful construct to have. Please, please merge it :-)

saghul:
@bnoordhuis I just rebased the patch with latest master and force pushed.

src/unix/core.c
```diff
@@ -286,12 +286,17 @@ int uv_run(uv_loop_t* loop, uv_run_mode mode) {
...
+      return uv__loop_alive(loop);
+    }
     uv__update_time(loop);
     uv__run_timers(loop);
     uv__run_idle(loop);
     uv__run_prepare(loop);
     uv__run_pending(loop);
-    uv__io_poll(loop, (mode & UV_RUN_NOWAIT ? 0 : uv_backend_timeout(loop)));
+    uv__io_poll(loop, ((mode & UV_RUN_NOWAIT ||
+                       loop->stop_flag) ? 0 : uv_backend_timeout(loop)));
```
bnoordhuis: I'd rather the conditional is split off into one or more separate statements. Ditto for the conditional in src/win/core.c. Otherwise LGTM.

saghul: You mean something like this?

```
int timeout;

...

if (mode & UV_RUN_NOWAIT || loop->stop_flag)
    timeout = 0;
else
    timeout = uv_backend_timeout(loop);
uv__io_poll(loop, timeout);
```

bnoordhuis: Yes, but with fixed indentation. :-)

bnoordhuis:
Landed in 4b95748, 492efb9, bb3d1e2 and 7988012.

PR #677, opened 2013-01-05
Windows build: support vs2012, fix shared

gigi81:
Added support for Visual Studio 2012(updated gyp to the latest available
version)
Fixed shared build on windows (run-tests was failing)


uv.gyp
```diff
@@ -353,6 +353,51 @@
             '_XOPEN_SOURCE=500',
           ],
         }],
+        ['OS=="win" and library=="shared_library"', {
```
bnoordhuis, line 356: Why is this section necessary?

gigi81, line 356: Without that section when building the shared(dll) version of the library you get this error:

test-spawn.obj : error LNK2019: unresolved external symbol _uv_utf8_to_utf16 referenced in function _run_test_environment_creation [C:\Users\Luigi\Documents\GitHub\libuv\run-tests.vcxproj]
test-spawn.obj : error LNK2019: unresolved external symbol _make_program_args referenced in function _run_test_argument_escaping [C:\Users\Luigi\Documents\GitHub\libuv\run-tests.vcxproj]
test-spawn.obj : error LNK2019: unresolved external symbol _quote_cmd_arg referenced in function _run_test_argument_escaping [C:\Users\Luigi\Documents\GitHub\libuv\run-tests.vcxproj]
test-spawn.obj : error LNK2019: unresolved external symbol _make_program_env referenced in function _run_test_environment_creation [C:\Users\Luigi\Documents\GitHub\libuv\run-tests.vcxproj]

bnoordhuis, line 356: Isn't that a matter of fixing the sources list in the normal OS=="win" section? This seems like a rather massive duplication.

gigi81, line 356: It's the cleanest solution I can think of.
This section is only needed when building the dll on windows. Other solution would involve to change multiple sections that don't actually need this change. The change is a single block (that can be easily removed if the future if not needed anymore) so I think it's really clean.
As long as the test program is using library staff that is not exported by the library (because not needed) this change is required to make the build work.

gigi81:
@bnoordhuis any update? planning to merge this? at lest merge the bat file. even if the dll build will be left broken at least we have a build on vs2012

bnoordhuis:
@piscisaureus should review it (he works on the Windows side the most) but he's rather busy.

saghul:
VS2012 support already landed long ago, please one a new pull request if there is anything else to be fixed there, thanks!

PR #689, opened 2013-01-18
Fix incosistent error code handling in tcp_bind.

txdv:
Just like tcp_connect and udp_bind it should return an EINVAL when the
handle is of an invalid type or when the network address is faulty.

The error codes are inconsistent when checking for bad addresses:

https://github.com/joyent/libuv/blob/master/test/test-tcp-connect-error.c#L60
https://github.com/joyent/libuv/blob/master/test/test-tcp-bind-error.c#L131

The first returns EINVAL, the other EFAULT. I discussed with bnoordhuis this on the channel, it shoulds return EINVAL, because EFAULT means bad memory.

Furthermore, uv_tcp/udp_bind and uv_tcp_connect does checks on the handle and address but uv_udp_send doesn't. Should it check for a valid handle and a valid address too?


bnoordhuis:
Thanks, landed in 017e2d5.

PR #691, opened 2013-01-20
gyp: do not declare POSIX_C_SOURCE on freebsd

indutny:
On FreeBSD sys/cdefs.h contains following lines:

```
#if defined(_POSIX_SOURCE) && !defined(_POSIX_C_SOURCE)
#define _POSIX_C_SOURCE     198808
#endif
#ifdef _POSIX_C_SOURCE
  /* omitted */
#else

#if defined(_ANSI_SOURCE)   /* Hide almost everything. */
#define __POSIX_VISIBLE     0
#define __XSI_VISIBLE       0
#define __BSD_VISIBLE       0
#define __ISO_C_VISIBLE     1990
#elif defined(_C99_SOURCE)  /* Localism to specify strict C99 env. */
#define __POSIX_VISIBLE     0
#define __XSI_VISIBLE       0
#define __BSD_VISIBLE       0
#define __ISO_C_VISIBLE     1999
#else               /* Default environment: show everything. */
#define __POSIX_VISIBLE     200809
#define __XSI_VISIBLE       700
#define __BSD_VISIBLE       1
#define __ISO_C_VISIBLE     1999
#endif
#endif
```

If __BSD_VISIBLE is 0, setgroups/getgroups/uchar and etc won't be
available for external users of libuv (like node.js).

/cc @bnoordhuis


bnoordhuis:
Thanks, Fedor. I decided to take a slightly different approach in 7841f77 - only export _POSIX_C_SOURCE on Linux because it's needed there but nowhere else.

indutny:
@bnoordhuis good catch!

PR #693, opened 2013-01-22
Errorcode fix.

txdv:
A wrong membership should return an EINVAL.

This will fix the last misplaced EFAULT error setting.


bnoordhuis:
Thanks, landed in bdb498f.

PR #698, opened 2013-01-29
build: separate configuration and rules

errordeveloper:
_a)_ it's a good thing to separate these
_b)_ one can do this:

```
# cat libuv-config.mk
include config-common.mk

%:
    @echo $($@)
# make -I import/johndoe/libuv -f libuv-config.mk LDFLAGS
```

would it be of interest to add actual _`libuv-config.mk`_ from above?


bnoordhuis:
I'm afraid I landed a couple of commits just now that conflict with your PR.

> make -I import/johndoe/libuv -f libuv-config.mk LDFLAGS

When / why would you want to do that? I'm not sure I follow.

errordeveloper:
That's basically kind of intending to serve the same purpose as `libfoo-config --ldflags --cflags` does, however the original reason for digging into this was to fix an issue where on Darwin I have had to pass `-framework CoreServices` in order to compile my code with `uv.a`.
It actually looks like that was an issue in _v0.8_ and it's not in _master_ :smile:

errordeveloper:
@bnoordhuis it's fine by me if you wish to reject this as _b_ was a slightly odd use case, and I'm not too strong on _a_ after all ...

So as of _master_ (at least) one shouldn't be able to compile with `cc myprog.c libuv/uv.a`, right?

errordeveloper:
I'm taking it back, I still have to pass `-framework CoreServices`.

So the point _b_ of this PR still stands strong.

The idea of doing it this way is quite simple:
1. one needs to build something that uses _libuv_
2. the only tools at had is _make_ (no _cmake_, _gyp_ or anything of that kind) 
3. no code should be replicated from _libuv_ makefiles, as it's hard to maintain

I have only tried this on Darwin, but I suspect all the other platform-specific `LDFLAGS` will get on the way and I'd need to copy quite few lines from _`config-unix.mk` & `config-mingw.mk`_. And this PR solves the problem upstream with just splitting makefiles and no effective change of build process.

bnoordhuis:
Closing, this PR no longer applies after the switch to autotools. Thanks though.

PR #700, opened 2013-02-06
win: Avoid using subversion to download gyp

saghul:
Avoid another build time dependency, who needs subversion these days anyway? :-)


bnoordhuis:
Landed in 5fa6903.

PR #701, opened 2013-02-07
unix, windows: fix order of timer callback

shigeki:
When two timer handles have the same timeout value, callbacks should follow the order of start. `uv__is_active()` is the best to compare handles to work with it. Please review this.

This fix is requested by @saghul in  https://github.com/shigeki/libuv/commit/50d55ed3f5a363ba38c1a12efe2504049627f6fe#commitcomment-2568588 . 
The comment from Ben, https://github.com/shigeki/libuv/commit/50d55ed3f5a363ba38c1a12efe2504049627f6fe#commitcomment-1761937,  was very valuable for me to rewrite this patch but I missed it at that time. Sorry, Ben. 

[Edit] The patch is modified so as to use timer_counter


piscisaureus:
Will this work? I find it rather odd because:

(a) Timers in the tree are always active, and.
(b) if inactive timers were there, then removing a timer from the tree might fail because the compare function no longer provides an absolute ordering, so the wrong branch may be traversed.

@bnoordhuis This was your idea? :-p

shigeki:
@piscisaureus Oh, No. It's my own ideal. Ben suggest me for to think of a->loop->timer_being_inserted. Hmm, I'd better to use counter instread of this.

shigeki:
@piscisaureus Per IRC communication, I've changed this so as to use loop->timer_couner. Please review this.

bnoordhuis:
LGTM. Thanks Shigeki, landed in fadfeaf.

saghul:
Awesome! thanks @shigeki!

PR #703, opened 2013-02-08
stream: use kevent() information before accept()

indutny:
Limit number of syscalls by using backlog length information provided by
kevent().

Benchmarks are coming soon.

/cc @bnoordhuis


indutny:
@bnoordhuis landed with fixes 4deb3bf.

indutny:
Sorry, force pushed with fixed indentation: c15d4a7

PR #705, opened 2013-02-10
unix, win:  add netmask to uv_interface_address

wanderview:
Include the netmask when returning information about the OS network
interfaces.  The end goal is to get this back to node's `os.networkInterfaces()`
function.

This commit provides implementations for windows and those unix platforms
using getifaddrs().

AIX was not implemented because it requires the use ioctls and I do not
have an AIX development/test environment.  The windows code was developed
using mingw on winxp as I do not have access to visual studio.

Tested on darwin (ipv4/ipv6) and winxp (ipv4 only).  Needs testing on
newer windows using ipv6 and other unix platforms.

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -1437,6 +1437,10 @@ struct uv_interface_address_s {
     struct sockaddr_in address4;
     struct sockaddr_in6 address6;
   } address;
+  union {
+    struct sockaddr_in netmask4;
+    struct sockaddr_in6 netmask6;
+  } netmask;
 };
 
 UV_EXTERN char** uv_setup_args(int argc, char** argv);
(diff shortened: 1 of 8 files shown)


bnoordhuis:
UNIX side of things LGTM.

@piscisaureus will have to sign off on the Windows changes but he probably won't be around for the next few days.

wanderview:
Thanks for the review.  Note, I'd be willing to look at AIX if there was a VM or server I could use.  How do you normally test AIX?

bnoordhuis:
We don't test AIX at all. Someone else wrote the port, we don't really support it.

wanderview:
@bnoordhuis If the windows code will take longer to review, would it make sense to just zero the values for now as you suggested in #730?

wanderview:
Ping @piscisaureus and @bnoordhuis now that we've gotten past the push for node v0.10.0.  Willing to do more work on this if requested.

Thanks!

bnoordhuis:
Thanks Ben, landed in 14aa615.

PR #706, opened 2013-02-12
Update src/unix/linux-syscalls.h

bsnote:
replace hardcoded O_NONBLOCK and O_CLOEXEC flags with ones from glibc

This particularly fixes uv_fs_event_init function for MIPS platform (it returned -1) and probably for other platforms as well (at least enables O_NONBLOCK feature) since 0x800 is incorrect value anyway.


bnoordhuis:
The reason we have those defines is that you cannot depend on them being available in the glibc headers (notably O_CLOEXEC, it wasn't added until glibc 2.7 IIRC).

I've fixed the values for MIPS and a couple of other archs in b6f72e5.

bsnote:
It makes sense. Thanks!

PR #708, opened 2013-02-15
Add common uv_udp_* error checking.

txdv:
This makes uv_udp_\* error code checking independent of the platform.
Do not pull immediately, currently this is only the linux version.


src/uv-common.c
```diff
@@ -264,6 +262,46 @@ int uv_tcp_connect6(uv_connect_t* req,
 }
 
 
+int uv_udp_send(uv_udp_send_t* req, uv_udp_t* handle,
+                uv_buf_t bufs[], int bufcnt, struct sockaddr_in addr,
+                uv_udp_send_cb send_cb) {
```
bnoordhuis: Style: each argument on its own line.

txdv: the methods below too?

bnoordhuis: Yep.

bnoordhuis:
Landed in c5101ae, thanks.

PR #709, opened 2013-02-15
WIP Common timer functionality

txdv:
Do not pull yet, wnidows version not here.
I am going to squash the commits later on, just comment on them for now.


txdv:
I have a problem with uv_timer_set_repeat and uv_timer_get.
Both use asserts instead of returning error codes.
Now I can change uv_timer_set_repeat easily to return an int (0 for success, other for error).

But what should I do with uv_timer_get_repeat?

saghul:
Since you'd prevent the timer from getting anything < 0 on set, maybe get can return -1 on error or the actual value otherwise?

txdv:
Yeah, but we loose the errorcode, which is in this is always invalid argument.
Maybe

``` c
int uv_timer_get_repeat(uv_handle_t *handle, int64_t* value)
```

seems like an overkill to me.

saghul:
Hum, no you wouldn't. If there is an error you store it in the loop so it can be retrieved with uv_last_error, and return -1, thus telling the user that he should check it on the loop. This is how uv_fs_read works for example.

txdv:
Seems legit.

PR #710, opened 2013-02-15
Adhere to the naming conventions in uv_timer_* functions.

txdv:
In the src/*/timer.c code the first argument is called handle, not timer.
We should be consistent in the interface definition file.

This one is good to pull, please review fast.


bnoordhuis:
Landed in 7480974, thanks.

PR #712, opened 2013-02-15
Return the errorcode provided by uv_set_artificial_error.

txdv:
This reduces the line count.
This is the method suggested by Ben and I think it is a good method, because it reduces the line count in an obvious situation.


bnoordhuis:
Landed in 6bf1a56, thanks.

PR #713, opened 2013-02-16
windows: pass ENOENT on to process exit callback

sblom:
This is part of the fix to https://github.com/joyent/node/issues/4674.

The other part is https://github.com/joyent/node/pull/4780.


sblom:
Looks like we're already doing this.

PR #714, opened 2013-02-20
Safe allocation of memory for uv_fs_t

bsnote:
Recently I faced a problem of using functions which have `uv_fs_t` parameter. For example:

``` c++
uv_fs_t* req = (uv_fs_t*)malloc(sizeof(uv_fs_t));
uv_fs_stat(uv_default_loop(), req, filename, stat_cb);
```

I checked this code using valgrind. It gives errors like

```
==28480== Invalid write of size 4
==28480==    at 0x805AA3A: uv_fs_req_init (fs.c:78)
==28480==    by 0x805B738: uv_fs_stat (fs.c:373)
...
==28480==  Address 0x402aab4 is 4 bytes after a block of size 136 alloc'd
```

It turned out that the size of `uv_fs_t` in my code (136) is different from the size of `uv_fs_t` in libuv (144). Particularly the size of `statbuf` member differs, which is of type `struct stat`: 88 in my code vs 96 in libuv. I found `struct stat` definition in `<bits/stat.h>`:

``` c++
struct stat
  {
...
#if defined __x86_64__ || !defined __USE_FILE_OFFSET64
    __off_t st_size;      /* Size of file, in bytes.  */
#else
    __off64_t st_size;      /* Size of file, in bytes.  */
#endif
...
#if defined __x86_64__  || !defined __USE_FILE_OFFSET64
    __blkcnt_t st_blocks;   /* Number 512-byte blocks allocated. */
#else
    __blkcnt64_t st_blocks;   /* Number 512-byte blocks allocated. */
#endif
...
  };
```

As can be seen, the size of `struct stat` differs because of __USE_FILE_OFFSET64 definition. It is enabled in libuv by including `-D_FILE_OFFSET_BITS=64` compiler option but it is not enabled in my code. I can solve my problem by including `-D_FILE_OFFSET_BITS=64` too but I think it's tricky. It looks like a common problem and the solution is far from being obvious.

I am proposing two new public functions:
- uv_fs_t\* uv_fs_alloc()
- void uv_fs_free(uv_fs_t*)

As far as I understand `uv_buf_init()` solves a similar problem though it doesn't allocate memory and returns struct by value.

What do you think?

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -484,6 +484,15 @@ UV_EXTERN void uv_close(uv_handle_t* handle, uv_close_cb close_cb);
  */
 UV_EXTERN uv_buf_t uv_buf_init(char* base, unsigned int len);
 
+/*
+ * Memory allocator for uv_fs_t.
+ * The size of the uv_fs_t struct depends on _FILE_OFFSET_BITS definition
+ * value which can be different in user code. This function ensures that all
+ * objects of uv_fs_t type have the same size and therefore can be used
+ * safely. The user is responsible for freeing returned object.
+ */
+UV_EXTERN uv_fs_t* uv_fs_alloc();
+
 
 /*
  * Utility function. Copies up to `size` characters from `src` to `dst`
@@ -1468,7 +1477,8 @@ UV_EXTERN void uv_free_interface_addresses(uv_interface_address_t* addresses,
  * thread pool) and call the specified callback in the specified loop after
  * completion. If the user gives NULL as the callback the blocking system
  * call will be called synchronously. req should be a pointer to an
- * uninitialized uv_fs_t object.
+ * uninitialized uv_fs_t object. Memory for the object must always be
+ * allocated via uv_fs_alloc().
  *
  * uv_fs_req_cleanup() must be called after completion of the uv_fs_
  * function to free any internal memory allocations associated with the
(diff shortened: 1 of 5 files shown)


saghul:
How are you compiling libuv? AFAIS either if you compile it with the makefiles or GYP it will have _FILE_OFFSET_BITS=64 defined.

indutny:
My opinion, if you're not using same flags as libuv does you may end up with some very interesting errors. Also, adding routines for allocation of structs doesn't seem to be a proper thing for me, especially considering that this structures are often embedded in other structures.

bsnote:
@indutny It's not a solution I am proud of. I'd be glad to find a better one. Probably `struct stat` should not be included in `uv_fs_t` but rather abstracted in some way.

I don't think that it's a common rule of thumb to include same flags as a third party library does. There can be many dependencies with inconsistent flags.

At least this caveat should be documented. I spent a lot of time trying to figure out what is wrong with `free(req)` statement. It would be nice if we could secure new users.

bsnote:
@indutny Are there any other flags you are talking about besides `-D_LARGEFILE_SOURCE` and `-D_FILE_OFFSET_BITS=64`?

saghul:
I believe GNU_SOURCE, there is a mention about it in the README.

bsnote:
@saghul good point, thanks

bnoordhuis:
Thanks but I won't take this patch for the reasons pointed out above. I'll update the documentation and have it mention _LARGEFILE_SOURCE and _FILE_OFFSET_BITS explicitly.

PR #715, opened 2013-02-21
speed up net write if socket buffer large enough

freedaxin:
add uv_try_write() to libuv, try to write as many bytes as possible
into socket buffer for nonblocking socket, if all data is written
successfully, async uv write can be avoid, thereby saving the
consumption of uv context and queue operations.

I've opened an issue in node, where explained the reason in detail:
https://github.com/joyent/node/issues/4699

And the corresponding patch for node:
https://github.com/freedaxin/node/commit/d0718d00a231a69c78862bfadab0c7b7b5c59122

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -594,6 +594,15 @@ UV_EXTERN int uv_write(uv_write_t* req, uv_stream_t* handle,
 UV_EXTERN int uv_write2(uv_write_t* req, uv_stream_t* handle, uv_buf_t bufs[],
     int bufcnt, uv_stream_t* send_handle, uv_write_cb cb);
 
+/*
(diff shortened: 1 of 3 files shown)


saghul:
This looks similar to what you were tying recently @indutny, maybe we can close this since it wasn't so helpful after all?

indutny:
Well, it won't work at windows, so I decided that it isn't worth having it. Anyway, @freedaxin I wonder how much improvement did it gave to you?

freedaxin:
@indutny, I got a improvement about 30% in my project. Maybe the patch is not perfect, but the approach is right. I wonder what's your "tying recently"?

indutny:
https://github.com/joyent/libuv/pull/1028 , but it doesn't and won't work on windows :)

PR #716, opened 2013-02-21
Gentle check for send_handle on uv_write2

saghul:
As discussed on the mailing list.

I added an extra check, which checks for the send_handle fd and fails early with EBADF. The assert is still there, but I guess the only case when it would blow up is if something really weird happens.

I went for checking the fd in stream and send_handle the same, instead of checking for stream and asserting for send_handle.

Since I was there, I removed a few lines of code by returning directly the value from uv__set_sys_error.

### DIFF
--- a/src/unix/stream.c
+++ b/src/unix/stream.c
@@ -1193,16 +1193,12 @@ int uv_write2(uv_write_t* req,
       stream->type == UV_TTY) &&
       "uv_write (unix) does not yet support other types of streams");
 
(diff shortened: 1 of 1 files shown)


bnoordhuis:
Thanks Saul, landed with slightly reworded commit logs in 3f47a2d and 9a8db3c.

saghul:
That was quick! Thanks, Ben.

PR #721, opened 2013-02-26
Small fixes.

ghost:
Changes:
- _inet.c:64_ `l` is unsigned so checking for `<= 0` is pointless.
- _unix/getaddrinfo.c:152_ `buf` is never free'd.
- _win/fs.c:1051_ `buf` is never free'd.
- _win/tty.c:99_ `handle` is assigned a value in line 103.
- _win/tty.c:989_ `FOREGROUND_RED` is bitwise-OR'd with `FOREGROUND_RED`.

I'm not really that familiar with the code-base so excuse any stupidity but
these are a few small things I noticed.


bnoordhuis:
> unix/getaddrinfo.c:152 buf is never free'd.

That's incorrect - the memory _is_ freed but the logic is rather subtle - so please revert that.

I'll leave it to @piscisaureus to review the other changes.

src/inet.c
```diff
@@ -61,7 +61,7 @@ static uv_err_t inet_ntop4(const unsigned char *src, char *dst, size_t size) {
 #else
   l = _snprintf(tmp, sizeof(tmp), fmt, src[0], src[1], src[2], src[3]);
 #endif
-  if (l <= 0 || l >= size) {
+  if (l == 0 || l >= size) {
```
piscisaureus, line 64: In my book `snprintf` returns an int and can actually return a negative value on failure. So the correct fix would be to make `l` of type `int` and not `size_t`.

src/win/fs.c
```diff
@@ -1048,6 +1048,8 @@ static void fs__sendfile(uv_fs_t* req) {
     }
   }
 
+  free(buf);
```
piscisaureus, line 1051: Seems right.

piscisaureus:
Thanks. I landed the pieces I agree about in 7570a35 and 603915d.

PR #722, opened 2013-02-26
build: add support for Visual Studio 2012

seishun:
This is pretty straightforward. The shared build (`vcbuild shared`) fails on tests, but that's a problem with Visual Studio 2010 as well, so it's not within the scope of this pull request.


seishun:
Signed.

piscisaureus:
Perfect! Landed in bc5fa79.

(And I wonder why github ignored the `closes: #722` clause...)

PR #724, opened 2013-02-26
Added -fPIC to CFLAGS

csaoh:
Without this, make fails on x64 machines.

Maybe there's a better place to put it, though ?

### DIFF
--- a/config-unix.mk
+++ b/config-unix.mk
@@ -20,7 +20,7 @@
 
 E=
 CSTDFLAG=--std=c89 -pedantic -Wall -Wextra -Wno-unused-parameter
-CFLAGS += -g
+CFLAGS += -g -fPIC
 CPPFLAGS += -I$(SRCDIR)/src
 LDFLAGS=-lm
 


bnoordhuis:
What are you doing that requires -fPIC unconditionally? Note that `make libuv.so` builds a shared object with -fPIC.

bnoordhuis:
I should add that there is no way I'm merging this as-is - PIC code on i386 (but not x86_64) is massively slower than non-PIC code.

csaoh:
alright, sorry. is there a way to only add it to x64 machines then ?

bnoordhuis:
Well, what do you need it for? If you're building an archive, -fPIC is useless unless you link it into a .so later on. And if you're doing that, you can just build with `make libuv.a CFLAGS=-fPIC`.

csaoh:
So far, everytime I try to build it on a x64 machine (which includes Travis CI machines), in order to make it work with a php binding, i have the following message "/usr/bin/ld: libuv/libuv.a(async.o): relocation R_X86_64_32 against `.rodata' can not be used when making a shared object; recompile with -fPIC", so I thought it needed to be added, at least to x64 compilation.

But I didn't know you could simply add CFLAGS in this way, so closing the PR now. Sorry again.

PR #727, opened 2013-02-28
build, windows: return an error code on failure

schlamar:


bnoordhuis:
Thanks, landed in 0ad46bd.

PR #729, opened 2013-03-01
Add interval and count to keep-alive

jameshartig:
The default probe interval and failure counts set by the OS are way too extreme and if you have a real-time socket server, you're going to want to customize these so you can have a much lower keep-alive interval and count so you're notified of disconnected sockets sooner.

Unfortunately, Windows doesn't let you set count, but using SIO_KEEPALIVE_VALS we can set interval (see: http://msdn.microsoft.com/en-us/library/windows/desktop/dd877220(v=vs.85).aspx). The defaults for the OS are 2 hour timeouts and 1 second intervals. I was hesitant about setting the default 1 second because that's pretty insane but decided just to leave it.

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -653,11 +653,16 @@ UV_EXTERN int uv_tcp_nodelay(uv_tcp_t* handle, int enable);
 /*
  * Enable/disable TCP keep-alive.
  *
- * `delay` is the initial delay in seconds, ignored when `enable` is zero.
+ * Options below are ignored when `enable` is zero.
+ * `delay` is the initial delay in seconds. If 0, value is ignored.
+ * `interval` is the interval in seconds after initial probe. If 0, value is ignored.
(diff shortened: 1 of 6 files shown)


jameshartig:
According to the node docs "Setting 0 for initialDelay will leave the value unchanged from the default (or previous) setting.". So I've added the delay back to the if checks in tcp.c to keep that true.

hertzg:
Any updates on this?

saghul:
@hertzg Looks like this fell through the cracks long time ago. The problem with this patch now is that it changes the API and thus it cannot land in v1.x. That aside, the values are not stored and thus lost if the socket was not yet created.

There is a proposal to create sockets early: https://github.com/libuv/leps/pull/5 which would simplify the problem, because the values can always be set.

Either way, libuv moved over to https://github.com/libuv/libuv, so please open a new issue there. Thanks!

PR #730, opened 2013-03-01
unix, test: add netmask to interface information.

upsuper:
Including implementation for darwin, linux, netbsd, and sunos, but only the code for darwin has been tested because of the lack of test environment. Test code is included.


include/uv.h
```diff
@@ -1441,6 +1441,10 @@ struct uv_interface_address_s {
     struct sockaddr_in address4;
     struct sockaddr_in6 address6;
   } address;
+  union {
+    struct sockaddr_in address4;
+    struct sockaddr_in6 address6;
+  } netmask;
```
bnoordhuis, line 1447: It should at the very least be zeroed on Windows so it doesn't contain junk data.

wanderview:
This also looks like a duplicate of #705 which includes a windows implementation.

upsuper:
@wanderview Yes, you are right.

PR #734, opened 2013-03-05
Windows TTY fixes

piscisaureus:
@sblom Could you review this please?


sblom:
Required a trip to MSDN to discern that never using `ENABLE_EXTENDED_FLAGS` was the key to the change. Seems 100% correct.

LGTM.

piscisaureus:
Thanks for reviewing @sblom

PR #735, opened 2013-03-06
test: add tap output for windows

tjfontaine:
In lieu of having fgets this will strchr the buffer for `\n` and when it finds one adds an `#` afterward


txdv:
was is this

tjfontaine:
@txdv see testanything.org

@piscisaureus force pushed

piscisaureus:
Thanks! Landed in 2f84a57

PR #739, opened 2013-03-12
WIP: uv_fs_[fl]stat should support subsecond resolution

tjfontaine:
This adds uv_stat_t with a consistent interface across platforms, and on platforms where it's supported ->st_mtim.tv_nsec will be populated

currently only for unix, and doesn't check if we're on a platform where it's not supported.

### DIFF
--- a/include/uv-private/uv-unix.h
+++ b/include/uv-private/uv-unix.h
@@ -121,7 +121,6 @@ typedef struct {
 
 typedef int uv_file;
 typedef int uv_os_sock_t;
-typedef struct stat uv_statbuf_t;
 
 #define UV_ONCE_INIT PTHREAD_ONCE_INIT
 
--- a/include/uv-private/uv-win.h
+++ b/include/uv-private/uv-win.h
@@ -207,8 +207,6 @@ typedef struct uv_buf_t {
 
 typedef int uv_file;
 
-typedef struct _stati64 uv_statbuf_t;
-
 typedef SOCKET uv_os_sock_t;
 
 typedef HANDLE uv_thread_t;
--- a/include/uv.h
+++ b/include/uv.h
@@ -350,6 +350,29 @@ typedef void (*uv_getaddrinfo_cb)(uv_getaddrinfo_t* req,
                                   int status,
                                   struct addrinfo* res);
 
+typedef struct {
+  long tv_sec;
+  long tv_nsec;
+} uv_timespec_t;
+
+
+typedef struct {
+  uint64_t st_dev;
+  uint64_t st_mode;
+  uint64_t st_nlink;
+  uint64_t st_uid;
+  uint64_t st_gid;
+  uint64_t st_rdev;
+  uint64_t st_ino;
+  uint64_t st_size;
+  uint64_t st_blksize;
+  uint64_t st_blocks;
+  uv_timespec_t st_atim;
+  uv_timespec_t st_mtim;
+  uv_timespec_t st_ctim;
+} uv_stat_t;
+
+
 /*
 * This will be called repeatedly after the uv_fs_event_t is initialized.
 * If uv_fs_event_t was initialized with a directory the filename parameter
@@ -361,8 +384,8 @@ typedef void (*uv_fs_event_cb)(uv_fs_event_t* handle, const char* filename,
 
 typedef void (*uv_fs_poll_cb)(uv_fs_poll_t* handle,
                               int status,
-                              const uv_statbuf_t* prev,
-                              const uv_statbuf_t* curr);
+                              const uv_stat_t* prev,
+                              const uv_stat_t* curr);
 
 typedef void (*uv_signal_cb)(uv_signal_t* handle, int signum);
 
@@ -1519,7 +1542,7 @@ struct uv_fs_s {
   void* ptr;
   const char* path;
   uv_err_code errorno;
-  uv_statbuf_t statbuf;  /* Stores the result of uv_fs_stat and uv_fs_fstat. */
+  uv_stat_t statbuf;  /* Stores the result of uv_fs_stat and uv_fs_fstat. */
   UV_FS_PRIVATE_FIELDS
 };
 
@@ -1646,7 +1669,7 @@ UV_EXTERN int uv_fs_poll_init(uv_loop_t* loop, uv_fs_poll_t* handle);
  * or the error reason changes).
  *
  * When `status == 0`, your callback receives pointers to the old and new
- * `uv_statbuf_t` structs. They are valid for the duration of the callback
+ * `uv_stat_t` structs. They are valid for the duration of the callback
  * only!
  *
  * For maximum portability, use multi-second intervals. Sub-second intervals
--- a/src/unix/fs.c
+++ b/src/unix/fs.c
@@ -497,6 +497,69 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
   return r;
 }
 
+static inline void uv__to_stat(struct stat* src, uv_stat_t* dst) {
+  dst->st_dev = src->st_dev;
+  dst->st_mode = src->st_mode;
+  dst->st_nlink = src->st_nlink;
+  dst->st_uid = src->st_uid;
+  dst->st_gid = src->st_gid;
+  dst->st_rdev = src->st_rdev;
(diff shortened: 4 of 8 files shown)


include/uv.h
```diff
@@ -350,6 +350,25 @@ typedef void (*uv_getaddrinfo_cb)(uv_getaddrinfo_t* req,
...
+  int st_nlink;
+  int st_uid;
+  int st_gid;
+  int st_rdev;
+  int64_t st_ino;
+  int64_t st_size;
+  uv_timespec_t st_atim;
+  uv_timespec_t st_mtim;
+  uv_timespec_t st_ctim;
+#if defined(__POSIX__)
```
bnoordhuis: Drop the conditional and zero the values on Windows?

Most of the fields in this struct should probably be unsigned long or uint64_t. st_dev and st_rdev, for example, are 64 bits on 64 bits Linux. (Though it's up for debate how plausible it is to have over 2^32 devices.)

include/uv-private/uv-unix.h
```diff
@@ -121,7 +121,7 @@ struct uv__work {
 
 typedef int uv_file;
 typedef int uv_os_sock_t;
-typedef struct stat uv_statbuf_t;
+typedef struct timespec uv_timespec_t;
```
bnoordhuis: I'm undecided if it's better to use our own timespec struct. It's (at least in theory) possible for compiler flags to change `sizeof(struct timespec)`.

include/uv-private/uv-win.h
```diff
@@ -207,7 +207,10 @@ typedef int (WSAAPI* LPFN_WSARECVFROM)
 
 typedef int uv_file;
 
-typedef struct _stati64 uv_statbuf_t;
+typedef struct {
+  time_t tv_sec;
```
bnoordhuis: Any particular reason to use time_t instead of long?

src/unix/fs.c
```diff
@@ -497,6 +497,66 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
...
+  dst->st_atim = src->st_atimespec;
+  dst->st_mtim = src->st_mtimespec;
+  dst->st_ctim = src->st_ctimespec;
+#elif defined(__POSIX__) /* TODO this isn't entirely accurate */
+  dst->st_atim = src->st_atim;
+  dst->st_mtim = src->st_mtim;
+  dst->st_ctim = src->st_ctim;
+#else
+  dst->st_atim.tv_sec = src->st_atime;
+  dst->st_atim.tv_nsec = 0;
```
bnoordhuis: Style: `type*` instead of `type *`

src/unix/internal.h
```diff
@@ -181,10 +181,12 @@ void uv__work_submit(uv_loop_t* loop,
 int uv__platform_loop_init(uv_loop_t* loop, int default_loop);
 void uv__platform_loop_delete(uv_loop_t* loop);
 
+/* fs */
+void uv__fs_event_close(uv_fs_event_t* handle);
```
bnoordhuis: Que?

tjfontaine: oops forgot to revert this part, I will fix that

tjfontaine:
I've updated this, with the suggested changes, there's also a version that needs to be reviewed by @piscisaureus which must offend his taste, because it offends mine

src/unix/fs.c
```diff
@@ -497,6 +497,70 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
...
+  dst->st_blocks = src->st_blocks;
+
+#if defined(__APPLE__)
+  dst->st_atim.tv_sec = src->st_atimespec.tv_sec;
+  dst->st_atim.tv_nsec = src->st_atimespec.tv_nsec;
+  dst->st_mtim.tv_sec = src->st_mtimespec.tv_sec;
+  dst->st_mtim.tv_nsec = src->st_mtimespec.tv_nsec;
+  dst->st_ctim.tv_sec = src->st_ctimespec.tv_sec;
+  dst->st_ctim.tv_nsec = src->st_ctimespec.tv_nsec;
+#elif defined(_POSIX_C_SOURCE) /* TODO this isn't entirely accurate */
```
bnoordhuis: `#elif defined(_BSD_SOURCE) || defined(_SVID_SOURCE)`?

src/unix/fs.c
```diff
@@ -497,6 +497,70 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
   return r;
 }
 
+/* TODO this should be a macro? */
```
bnoordhuis: No.

src/win/fs.c
```diff
@@ -808,7 +807,19 @@ void fs__readdir(uv_fs_t* req) {
 }
 
 
-INLINE static int fs__stat_handle(HANDLE handle, uv_statbuf_t* statbuf) {
+/* TODO this will be a macro */
+INLINE static uv_timespec_t fs__filetime_to_timespec(FILETIME *filetime) {
+  uv_timespec_t t;
+  uint64_t total = *(uint64_t*)filetime;
```
bnoordhuis: Aliases in a horribly way. Won't work on 64 bits ARM (assuming Windows and libuv will run on that someday.)

piscisaureus: I'm actually fine with this.

src/win/fs.c
```diff
@@ -846,9 +860,9 @@ INLINE static int fs__stat_handle(HANDLE handle, uv_statbuf_t* statbuf) {
       ((_S_IREAD|_S_IWRITE) >> 6));
   }
 
-  statbuf->st_mtime = FILETIME_TO_TIME_T(info.ftLastWriteTime);
-  statbuf->st_atime = FILETIME_TO_TIME_T(info.ftLastAccessTime);
-  statbuf->st_ctime = FILETIME_TO_TIME_T(info.ftCreationTime);
+  statbuf->st_mtim = fs__filetime_to_timespec(&info.ftLastWriteTime);
```
piscisaureus: What was the reason for adding this function rather than updating the FILETIME_TO_TIME_T macro?

tjfontaine: I was worried that I would have to debug semantics on windows and thought it would be easier for setting a break point on it, my intention was to change the macro

src/unix/fs.c
```diff
@@ -497,6 +497,69 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
...
+  dst->st_blocks = src->st_blocks;
+
+#if defined(__APPLE__)
+  dst->st_atim.tv_sec = src->st_atimespec.tv_sec;
+  dst->st_atim.tv_nsec = src->st_atimespec.tv_nsec;
+  dst->st_mtim.tv_sec = src->st_mtimespec.tv_sec;
+  dst->st_mtim.tv_nsec = src->st_mtimespec.tv_nsec;
+  dst->st_ctim.tv_sec = src->st_ctimespec.tv_sec;
+  dst->st_ctim.tv_nsec = src->st_ctimespec.tv_nsec;
+#elif defined(_BSD_SOURCE) || defined(_SVID_SOURCE)
```
tjfontaine: @bnoordhuis this isn't catching on sunos, would it be inappropriate if I add `defined(_XOPEN_SOURCE)` or is that too broad?

bnoordhuis: I don't think it'll hurt. If it causes build breakage on some platforms, we'll fix it.

src/win/fs.c
```diff
@@ -77,8 +77,20 @@
...
 #define FILETIME_TO_TIME_T(filetime)                                        \
-   ((*((uint64_t*) &(filetime)) - 116444736000000000ULL) / 10000000ULL);
+   (FILETIME_TO_UINT(filetime) / 10000000ULL);
+
+#define FILETIME_TO_TIME_NS(filetime, secs)                                 \
+   ((FILETIME_TO_UINT(filetime) - (secs * 10000000ULL)) * 100);
+
+#define FILETIME_TO_TIMESPEC(ts, filetime)                                  \
+   do {                                                                     \
+     ts.tv_sec = FILETIME_TO_TIME_T(filetime);                              \
```
piscisaureus: Wrap `ts` in brackets, e.g. `(ts).tv_sec`

PR #744, opened 2013-03-14
IPv6: support link-local addresses

### LINKED ISSUE #271: IPv6: support link-local addresses
@tshinnic pointed out that on at least Linux link-local addresses require that the interface is specified. In other words, we need to:
1. Parse the address in `uv_ip6_addr()`
2. Look up the interface ID with `if_nametoindex()` or `ioctl(SIOCGIFINDEX)`
3. Set `sin6_scope_id` to the interface ID to use

For the record, a qualified link-local address looks like this: `fe80::20c:29ff:fe31:8d4e%eth0`

bajtos:
Modified uv_ip6_addr() to fill sin6_scope_id for link-local addresses (Linux and Windows only).

Fixes #271


src/uv-common.c
```diff
@@ -170,16 +173,41 @@ struct sockaddr_in uv_ip4_addr(const char* ip, int port) {
   return addr;
 }
 
-
```
bnoordhuis, line 173: Style: two newlines between functions (i.e. don't remove this line.)

src/uv-common.c
```diff
@@ -170,16 +173,41 @@ struct sockaddr_in uv_ip4_addr(const char* ip, int port) {
...
 struct sockaddr_in6 uv_ip6_addr(const char* ip, int port) {
   struct sockaddr_in6 addr;
+  char * address_part = NULL;
+  const char* zone_index = NULL;
```
bnoordhuis: Style: split declaration and assignment.

src/uv-common.c
```diff
@@ -170,16 +173,41 @@ struct sockaddr_in uv_ip4_addr(const char* ip, int port) {
...
+      ip = address_part;
+      /* else report error, but there is no way how to signal error to the caller :( */
+
+    zone_index++; /* skip '%' */
+    addr.sin6_scope_id = if_nametoindex(zone_index);
+    /* TODO - check that result is not zero (indicating unknown interface) */
+  }
+#else
+  /* TODO - implement sin6_scope_id for other platforms
+   * Windows IPv6 stack uses numeric zone indexes, e.g., fe80::3%1. The index is determined by the interface number;
```
bnoordhuis: Style: long line, wrap at 80 columns.

src/uv-common.c
```diff
@@ -170,16 +173,41 @@ struct sockaddr_in uv_ip4_addr(const char* ip, int port) {
...
 
   memset(&addr, 0, sizeof(struct sockaddr_in6));
-
   addr.sin6_family = AF_INET6;
   addr.sin6_port = htons(port);
+
+#ifdef __linux__
+  zone_index = strchr(ip, '%');
+  if (zone_index != NULL) {
+    address_part = strndup(ip, zone_index - ip);
```
bnoordhuis: Don't use a dynamic allocation here, we try to avoid that wherever possible in libuv. Use an on-stack buffer that is large enough to hold the address.

test/test-ip6-addr.c
```diff
@@ -0,0 +1,89 @@
...
+void stringify_address(char* address) {
+    int i, j;
+    ASSERT(strlen(address)==32);
+    address[40] = '\0';
+    for (i=7; i>=0; i--) {
+      for (j=0; j<4; j++)
+        address[i*5 + j] = address[i*4 + j];
+      if (i > 0)
+        address[i*5 - 1] = ':';
+    }
```
bnoordhuis: Style: indentation.

test/test-ip6-addr.c
```diff
@@ -0,0 +1,89 @@
...
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <stdio.h>
+#include <string.h>
+
+typedef void (*iface_info_cb)(const char* ip6_addr, const char* device_name, int iface_index, int scope_value);
+
+void stringify_address(char* address) {
+    int i, j;
+    ASSERT(strlen(address)==32);
```
bnoordhuis: Style: spaces around the ==.

test/test-ip6-addr.c
```diff
@@ -0,0 +1,89 @@
...
+#include "task.h"
+
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <stdio.h>
+#include <string.h>
+
+typedef void (*iface_info_cb)(const char* ip6_addr, const char* device_name, int iface_index, int scope_value);
+
+void stringify_address(char* address) {
```
bnoordhuis: Style: two newlines between functions and the top declarations.

test/test-ip6-addr.c
```diff
@@ -0,0 +1,89 @@
...
+    address[40] = '\0';
+    for (i=7; i>=0; i--) {
+      for (j=0; j<4; j++)
+        address[i*5 + j] = address[i*4 + j];
+      if (i > 0)
+        address[i*5 - 1] = ':';
+    }
+}
+
+void foreach_network_interface(iface_info_cb iface_cb) {
```
bnoordhuis: There is a libuv function for that: uv_interface_addresses()

test/test-ip6-addr.c
```diff
@@ -0,0 +1,89 @@
...
+
+  LOGF("Got scope_id 0x%02x\n", addr.sin6_scope_id);
+  ASSERT(iface_index == addr.sin6_scope_id);
+}
+
+TEST_IMPL(ip6_addr_link_local) {
+  foreach_network_interface(&test_ip6_addr_scope);
+  return 0;
+}
+
```
bnoordhuis: Style: no newline at EOF.

build.mk
```diff
@@ -136,6 +136,7 @@ TESTS= \
 	test/test-udp-send-and-recv.o \
 	test/test-util.o \
 	test/test-walk-handles.o \
+	test/test-ip6-addr.o \
```
bnoordhuis: They're sorted alphabetically (but not in uv.gyp for some reason. You can leave that one as-is.)

test/test-ip6-addr.c
bajtos: I don't understand which part is wronly indented? The function will be removed, but I'd like to know the correct style.

bnoordhuis: All of it. :-) Use two spaces, not four.

src/uv-common.c
```diff
@@ -170,14 +173,38 @@ struct sockaddr_in uv_ip4_addr(const char* ip, int port) {
...
 
   memset(&addr, 0, sizeof(struct sockaddr_in6));
 
   addr.sin6_family = AF_INET6;
   addr.sin6_port = htons(port);
+
+#if defined(__linux__) || defined(_WIN32)
+  zone_index = strchr(ip, '%');
+  if (zone_index != NULL) {
+    assert(zone_index - ip < 40);
```
bnoordhuis: sizeof(address_part)

src/uv-common.c
```diff
@@ -170,14 +173,38 @@ struct sockaddr_in uv_ip4_addr(const char* ip, int port) {
...
+    strncpy(address_part, ip, zone_index - ip);
```
bnoordhuis, line 193: Doesn't zero-terminate the string when the address is > sizeof(address_part). Don't rely on the assert() to catch that.

test/test-ip6-addr.c
```diff
@@ -0,0 +1,83 @@
...
+#include <stdio.h>
+#include <string.h>
+
+#ifdef __linux__
+# include <sys/socket.h>
+# include <net/if.h>
+#endif
+
+typedef void (*iface_info_cb)(const char* ip6_addr, const char* device_name,
+     unsigned iface_index);
```
bnoordhuis: Style: when arguments are split over multiple lines, they should line up (and one argument per line.)

test/test-ip6-addr.c
```diff
@@ -0,0 +1,83 @@
...
+
+void call_iface_info_cb(iface_info_cb iface_cb, char const* iface_name, struct sockaddr_in6 const* address) {
```
bnoordhuis: Long line, wrap at 80 columns. Happens in a few other places too.

test/test-ip6-addr.c
```diff
@@ -0,0 +1,83 @@
...
+
+void foreach_ip6_interface(iface_info_cb iface_cb) {
+  uv_err_t result;
+  int count, ix;
+  uv_interface_address_t* addresses;
+
+  result = uv_interface_addresses(&addresses, &count);
+  ASSERT(result.code == UV_OK);
+
+  for (ix=0; ix<count; ix++) {
```
bnoordhuis: Style: spaces around assignment/comparison operators.

test/test-ip6-addr.c
```diff
@@ -0,0 +1,83 @@
...
+    if (addresses[ix].address.address4.sin_family != AF_INET6) continue;
+    call_iface_info_cb(iface_cb, addresses[ix].name, &addresses[ix].address.address6);
```
bnoordhuis: Merge that function? It's only used here.

EDIT: I guess you could just inline everything into TEST_IMPL(ip6_addr_link_local).

test/test-ip6-addr.c
```diff
@@ -0,0 +1,83 @@
...
+  char string_address[INET6_ADDRSTRLEN];
+  uv_err_t result;
+
+  result = uv_inet_ntop(AF_INET6, &address->sin6_addr, string_address, INET6_ADDRSTRLEN);
+  ASSERT(result.code == UV_OK);
+
+  iface_cb(string_address, iface_name, address->sin6_scope_id);
+}
+
+void foreach_ip6_interface(iface_info_cb iface_cb) {
```
bnoordhuis, line 52: Style: two newlines between functions.

test/test-list.h
```diff
@@ -204,6 +204,9 @@
 TEST_DECLARE   (poll_duplex)
 TEST_DECLARE   (poll_unidirectional)
 TEST_DECLARE   (poll_close)
+#if defined(__linux__) || defined(_WIN32)
```
bnoordhuis: Preferably don't #ifdef things in test-list.h. Do it in the test itself and return 0 when we're on an unsupported platform (with maybe a printed message telling the user the test is skipped.)

test/test-ip6-addr.c
```diff
@@ -0,0 +1,83 @@
...
+  uv_err_t result;
+  int count, ix;
+  uv_interface_address_t* addresses;
+
+  result = uv_interface_addresses(&addresses, &count);
```
bnoordhuis, line 57: Call uv_free_interface_addresses() when you're done with it.

src/uv-common.c
```diff
@@ -170,14 +173,38 @@ struct sockaddr_in uv_ip4_addr(const char* ip, int port) {
   return addr;
 }
 
+static unsigned uv__parse_zone_index(const char* zone_index) {
+#ifdef __linux__
+  return if_nametoindex(zone_index);
+#elif defined(_WIN32)
+  return atoi(zone_index);
+#endif
+}
```
bnoordhuis: I suggest you inline this into the `#if defined(__linux__) || defined(_WIN32)` block further down below. It generates 'unused function' and 'no return value' warnings on non-Linux/Windows platforms.

test/test-ip6-addr.c
bajtos: I disagree on this point. I put the code in a standalone function so that the test method is more readable. The same applies for call_iface_info_cb(). I strongly believe that many small functions make better/cleaner code, which is also easier to maintain.

Why would you want to inline/merge the function?

test/test-list.h
bajtos: OK, I'll do the ifdef in the test method. The reason for ifdefing in test-list was that there are few other tests turned on/off this way.

BTW, I don't think there is an easy way how to print "test skipped" message. Output of LOG macro is printed only when the test fails.

test/test-ip6-addr.c
bajtos: This rule is not consistently obeyed in the code base (see e.g. uv_read2_cb in include/uv.h), but I will fix it per your suggestion.

piscisaureus: Agree with @bajtos. I use the double-indent style for function signatures (but not function calls) too.

bajtos: @piscisaureus: thanks :)

I'll leave it as it is for now, we can always change it later.

src/uv-common.c
bajtos: I don't like that idea because the zone index handling is already awfully long. I defined a new macro `UV_PLATFORM_HAS_IP6_LINK_LOCAL_ADDRESS` and protected the helper function with `#ifdef` instead.

Perhaps we could move `uv__parse_zone_index` to platform-specific files (e.g. [include/uv-private/uv-win.h](https://github.com/joyent/libuv/blob/master/include/uv-private/uv-win.h)) in order to keep [src/uv-common.c](https://github.com/joyent/libuv/blob/master/src/uv-common.c) cleaner, possibly defining the helper function as a macro instead?

What do you think?

test/test-ip6-addr.c
bajtos, line 57: This has been fixed in the last commit (github did not pick the change in this diff view).

bajtos:
Hello, what's the status of this pull request? What changes are necessary to finish this work and have the request pulled to master?

saghul:
Looks like if_nametoindex is also supported on OSX. https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man3/if_nametoindex.3.html not sure about other BSDs though.

bajtos:
I fixed remaining issues, added Darwin implementation and squashed the changes into a single commit.

@bnoordhuis @piscisaureus could you please review?

bnoordhuis:
Landed on July 4 in fd45f87.

PR #745, opened 2013-03-15
stream: skip try_select check for incoming streams

indutny:
This check is redundant for incoming TCP connection, and we're just
loosing performance on it. Skip it in the cases, where we already know
that fd works fine with kqueue.

/cc @bnoordhuis @piscisaureus


bnoordhuis:
This is the wrong approach. The only place where 'exotic' file descriptors come in is in uv_pipe_open() and uv_tty_init() so that's where the 'is this a file descriptor that works with kqueue?' checks should be.

indutny:
That's not really true statement, stream might be created using existing fd.

bnoordhuis:
Yes, in those two functions I mentioned. There's also uv_tcp_open() but the implicit contract is that you pass it a socket so that one doesn't need special handling. That about covers it, doesn't it?

src/unix/pipe.c
```diff
@@ -155,6 +155,16 @@ void uv__pipe_close(uv_pipe_t* handle) {
 
 
 int uv_pipe_open(uv_pipe_t* handle, uv_file fd) {
+#if defined(__APPLE__)
+  {
+    int r;
+
+    r = uv__stream_try_select((uv_stream_t*) handle, &fd);
```
bnoordhuis:
``` c
if (uv__stream_try_select((uv_stream_t*) handle, &fd))
  return -1;
```

?

indutny:
All nits fixed, anything else?

bnoordhuis:
Nice work, Fedor. Landed in b45a74f.

PR #749, opened 2013-03-18
tcp: add flags argument to uv_tcp_bind[6]()

indutny:
Possible flag values 0 or UV_TCP_IPV6ONLY.

This is important for https://github.com/indutny/node/compare/feature-dualstack. And this PR requires changes on windows side.

/cc @bnoordhuis @piscisaureus.


indutny:
Probably, it should be the other way there. Instead of providing flag to disable dual stack, we should provide flag to enable it.

indutny:
@bnoordhuis updated.

src/win/tcp.c
```diff
@@ -266,6 +266,19 @@ static int uv__bind(uv_tcp_t* handle,
     }
   }
 
+#ifdef IPV6_V6ONLY
+  if (family == AF_INET6) {
+    int on;
+
+    on = (tcp->flags & UV_HANDLE_TCP_DUALSTACK) ? 0 : 1;
```
tjfontaine: src\win\tcp.c(273): error C2065: 'tcp' : undeclared identifier [g:\jenkins\workspace\libuv-pullrequest\label\windows\libuv.vcxproj]

saghul:
@indutny is this still relevant?

txdv:
This patch might not, but generally Dual Stack is still relevant.

saghul:
@txdv sure. I meant this patch, since it was partially reviewed, and looks like rebasing against current master wouldn't be much trouble.

txdv:
This adds uv_tcp_dualstack(handle, enable) as a function while the udp variant is a flag set in uv_udp_bind (UV_UDP_IPV6ONLY).

I think the direction of making it an optional dualstack instead of optional ipv6only is right, but we need a consistent API.

indutny:
@txdv I disagree, it should be optional ipv6only. This is a brave new world, man! Regarding this patch, I think it is still relevant. One small fix on windows side, renaming `tcp` to `handle` should make it work.

txdv:
@indutny Is dualstack on by default or off by default? Because the name dualstack implies (to me personally) that it is always off and you have to turn it on.

saghul:
@indutny API-wise, wouldn't it make sense to have:

```
int uv_tcp_bind(uv_tcp_t* handle const struct sockaddr* addr, unsigned int flags)
```

like @txdv suggested? So, we'd be dualstack by default, unless the flags indicate that IPv6-only should be used.

indutny:
@saghul agree

txdv:
NO! I suggest that dual stack is off by default and has to be turned on with a UV_DUALSTACK const. Well, I actually don't care as long as it gets into the lib.

indutny:
@saghul @txdv force pushed branch!

txdv:
Looks like a mirror version of udp so looks good to me.

txdv:
Maybe rename the title of the commit, because there is no uv_tcp_dualstack anymore.

indutny:
Right, will do right before landing it.

include/uv.h
```diff
@@ -783,6 +783,11 @@ UV_EXTERN int uv_tcp_keepalive(uv_tcp_t* handle,
  */
 UV_EXTERN int uv_tcp_simultaneous_accepts(uv_tcp_t* handle, int enable);
 
+enum uv_tcp_flags {
+  /* Disables dual stack mode. Used with uv_tcp_bind6(). */
```
saghul, line 787: "Used with uv_tcp_bind, when an IPv6 address is used", `uv_tcp_bind6` is no longer a thing :-)

saghul:
LGTM, just a minor nit. We don't seem to have any test for this, since all tests seem to pass 0 as flags. Do you think we can have a test similar to test-udp-ipv6, which actually tests the functionality?

indutny:
@saghul sounds like a plan.

indutny:
Landed in 8f15aae52f79 , waiting for @txdv to help me with tests! ;)

saghul:
Closing since this was already merged, I created https://github.com/joyent/libuv/issues/1176, when someone has time a test would be appreciated :-)

mmalecki:
So I started working on #1176, but ended up being a bit confused about this pull request. As I understand, for bound sockets, `IPV6_V6ONLY` disables the functionality of accepting IPv4 clients. However, we don't seem to provide functionality to disable connecting to IPv4 endpoints (done with the same flag), as we only accept the `UV_TCP_IPV6ONLY` flag in `uv_tcp_bind`.

Is this a valid concern? If so, should we perhaps expose a `uv_tcp_ipv6only` function, similar to how we do with `uv_tcp_nodelay`?

saghul:
@mmalecki if you bind using that flag and then try to connect to an IPv4 address you should get an error. Are you observing otherwise?

saghul:
Also, not sure how much having a toggleable would help, given: http://stackoverflow.com/questions/5587935/cant-turn-off-socket-option-ipv6-v6only

mmalecki:
@saghul hm, I haven't tested it like that because it seemed counter-intuitive to call `bind` when connecting. I'll verify that this works and use it in the test then, thanks!

PR #753, opened 2013-03-26
unix, windows: don't read/recv if buf.len==0

### LINKED ISSUE #752: uv__read should stop reading when alloc_cb can't allocate
In my project I am using a static buffer that has a reasonable size but if it's full I just like to close the connection and reconnect.  This is, however, not possible with uv__read right now because alloc_cb failure will just raise an assert, and the calls to uv_read_stop() and uv_close() that were initiated by my alloc_cb are never executed.

bnoordhuis:
Make it possible for the libuv user to handle out of memory conditions
gracefully. When alloc_cb() returns a buffer with len==0, call the read
or recv callback with nread=-1 and error=UV_ENOBUFS. It's up to the
user to close the handle.

Fixes #752.

Suggested reviewer: @piscisaureus

I tried to exercise due diligence when going over the Windows code but it's completely untested. View it as a starting point for you to improve on. :-)

### DIFF
--- a/src/win/tty.c
+++ b/src/win/tty.c
@@ -348,8 +348,11 @@ static void uv_tty_queue_read_line(uv_loop_t* loop, uv_tty_t* handle) {
   memset(&req->overlapped, 0, sizeof(req->overlapped));
 
   handle->read_line_buffer = handle->alloc_cb((uv_handle_t*) handle, 8192);
+  if (handle->read_line_buffer.len == 0) {
+    handle->read_cb(handle, UV_ENOBUFS, handle->read_line_buffer);
+    return;
+  }
   assert(handle->read_line_buffer.base != NULL);
-  assert(handle->read_line_buffer.len > 0);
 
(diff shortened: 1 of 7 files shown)


src/win/tty.c
```diff
@@ -346,8 +346,12 @@ static void uv_tty_queue_read_line(uv_loop_t* loop, uv_tty_t* handle) {
   memset(&req->overlapped, 0, sizeof(req->overlapped));
 
   handle->read_line_buffer = handle->alloc_cb((uv_handle_t*) handle, 8192);
+  if (handle->read_line_buffer.len == 0) {
+    uv__set_artificial_error(handle->loop, UV_ENOBUFS);
+    handle->read_cb(handle, -1, buf);
```
tjfontaine: buf is undefined, this should be `handle->read_line_buffer` instead of buf?

bnoordhuis: Yes, it should. Good catch. Addressed in bnoordhuis/libuv@9896947.

PR #755, opened 2013-03-26
unix, windows: remove ngx-queue.h

bnoordhuis:
Avoids an extra #include in public headers and stops the ngx_queue_*
types and macros from leaking into user code.

Suggested reviewer: @piscisaureus


piscisaureus:
LGTM. But what was exactly the motivation for this again? It doesn't make so much of a difference -- now the queue has to be an array.

bnoordhuis:
The main benefit is that we no longer leak ngx_queue_\* things into the namespace of the libuv user. From an implementation point of view, nothing really changes. Landed in 0635e29.

bnoordhuis:
BTW, I might do the same thing with tree.h someday.

piscisaureus:
That would make me very sad. I love tree.h

Is #undef not an option? They are all macros, so we could just #undef them.

Van: Ben Noordhuis [mailto:notifications@github.com] 
Verzonden: Wednesday, March 27, 2013 12:14 AM
Aan: joyent/libuv
CC: Bert Belder
Onderwerp: Re: [libuv] unix, windows: remove ngx-queue.h (#755)

BTW, I might do the same thing with tree.h someday.

shigeki:
Oh, rest in peace and thank you ngx_queue.

PR #756, opened 2013-03-28
unix: Don't clear handle flags after closing UDP handle

saghul:


bnoordhuis:
Thanks, landed in a9a23dc.

PR #761, opened 2013-04-01
Introducing uv_null_check(). A revolutionary way to catch those pesky null pointers.

joshthecoder:
Just pass your pointers to uv_null_check() to check if they are NULL.
If the pointer is in fact NULL, this handy function will send a very useful
message to a remote SOS server and then kindly crash your application for you.

Checkout the "null_check" test. I have even included a Node based SOS server.

So begin this month of April by refactoring your servers to use this handy new API. :)

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -222,6 +222,15 @@ typedef enum {
   UV_RUN_NOWAIT
 } uv_run_mode;
 
+/*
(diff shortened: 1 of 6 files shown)


txdv:
LGTM

bnoordhuis:
Sorry, merge window's closed. Maybe next year.

PR #762, opened 2013-04-02
lint

mscdex:


bnoordhuis:
We don't do lint here at joyent/libuv but your PR did catch a syntax error that I've fixed in 3a8c398.

@piscisaureus The missing free() in src/win/fs.c looks like a genuine bug to me.

mscdex:
So unused/useless code is ok? :confused:

piscisaureus:
I'm still reviewing but so far I've seen good improvements. Hang on.

src/win/poll.c
```diff
@@ -319,11 +319,7 @@ static DWORD WINAPI uv__slow_poll_thread_proc(void* arg) {
   assert(handle->type == UV_POLL);
   assert(req->type == UV_POLL_REQ);
 
-  if (req == &handle->poll_req_1) {
-    events = handle->submitted_events_1;
-  } else if (req == &handle->poll_req_2) {
-    events = handle->submitted_events_2;
-  } else {
+  if (req != &handle->poll_req_1 && req != &handle->poll_req_2) {
```
piscisaureus: Good catch on the superfluous `events` var and assignment, but I'm sure that the construct `if (!condition) assert(0)` wasn't in the style guide either. I'd just drop the check altogether.

bnoordhuis:
> So unused/useless code is ok?

If it serves a purpose, like being documentation for the next programmer. As an example:

```
-  /* We've got a copy, don't touch the original any more. */
-  name = NULL;
```

That tells me or whoever comes after me that name should not be used after that point.

src/win/tty.c
```diff
@@ -96,10 +96,9 @@ void uv_console_init() {
 
 
 int uv_tty_init(uv_loop_t* loop, uv_tty_t* tty, uv_file fd, int readable) {
-  HANDLE handle = INVALID_HANDLE_VALUE;
+  HANDLE handle = (HANDLE) _get_osfhandle(fd);
```
piscisaureus: Disagree on moving the assignment to the declaration. Just drop the initialization with INITIAL_HANDLE_VALUE.

src/win/util.c
```diff
@@ -579,7 +579,7 @@ uv_err_t uv_uptime(double* uptime) {
 
 
 uv_err_t uv_cpu_info(uv_cpu_info_t** cpu_infos, int* count) {
-  SYSTEM_PROCESSOR_PERFORMANCE_INFORMATION* sppi;
+  SYSTEM_PROCESSOR_PERFORMANCE_INFORMATION* sppi = NULL;
```
piscisaureus, line 582: why?

src/inet.c
```diff
@@ -61,7 +61,7 @@ static uv_err_t inet_ntop4(const unsigned char *src, char *dst, size_t size) {
 #else
   l = _snprintf(tmp, sizeof(tmp), fmt, src[0], src[1], src[2], src[3]);
 #endif
-  if (l <= 0 || l >= size) {
+  if (l == 0 || l >= size) {
```
piscisaureus: disagree. `_snprintf` returns `int` on windows.

bnoordhuis:
Also, I hate getting sidetracked by lint commits when I'm doing code archeology.

src/win/tty.c
mscdex: fine by me

src/unix/getaddrinfo.c
```diff
@@ -88,7 +88,7 @@ static void uv__getaddrinfo_done(struct uv__work* w, int status) {
   else if (req->retcode == EAI_NODATA)
     uv__set_sys_error(req->loop, ENOENT);
 #elif defined(__sun)
-  if (req->retcode == EAI_MEMORY && hostlen >= MAXHOSTNAMELEN) {
```
piscisaureus, line 91: How did this ever work? Nobody ever compiled on __sun ?

src/win/util.c
mscdex, line 582: Due to a possible early bailout on [this line](https://github.com/joyent/libuv/blob/master/src/win/util.c#L603)? There is an `if (sppi)` check right after the :out label and sppi would not necessarily be initialized to null?

src/inet.c
mscdex: isn't size_t unsigned?

src/win/poll.c
mscdex: fine by me

src/win/util.c
piscisaureus, line 582: Good catch!

mscdex:
> > So unused/useless code is ok?
> 
> If it serves a purpose, like being documentation for the next programmer. As an example:
> 
> ```
> -  /* We've got a copy, don't touch the original any more. */
> -  name = NULL;
> ```
> 
> That tells me or whoever comes after me that name should not be used after that point.

Ok, so leaving just the comment in this particular case would be fine?

bnoordhuis:
No. We're poisoning the variable. If someone does inadvertently dereference the pointer, s/he will find out soon enough because it'll segfault.

mscdex:
Updated commit as per suggestions.

src/inet.c
mscdex: i've reverted this and instead changed `l` to be `int` instead of `size_t`.

piscisaureus:
Thanks Brian, all the issues you found are addressed in the following commits. Some "useless" code removal wasn't landed because it serves a purpose.

https://github.com/joyent/libuv/commit/4d4f1496d9b600dfd3e4c4f0040d7b4a0d41043c
https://github.com/joyent/libuv/commit/9021dbcd8d1dc3b1c6cca82f46d5e774088d4ae6
https://github.com/joyent/libuv/commit/a9bce29f0c98187fcf09d1ba9dcd4eb5548fb30e
https://github.com/joyent/libuv/commit/7570a35b70a33258e4c36e080e900f30c3213697
https://github.com/joyent/libuv/commit/603915dd0081ec2f34d612291a1aeb909f6f2b76

PR #765, opened 2013-04-05
doc,unix: uv_udp_open supports all datagrams

kristrev:
On Unix, uv_udp_open can be used with any socket as long as the socket follows the datagram contract. This means that any datagram-socket, like for example netlink or raw sockets, can be used with libuv. I think this is very useful and expands the possible use cases for libuv, and should be documented.

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -787,6 +787,12 @@ UV_EXTERN int uv_udp_init(uv_loop_t*, uv_udp_t* handle);
 
 /*
  * Opens an existing file descriptor or SOCKET as a udp handle.
+ *
+ * Unix only:  
+ *  The only requirement of the sock argument is that it follows the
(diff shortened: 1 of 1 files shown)


txdv:
squash into 1 commit

kristrev:
Thanks for letting me know! Thought I had done it, but apparently not :)

kristrev:
Fixed the messages from Jenkins. Please let me know if there is something else to improve/fix, and if I should create a new pull request with correct commit message.

bnoordhuis:
@kristrev Did you sign the CLA? I can't find you in the list.

Re: commit log, I'd probably prefix it with `include: document etc.`

kristrev:
@bnoordhuis Thanks, I forgot that. I have signed the CLA and updated the commit.

bnoordhuis:
Thanks Kristian, landed in 1e8fe45.

PR #766, opened 2013-04-06
doc: add reference to Visual Studio Express in README.

wanderview:
Until recently I was under the impression MinGW was the only option for building on windows if you did not own a Visual Studio license.  @isaacs set me straight and told me about VS Express.

Since others coming from a unix background may not be aware of this either, consider adding a reference to Visual Studio Express in the README.


wanderview:
Updated reference to Visual Studio 2010 which runs on winxp as suggested by @mscdex.

bnoordhuis:
Thanks, landed in 8ef07cc.

PR #768, opened 2013-04-07
build: Update build.mk

staticfloat:
Clean `make test` build objects properly.  I ran into this when packaging `libuv` on debian, as when packaging it is important to be able to clean and return to a pristine state.


staticfloat:
Not sure what subsystem to file this under..... is there a `build:` subsystem?

indutny:
Yes, prefixing it with `build:` should work.

staticfloat:
Done.

bnoordhuis:
Landed in d67d713, thanks.

PR #770, opened 2013-04-08
stream: account _SC_IOV_MAX

indutny:
Write no more than `sysconf(_SC_IOV_MAX)` chunks with `writev()` at
once, otherwise `writev()` returns EINVAL.

https://github.com/joyent/node/pull/5246 depends on it.

/cc @bnoordhuis


src/unix/stream.c
```diff
@@ -742,6 +755,10 @@ static void uv__write(uv_stream_t* stream) {
   iov = (struct iovec*) &(req->bufs[req->write_index]);
   iovcnt = req->bufcnt - req->write_index;
 
+  /* Limit iov count to avoid EINVALs from writev() */
+  if (iovcnt > uv__iov_max)
+    iovcnt = uv__iov_max;
```
bnoordhuis, line 760: `#include <limits.h>` and check for `iovcnt > IOV_MAX`? There's not much point in using sysconf() here, IOV_MAX is not configurable on systems we support.

(EDIT: Truth be told, I don't know of any system where it's configurable.)

test/test-tcp-writealot.c
```diff
@@ -26,8 +26,8 @@
 
 
 #define WRITES            3
-#define CHUNKS_PER_WRITE  3
-#define CHUNK_SIZE        10485760 /* 10 MB */
+#define CHUNKS_PER_WRITE  4096
+#define CHUNK_SIZE        10 * 1024 /* 10 kb */
```
bnoordhuis, line 30: Never ever use complex expressions in a macro without wrapping them in parentheses.

indutny:
Thanks, landed in 9e87974.

PR #772, opened 2013-04-09
unix: dtrace probes for tick-start and tick-stop

tjfontaine:
Adds probes for start of tick and end of tick.

The Makefile change isn't as pretty as I would have liked.

Disabling thin archives was required to do a gyp build on smartos, adding allextract was necessary in for gyp builds to make sure run-tests and run-benchmarks actually got the dtrace symbols.

### DIFF
--- a/config-unix.mk
+++ b/config-unix.mk
@@ -33,6 +33,10 @@ RUNNER_SRC=test/runner-unix.c
 RUNNER_CFLAGS=$(CFLAGS) -I$(SRCDIR)/test
 RUNNER_LDFLAGS=-L"$(CURDIR)" -luv -Xlinker -rpath -Xlinker "$(CURDIR)"
 
+HAVE_DTRACE=
+DTRACE_OBJS=
+DTRACE_HEADER=
+
 OBJS += src/unix/async.o
 OBJS += src/unix/core.o
 OBJS += src/unix/dl.o
@@ -58,11 +62,14 @@ OBJS += src/inet.o
 OBJS += src/version.o
 
 ifeq (sunos,$(PLATFORM))
+HAVE_DTRACE=1
 CPPFLAGS += -D__EXTENSIONS__ -D_XOPEN_SOURCE=500
 LDFLAGS+=-lkstat -lnsl -lsendfile -lsocket
 # Library dependencies are not transitive.
 RUNNER_LDFLAGS += $(LDFLAGS)
 OBJS += src/unix/sunos.o
+OBJS += src/unix/dtrace.o
+DTRACE_OBJS += src/unix/core.o
 endif
 
 ifeq (aix,$(PLATFORM))
@@ -72,6 +79,7 @@ OBJS += src/unix/aix.o
 endif
 
 ifeq (darwin,$(PLATFORM))
+HAVE_DTRACE=1
 CPPFLAGS += -D_DARWIN_USE_64_BIT_INODE=1
 LDFLAGS += -framework Foundation \
            -framework CoreServices \
@@ -96,6 +104,7 @@ OBJS += src/unix/linux-core.o \
 endif
 
 ifeq (freebsd,$(PLATFORM))
+HAVE_DTRACE=1
 LDFLAGS+=-lkvm
 OBJS += src/unix/freebsd.o
 OBJS += src/unix/kqueue.o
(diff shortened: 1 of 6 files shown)


tjfontaine:
The Makefile changes, while they work on my smartos install, seem to not like the jenkins smartos slave

tjfontaine:
ok @bnoordhuis I think most of your tweaks have been solved, I put the check for HAVE_DTRACE in internal.h let me know if you want it somewhere else

config-unix.mk
```diff
@@ -162,7 +177,16 @@ test/%.o: test/%.c include/uv.h test/.buildstamp
 	$(CC) $(CSTDFLAG) $(CPPFLAGS) $(CFLAGS) -c $< -o $@
 
 clean-platform:
-	$(RM) test/run-{tests,benchmarks}.dSYM $(OBJS) $(OBJS:%.o=%.pic.o)
+	$(RM) test/run-{tests,benchmarks}.dSYM $(OBJS) $(OBJS:%.o=%.pic.o) src/unix/uv-dtrace.h
 
 %.pic.o %.o:  %.m
 	$(OBJC) $(CPPFLAGS) $(CFLAGS) -c $^ -o $@
+
+src/unix/uv-dtrace.h:
```
bnoordhuis: Shouldn't this depend on src/unix/uv-dtrace.d?

(EDIT: Ditto for the src/unix/dtrace.o and src/unix/dtrace.pic.o targets.)

config-unix.mk
```diff
@@ -58,11 +63,14 @@ OBJS += src/inet.o
...
 
 ifeq (sunos,$(PLATFORM))
+HAVE_DTRACE=1
 CPPFLAGS += -D__EXTENSIONS__ -D_XOPEN_SOURCE=500
 LDFLAGS+=-lkstat -lnsl -lsendfile -lsocket
 # Library dependencies are not transitive.
 RUNNER_LDFLAGS += $(LDFLAGS)
 OBJS += src/unix/sunos.o
+DTRACE_OBJS += src/unix/core.o
+DTRACE_OUT=src/unix/dtrace.o
```
bnoordhuis: Maybe I'm missing something but why don't you append it to OBJS? That would let you drop DTRACE_OUT altogether, right?

uv.gyp
```diff
@@ -1,4 +1,9 @@
 {
+  'variables': {
+    'uv_use_dtrace%': 'false',
+    'uv_prodir': '',
```
bnoordhuis: What is the purpose of uv_prodir?

tjfontaine: the expansion of <(PRODUCT_DIR) and following used in the `dtrace -G` stanzas won't expand to the right location because it's dependent on where the top level project includes it, so you need to inform the dependent project of that path, see tjfontaine/node@4f9cd83

src/unix/internal.h
```diff
@@ -256,4 +256,11 @@ static void uv__update_time(uv_loop_t* loop) {
   loop->time = uv__hrtime() / 1000000;
 }
 
+#ifdef HAVE_DTRACE
+#include "uv-dtrace.h"
+#else
+#define UV_TICK_START(arg0, arg1)
+#define UV_TICK_STOP(arg0, arg1)
+#endif
+
```
tjfontaine, line 265: This ends up being pretty loud because of the warnings because of how dtrace macros have $

config-unix.mk
```diff
@@ -132,12 +141,17 @@ else
...
 endif
 
+ifeq ($(HAVE_DTRACE), 1)
+DTRACE_HEADER=src/unix/uv-dtrace.h
+CFLAGS += -DHAVE_DTRACE
+endif
+
 libuv.a: $(OBJS)
-	$(AR) rcs $@ $^
+	$(AR) rcs $@ $(OBJS)
```
bnoordhuis: Undo?

config-unix.mk
```diff
@@ -132,12 +141,17 @@ else
...
 
 libuv.$(SOEXT):	override CFLAGS += -fPIC
 libuv.$(SOEXT):	$(OBJS:%.o=%.pic.o)
-	$(CC) -shared -o $@ $^ $(LDFLAGS)
+	$(CC) -shared -o $@ $(OBJS:%.o=%.pic.o) $(LDFLAGS)
```
bnoordhuis: Ditto.

uv.gyp
bnoordhuis: Okay, noted. Maybe add a comment and give it a more descriptive name.

config-unix.mk
tjfontaine: damn I knew I forgot something

bnoordhuis:
Thanks TJ, landed.

PR #780, opened 2013-04-16
unix, win: Added uv_is_refd function

saghul:
I'm considering making ref a property in pyuv instead of having ref/unref functions, and in order to make it consistent I'd need to make it return the current value if queried. This function adds the ability to do that.


bnoordhuis:
On the one hand, it's something that you can perfectly track yourself. On the other hand, it's a minor addition and probably quite useful. A tentative +0.5 from me. uv_is_refd() is a really ugly name though, what about e.g. uv_has_ref()?

@piscisaureus Opinions?

saghul:
I'm fine with the name change :-)

PR #783, opened 2013-04-17
test, sunos: disabled process_title test.

bajtos:
Disabled unit test failing due to missing implementation
of uv_(set|get)_process_title for Sun OS (SmartOS).

Based on discussion with @tjfontaine, such implementaton is difficult
if possible at all and it won't be done anytime soon. Thus there is
no point in keeping the failing test around.


bnoordhuis:
Landed in 0564ee4, thanks.

PR #784, opened 2013-04-17
build: avoid double / in paths for dtrace

tjfontaine:
GYP doesn't resolve or normalize the paths, resulting in the action not
firing if the path contains // so take care to avoid that


bnoordhuis:
Landed in 2d89b4b, thanks.

PR #785, opened 2013-04-18
windows: deal with the fact that GetTickCount might lag

piscisaureus:
We use GetQueuedCompletionStatus(Ex) to sleep the thread until the next
timer expires (provided that no other events happen before that).
However after waking up from a sleep the GetTickCount() return value may
not immediately reflect that some time has passed. This happens because
gqcs can sometimes sleep for periods shorter than the GetTickCount clock
resulution. This patch changes time tracking so the amount of time
waited by gqcs is taken into account.

This has the following advantages:
- Excessive loop iterations are avoided.
- Small timeouts are fired more precisely.
- The `loop-stop` test that used to be flaky on Windows now passes
  consistently.

@bnoordhuis or @sblom -> review?

### DIFF
--- a/src/win/timer.c
+++ b/src/win/timer.c
@@ -29,19 +29,38 @@
 
 
 void uv_update_time(uv_loop_t* loop) {
-  DWORD ticks = GetTickCount();
+  DWORD ticks;
+  ULARGE_INTEGER time;
 
-  /* The assumption is made that LARGE_INTEGER.QuadPart has the same type */
-  /* loop->time, which happens to be. Is there any way to assert this? */
-  LARGE_INTEGER* time = (LARGE_INTEGER*) &loop->time;
+  ticks = GetTickCount();
 
-  /* If the timer has wrapped, add 1 to it's high-order dword. */
+  /* GetTickCount() can conceivably wrap around, so when the current tick */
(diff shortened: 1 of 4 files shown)


src/win/timer.c
```diff
@@ -29,19 +29,38 @@
...
-  /* If the timer has wrapped, add 1 to it's high-order dword. */
+  /* GetTickCount() can conceivably wrap around, so when the current tick */
+  /* count is lower than the last tick count, we'll assume it has wrapped. */
   /* uv_poll must make sure that the timer can never overflow more than */
   /* once between two subsequent uv_update_time calls. */
-  if (ticks < time->LowPart) {
-    time->HighPart += 1;
-  }
-  time->LowPart = ticks;
+  time = *(ULARGE_INTEGER*) &loop->time;
```
bnoordhuis, line 41: Doesn't msvc complain loudly about aliasing here?

bnoordhuis:
I don't know, I still think it's a deeply evil hack.

src/win/timer.c
piscisaureus, line 41: No, complaining about aliasing is a gcc-ism :)

piscisaureus:
@sblom, can I have your opinion here as well. Terrible hack or acceptable solution?

PR #786, opened 2013-04-18
Improvement of tests for unimplemented functionality

bajtos:
At the moment there are two ways how to handle the situation when certain functionality is implemented for some platforms only:
1. Leave the test failing.
2. Skip the test (don't check anything and return success), optionally logging a explanation (which is actually silently discarded by test runner).

Neither option is good. I am proposing a different solution that would allow us to clearly distinguish test for missing functionality, i.e. answering the following questions:
1. Which tests are failing because of missing platform-specific implementation?
2. Why is the functionality missing (short explanation next to the failing test).

The solution is based on TAP directives SKIP and TODO - see [this page](http://podwiki.hexten.net/TAP/TAP.html?page=TAP#DIRECTIVES) for more details.

The code in test/runner.c is not finished yet, it will require some polishing. Also Windows version of process_read_last_line() is not implemented yet.

@tjfontaine Could you please run this PR on build server, so that we can see how TAP runner handles SKIP and TODO directives?

@bnoordhuis @piscisaureus What do you thing about this solution? Is it something you would accept into libuv and use in your tests?


bnoordhuis:
> What do you thing about this solution? Is it something you would accept into libuv and use in your tests?

I can live with that.

> Commit message must indicate the subsystem this commit changes

@tjfontaine Jenkins bug? The commit log starts with 'test: '.

bajtos:
@bnoordhuis I think Nodejs-Jenkins picked up the pull request before I force-pushed amended commit with corrected log.

tjfontaine:
Completed, see
- http://jenkins.nodejs.org/job/libuv-pullrequest/label=smartos/lastCompletedBuild/tapResults/?
- http://jenkins.nodejs.org/job/libuv-pullrequest/label=smartos/lastCompletedBuild/tapTestReport/?

bajtos:
I have finished the work and squashed commits into a single one. Could you please review the changes and accept/merge the pull request?

test/runner-unix.c
```diff
@@ -286,6 +286,33 @@ int process_copy_output(process_info_t *p, int fd) {
 }
 
 
+/* Copy the last line of the stdio output buffer to `buffer` */
+int process_read_last_line(process_info_t *p, char * buffer,
+     size_t buffer_len) {
```
bnoordhuis: Style: if the arguments don't fit on a single line, they need to line up with each argument on a separate line (except for the first one). That is:

``` c
int process_read_last_line(process_info_t *p,
                           char* buffer,
                           size_t buffer_len) {
```

Also, note that the convention is `char* ptr`. (See the line below.)

test/runner-unix.c
```diff
@@ -286,6 +286,33 @@ int process_copy_output(process_info_t *p, int fd) {
...
+
+  int r = fseek(p->stdout_file, 0, SEEK_SET);
+  if (r < 0) {
+    perror("fseek");
+    return -1;
+  }
+
+  buffer[0] = '\0';
+
+  while (fgets(buffer, buffer_len, p->stdout_file) != NULL ) {
```
bnoordhuis: Style: no space before closing parenthesis.

test/runner-win.c
```diff
@@ -248,6 +248,42 @@ int process_copy_output(process_info_t *p, int fd) {
 }
 
 
+int process_read_last_line(process_info_t *p, char * buffer,
+     size_t buffer_len) {
+  DWORD read;
+  char buf[1024];
+  char *line, *start, *buffer_end;
```
bnoordhuis: Style: each local variable should preferably go on its own line. That is:

``` c
char* line;
char* start;
char* buffer_end;
```

If you never mutate the memory a pointer points to, make it const.

test/runner-win.c
```diff
@@ -248,6 +248,42 @@ int process_copy_output(process_info_t *p, int fd) {
...
+  size_t count;
+
+  if (SetFilePointer(p->stdio_out, 0, 0, FILE_BEGIN) == INVALID_SET_FILE_POINTER)
+    return -1;
+
+  buffer_end = buffer;
+
+  while (ReadFile(p->stdio_out, (void*)&buf, sizeof(buf), &read, NULL) &&
```
bnoordhuis: Superfluous cast to void*. Also, &buf == buf so we don't usually qualify it further.

test/runner.c
```diff
@@ -31,12 +31,16 @@
...
+    int skipped, const char* name) {
+  int progress;
+
   if (total == 0)
     total = 1;
 
-  LOGF("[%% %3d|+ %3d|- %3d]: %s", (int) ((passed + failed) / ((double) total) * 100.0),
-      passed, failed, name);
+  progress = (int) ((passed + failed + todos) / ((double) total) * 100.0);
+  LOGF("[%% %3d|+ %3d|- %3d|T %3d|S %3d]: %s", progress,
```
bnoordhuis: Style: the long arguments rule applies here too: one per line.

test/runner.c
```diff
@@ -106,13 +112,16 @@ int run_tests(int timeout, int benchmark_output) {
...
     }
 
-    if (run_test(task->task_name, timeout, benchmark_output, current) == 0) {
-      passed++;
-    } else {
-      failed++;
+    test_result = run_test(task->task_name, timeout, benchmark_output, current);
+    switch (test_result)
+    {
+      case TEST_OK: passed++; break;
```
bnoordhuis: Style: brace on the same line as the switch keyword, don't indent the case keywords. (We're not overly consistent with that second rule but most of the violations are old code.)

test/task.h
```diff
@@ -119,4 +119,22 @@
 /* Format big numbers nicely. WARNING: leaks memory. */
 const char* fmt(double d);
 
+enum test_status {
+  TEST_OK = 0,
+  TEST_TODO = 64,
+  TEST_SKIP = 65
```
bnoordhuis: Is there a logic behind this numbering scheme?

test/task.h
```diff
@@ -119,4 +119,22 @@
...
+};
+
+#define RETURN_OK return TEST_OK
+
+#define RETURN_TODO(explanation) do { \
```
bnoordhuis: Style: backslashes go on the 79th column.

bnoordhuis: And the preferred form is this (imagine proper indented backslashes here):

``` c
#define FOO(bla) \
  do {
    bar();
    baz();
  }
  while (0)
```

test/task.h
bajtos: Yes, 0 is already used in existing test cases. The other two numbers (64 and 65) are sort of arbitrary, however they must not clash with other exit codes (>128 are used for crashes on Unix, small numbers might be potentially used in existing tests - although I didn't check).

test/runner-unix.c
bajtos: Well we should really come to a shared agreement on this item. @piscisaureus prefers the style I used, this style is used in other places in libuv too.

test/runner-win.c
bajtos: This is same style as used by a function above:

```
int process_copy_output(process_info_t *p, int fd) {
  DWORD read;
  char buf[1024];
  char *line, *start;
```

It seems like another inconsistent rule? (/cc @piscisaureus)

test/task.h
bnoordhuis: None of the tests do except fail_always but that one is disabled. (It's a sanity check for the test runner itself.)

I don't really care about the numbering but add a comment that explains the reasoning behind it. Either that, or use the default values.

test/runner-unix.c
bnoordhuis: I invoke the unwritten first rule of open source: In case of doubt, he with the most commits gets to make the call. Sorry Bert, overruled.

test/task.h
```diff
@@ -119,4 +119,25 @@
...
 const char* fmt(double d);
 
+/* Reserved test exit codes. */
+enum test_status {
+  TEST_OK,
+  TEST_TODO,
+  TEST_SKIP
+};
+
+#define RETURN_OK return TEST_OK
```
piscisaureus: Use RETURN_OK() instead of RETURN_OK

test/runner-unix.c
```diff
@@ -286,6 +286,34 @@ int process_copy_output(process_info_t *p, int fd) {
 }
 
 
+/* Copy the last line of the stdio output buffer to `buffer` */
+int process_read_last_line(process_info_t *p,
+                           char * buffer,
+                           size_t buffer_len) {
+  char * ptr;
```
bnoordhuis: Style: char*

test/runner-unix.c
```diff
@@ -286,6 +286,34 @@ int process_copy_output(process_info_t *p, int fd) {
 }
 
 
+/* Copy the last line of the stdio output buffer to `buffer` */
+int process_read_last_line(process_info_t *p,
+                           char * buffer,
```
bnoordhuis: Here too.

test/runner-win.c
```diff
@@ -248,6 +248,44 @@ int process_copy_output(process_info_t *p, int fd) {
...
+  const char *start;
+  char *buffer_end;
+  size_t count;
+
+  if (SetFilePointer(p->stdio_out, 0, 0, FILE_BEGIN) == INVALID_SET_FILE_POINTER)
+    return -1;
+
+  buffer_end = buffer;
+
+  while (ReadFile(p->stdio_out, &buf, sizeof(buf), &read, NULL) && read > 0) {
```
piscisaureus: Distinguish ReadFile failure from EOF

test/runner.c
```diff
@@ -106,13 +117,15 @@ int run_tests(int timeout, int benchmark_output) {
...
+      log_progress(total, passed, failed, todos, skipped, task->task_name);
     }
 
-    if (run_test(task->task_name, timeout, benchmark_output, current) == 0) {
-      passed++;
-    } else {
-      failed++;
+    test_result = run_test(task->task_name, timeout, benchmark_output, current);
+    switch (test_result) {
+    case TEST_OK: passed++; break;
```
piscisaureus, line 135: indentation

test/runner.c
```diff
@@ -31,12 +31,21 @@
 int tap_output = 0;
 
 
-static void log_progress(int total, int passed, int failed, const char* name) {
+static void log_progress(int total, int passed, int failed, int todos,
+    int skipped, const char* name) {
```
bnoordhuis: Style: arguments should line up.

test/runner.c
```diff
@@ -121,13 +134,48 @@ int run_tests(int timeout, int benchmark_output) {
...
 
 
+void log_tap_result(int test_count, const char * test, int status,
+    process_info_t * process) {
+  const char * result;
+  const char * directive;
+  char reason[1024];
+
+  switch (status) {
+  case TEST_OK:
```
piscisaureus, line 163: indent

test/runner.c
bajtos, line 135: If I remember correctly, I was told by @bnoordhuis to not indent `case` statements.

Seems that I am not consistent about this rule (see `log_tap_result()` below).

What style shall I use then?

test/runner.c
```diff
@@ -78,7 +87,7 @@ const char* fmt(double d) {
 
 
 int run_tests(int timeout, int benchmark_output) {
-  int total, passed, failed, current;
+  int total, passed, failed, todos, skipped, current, test_result;
```
bnoordhuis: While you're here, can you put every declaration on its own line?

test/runner.c
```diff
@@ -121,13 +134,48 @@ int run_tests(int timeout, int benchmark_output) {
...
   }
 
   return failed;
 }
 
 
+void log_tap_result(int test_count, const char * test, int status,
+    process_info_t * process) {
+  const char * result;
+  const char * directive;
```
bnoordhuis: Style issues.

test/task.h
bnoordhuis: And wrap it in a do/while so `RETURN_OK(), 42` results in a build error.

test/task.h
```diff
@@ -119,4 +119,25 @@
 /* Format big numbers nicely. WARNING: leaks memory. */
 const char* fmt(double d);
 
+/* Reserved test exit codes. */
+enum test_status {
+  TEST_OK,
```
piscisaureus: you probably want to make `TEST_OK = 0` explicit

test/runner.c
```diff
@@ -121,13 +134,48 @@ int run_tests(int timeout, int benchmark_output) {
...
   if (!benchmark_output && !tap_output) {
-    log_progress(total, passed, failed, "Done.\n");
+    log_progress(total, passed, failed, todos, skipped, "Done.\n");
   }
 
   return failed;
 }
 
 
+void log_tap_result(int test_count, const char * test, int status,
```
piscisaureus: Style: use `const char* var`.

test/runner.c
```diff
@@ -121,13 +134,48 @@ int run_tests(int timeout, int benchmark_output) {
...
+    process_info_t * process) {
```
piscisaureus: ditto

test/runner.c
```diff
@@ -121,13 +134,48 @@ int run_tests(int timeout, int benchmark_output) {
...
+  const char * result;
```
piscisaureus: ditto

test/runner.c
bnoordhuis, line 135: @bajtos is right, we're not being consistent here. We don't indent case keywords in uv-unix and node.js so I move we stick to that.

bnoordhuis, line 135: On that note, we should probably come up with a common style guide for uv-unix and uv-win.

bajtos, line 135: Having a way to verify coding-style conformance (like running `make clint`)
would be event better.

On 8 May 2013 10:29, Ben Noordhuis notifications@github.com wrote:

> In test/runner.c:
> 
> > ```
> >  }
> > ```
> > -    if (run_test(task->task_name, timeout, benchmark_output, current) == 0) {
> > -      passed++;
> > -    } else {
> > -      failed++;
> > -    test_result = run_test(task->task_name, timeout, benchmark_output, current);
> > -    switch (test_result) {
> > -    case TEST_OK: passed++; break;
> 
> On that note, we should probably come up with a common style guide for
> uv-unix and uv-win.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/joyent/libuv/pull/786/files#r4138871
> .

## 

http://about.me/bajtos

piscisaureus, line 135: Google style guides

bnoordhuis, line 135: Then you have some cleaning up to do in src/win: weird comment style, function arguments don't line up, etc.. :-)

I'm okay with switching to google's style guide even though I hate lint patches with a passion.

test/runner-unix.c
bajtos: Fixed in 9ca3849730dbccb2f1ab84dcc0bb68754e61f4cf.

test/runner-unix.c
bajtos: Fixed in 9ca3849730dbccb2f1ab84dcc0bb68754e61f4cf.

test/runner.c
bajtos: Fixed in 9ca3849730dbccb2f1ab84dcc0bb68754e61f4cf.

test/runner.c
bajtos: Fixed in 9ca3849730dbccb2f1ab84dcc0bb68754e61f4cf.

test/runner.c
bajtos, line 135: I'll leave this unchanges as per Ben's comment.

test/runner.c
bajtos, line 163: I'll leave this unchanges as per Ben's comment above.

test/runner.c
bajtos: Fixed in 9ca3849730dbccb2f1ab84dcc0bb68754e61f4cf.

test/runner.c
bajtos: Fixed in 9ca3849730dbccb2f1ab84dcc0bb68754e61f4cf.

test/runner.c
bajtos: Fixed in 9ca3849730dbccb2f1ab84dcc0bb68754e61f4cf.

test/runner.c
bajtos: Fixed in 9ca3849730dbccb2f1ab84dcc0bb68754e61f4cf.

test/task.h
bajtos: Fixed in 9ca3849730dbccb2f1ab84dcc0bb68754e61f4cf.

test/task.h
bajtos: Fixed in 9ca3849730dbccb2f1ab84dcc0bb68754e61f4cf.

test/runner.c
piscisaureus, line 135: http://google-styleguide.googlecode.com/svn/trunk/cppguide.xml?showone=Loops_and_Switch_Statements#Loops_and_Switch_Statements

bajtos:
All issues are fixed, could you please review?

@piscisaureus The code you sent me for process_read_last_line() works well, see the second commit.

test/runner.h
```diff
@@ -143,6 +143,10 @@ int run_test(const char* test,
 /* Copy the contents of the stdio output buffer to `fd`. */
 int process_copy_output(process_info_t *p, int fd);
 
+/* Copy the last line of the stdio output buffer to `buffer` */
+int process_read_last_line(process_info_t *p, char * buffer,
+     size_t buffer_len);
```
bnoordhuis: Style.

test/runner.c
```diff
@@ -121,13 +144,50 @@ int run_tests(int timeout, int benchmark_output) {
...
+    result = "ok";
+    directive = " # SKIP ";
+    break;
+  default:
+    result = "not ok";
+    directive = "";
+  }
+
+  if ((status == TEST_SKIP || status == TEST_TODO)
+       && process_output_size(process) > 0) {
```
bnoordhuis: && goes on the previous line.

test/task.h
```diff
@@ -119,4 +119,28 @@
...
+/* Reserved test exit codes. */
+enum test_status {
+  TEST_OK = 0,
+  TEST_TODO,
+  TEST_SKIP
+};
+
+#define RETURN_OK ()                                                          \
+  do {                                                                        \
+  return TEST_OK;                                                             \
```
bnoordhuis: Indentation.

bnoordhuis: Oh, and no spaces between RETURN_OK and ().

bnoordhuis:
Landed in 2c21050, thanks!

PR #788, opened 2013-04-21
Make the library more FFI-friendly

za-creature:
Hello,

I'm trying to port pyuv to ctypes so that it'll run in pypy as well as cpython. However, I'm having some real trouble importing the C structs: there seem to be quite a number of conditional defines that are really difficult to replicate in a non-C language.

The proposed patch attempts to fix this by creating wrapper functions on top of each structure's members.


saghul:
hi @za-creature I have wanted to do this for a while, but didn't find the time yet :-S Someone started a cffi based port here: https://github.com/inercia/pyuv which doesn't need this changes to libuv. Given that cffi appears to be the way forward instead of ctypes, is something like this really needed? I believe others have implemented FFI bindings for libuv already without the need for it.

bnoordhuis:
Thanks but I won't take this patch. It has technical shortcomings (the functions are non-static, meaning you'll get symbol redefinition errors when more than one source file includes uv.h) but it also strikes me as solving it in the wrong place. @saghul already mentioned cffi. If you want to use ctypes, I suggest adding a small native module that exposes the fields you want.

za-creature:
@saghul You are right, in a strict sense, this is not required. The structures _could_ be defined in both ctypes and cffi without any changes to libuv. I'm using ctypes because, quite honestly, I had no idea cffi even existed and http://doc.pypy.org/en/latest/extending.html suggested ctypes as the prime candidate.
@bnoordhuis I'm sorry for pressing the issue, but while I'm not going to pretend that this is a democracy, I still feel like this was closed a bit early. Technical issues aside (making everything static is a 5 minute job), I disagree that this is not libuv's issue: defining and testing all the possible structures on all platforms is a far more time consuming task than updating libuv, and I'm not just talking about python here. I'm talking about any language that is not written in C to begin with (the 'add a native module' is not a solution for PyPy, as it's written in Python). More to the point, functions that help FFI already exist in libuv (I'm talking about uv_req_size and uv_handle_size), and as far as I can tell, their sole purpose is being a cross-platform sizeof operator. I'm not insisting on getting this pushed (actually, I'm quite biased against it, as I haven't written C for quite some time), all I'm saying is that something should be done, and that it should be done in libuv.

bnoordhuis:
> defining and testing all the possible structures on all platforms is a far more time consuming task than updating libuv,

There's no need for that. You can either ship a small C wrapper that exposes the fields you want or use a little ctypes fu like below:

``` python
from ctypes import *
libuv = CDLL('/path/to/libuv.so')
uv_handle_size = libuv.uv_handle_size
uv_handle_size.argtypes = [c_int]
uv_handle_size.restype = c_size_t
UV_TCP = 12
class uv_tcp_t(Structure): pass
uv_tcp_t._fields_ = [('data', c_void_p), ('handle', c_byte * uv_handle_size(UV_TCP))]
```

Then, in your callback you simply subtract `sizeof(c_void_p)` from the handle address and you have your uv_tcp_t wrapper back. The 'data' field can be anything, including a pointer to a python object. Just make sure you get the address calculation right.

(In case you're wondering what's going on, it's the python equivalent of container_of.)

za-creature:
I didn't think about shipping an extra C library that exports additional symbols, thanks! It's actually a pretty clean solution as it doesn't bloat the libuv core, while being being forwards-compatible with future releases at the same time.

As for the second solution, I considered it after noticing that the user data field (which is arguably the only important one in my case) is always the first field in every structure, but then I ran into uv_fs_t, at which point ctypes gave me the one finger salute right back.

Would you be willing to accept a smaller patch that only adds uv_fs_errorno, uv_fs_result and uv_fs_statbuf ? (same as before, the problem is the UV_REQ_PRIVATE_FIELDS which is ifdef'-ed)

bnoordhuis:
> Would you be willing to accept a smaller patch that only adds uv_fs_errorno, uv_fs_result and uv_fs_statbuf ?

Well... maybe. Let me think about it for a bit. If you're taking the C wrapper approach you won't need it though.

PR #789, opened 2013-04-24
stream: fix osx select hack

indutny:
After latest twiddling, `stream->io_watcher.fd` doesn't contain a real
stream's file descriptior anymore. The one from `select_state` should be
used instead.

Zero-initialize `select_state->event` field.

---

@bnoordhuis, I'm really really sorry for this, but it really makes things work. (i.e. right now `node < /dev/tty` is fucked again. I promise you to eventually write test for this stuff, no matter how complicated it will be.


bnoordhuis:
Landed in 2400716.

PR #791, opened 2013-04-26
stream: fix small nit in select hack, add test

indutny:
/cc @bnoordhuis


test/test-osx-select.c
```diff
@@ -0,0 +1,84 @@
...
+
+#ifdef __APPLE__
+
+#include "sys/ioctl.h"
+#include "string.h"
+
+uv_buf_t alloc_cb(uv_handle_t* handle, size_t size) {
+  uv_buf_t res;
+  res.len = 1024;
+  res.base = malloc(res.len);
```
saghul: Maybe use a statically allocated buffer instead? IIRC @bnoordhuis once told me to do that in tests.

indutny: oh, right!

indutny: force pushed.

src/unix/stream.c
```diff
@@ -245,7 +245,6 @@ static void uv__stream_osx_select_cb(uv_async_t* handle, int status) {
   s->events = 0;
   uv_mutex_unlock(&s->mutex);
 
-  assert(events != 0);
```
bnoordhuis, line 248: I put that assert there for a reason, Fedor. If events==0, it means the other thread wakes up the event loop unnecessarily.

test/test-osx-select.c
```diff
@@ -0,0 +1,84 @@
...
+#include "task.h"
+
+#ifdef __APPLE__
+
+#include "sys/ioctl.h"
+#include "string.h"
+
+static char buf[1024];
+
+uv_buf_t alloc_cb(uv_handle_t* handle, size_t size) {
```
bnoordhuis: Function should be static. I suggest you move buf into the function body.

test/test-osx-select.c
```diff
@@ -0,0 +1,84 @@
...
+  uv_buf_t res;
+  res.len = ARRAY_SIZE(buf);
+  res.base = buf;
+  return res;
+}
+
+static int read_count = 0;
+
+void read_cb(uv_stream_t* stream, ssize_t nread, uv_buf_t buf) {
```
bnoordhuis: Ditto.

test/test-osx-select.c
```diff
@@ -0,0 +1,84 @@
...
+static char buf[1024];
+
+uv_buf_t alloc_cb(uv_handle_t* handle, size_t size) {
+  uv_buf_t res;
+  res.len = ARRAY_SIZE(buf);
+  res.base = buf;
+  return res;
+}
+
+static int read_count = 0;
```
bnoordhuis: All static variables should go at the top.

test/test-osx-select.c
```diff
@@ -0,0 +1,84 @@
...
+void read_cb(uv_stream_t* stream, ssize_t nread, uv_buf_t buf) {
+  fprintf(stdout, "got data %d\n", ++read_count);
+
+  if (read_count == 3) {
+    uv_read_stop(stream);
+    uv_close((uv_handle_t*) stream, NULL);
+  }
+}
+
+TEST_IMPL(osx_select) {
```
bnoordhuis, line 48: Two newlines between functions.

test/test-osx-select.c
```diff
@@ -0,0 +1,84 @@
...
+  int r;
+  int fd;
+  size_t i;
+  const char* str;
+  uv_tty_t tty;
+  uv_loop_t* loop = uv_default_loop();
```
bnoordhuis: Use loop consistently or remove it.

test/test-osx-select.c
```diff
@@ -0,0 +1,84 @@
...
+
+  fd = open("/dev/tty", O_RDONLY, 0);
```
bnoordhuis: Minor nit: the third argument is not necessary.

test/test-osx-select.c
```diff
@@ -0,0 +1,84 @@
...
+  r = uv_tty_init(uv_default_loop(), &tty, fd, 1);
+  ASSERT(r == 0);
+
+  uv_read_start((uv_stream_t*) &tty, alloc_cb, read_cb);
+
+  // Emulate user-input
+  str = "got some input\n"
+        "with a couple of lines\n"
+        "feel pretty happy\n";
+  for (i = 0; i < strlen(str); i++) {
```
bnoordhuis: Baby Jesus weeps...

uv.gyp
```diff
@@ -387,6 +387,11 @@
             '_XOPEN_SOURCE=500',
           ],
         }],
+        [ 'OS=="mac"', {
+          'sources': [
+            'test/test-osx-select.c',
```
bnoordhuis: Maybe compile this unconditionally? Consistent with build.mk.

build.mk
```diff
@@ -136,6 +136,7 @@ TESTS= \
 	test/test-udp-send-and-recv.o \
 	test/test-util.o \
 	test/test-walk-handles.o \
+	test/test-osx-select.o \
```
bnoordhuis: Should be sorted alphabetically.

test/test-list.h
```diff
@@ -220,6 +220,9 @@
 TEST_DECLARE   (we_get_signals)
 TEST_DECLARE   (signal_multiple_loops)
 #endif
+#ifdef __APPLE__
+TEST_DECLARE   (osx_select)
+#endif
```
tjfontaine, line 225: I was thinking that if @bajtos skip/todo (#786) stuff was brought it in it should be always be available but SKIP'd on platforms where it isn't used

test/test-osx-select.c
```diff
@@ -0,0 +1,84 @@
...
+ * IN THE SOFTWARE.
+ */
+
+#include "uv.h"
+#include "task.h"
+
+#ifdef __APPLE__
+
+#include "sys/ioctl.h"
+#include "string.h"
```
bnoordhuis: Use proper `<include>` directives here.

test/test-list.h
bnoordhuis, line 225: Agreed but this particular test won't compile on Windows.

test/test-osx-select.c
```diff
@@ -0,0 +1,84 @@
...
+  return res;
+}
+
+static int read_count = 0;
+
+void read_cb(uv_stream_t* stream, ssize_t nread, uv_buf_t buf) {
+  fprintf(stdout, "got data %d\n", ++read_count);
+
+  if (read_count == 3) {
+    uv_read_stop(stream);
```
bnoordhuis: Shouldn't be necessary, right? uv_close() implicitly stops reading.

test/test-list.h
tjfontaine, line 225: that's what the #ifdef in the test is for?

bnoordhuis, line 225: You still need a `TEST_IMPL(osx_select) { ... }` but that's behind an `#ifdef __APPLE__` guard now. But yes, it's possible with some code reshuffling.

src/unix/stream.c
indutny, line 248: Race condition, man, race condition. I don't want to lock for too much time and its just simplier to invoke this function a little bit more often, than it should.

indutny:
Fixed all nits except `events != 0`

src/unix/stream.c
bnoordhuis, line 248: I was going to point out that there cannot be a race condition because all loads and stores to s->events are guarded by a mutex but I see now that uv_is_readable() and uv_is_writable() are called uv__stream_osx_select() outside of any critical sections... That should be addressed.

indutny, line 248: No, actually race happens on `uv_async_send` call which is performed outside mutex.

src/unix/stream.c
```diff
@@ -204,10 +204,10 @@ static void uv__stream_osx_select(void* arg) {
 
     uv_mutex_lock(&s->mutex);
     s->events |= events;
-    uv_mutex_unlock(&s->mutex);
 
     if (events != 0)
```
bnoordhuis: Move the lock/send/unlock into this block? There is no point or'ing when events==0.

There should probably also be a `assert(events != 0 || FD_ISSET(s->int_fd, &sread))` check here.

indutny: sure, ok.

indutny: force pushed.

bnoordhuis:
Thanks Fedor, landed in ac4e7e7.

andrew-d:
Hi there,

Are there plans to apply this to libuv master?  I'm running into the same issue (see the reference to @saghul's evergreen project, above, and joyent/node#5353), and from what I can tell, libuv master is [version 0.11.2](https://github.com/joyent/libuv/blob/master/src/version.c#L30), versus the branch this was applied to being [version 0.10.6](https://github.com/joyent/libuv/blob/v0.10/src/version.c#L33).  I've tried fiddling with this to get it working on the newer code, but I'm not experienced enough with libuv to do so.

Also, same question for #794 , since it looks like both are required to solve the problem.

Thank you!

saghul:
@andrew-d I believe both fixes landed on libuv 0.10.x, so using pyuv 0.10.x and raising the libuv revision will do. I plan to make a new pyuv release later today, btw.

andrew-d:
Yep, works just fine.  Still kinda curious if it'll make it into 0.11.x at some point, my code works now.  Thanks :-)

bnoordhuis:
I've merged v0.10 into master in ebdeaed.

PR #793, opened 2013-04-30
Fix/osx hack harder synchronization restrictions

indutny:
/cc @bnoordhuis


indutny:
Oh, stupid github, wrong base.

PR #794, opened 2013-04-30
stream: use harder sync restrictions for osx-hack

indutny:
Synchronize harder to avoid excessive spins, invokations of async
callback and sporadic assertion failures on double-call of async
callback.

/cc @bnoordhuis


src/unix/stream.c
```diff
@@ -204,11 +204,13 @@ static void uv__stream_osx_select(void* arg) {
 
     assert(events != 0 || FD_ISSET(s->int_fd, &sread));
     if (events != 0) {
-      uv_mutex_lock(&s->mutex);
-      s->events |= events;
+      s->events = events;
```
bnoordhuis, line 207: Use ACCESS_ONCE here.

src/unix/stream.c
```diff
@@ -242,10 +244,9 @@ static void uv__stream_osx_select_cb(uv_async_t* handle, int status) {
   stream = s->stream;
 
   /* Get and reset stream's events */
-  uv_mutex_lock(&s->mutex);
   events = s->events;
   s->events = 0;
```
bnoordhuis, line 248: And here.

indutny:
Landed in 67f9b91

PR #798, opened 2013-05-11
build: If user already has $PLATFORM defined, convert to lower case

staticfloat:
Useful for when users already have `$PLATFORM` set to `uname -s`, but on e.g. OSX `uname -s` is equal to "Darwin", therefore the build is missing some symbols.


staticfloat:
Shrunk commit message to 49 characters

bnoordhuis:
Good call. Landed in 96a2df8, thanks.

PR #805, opened 2013-05-21
uv: support for android build

linusmartensson:
Adds targets and fixes to ensure building for android works.

To build on android, refer to the android standalone toolchain.

### DIFF
--- a/src/unix/linux-core.c
+++ b/src/unix/linux-core.c
@@ -36,7 +36,10 @@
 #include <fcntl.h>
 #include <time.h>
 
+#ifndef ANDROID
 #define HAVE_IFADDRS_H 1
+#endif
+
 #ifdef __UCLIBC__
 # if __UCLIBC_MAJOR__ < 0 || __UCLIBC_MINOR__ < 9 || __UCLIBC_SUBLEVEL__ < 32
 #  undef HAVE_IFADDRS_H
--- a/src/unix/linux-syscalls.h
+++ b/src/unix/linux-syscalls.h
@@ -28,6 +28,7 @@
 #include <stdint.h>
 #include <signal.h>
 #include <sys/types.h>
+#include <sys/time.h>
 #include <sys/socket.h>
 
 #if defined(__alpha__)
--- a/src/unix/pthread-fixes.c
+++ b/src/unix/pthread-fixes.c
@@ -1,0 +1,80 @@
+/* Copyright (c) 2013, Sony Mobile Communications AB
+ * Copyright (c) 2012, Google Inc.
+   All rights reserved.
+
+   Redistribution and use in source and binary forms, with or without
+   modification, are permitted provided that the following conditions are
+   met:
+
+     * Redistributions of source code must retain the above copyright
+   notice, this list of conditions and the following disclaimer.
+       * Redistributions in binary form must reproduce the above
+   copyright notice, this list of conditions and the following disclaimer
+   in the documentation and/or other materials provided with the
+   distribution.
+       * Neither the name of Google Inc. nor the names of its
+   contributors may be used to endorse or promote products derived from
+   this software without specific prior written permission.
+
+   THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+   "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+   LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+   A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+   OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+   SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+   LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+   DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+   THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+   (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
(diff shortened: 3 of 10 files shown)


lucab:
Hi, as part of rust we did the same android porting; the code is here https://github.com/lucab/libuv/commit/f71151f54aabef0be62b858ca239775326b4a88d
I was going to submit a PR, when spotted this one and didn't proceed with mine.

Looking at the code, it looks like it is almost identical (yours also a bit cleaner), so I'll probably drop mine. 
What's the status of having this PR merged? (So that rust can go back to mainline libuv...)

src/unix/linux-syscalls.h
```diff
@@ -28,6 +28,9 @@
 #include <stdint.h>
 #include <signal.h>
 #include <sys/types.h>
+#ifdef ANDROID
+#include <time.h>
```
bnoordhuis: I'm guessing that's for struct timespec? You can include it unconditionally, I was relying on timespec being pulled in implicitly.

src/unix/linux-core.c
```diff
@@ -36,7 +36,10 @@
 #include <fcntl.h>
 #include <time.h>
 
+#ifndef ANDROID
 #define HAVE_IFADDRS_H 1
+#endif
```
bnoordhuis, line 41: Doesn't android have ifaddrs.h?

EDIT: Okay, looks like it doesn't. I guess we could implement getifaddrs() ourselves through ioctl(SIOCGIFCONF) and AF_NETLINK sockets but meh, I don't think many people care. I'll wait for the bug reports to come in first. :-)

src/unix/stream.c
```diff
@@ -705,8 +705,20 @@ static void uv__write(uv_stream_t* stream) {
   iovcnt = req->bufcnt - req->write_index;
 
   /* Limit iov count to avoid EINVALs from writev() */
+#if defined(ANDROID)
+  /*
+   * Android does not have this value predefined,
+   * we need to retrieve it using the syscnf interface.
```
bnoordhuis: That sounds unlikely. Some quick grepping through an android source tree suggests it's there. Could it be there's a compiler flag or define hiding it?

src/unix/thread.c
```diff
@@ -26,9 +26,9 @@
 #include <assert.h>
 #include <errno.h>
 
-#if defined(__APPLE__) && defined(__MACH__)
+#if (defined(__APPLE__) && defined(__MACH__)) || defined(ANDROID)
 #include <sys/time.h>
-#endif /* defined(__APPLE__) && defined(__MACH__) */
+#endif /* (defined(__APPLE__) && defined(__MACH__)) || defined(ANDROID) */
```
bnoordhuis: Feel free to include that header unconditionally. I'm not sure why it's wrapped in an #ifdef guard in the first place.

bnoordhuis:
Couple of nits but mostly LGTM. If you fix them up, I'll land it.

src/unix/stream.c
linusmartensson: Unfortunately, it's not available in the bionic version of libc at all, which is what we use from the NDK. The alternative to fetching it dynamically would be to provide a default value on ifndef IOV_MAX, but I felt a dynamic fetch via sysconf was more flexible.

There's references to a default definition in some prebuilt components here, but this is as far as I can tell a fallback system for gcc internals:
http://androidxref.com/4.2.2_r1/search?q=IOV_MAX&defs=&refs=&path=&hist=&project=prebuilts

Other external packages that rely on the variable define it themselves:
http://androidxref.com/4.2.2_r1/search?q=IOV_MAX&defs=&refs=&path=&hist=&project=external

bnoordhuis: Hm, okay. Can you move this code into a separate function called e.g. getiovmax? I suggest replacing the `#if defined(ANDROID)` with `#if !defined(IOV_MAX)`.

src/unix/linux-core.c
linusmartensson, line 41: Double-checked, and no, it's not available in the SDK. The dhcpd-implementation in the android sources (http://androidxref.com/4.2.2_r1/xref/external/dhcpcd/) provides it's own implementation of ifaddrs based on traversal of sysfs. I presume it could be copied to NodeJS under the BSD/Apache license referenced in that library, but this should probably be a separate patch.

src/unix/stream.c
piscisaureus: Would it be better to make `iovmax` of type `sig_atomic_t` ?

bnoordhuis: Why? Because of concurrent loads / stores?  That doesn't really matter, at worst you make a few superfluous calls to sysconf().  Besides, sig_atomic_t is an int on all platforms and architectures. :-)

src/unix/thread.c
linusmartensson: Done in latest patch version.

src/unix/stream.c
linusmartensson: Moved to separate function, and I added a final fallback on 1024 if neither _SC_IOV_MAX_ nor IOV_MAX was available.

src/unix/pthread-fixes.c
```diff
@@ -0,0 +1,84 @@
...
+   SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+   LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+   DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+   THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+   (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+   OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+   This contains Pthread-related functions not provided by the Android NDK
+   but required by the Breakpad unit test. The functions are inlined here
+   in a C++ anonymous namespace in order to keep the build files simples.
```
piscisaureus: Remove the comment about C++ namespace and breakpad?

uv.gyp
```diff
@@ -20,7 +20,8 @@
         'conditions': [
           ['OS=="solaris"', {
             'cflags': [ '-pthreads' ],
-          }, {
+          }],
+          ['OS!="android"', {
```
bnoordhuis, line 24: Just realized that this breaks the sunos gyp build. I'll see if I can fix it up.

bnoordhuis:
Thanks Linus, landed with some fix-ups in fc6a2ad.

PR #809, opened 2013-05-25
unix: appease warning about non-standard `inline`

chisophugis:
Clang warns about using `inline`, which is not technically allowed in
C89 mode (libuv compiles with `-std=c89`). It's probably best to leave
it to the compiler to do the inlining anyway.

The exact warnings that Clang emits are:

```
../../deps/uv/src/unix/fs.c:500:8: warning: extension used [-Wlanguage-extension-token]
static inline void uv__to_stat(struct stat* src, uv_stat_t* dst) {
       ^
1 warning generated.
```

```
../../deps/uv/src/unix/signal.c:119:1: warning: extension used [-Wlanguage-extension-token]
inline static uv_signal_t* uv__signal_first_handle(int signum) {
^
1 warning generated.
```


chisophugis:
Derp, used my other email address, which is not the one I signed the CLA with (I contributed this tiny patch to node a while back https://github.com/joyent/node/pull/5332). Should be good now.

bnoordhuis:
Thanks, landed in c396486.

Issue #796, opened 2013-05-06
Inconsistent execution of close callbacks


saghul:
I know there were some discussions about this in the past, but somehow this never affected me, until today :-S

Here is an example program written with pyuv:

```
import pyuv

def cb1(handle):
    print "cd1"

def cb2(handle):
    print "cb2"
    handle.close()

def check1(handle):
    print "check!"


loop = pyuv.Loop()

c1 = pyuv.Check(loop)
c1.start(check1)
t1 = pyuv.Timer(loop)
t1.start(cb1, 100, 0)
t2 = pyuv.Timer(loop)
t2.start(cb2, 0.01, 0)

loop.run()
```

On unix this program will print:

```
check!
cb2
check!
# wait for 100 seconds here
...
```

whereas on Windows, it will print:

```
check!
cb2
# wait for 100 seconds here
...
```

The reason seems to be that on Windows endgames are processed _before_ blocking for i/o, but on unix they are processed after. This means that on unix the next loop iteration will not block for i/o (hence the check handle will run) but on windows it will block, because the conditions not to block are not met.

This is a synthetic example, in my real use case (the evergreen project), I have some sort of event loop driven scheduling system, where timers just insert their callbacks in a queue (when they kick), which is processed in a check handle, so on windows, this means I'm in trouble. (hopefully I explained myself correctly here)

If I move the call to `uv_process_endgames` to after running the check handles things work as I expected, though I haven't checked if there are any further implications. So I'd like the windows implementation to behave like unix on this one :-)

Thoughts @piscisaureus @bnoordhuis ?

bnoordhuis:
@piscisaureus?

PR #814, opened 2013-05-29
win: Run close callbacks after polling for i/o

saghul:
Fix for #796

/cc @piscisaureus

### DIFF
--- a/build.mk
+++ b/build.mk
@@ -68,6 +68,7 @@ TESTS= \
 	test/test-barrier.o \
 	test/test-callback-order.o \
 	test/test-callback-stack.o \
+	test/test-close-order.o \
 	test/test-condvar.o \
 	test/test-connection-fail.o \
 	test/test-cwd-and-chdir.o \
--- a/src/win/core.c
+++ b/src/win/core.c
@@ -278,8 +278,6 @@ int uv_run(uv_loop_t *loop, uv_run_mode mode) {
     }
 
     uv_process_reqs(loop);
-    uv_process_endgames(loop);
-
     uv_prepare_invoke(loop);
 
     (*poll)(loop, loop->idle_handles == NULL &&
@@ -291,6 +289,8 @@ int uv_run(uv_loop_t *loop, uv_run_mode mode) {
                   !(mode & UV_RUN_NOWAIT));
 
     uv_check_invoke(loop);
+    uv_process_endgames(loop);
+
     r = uv__loop_alive(loop);
     if (mode & (UV_RUN_ONCE | UV_RUN_NOWAIT))
       break;
--- a/test/test-close-order.c
+++ b/test/test-close-order.c
@@ -1,0 +1,80 @@
+/* Copyright Joyent, Inc. and other Node contributors. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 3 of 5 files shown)


bnoordhuis:
LGTM, not even a style issue to complain about.

/cc @piscisaureus

saghul:
I solved the problem I had in #796 by using an idle handle, however I still think close callback execution should be consistent across platforms. Is there any problem with this approach @piscisaureus?

If we can live with the inconsistency, then feel free to close this, no problem.

piscisaureus:
You're such a bunch of conceited kids :-)

Okay, land it. But don't say I didn't warn you:
- libuv still can't make any guarantees about when the close callback actually runs. It might take many iterations. The fact that the test passes is that the uv_timer_t implementation is not quite as much a black box as actual i/o handles.
-  It might make libuv less efficient. It's unlikely to show up in benchmarks since it shouldn't affect busy processes. But after this change we'll end up always doing a zero-timeout poll before calling close callbacks. That doesn't affect high-throughput applications since the zero-timeout poll would likely pick up actual events. If the loop isn't so busy, it'll use a tiny bit of your laptop's battery, burn your lap a bit more, and cause a little bit more CO2 to be released into the atmosphere.

## How much CO2 exactly?

That's hard to compute. Let's try to calculate it coarsely.

A "fully loaded" modern quad-core cpu consumes about 100 watts (of which approximately half is dissipated as heat), so that's 25 watts per core. Going by the `loop_count` benchmark libuv can do about 1 million "empty" loop iterations per second, and during this benchmark one core on my machine is completely saturated by it.

Let's assume that the core wouldn't use any energy if it was idle. That probably isn't completely true, but there are also effects that move the scale in the other direction (e.g. the additional heat production might trigger the fan to speed up; other cores might run at a higher frequency because one is saturated, etc).

That suggests that the marginal energy consumption would be about 25 w / 1_10^-6 s/iteration = 2.5_10^-5 j/iteration.

CO2 dissipation per joule of electricity produced depends of course on how it is produced. Transport and storage losses also play a role, however let's not make it too complicated. According to [this document](http://www.mfe.govt.nz/publications/climate/electricity-emissions-factor-reports/electricity-factor-aug03.pdf) a typical fossil-fuel burning power plant emits approximately between 500 and 1000 kilograms of CO2 per megawatt-hour, so let's assume 750 Kg/MWh, which is equivalent to about 200 Kg/GJ or 2*10^-7 g/j.

We can just multiply the two numbers we found to compute that 1 empty libuv loop iteration amounts to about 5*10^-12 grams (aka 5 picograms) of CO2 being released into the atmosphere.

Now let's consider the molar mass of CO2 (44.01 g/mol), and it's a piece of cake to conclude that the exhaust flow is about 1.1_10^-13 mol CO2 / iteration. Since a mol is defined as exactly 6.022_10^23 molecules, the number of molecules released per loop iteration would be a little more than 6*10^10.

Yes, you read that right!

60 billion CO2 molecules per loop iteration end up in the atmosphere!

And since closing a watcher now takes an additional loop iteration, that's _60 billion CO2 molecules per close callback_.

This, @saghul, is the harm you have inflicted on humanity.

saghul:
Oh my!

> You're such a bunch of conceited kids :-)
> 
> Okay, land it. But don't say I didn't warn you:
> 
> libuv still can't make any guarantees about when the close callback actually runs. It might take many iterations. The fact that the test passes is that the uv_timer_t implementation is not quite as much a black box as actual i/o handles.

But even if it takes more than one iteration, they wouldn't run before _this_ iteration has ended, which was what I was trying to fix (and hopefully I got right).

> It might make libuv less efficient. It's unlikely to show up in benchmarks since it shouldn't affect busy processes. But after this change we'll end up always doing a zero-timeout poll before calling close callbacks. That doesn't affect high-throughput applications since the zero-timeout poll would likely pick up actual events. If the loop isn't so busy, it'll use a tiny bit of your laptop's battery, burn your lap a bit more, and cause a little bit more CO2 to be released into the atmosphere.

That's what the unix side does and even if it's particularly hot today in Amsterdam, I don't think libuv has anything to do with it :-P

> How much CO2 exactly?
> That's hard to compute. Let's try to compute it coarsely.

[snip]

> This, @saghul, is the harm you have inflicted on humanity.

Don't worry, we'll get some tickets to go to Elysium, once Earth is ruined because of this :-)

saghul:
Rebased and force pushed.

saghul:
This will land in master only, continuing on https://github.com/joyent/libuv/pull/922

PR #815, opened 2013-05-29
Fix the usage of sendfile on OSX to respect the caller's length request.

wynnw:
- The osx version of sendfile takes the length from the &len pointer, so initialize it with the req->len
  instead of just using 0. This was causing the sendfile call to always send to the end of the file,
  making it impossible to send a middle section of a file.


src/unix/fs.c
```diff
@@ -436,11 +436,12 @@ static ssize_t uv__fs_sendfile(uv_fs_t* req) {
      * non-blocking mode and not all data could be written. If a non-zero
      * number of bytes have been sent, we don't consider it an error.
      */
-    len = 0;
 
 #if defined(__FreeBSD__)
+    len = 0;
     r = sendfile(in_fd, out_fd, req->off, req->len, NULL, &len, 0);
 #else
+    len = req->len;  /* need to specify how many bytes to write - don't assume the caller wants to send to the EOF */
```
bnoordhuis, line 444: Style: long line, wrap at 80 columns. Otherwise LGTM.

wynnw:
Created a new pull request that fixes the above issues, closing this one...

PR #816, opened 2013-05-29
darwin: fix the use of sendfile to respect the caller's length parameter

wynnw:
The darwin sendfile implementation uses the &len parameter as input and
output. The code was sending 0 (not using the value of req->len) so
the behavior wasn't what the caller was expecting. This makes sure to
initialize len with req->len to ensure that the caller can send portions
of a file.

(This is my second pull request - the first one's commit message didn't follow the right format, and the line width's were wrong. I've deleted the first pull request)


wynnw:
trying a new pull request to fix the commit message log - closing this.

bnoordhuis:
@wynnw Next time, just force-push and post an 'all fixed' comment on the pull request.

wynnw:
will do - sorry for the extra work.

PR #817, opened 2013-05-29
 darwin: fix sendfile call to respect the caller's length parameter

wynnw:
The darwin sendfile implementation uses the &len parameter as input
 and output. The code was sending 0 (not using the value of
 req->len) so the behavior wasn't what the caller was expecting.
 This makes sure to initialize len with req->len to ensure that the
 caller can send portions of a file (not always everything to the
 end of the file).

(sorry for the churn - messed up the commit message format again)


bnoordhuis:
LGTM but can you sign the CLA? Thanks.

wynnw:
Just signed the CLA. Thanks for reviewing.

bnoordhuis:
Thanks Wynn, landed in b4c658c.

PR #820, opened 2013-05-31
unix: Support Android libuv standalone build

linusmartensson:
Tested most of my compilation in the previous patch on NodeJS
and extracted the patches from there. This additional patch ensures 
libuv will be capable of building standalone as well, both with gyp
and Makefiles.

Build documentation was added to the README.md file.

Some tests are failing, and I have not heavily investigated
the reasons. The failures are generally on error-conditions, and are
likely related to differences between fully POSIX-compatible
systems and android. Resolving these failures should logically
be done in separate bug reports and is currently out of scope
for my work on uv and node.

### DIFF
--- a/config-unix.mk
+++ b/config-unix.mk
@@ -19,7 +19,7 @@
 # IN THE SOFTWARE.
 
 E=
-CFLAGS += -g -Wall -Wextra -Wno-unused-parameter -std=c89
+CFLAGS += -g -Wall -Wextra -Wno-unused-parameter
 CPPFLAGS += -I$(SRCDIR)/src
 LDFLAGS=-lm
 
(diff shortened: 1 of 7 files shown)


include/uv-private/uv-unix.h
```diff
@@ -37,8 +37,17 @@
...
-#ifdef ANDROID
+#ifdef __ANDROID__
 #include "pthread-fixes.h"
+
+/*More missing defines on android*/
+#ifndef S_IREAD
+#define S_IREAD 0
+#endif
+#ifndef S_IWRITE
+#define S_IWRITE 0
```
bnoordhuis: I assume that's to patch up the test suite? That should be addressed in a different way; it doesn't belong in a public header and besides, S_READ and S_IWRITE are obsolete BSD-isms anyway. I think the reason we use them is that Windows doesn't have S_IRUSR and S_IWUSR.

See #821 - once it checks out on Windows, I'll land it in master. Can you rebase and remove this afterwards?

config-unix.mk
```diff
@@ -104,6 +104,18 @@ OBJS += src/unix/linux-core.o \
         src/unix/proctitle.o
 endif
 
+ifeq (android,$(PLATFORM))
+CFLAGS += -D_GNU_SOURCE -std=c99
```
bnoordhuis: Why -std=c99?

include/uv-private/uv-unix.h
linusmartensson: Will do

config-unix.mk
linusmartensson: The android toolchain utilizes inline, which is not available in c89. 
c99 was my knee-jerk response to that. I checked now, and I can also leave this line out completely for a working gnu90 compilation. I'll do that instead.

bnoordhuis:
Landed in 3fdd2a1, thanks!

PR #821, opened 2013-05-31
test: don't use S_IREAD and S_IWRITE

bnoordhuis:
They're BSD-isms and obsolete ones at that. Replace with S_IRUSR and
S_IWUSR. Alias as _S_IREAD and _S_IWRITE on Windows because the '90s
never ended in Redmond, WA.


bnoordhuis:
Landed in ede8318.

PR #822, opened 2013-05-31
uv: Fix the build breakage.

linusmartensson:
Accidentally added a variable that was defined conditionally in the gyp config.

Might be better ways to resolve this, if so, please modify.


bnoordhuis:
Thanks, but I landed an alternative approach in eb1f60c.

PR #824, opened 2013-06-03
Win: EADDRNOTAVAIL when all the ports are in use

Raulmsm:
When several connections are opened to the same host on the same port
on a very short time-span, all the ports will be exhausted and the
error returned was ENOBUFS. I think this error is misleading as I
immediately thought it was related to memory exhaustion on socket
buffers. A google search showed that there are more people assuming the
same.

The error I would expect for this case would be EADDRNOTAVAIL. While it
is still not perfect, I believe it's a stronger indication of the
underlying problem.


tjfontaine:
You do **NOT** need to close and open new issues to make the jenkins bot happy.

This particular issue of the commit message can be fixed with `git commit --amend && git push origin +no-available-ports-error`

Raulmsm:
@tjfontaine Thanks for the help!

Raulmsm:
@piscisaureus

Actually I went for WSAEADDRINUSE instead of WSAEADDRNOTAVAIL when you run out of ports.

Why WSAEADDRINUSE for STATUS_TOO_MANY_ADDRESSES: Looking at winsock reference, the reference states that "When bind is called with a wildcard address (involving ADDR_ANY), a WSAEADDRINUSE error could be delayed until the specific address is committed. ", this leads me to assume that if no port is eventually found available, we're supposed to return WSAEADDRINUSE. 

Why WSAEADDRNOTAVAIL for STATUS_CONFLICTING_ADDRESSES: On this one I'm not so sure as I couldn't find that much information on this apart from the basic error code/code comment, but it seems that it happens when you intend to use an address that is out of the given address range, which seems coherent with this error.

saghul:
@piscisaureus any objections to this? Sounds reasonable IMHO.

saghul:
I take that as a yes :-) Thanks @Raulmsm, landed in https://github.com/joyent/libuv/commit/f326820f55bae77b7d5f91273e51cce796d841c5.

PR #825, opened 2013-06-05
Android build fix

linusmartensson:
Since OS is transformed from android to linux, we need to ensure that conditions
specifying OS==linux are prevented from executing by additionally specifying
'and android_build=0'. 

This was done on most conditions in bnoordhuis patch, but I added a couple missed ones that broke the android build.

### DIFF
--- a/common.gypi
+++ b/common.gypi
@@ -1,6 +1,5 @@
 {
   'variables': {
-    'android_build%': 0,
     'visibility%': 'hidden',         # V8's visibility setting
     'target_arch%': 'ia32',          # set v8's target architecture
     'host_arch%': 'ia32',            # set v8's host architecture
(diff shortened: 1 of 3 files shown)


common.gypi
```diff
@@ -143,7 +143,7 @@
             'cflags': [ '-m32' ],
             'ldflags': [ '-m32' ],
           }],
-          [ 'OS=="linux"', {
+          [ 'OS=="linux" and android_build == 0', {
             'cflags': [ '-ansi' ],
```
bnoordhuis, line 146: What does -ansi break (and why)?

uv.gyp
```diff
@@ -51,7 +51,7 @@
           ['OS == "mac"', {
             'defines': [ '_DARWIN_USE_64_BIT_INODE=1' ],
           }],
-          ['OS == "linux"', {
+          ['OS == "linux" and android_build == 0', {
             'defines': [ '_POSIX_C_SOURCE=200112' ],
```
bnoordhuis, line 54: Ditto. Should be harmless, right?

bnoordhuis:
I'm okay with the patch but please update the commit log. Wasn't there a Makefile fix-up as well? -lrt?

linusmartensson:
This is, I hope, a bit clearer.

bnoordhuis:
Thanks, landed in ff3ee84.

PR #828, opened 2013-06-06
build: make HAVE_DTRACE=0 should disable dtrace

tjfontaine:
actually if they set HAVE_DTRACE to anything but 1 it won't try and build with dtrace


bnoordhuis:
Landed in f84becc, thanks.

PR #830, opened 2013-06-07
windows: Enable blocking pipe for stdio.

HenryRawas:
Part of fix for node issue #3584.
If blocking set on pipe, then do blocking write.


sblom:
Related to https://github.com/joyent/node/pull/5595

saghul:
Looks like this landed in https://github.com/joyent/libuv/commit/92040eb7123e892261c8da71c3eedb97c00e4b88, closing.

PR #831, opened 2013-06-09
unix, windows: improve getaddrinfo error reporting

bnoordhuis:
- Map EAI_\* constants to UV_EAI_\* constants for portability reasons.
- Have uv_getaddrinfo() hand the callback a UV_EAI_\* error code.
- Add uv_getaddrinfo_strerror() that returns a string representation
  of the UV_EAI_\* error code.

uv_getaddrinfo() can now report the full gamut of EAI_\* errors. It no
longer touches loop->last_err.

Caveats: you should not use uv_last_error() anymore in your getaddrinfo
callback. Don't pass the error code to uv_err_name() or uv_strerror(),
you'll get bogus results (or hit an assertion.)

Reviewer: @piscisaureus

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -64,12 +64,35 @@ extern "C" {
 # include "uv-private/uv-win.h"
 #endif
 
+/* libuv maps EAI_* error codes to its own set of error codes to work around
+ * the fact that:
+ *
+ *  a) not all platforms define all the EAI_* constants, and
+ *
+ *  b) the EAI_* error codes are > 0 on some platforms but < 0 on other
+ *     platforms. We always want them to be < 0 for easy error checking
+ *     in uv_getaddrinfo() callbacks.
+ */
+#define UV_EAI_ADDRFAMILY (-1)
+#define UV_EAI_AGAIN      (-2)
+#define UV_EAI_BADFLAGS   (-3)
+#define UV_EAI_CANCELED   (-4)
+#define UV_EAI_FAIL       (-5)
+#define UV_EAI_FAMILY     (-6)
+#define UV_EAI_MEMORY     (-7)
+#define UV_EAI_NODATA     (-8)
+#define UV_EAI_NONAME     (-9)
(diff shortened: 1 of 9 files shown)


bnoordhuis:
Note to self: add a ReleaseNotes file that documents the caveats and large sweeping changes in each major release.

piscisaureus:
> Note to self: add a ReleaseNotes file that documents the caveats and large sweeping changes in each major release.

Maybe it's better to rethink the conept of changelog a little bit. We could put all API changes first and label them as such.

saghul:
I'll throw my 2 cents here :-) IMHO it would be nice if no extra function was needed for the getaddrinfo errors and have a single consistent way of getting the error description for a given errno. What about starting EAI errnos in -1000? They would still be negative and non confusable with others, and there would be no need for the extra function, apparently.

In case this is not possible, would it be possible to define the EAI errnos with the same preprocessor trickery as the other ones? It helps create bindings in C, since I don't need to define them one by one: https://github.com/saghul/pyuv/blob/master/src/errno.c#L78

bnoordhuis:
> I'll throw my 2 cents here :-) IMHO it would be nice if no extra function was needed for the getaddrinfo errors and have a single consistent way of getting the error description for a given errno. What about starting EAI errnos in -1000? They would still be negative and non confusable with others, and there would be no need for the extra function, apparently.

It will probably move into that direction. This PR is a prelude to https://github.com/bnoordhuis/libuv/compare/errno which removes uv_last_error() altogether (and makes libuv return or pass errors everywhere.)

The reason there is a uv_getaddrinfo_strerror() for now, is that for EAI_SYSTEM errors, I want to encode the errno in the error code.  I held off on doing that in this PR because I'm not sure yet how it's going to interact with the errno branch.

saghul:
Looks like an interesting change! I'll keep an eye on this, thanks for explaining Ben.

bnoordhuis:
Closing, this is being redone in the errno branch in a slightly different way.

PR #834, opened 2013-06-13
linux: add support for MIPS (fix uv_cpu_info())

bsnote:


src/unix/linux-syscalls.c
```diff
@@ -37,7 +37,17 @@
 # else
 #  define UV_SYSCALL_BASE 0x900000
 # endif
-#endif /* __arm__ */
+#endif	/* __arm__ */
+
+#if defined(__mips__)
+# if _MIPS_SIM == _MIPS_SIM_ABI32
+#  define __NR_Linux 4000
```
bnoordhuis: This conflicts with the #define from unistd.h, doesn't it?

src/unix/linux-core.c
```diff
@@ -438,7 +438,7 @@ static int read_models(unsigned int numcpus, uv_cpu_info_t* ci) {
   model_idx = 0;
   speed_idx = 0;
 
-#if defined(__arm__) || defined(__i386__) || defined(__x86_64__)
+#if defined(__arm__) || defined(__i386__) || defined(__mips__) || defined(__x86_64__)
```
bnoordhuis: Long line, please wrap at 80 columns like this:

``` c
#if defined(__arm__) || \
    defined(__i386__) || \
// etc.
```

bsnote:
Thanks Ben. Fixed now.

bnoordhuis:
Landed in 3d251ad, thanks!

bnoordhuis:
Wait, make that 5096f1e. Landing it in v0.10 rather than master.

Issue #835, opened 2013-06-15
windows 64: QUEUE_DATA causes "truncates pointer" warning in offsetof computation


vtjnash:
```
In file included from src/uv-common.h:41:0,
                 from src/uv-common.c:23:
src/uv-common.c: In function 'uv_walk':
src/queue.h:29:32: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
   ((type *) ((char *) (ptr) - ((long) &((type *) 0)->field)))
                                ^
src/uv-common.c:368:9: note: in expansion of macro 'QUEUE_DATA'
     h = QUEUE_DATA(q, uv_handle_t, handle_queue);
         ^
src/uv-common.c: In function 'uv__print_handles':
src/queue.h:29:32: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
   ((type *) ((char *) (ptr) - ((long) &((type *) 0)->field)))
                                ^
src/uv-common.c:385:9: note: in expansion of macro 'QUEUE_DATA'
     h = QUEUE_DATA(q, uv_handle_t, handle_queue);
```

PR #836, opened 2013-06-15
queue: fix pointer truncation on LLP64 platforms

piscisaureus:
Fix for #835. Making travis run the tests.


bnoordhuis:
src/queue.h in node.js probably needs upgrading too.

PR #840, opened 2013-06-26
Adding install target to makefile

navaneeth:
Before this commit, there was no way to install libuv other than
manually copying required files to the path. This commit introduces an
install target which installs the libuv.a, libuv.so and header files to
the install prefix.

```
make
make install
```

The above will use the default prefix `/usr/local`

```
make
make install PREFIX=/usr
```

The above will install to `/usr`

Shared libraries will be installed only if it is available.

### DIFF
--- a/build.mk
+++ b/build.mk
@@ -130,6 +130,7 @@ TESTS= \
 	test/test-threadpool.o \
 	test/test-threadpool-cancel.o \
 	test/test-timer-again.o \
+	test/test-timer-from-check.o \
 	test/test-timer.o \
 	test/test-tty.o \
 	test/test-udp-dgram-too-big.o \
@@ -142,7 +143,7 @@ TESTS= \
 	test/test-util.o \
 	test/test-walk-handles.o \
 
-all: libuv.a
+.PHONY: all bench clean clean-platform distclean test
 
 run-tests$(E): test/run-tests.o test/runner.o $(RUNNER_SRC) $(TESTS) libuv.a
 	$(CC) $(CPPFLAGS) $(RUNNER_CFLAGS) -o $@ $^ $(RUNNER_LIBS) $(RUNNER_LDFLAGS)
@@ -151,10 +152,6 @@ run-benchmarks$(E): test/run-benchmarks.o test/runner.o $(RUNNER_SRC) $(BENCHMAR
 	$(CC) $(CPPFLAGS) $(RUNNER_CFLAGS) -o $@ $^ $(RUNNER_LIBS) $(RUNNER_LDFLAGS)
 
 test/echo.o: test/echo.c test/echo.h
-
-
-.PHONY: clean clean-platform distclean test bench
-
 
 test: run-tests$(E)
 	$(CURDIR)/$<
--- a/common.gypi
+++ b/common.gypi
@@ -36,6 +36,7 @@
         },
         'xcode_settings': {
           'GCC_OPTIMIZATION_LEVEL': '0',
+          'OTHER_CFLAGS': [ '-Wno-strict-aliasing' ],
         },
         'conditions': [
           ['OS != "win"', {
--- a/config-unix.mk
+++ b/config-unix.mk
@@ -28,7 +28,7 @@ CPPFLAGS += -D_FILE_OFFSET_BITS=64
 
 RUNNER_SRC=test/runner-unix.c
 RUNNER_CFLAGS=$(CFLAGS) -I$(SRCDIR)/test
-RUNNER_LDFLAGS=-L"$(CURDIR)" -luv
+RUNNER_LDFLAGS=
 
 DTRACE_OBJS=
 DTRACE_HEADER=
@@ -164,6 +164,13 @@ SO_LDFLAGS = -Wl,-soname,libuv.so.0.11
 endif
 
 RUNNER_LDFLAGS += $(LDFLAGS)
+
+all:
+	# Force a sequential build of the static and the shared library.
+	# Works around a make quirk where it forgets to (re)build either
+	# the *.o or *.pic.o files, depending on what target comes first.
+	$(MAKE) -f $(SRCDIR)/Makefile libuv.a
+	$(MAKE) -f $(SRCDIR)/Makefile libuv.$(SOEXT)
 
 libuv.a: $(OBJS)
 	$(AR) rcs $@ $^
--- a/src/win/process.c
+++ b/src/win/process.c
@@ -225,7 +225,7 @@ static WCHAR* search_path_join_test(const WCHAR* dir,
   attrs = GetFileAttributesW(result);
 
   if (attrs != INVALID_FILE_ATTRIBUTES &&
-     !(attrs & (FILE_ATTRIBUTE_DIRECTORY | FILE_ATTRIBUTE_REPARSE_POINT))) {
+      !(attrs & FILE_ATTRIBUTE_DIRECTORY)) {
     return result;
   }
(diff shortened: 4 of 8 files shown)


build.mk
```diff
@@ -167,3 +171,9 @@ clean distclean: clean-platform
 		test/run-tests.o test/run-benchmarks.o \
 		test/runner.o run-tests$(E) test/run-benchmarks$(E) \
 		$(BENCHMARKS) $(TESTS) $(RUNNER_LIBS)
+
+install: libuv.a
```
bnoordhuis: Suggestion: depend on libuv.$(SOEXT) as well?

build.mk
```diff
@@ -167,3 +171,9 @@ clean distclean: clean-platform
...
+	install libuv.a $(PREFIX)/lib
```
bnoordhuis: Doesn't this install libuv.a with mode=0755?

build.mk
```diff
@@ -167,3 +171,9 @@ clean distclean: clean-platform
...
+	if [ -a libuv.$(SOEXT) ]; then install libuv.$(SOEXT) $(PREFIX)/lib; fi;
+	install $(SRCDIR)/include/uv.h $(PREFIX)/include
+	cp -r $(SRCDIR)/include/uv-private $(PREFIX)/include/
```
bnoordhuis: Use `install -d` rather than `cp -r`?

build.mk
navaneeth: @bnoordhuis I have tried that initially, but later I found that just doing `make` will build only `libuv.a` not `libuv.so`. So since user explicitly build shared library, I think install shouldn't depend on it either. 

Also, when adding `libuv.a` and `libuv.$(SOEXT)` as dependencies, `libuv.$(SOEXT)` will fail. It looks like, `libuv.a`'s build is changing directories and when doing `libuv.$(SOEXT)` immediately, it fails.

build.mk
navaneeth: Yes, by default `install` will use 0755. Should I explicitly set it? Do you think it should set to 0700?

build.mk
navaneeth: `install -d` has different options in different platforms. In OSX I couldn't get it working. This [stackoverflow](http://stackoverflow.com/questions/417236/whats-the-best-way-to-move-a-directory-into-place-in-a-makefile-install) link says `cp -r` can be used instead.

jbergstroem: I think separating static and shared libraries installation would also be beneficial here.

build.mk
bnoordhuis: > So since user explicitly build shared library, I think install shouldn't depend on it either.

Then the install target shouldn't install the shared library. Having it install libuv.$(SOEXT) without depending on it means it can install a stale version of the shared library.

> Also, when adding libuv.a and libuv.$(SOEXT) as dependencies, libuv.$(SOEXT) will fail. It looks like, libuv.a's build is changing directories and when doing libuv.$(SOEXT) immediately, it fails.

That's not exactly what happens but yes, that's a bug. I'll see if I can fix it.

build.mk
bnoordhuis: 0644 for the static library and the headers, 0755 for the shared library and directories.

build.mk
bnoordhuis: -d is pretty universal, maybe you're thinking of -D? Usage is like this:

```
# create directories first
install -m 755 -d /foo/bar
install -m 644 baz /foo/bar
```

build.mk
navaneeth: > Then the install target shouldn't install the shared library. Having it install libuv.$(SOEXT) without depending on it means it can install a stale version of the shared library.

Fair point. 

So I will separate this into two targets, one installs the shared library and the other installs the static library. 

```
make install # default, and installs the libuv.a, depends on libuv.a
make installshared #installs shared library, depends on libuv.$(SOEXT)
```

If this sounds good, I will change the implementation and add the commits here.

build.mk
navaneeth: Alright. Got it. So it should be done for all the files in `include/uv-private` right?

```
install -m 644 include/uv-private/pthread-fixes.h $(PREFIX)/include/uv-private  
install -m 644 include/uv-private/stdint-msvc2008.h $(PREFIX)/include/uv-private  
....
....
```

build.mk
bnoordhuis: Well, I kind of like the idea of installing both the static and the shared library.  I'll land a patch in a few that should fix building both at the same time and will add a dependency to the 'all' target on libuv.$(SOEXT).

build.mk
bnoordhuis: Yep, that's correct.

build.mk
navaneeth: Alright. Thanks. I will wait till you push that. Will resubmit the PR after the changes.

bnoordhuis: Landed in 88a2c7f. Took a little longer due to a make quirk.

navaneeth: @bnoordhuis  Looks like it still has issues when adding `libuv.a` and `libuv.$(SOEXT)` as dependencies to `install`.

```
install: libuv.a libuv.$(SOEXT)
   ...
```

```
make clean
make install
```

Above fails with error, 

```
clang: error: no such file or directory: 'src/unix/async.pic.o'
clang: error: no such file or directory: 'src/unix/core.pic.o'
clang: error: no such file or directory: 'src/unix/dl.pic.o'
clang: error: no such file or directory: 'src/unix/error.pic.o'
clang: error: no such file or directory: 'src/unix/fs.pic.o'
clang: error: no such file or directory: 'src/unix/getaddrinfo.pic.o'
clang: error: no such file or directory: 'src/unix/loop.pic.o'
clang: error: no such file or directory: 'src/unix/loop-watcher.pic.o'
clang: error: no such file or directory: 'src/unix/pipe.pic.o'
clang: error: no such file or directory: 'src/unix/poll.pic.o'
clang: error: no such file or directory: 'src/unix/process.pic.o'
clang: error: no such file or directory: 'src/unix/signal.pic.o'
clang: error: no such file or directory: 'src/unix/stream.pic.o'
clang: error: no such file or directory: 'src/unix/tcp.pic.o'
clang: error: no such file or directory: 'src/unix/thread.pic.o'
clang: error: no such file or directory: 'src/unix/threadpool.pic.o'
clang: error: no such file or directory: 'src/unix/timer.pic.o'
clang: error: no such file or directory: 'src/unix/tty.pic.o'
clang: error: no such file or directory: 'src/unix/udp.pic.o'
clang: error: no such file or directory: 'src/fs-poll.pic.o'
clang: error: no such file or directory: 'src/uv-common.pic.o'
clang: error: no such file or directory: 'src/inet.pic.o'
clang: error: no such file or directory: 'src/version.pic.o'
clang: error: no such file or directory: 'src/unix/darwin.pic.o'
clang: error: no such file or directory: 'src/unix/kqueue.pic.o'
clang: error: no such file or directory: 'src/unix/fsevents.pic.o'
clang: error: no such file or directory: 'src/unix/proctitle.pic.o'
clang: error: no such file or directory: 'src/unix/darwin-proctitle.pic.o'
make: *** [libuv.dylib] Error 1
```

bnoordhuis: What happens when you depend on 'all'? It forces in-order builds of libuv.a and libuv.so, something I couldn't get to work with order-only prerequisites (`a: | b c` style rules.)

navaneeth:
I have created a new PR #841 Closing this.

PR #841, opened 2013-06-27
build: Added install target to the makefile.

navaneeth:
Before this commit, there was no way to install libuv other than
manually copying required files to the path. This commit introduces
an install target which installs the libuv.a, libuv.so and header files
to the install prefix.

```
``
make
make install
``
```

The above will use the default prefix `/usr/local`

```
``
make
make install PREFIX=/usr
``
```

The above will install to `/usr`

### DIFF
--- a/build.mk
+++ b/build.mk
@@ -32,6 +32,10 @@ else
 SOEXT = so
 endif
 
+ifndef PREFIX
+PREFIX=/usr/local
+endif
+
 ifneq (,$(findstring mingw,$(PLATFORM)))
 include $(SRCDIR)/config-mingw.mk
 else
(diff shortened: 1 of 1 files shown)


bnoordhuis:
LGTM but can you sign [the CLA](http://nodejs.org/cla.html)? Thanks.

navaneeth:
@bnoordhuis Sure, I did that. Should I send the PDF copy as well?

bnoordhuis:
No need, just signing online is enough. But can you sign it with your full name? Unless Navaneeth is your full name, of course.

navaneeth:
@bnoordhuis Done.

bnoordhuis:
Thanks Navaneeth, landed in 624be53.

PR #842, opened 2013-06-27
unix, windows: add extra fields to uv_stat_t

saghul:
They are BSD-isms. st_flags and st_gen are zeroed on non-BSD systems and st_birthtim is set to st_ctim.


bnoordhuis:
Landed in f9e6029, thanks.

PR #843, opened 2013-06-30
build system improvements

hasufell:
see commits


build.mk
```diff
@@ -169,17 +174,25 @@ clean distclean: clean-platform
...
-	install -m 644 $(SRCDIR)/include/uv-private/pthread-fixes.h $(PREFIX)/include/uv-private
-	install -m 644 $(SRCDIR)/include/uv-private/stdint-msvc2008.h $(PREFIX)/include/uv-private
-	install -m 644 $(SRCDIR)/include/uv-private/tree.h $(PREFIX)/include/uv-private
-	install -m 644 $(SRCDIR)/include/uv-private/uv-bsd.h $(PREFIX)/include/uv-private
-	install -m 644 $(SRCDIR)/include/uv-private/uv-darwin.h $(PREFIX)/include/uv-private
-	install -m 644 $(SRCDIR)/include/uv-private/uv-linux.h $(PREFIX)/include/uv-private
-	install -m 644 $(SRCDIR)/include/uv-private/uv-sunos.h $(PREFIX)/include/uv-private
-	install -m 644 $(SRCDIR)/include/uv-private/uv-unix.h $(PREFIX)/include/uv-private
-	install -m 644 $(SRCDIR)/include/uv-private/uv-win.h $(PREFIX)/include/uv-private
+install:
```
bnoordhuis, line 177: Depend on the 'all' target. A no-deps install target makes it possible to install outdated shared/static libraries.

build.mk
```diff
@@ -169,17 +174,25 @@ clean distclean: clean-platform
...
+	$(INSTALL_DIR) $(DESTDIR)$(PREFIX)/$(LIBDIR)
+	if [ -e libuv.a ] ; then $(INSTALL_DATA) libuv.a $(DESTDIR)$(PREFIX)/$(LIBDIR) ; fi
```
bnoordhuis, line 179: Can you split this over multiple lines?

build.mk
```diff
@@ -169,17 +174,25 @@ clean distclean: clean-platform
...
+ifdef SONAME
+	if [ -e libuv.$(SOEXT) ] ; then \
+		$(INSTALL_PROGRAM) libuv.$(SOEXT) $(DESTDIR)$(PREFIX)/$(LIBDIR)/libuv.$(SOEXT)$(SONAME) ; \
+		ln -s libuv.$(SOEXT)$(SONAME) $(DESTDIR)$(PREFIX)/$(LIBDIR)/libuv.$(SOEXT) ; \
+	fi
+else
+	if [ -e libuv.$(SOEXT) ] ; then $(INSTALL_PROGRAM) libuv.$(SOEXT) $(DESTDIR)$(PREFIX)/$(LIBDIR) ; fi
```
bnoordhuis, line 186: Ditto.

config-unix.mk
```diff
@@ -160,10 +162,12 @@ endif
 
 ifneq (darwin,$(PLATFORM))
 # Must correspond with UV_VERSION_MAJOR and UV_VERSION_MINOR in src/version.c
-SO_LDFLAGS = -Wl,-soname,libuv.so.0.11
+SONAME=.0.11
+SO_LDFLAGS = -Wl,-soname,libuv.so$(SONAME)
```
bnoordhuis, line 166: What value does this add?

bnoordhuis:
Generally LGTM but please heed the suggestions from the Jenkins bot. Also, the convention is to prefix the first line of the commit log with 'build:' rather than 'MAKE:'

config-unix.mk
hasufell, line 166: You already assign a soname without creating the appropriate symlink, which is wrong. These lines allow me to check if SONAME is set in the install rules and create that link.

build.mk
hasufell, line 177: Depending on the all target will force building of the shared library even if I only intend to install the static one.

In order to not install outdated stuff there is the "clean" target. Letting the install target depend on "all" is not nice. Another way would be to create separate install rules such as "install", "install-all", "install-shared" etc

bnoordhuis, line 177: > Depending on the all target will force building of the shared library even if I only intend to install the static one.

That bothers me less than the possibility of installing an outdated copy of the library does.

> Another way would be to create separate install rules such as "install", "install-all", "install-shared" etc

That's acceptable as long as the 'install' target is equivalent to 'install-all'.

config-unix.mk
bnoordhuis, line 166: The Debian people were fine with the non-symlink approach. But whatever, it's fine by me.

bnoordhuis:
Thanks for submitting this pull request but I (grudgingly) landed ddd7e04 which moves us to autotools.  That should give you things like LIBDIR and DESTDIR for free.

hasufell:
Nice, I will test that some time.

PR #855, opened 2013-07-16
added more libs (patch from rphillips: https://gist.github.com/rphillips...

rphillips:


piscisaureus:
what's this all about?

rphillips:
we are in the process of updating gyp, and the gyp developers removed a bunch of the default libraries... I accidentally opened this PR against the upstream libuv repo.

PR #858, opened 2013-07-21
unix, windows, darwin : Implemented uv channels

navaneeth:
# Introduction

This commit attempts to implement channels (something very similar to what golang has) to `libuv`. 

A channel provides a mechanism for two concurrently executing functions to communicate by passing any value. This value will be passed to all the receivers that the channel has.  This simplifies message passing between worker threads.

Channel internally holds a mutex and synchronizes the messages it gets. Because of this, user don't have to acquire a lock when sending or receiving from channels.
# Example

``` c
#include "include/uv.h" 
#include <stdio.h>
#include <string.h>

static void worker(void* arg) {
    int i;
    char *string;
    uv_chan_t* chan = arg;

    for (i = 0; i < 10; i++) {
        string = malloc(sizeof(char) * 2);
        sprintf(string, "%d", i);
        uv_chan_send (chan, string);
    }

    uv_chan_send (chan, NULL);
}

int main() {
    uv_thread_t thread1, thread2;
    uv_chan_t chan;
    char *message;
    int count = 0;

    uv_chan_init (&chan);
    uv_thread_create(&thread1, worker, &chan);
    uv_thread_create(&thread2, worker, &chan);

    while(count < 2) {
        message = uv_chan_receive (&chan);
        if (message == NULL)
            count++;
        printf ("%s\n", message);
        free (message);
    }

    uv_thread_join(&thread1);
    uv_thread_join(&thread2);
    uv_chan_destroy (&chan);
    return 0;
}
```
# Future work

This is just a first cut version of the channels. We could also implement the following,
- Specifying maximum size of elements that a channel can hold. If more elements that this is queued, `uv_chan_send` will block until someone reads from it. 
- To avoid multiple threads updating same instance of data, data needs to be deep copied before sending to the channel. Currently everyone who uses `uv_chan_send` has to remember to deep copy the data. This can be simplified by adding a `copy_cb` to the `uv_chan_t` which gets called whenever `uv_chan_send` is called. This callback can deep copy data and return pointer to the new instance. This way, user has to configure it only while creating the channel and while sending, callback will take care of deep copying. 

This idea is largely inspired from the golang's channels. I believe this simplifies communication between threads.

Let me know the feedback.

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -221,7 +221,7 @@ typedef struct uv_work_s uv_work_t;
 /* None of the above. */
 typedef struct uv_cpu_info_s uv_cpu_info_t;
 typedef struct uv_interface_address_s uv_interface_address_t;
-
+typedef struct uv_chan_s uv_chan_t;
 
 typedef enum {
   UV_RUN_DEFAULT = 0,
@@ -1999,6 +1999,27 @@ union uv_any_req {
 };
 #undef XX
 
+struct uv_chan_s {
+  uv_mutex_t mutex;
+  uv_cond_t cond;
+  void* q[2];
+};
+
+/*
+ * A channel provides a mechanism for concurrently executing threads to communicate by passing values.
+ * Given a uv_chan_t instance, any thread can send or receive data. Data will be send to all the receivers
+ * in the channel. The first one who reads data will get it and all other receivers will still wait. User don't
+ * have to explicitly lock when sending or receiving data.
+ *
+ * Receivers always block until there is data to receive.
+ *
+ * A channel can be closed with uv_chan_destroy
+ *
+ */
+UV_EXTERN int uv_chan_init(uv_chan_t* chan);
+UV_EXTERN void uv_chan_send(uv_chan_t* chan, void* data);
+UV_EXTERN void* uv_chan_receive(uv_chan_t* chan);
+UV_EXTERN void uv_chan_destroy(uv_chan_t* chan);
 
 struct uv_loop_s {
   /* User data - use this for whatever. */
--- a/src/uv-common.c
+++ b/src/uv-common.c
@@ -472,3 +472,54 @@ int uv__getaddrinfo_translate_error(int sys_err) {
   assert(!"unknown EAI_* error code");
   abort();
 }
+
+typedef struct {
+  void* data;
+  uv_req_type type;
+  void* active_queue[2];
+} uv__chan_item_t;
+
+int uv_chan_init(uv_chan_t* chan) {
+  int r = uv_mutex_init (&chan->mutex);
+  if (r == -1)
+    return r;
+
+  QUEUE_INIT(&chan->q);
+
+  return uv_cond_init (&chan->cond);
+}
+
+void uv_chan_send(uv_chan_t* chan, void* data) {
+  uv__chan_item_t* item = malloc(sizeof(uv__chan_item_t));
+  item->data = data;
+
+  uv_mutex_lock (&chan->mutex);
+  QUEUE_INSERT_TAIL(&chan->q, &item->active_queue);
+  uv_cond_signal (&chan->cond);
+  uv_mutex_unlock (&chan->mutex);
+}
+
+void* uv_chan_receive(uv_chan_t* chan) {
+  uv__chan_item_t* item;
+  QUEUE* head;
+  void* data = NULL;
+
+  uv_mutex_lock (&chan->mutex);
+  while (QUEUE_EMPTY(&chan->q)) {
+    uv_cond_wait (&chan->cond, &chan->mutex);
+  }
+
+  head = QUEUE_HEAD (&chan->q);
+  item = QUEUE_DATA (head, uv__chan_item_t, active_queue);
+  data = item->data;
+  QUEUE_REMOVE (head);
+  free (item);
+  uv_mutex_unlock (&chan->mutex);
+  return data;
+}
+
+void uv_chan_destroy(uv_chan_t* chan) {
+  uv_cond_destroy (&chan->cond);
+  uv_mutex_destroy (&chan->mutex);
+}
+
--- a/Makefile.am
+++ b/Makefile.am
@@ -141,6 +141,7 @@ test_run_tests_SOURCES = test/blackhole-server.c \
                          test/test-udp-options.c \
                          test/test-udp-send-and-recv.c \
                          test/test-util.c \
+                         test/test-channel.c \
                          test/test-walk-handles.c
 test_run_tests_LDADD = libuv.la
 
--- a/checksparse.sh
+++ b/checksparse.sh
@@ -160,6 +160,7 @@ test/test-udp-options.c
 test/test-udp-send-and-recv.c
 test/test-util.c
 test/test-walk-handles.c
+test/test-channel.c
 "
 
 case `uname -s` in
--- a/test/test-channel.c
+++ b/test/test-channel.c
@@ -1,0 +1,94 @@
+/* Copyright Joyent, Inc. and other Node contributors. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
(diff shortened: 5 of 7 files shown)


navaneeth:
I wasn't very sure about which file to be used to implementing uv_chan methods. I have used `uv_common.c`. If this is incorrect, let me know. I will fix it and update the PR.

src/uv-common.c
```diff
@@ -472,3 +472,54 @@ int uv__getaddrinfo_translate_error(int sys_err) {
...
+} uv__chan_item_t;
+
+int uv_chan_init(uv_chan_t* chan) {
+  int r = uv_mutex_init (&chan->mutex);
+  if (r == -1)
+    return r;
+
+  QUEUE_INIT(&chan->q);
+
+  return uv_cond_init (&chan->cond);
```
indutny, line 489: `chan->mutex` should be deinitalized on failure here.

indutny, line 489: Also, code style: we ain't inserting spaces after function name, i.e. `uv_cond_init(&chan->cond)`.

src/uv-common.c
```diff
@@ -472,3 +472,54 @@ int uv__getaddrinfo_translate_error(int sys_err) {
...
+}
+
+void uv_chan_send(uv_chan_t* chan, void* data) {
+  uv__chan_item_t* item = malloc(sizeof(uv__chan_item_t));
+  item->data = data;
+
+  uv_mutex_lock (&chan->mutex);
+  QUEUE_INSERT_TAIL(&chan->q, &item->active_queue);
+  uv_cond_signal (&chan->cond);
```
indutny, line 498: ditto.

src/uv-common.c
```diff
@@ -472,3 +472,54 @@ int uv__getaddrinfo_translate_error(int sys_err) {
...
+  uv_mutex_unlock (&chan->mutex);
```
indutny, line 499: ditto here too, and everywhere else.

src/uv-common.c
```diff
@@ -472,3 +472,54 @@ int uv__getaddrinfo_translate_error(int sys_err) {
...
+}
+
+void* uv_chan_receive(uv_chan_t* chan) {
```
indutny, line 502: I think `timeout` argument should be pretty useful in many cases. Making it wait indefinitely isn't always a good idea.

src/uv-common.c
```diff
@@ -472,3 +472,54 @@ int uv__getaddrinfo_translate_error(int sys_err) {
...
+  uv__chan_item_t* item;
+  QUEUE* head;
+  void* data = NULL;
```
indutny, line 505: Initialize on separate line.

src/uv-common.c
```diff
@@ -472,3 +472,54 @@ int uv__getaddrinfo_translate_error(int sys_err) {
...
+  if (r == -1)
+    return r;
+
+  QUEUE_INIT(&chan->q);
+
+  return uv_cond_init (&chan->cond);
+}
+
+void uv_chan_send(uv_chan_t* chan, void* data) {
+  uv__chan_item_t* item = malloc(sizeof(uv__chan_item_t));
```
indutny, line 493: Initialize on separate line.

src/uv-common.c
```diff
@@ -472,3 +472,54 @@ int uv__getaddrinfo_translate_error(int sys_err) {
...
 }
+
+typedef struct {
+  void* data;
+  uv_req_type type;
+  void* active_queue[2];
+} uv__chan_item_t;
+
+int uv_chan_init(uv_chan_t* chan) {
+  int r = uv_mutex_init (&chan->mutex);
```
indutny, line 483: Initialize on separate line.

include/uv.h
```diff
@@ -221,7 +221,7 @@
 /* None of the above. */
 typedef struct uv_cpu_info_s uv_cpu_info_t;
 typedef struct uv_interface_address_s uv_interface_address_t;
-
+typedef struct uv_chan_s uv_chan_t;
```
indutny, line 224: Why removing empty line?

src/uv-common.c
```diff
@@ -472,3 +472,54 @@ int uv__getaddrinfo_translate_error(int sys_err) {
...
+  uv__chan_item_t* item;
+  QUEUE* head;
+  void* data = NULL;
+
+  uv_mutex_lock (&chan->mutex);
+  while (QUEUE_EMPTY(&chan->q)) {
+    uv_cond_wait (&chan->cond, &chan->mutex);
+  }
+
+  head = QUEUE_HEAD (&chan->q);
```
indutny, line 512: Excessive space between name and paren.

src/uv-common.c
```diff
@@ -472,3 +472,54 @@ int uv__getaddrinfo_translate_error(int sys_err) {
...
+  item = QUEUE_DATA (head, uv__chan_item_t, active_queue);
+  data = item->data;
+  QUEUE_REMOVE (head);
+  free (item);
+  uv_mutex_unlock (&chan->mutex);
```
indutny, line 517: I think you should remove `item` from queue first and unlock mutex as early as possible, otherwise other threads will stall while current one is calling `free` and doing various stuff.

indutny:
Except few style and minor nits, LGTM. @bnoordhuis what do you think about the feature itself?

src/uv-common.c
navaneeth, line 489: Good catch. I will update the commit with all the required changes..

src/uv-common.c
navaneeth, line 493: How is `malloc` failures handled in `libuv`? I was not sure about how it was handled, that why I omitted the return value.

indutny, line 493: oh, good catch, I missed it. I think it should return `-ENOMEM` in such case.

bnoordhuis:
Thanks, but I'm not sure if this is something we really want. Libuv is more about providing the building blocks that let you construct higher-level constructs rather than providing those higher-level constructs directly.

@piscisaureus WDYT?

navaneeth:
@bnoordhuis I understand. But I think this improves synchronization and simplifies thread communication. Otherwise everyone who uses `libuv` and needs this functionality has to re-implement the whole thing.

bnoordhuis:
> Otherwise everyone who uses libuv and needs this functionality has to re-implement the whole thing.

I don't disagree, but that argument applies to just about any feature.

piscisaureus:
+1 for a feature similar to `uv_async_send` but one that let's you pass an additional value as context, and that doesn't coalesce. If channels is that, I'm game.

Although implementing this in uv-common.c seems like the obvious thing to do, I have reasons to prefer otherwise. I/O completion ports pretty much _are_ channels; take a look at the docs for `PostQueuedCompletionStatus`. Implementing something on our own, including locking and all that, seems unreasonably inefficient.

navaneeth:
@piscisaureus Thanks for commenting. I checked the documentation for `PostQueuedCompletionStatus`and it is only for Windows. I couldn't find alternative for other platforms.

bnoordhuis:
If we go through with this (I'm personally -1 on it) then at the very least the send function needs to change from synchronous and blocking to asynchronous and non-blocking.

I also don't like how it mallocs memory.  IMO it should be the responsibility of the caller to provide storage, e.g. in the form of a uv_chan_item_t struct.

navaneeth:
@bnoordhuis Sorry for the delay to get back on this. I will get back after I make the `send` non-blocking. I will also address the feedback about allocating memory.

saghul:
@navaneeth any progress on this?

navaneeth:
@saghul I am working on this. Didn't get enough time to look into this. Will get to this in a week.

indutny:
@navaneeth ping ;)

navaneeth:
@indutny I have implemented the changes. Need to do some cleanup and testing. Will push as early as possible.

saghul:
I'm going to close this. I posted an RFC here: https://github.com/libuv/libuv/issues/5 which will result in the addition of a `uv_callback` request type which gives the same benefit as the idea of channels, but it's more integrated with the libuv model of requests.

Sorry we neglected to deal with this sooner :-(

PR #862, opened 2013-07-24
src: remove unused variables

mscdex:


bnoordhuis:
Thanks, landed in c021ee0.

PR #863, opened 2013-07-25
unix: Expose tcp start and stop externally

davglass:
Exposing these to allow hooking into and stopping tcp requests on a connection while allowing current connections to continue.

I didn't patch the Win32 files because I'm not a windows guy. I'll be happy to add it for feature parity if someone can give me a hand.

_This request will also have a counterpart in Node land to expose this API to JS_

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -765,6 +765,14 @@ UV_EXTERN int uv_tcp_connect(uv_connect_t* req, uv_tcp_t* handle,
 UV_EXTERN int uv_tcp_connect6(uv_connect_t* req, uv_tcp_t* handle,
     struct sockaddr_in6 address, uv_connect_cb cb);
 
+/*
+ * Used for turning off the acceptance of new TCP connections for
+ * handling a graceful shutdown.
+ */
+
+UV_EXTERN void uv_tcp_start(uv_tcp_t* tcp);
+UV_EXTERN void uv_tcp_stop(uv_tcp_t* tcp);
+
 /* uv_connect_t is a subclass of uv_req_t */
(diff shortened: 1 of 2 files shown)


bnoordhuis:
I confess I don't get it. What's wrong with uv_read_start() and uv_read_stop()?

davglass:
If I'm reading them right, which I may not be, using `uv_read_start/stop` would stop all reading of the streams, even current ones. Where what we are looking for is the ability to not accept any new connections, but still letting current connections complete.

bnoordhuis:
uv_read_start() and uv_read_stop() work on a per-stream basis.

If you have a server handle and want to stop accepting new connections without closing the handle, well, you can't really do that right now.

Your patch partially accomplishes that but in a somewhat ad-hoc and TCP-specific fashion.

What you want is a uv_listen_stop() function. It's relatively trivial to implement on Unices, probably less so on Windows because you have to worry about in-flight requests. @piscisaureus can fill you in on the details.

davglass:
Cool, I'm willing to do what I can to get this feature in. We have been using this feature internally for a while now and it does what we need (provide a graceful shutdown on a process).

davglass:
Hey @piscisaureus, we talked briefly about this at NodeConfEU and you mentioned that it could be solved by re-implementing the `pause` method.

Just pinging so you remember :)

Thanks

davglass:
Ping @piscisaureus again :)

saghul:
Closing this since the patch cannot land in its current form and there was no followup. Feel free to work on a patch and open a new pull request :-)

PR #864, opened 2013-07-25
Fix uv_hrtime() for win32

witchu:
sometimes the calculation less than expected.


src/win/util.c
```diff
@@ -461,8 +461,8 @@ uint64_t uv_hrtime(void) {
   /* performance counter frequency, and there may not be much headroom to */
   /* multiply by NANOSEC without overflowing, we use 128-bit math instead. */
   return ((uint64_t) counter.LowPart * NANOSEC / hrtime_frequency_) +
-         (((uint64_t) counter.HighPart * NANOSEC / hrtime_frequency_)
-         << 32);
+         (((uint64_t) counter.HighPart * NANOSEC / hrtime_frequency_) << 32) +
+         (((uint64_t) counter.HighPart * NANOSEC % hrtime_frequency_) * 4294967296 / hrtime_frequency_);
```
bnoordhuis, line 465: Long line, wrap at 80 columns. Secondly, for the sake of consistency, is there a reason to multiply rather than shift? Thirdly, for the sake of clarity, it'd be better to cache the common part of the calculation in a a variable.

piscisaureus:
Of course if there is a bug in uv_hrtime() it's important and we shouldn't leave this pending for so long. However I am very curious what values cause the computation to overflow.

@witchu Are you seeing problems with this on a real-world system? If so, could you share the performance counter and frequency values you're seeing? Also, out of curiosity, what kind of hardware is it?

witchu:
I found the problem when it run on my windows server. The performance
frequency go very high.

saghul:
@piscisaureus is this still relevant?

saghul:
No follow-up, closing.

PR #867, opened 2013-07-29
Make autotools build system work with MinGW

Keno:
Extracted from https://github.com/JuliaLang/libuv/tree/julia-uv0.10.13


bnoordhuis:
The change looks good to me but it seems to break the Solaris / SmartOS build (and the MSVC Windows build but I'm not sure if that's related.)

@tjfontaine Can you take a look?

bnoordhuis:
Thanks Keno, landed in e52aa6c.

PR #873, opened 2013-08-09
Read console fixes

jepickett:
Two of the unit tests for Node.JS close the stdin handle. ReadConsoleA in tty.c at line 335 will not return with failure when the handle is deleted, thus causing the Node unit test to hang while waiting for IO to complete.

PR #877, opened 2013-08-12
Read console fixes

jepickett:
Reworked pull request #873. Sending a return key to force ReadConsoleA to exit before closing the stdin handle. Fix only applies only for Win8+.


indutny:
Is it still relevant?

PR #880, opened 2013-08-14
fsevents: FSEvents is most likely not thread-safe

indutny:
Perform all operation with FSEventStream in the same thread, where it'll
be used.

/cc @bnoordhuis @piscisaureus

see https://github.com/joyent/node/issues/6040


indutny:
Wrong issue number?

bnoordhuis:
LGTM

indutny:
Thanks, landed in 303ae3b.

PR #885, opened 2013-08-17
Fixed error when building for iOS

elslooo:


src/unix/darwin-proctitle.c
```diff
@@ -22,8 +22,11 @@
 #include <dlfcn.h>
 
 #if !TARGET_OS_IPHONE
-# include <CoreFoundation/CoreFoundation.h>
-# include <ApplicationServices/ApplicationServices.h>
+#include <CoreFoundation/CoreFoundation.h>
+#include <ApplicationServices/ApplicationServices.h>
+#else
+#include <errno.h>
+#include <stdlib.h>
```
bnoordhuis, line 29: These two can be included unconditionally.  They're being pulled in implicitly, might as well make that explicit.

bnoordhuis:
See comment and please fix up the commit log, good commit logs are important.  If there is a build error, please include the compiler output in the log.  See `git log` for inspiration.  Thanks.

src/unix/darwin-proctitle.c
elslooo, line 29: The problem is that CoreFoundation already includes `errno` and `stdlib` itself, so if you decide not to include `CoreFoundation` for iOS, various error numbers and `NULL` are simply not defined.

bnoordhuis, line 29: Yeah, that's what I mean.  Just include them right before dlfcn.h and all should be well, right?

elslooo, line 29: Ah, I misunderstood. Will do!

bnoordhuis:
Tim, I landed a similar patch in b03192e.  Sorry, but someone else was reporting the same build error and I didn't want to wait any longer with fixing it.  Keep the pull requests coming though. :-)

PR #891, opened 2013-08-19
fix iOS compile error & fix warning

zhzhxtrrk:


bnoordhuis:
Thanks. I didn't take your patch (the commit log needs touching up and I didn't want to wait for a signed CLA) but I applied similar patches in b03192e and 12bad62. Thanks again.

PR #893, opened 2013-08-20
windows: Fixed compilation with Visual Studio 2008

saghul:
Fixes https://github.com/saghul/pyuv/issues/123

### DIFF
--- a/src/queue.h
+++ b/src/queue.h
@@ -16,7 +16,11 @@
 #ifndef QUEUE_H_
 #define QUEUE_H_
 
-#include <stdint.h>
+#if defined(_MSC_VER) && _MSC_VER < 1600
+# include "stdint-msvc2008.h"
+#else
+# include <stdint.h>
+#endif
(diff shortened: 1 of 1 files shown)


saghul:
I see we now use this if-include thing in 7 places. How about creating some uv-stdint.h file like:

```
#ifndef UV_STDINT_H_
#define UV_STDINT_H_

#if defined(_MSC_VER) && _MSC_VER < 1600
  # include "stdint-msvc2008.h"
#else
  # include <stdint.h>
#endif

#endif
```

/cc @bnoordhuis

bnoordhuis:
Landed a fix that removes some code instead of adding more in 8531046.  Thanks for pointing out the issue though. :-)

> I see we now use this if-include thing in 7 places. How about creating some uv-stdint.h file

That's not a bad idea but I would restrict its usage to src/win.  I don't really want to add another public header if I can avoid it.

PR #894, opened 2013-08-21
fsevents: use shared FSEventStream

indutny:
It seems that number of simultaneously opened FSEventStreams is limited
on OSX (i.e. you can have only fixed number of them on one running
system), getting past through this limit will cause `FSEventStreamCreate` to
return false and write following message to stderr:

```
(CarbonCore.framework) FSEventStreamStart: register_with_server: ERROR:
f2d_register_rpc() => (null) (-21)
```

To prevent this, we must use only one shared FSEventStream with a paths for
all uv_fsevent_t handles, and then filter out events for each handle
using this paths again.

See https://github.com/joyent/node/issues/5463


indutny:
Should be ABI compatible with v0.10 if I hadn't messed with anything :)

src/unix/fsevents.c
```diff
@@ -72,26 +100,21 @@ static void uv__cf_loop_signal(uv_loop_t* loop,
 
 #define UV__FSEVENTS_WALK(handle, block)                                      \
```
bnoordhuis: While we're here, the name of this macro is somewhat poorly chosen.  'Walk' implies it just iterates but it also mutates it.  Can you address that in a follow-up commit?

indutny: Sure, anything else to fix?

src/unix/fsevents.c
```diff
@@ -125,42 +148,35 @@ static void uv__fsevents_event_cb(ConstFSEventStreamRef streamRef,
...
+    /* Process and filter out events */
+    for (i = 0; i < numEvents; i++) {
+      /* Ignore system events */
+      if (eventFlags[i] & kFSEventsSystem)
+        continue;
+
+      path = paths[i];
+      len = strlen(path);
+
+      /* Filter out paths that are outside hanlde's request */
```
bnoordhuis: s/hanlde/handle/

src/unix/fsevents.c
```diff
@@ -125,42 +148,35 @@ static void uv__fsevents_event_cb(ConstFSEventStreamRef streamRef,
...
+      if (strstr(path, handle->realpath) != path)
```
bnoordhuis: Use strncmp() here. strstr() is potentially O(n*m).

src/unix/fsevents.c
```diff
@@ -169,79 +185,79 @@ static void uv__fsevents_event_cb(ConstFSEventStreamRef streamRef,
...
-
   latency = 0.15;
 
   /* Set appropriate flags */
   flags = kFSEventStreamCreateFlagFileEvents;
 
+  /*
+   * NOTE: While we definitely could remember last event id,
+   * this will not be a right solution here. Because we might be watching
+   * different paths after reschedule.
```
bnoordhuis: Reschedule? I'm not sure I understand this comment.

src/unix/fsevents.c
```diff
@@ -381,11 +506,13 @@ static void* uv__cf_loop_runner(void* arg) {
...
+  uv__cf_loop_state_t* state;
   QUEUE* item;
   QUEUE split_head;
   uv__cf_loop_signal_t* s;
 
   loop = arg;
+  state = loop->cf_state;
 
   uv_mutex_lock(&loop->cf_mutex);
   QUEUE_INIT(&split_head);
```
bnoordhuis: Apropos nothing, there is no reason for this to be inside the critical section.

src/unix/fsevents.c
```diff
@@ -169,79 +185,79 @@ static void uv__fsevents_event_cb(ConstFSEventStreamRef streamRef,
...
 #endif /* MAC_OS_X_VERSION_10_7 */
 
-    /* Do not emit events from subdirectories (without option set) */
-    pos = strchr(path, '/');
-    if ((handle->cf_flags & UV_FS_EVENT_RECURSIVE) == 0 &&
-        pos != NULL &&
-        pos != path + 1)
-      continue;
+      /* Do not emit events from subdirectories (without option set) */
+      pos = strchr(path, '/');
```
bnoordhuis: Don't call strchr() when UV_FS_EVENT_RECURSIVE is not set.

src/unix/fsevents.c
```diff
@@ -250,43 +266,113 @@ static void uv__fsevents_schedule(void* arg) {
...
+    curr = QUEUE_DATA(q, uv_fs_event_t, cf_member);
+
+    assert(curr->realpath != NULL);
+    paths[i] = CFStringCreateWithCString(NULL,
+                                         curr->realpath,
+                                         CFStringGetSystemEncoding());
+    if (paths[i] == NULL)
+      abort();
+  }
+  uv_mutex_unlock(&state->fsevent_mutex);
```
bnoordhuis, line 348: Does this critical section have to be this big? Can you move the allocation and the loop out of it somehow?

src/unix/fsevents.c
```diff
@@ -250,43 +266,113 @@ static void uv__fsevents_schedule(void* arg) {
...
+    /* Create new FSEventStream */
+    cf_paths = CFArrayCreate(NULL, (const void**) paths, path_count, NULL);
+    if (cf_paths == NULL)
+      abort();
+    uv__fsevents_create_stream(handle->loop, cf_paths);
+  }
 
   /* Notify main thread that we're done here */
-  uv_sem_post(&handle->cf_sem);
+  if ((handle->flags & (UV_CLOSING | UV_CLOSED)) != 0)
```
bnoordhuis: uv__is_closing?

src/unix/fsevents.c
```diff
@@ -70,28 +98,23 @@ static void uv__cf_loop_signal(uv_loop_t* loop,
...
+      (handle)->cf_event = NULL;                                              \
       uv_mutex_unlock(&(handle)->cf_mutex);                                   \
-      while (!QUEUE_EMPTY(&split_head)) {                                     \
-        curr = QUEUE_HEAD(&split_head);                                       \
+      while (event != NULL) {                                                 \
         /* Invoke callback */                                                 \
-        event = QUEUE_DATA(curr, uv__fsevents_event_t, member);               \
-        QUEUE_REMOVE(curr);                                                   \
         /* Invoke block code, but only if handle wasn't closed */             \
         if (((handle)->flags & (UV_CLOSING | UV_CLOSED)) == 0)                \
```
bnoordhuis: uv__is_closing?

src/unix/fsevents.c
```diff
@@ -250,43 +266,113 @@ static void uv__fsevents_schedule(void* arg) {
...
   int err;
 
-  if (loop->cf_loop != NULL)
+  if (loop->cf_state != NULL)
     return 0;
 
+  state = malloc(sizeof(*state));
+  if (state == NULL)
+    return -ENOMEM;
+  memset(state, 0, sizeof(*state));
```
bnoordhuis: Just curious, why not calloc()?

src/unix/fsevents.c
```diff
@@ -442,10 +572,13 @@ int uv__fsevents_init(uv_fs_event_t* handle) {
 
   /* Get absolute path to file */
   handle->realpath = realpath(handle->filename, NULL);
```
bnoordhuis, line 588: Calling realpath() with resolved_name=NULL is implementation-defined.  I'm reasonably sure older versions of OS X segfault when it's NULL.  (Though it's possible those don't support FSEvents in the first place.)

src/unix/fsevents.c
```diff
@@ -442,10 +572,13 @@ int uv__fsevents_init(uv_fs_event_t* handle) {
...
-  if (handle->realpath != NULL)
-    handle->realpath_len = strlen(handle->realpath);
+  if (handle->realpath == NULL)
+    return -errno;
+  handle->realpath_len = strlen(handle->realpath);
+
+  /* Initialize signly-linked list */
```
bnoordhuis: Typo.  Hip startup name though.

src/unix/fsevents.c
indutny: No reason, just added memset in separate pass. Will fix in follow-up commit.

src/unix/fsevents.c
```diff
@@ -459,35 +592,54 @@ int uv__fsevents_init(uv_fs_event_t* handle) {
   handle->cf_cb->flags |= UV__HANDLE_INTERNAL;
   uv_unref((uv_handle_t*) handle->cf_cb);
 
+  /* TODO(indutny): check return value */
```
bnoordhuis: Well, do it then. :-)

src/unix/fsevents.c
indutny, line 348: Not really...

indutny, line 348: Its not that big, actually... Consider it as we're doing this stuff in main thread.

src/unix/fsevents.c
indutny: Well, imagine you're watching `/a` and last event id was `1`, then you're stopping it and starting watching `/b`. One might think (as I did) that its a good idea to start watching from event # `1`, but `/b` might already be at event-id `10000` and thus a lot of irrelevant data will be emitted.

src/unix/fsevents.c
indutny, line 588: It seems to be working, as this code was here for a half-year (or more).

bnoordhuis, line 588: By the way, handle->realpath is leaked when the call to malloc() a few lines down fails.

src/unix/fsevents.c
bnoordhuis: Ah, okay. Can you update the comment with that? :-)

src/unix/fsevents.c
bnoordhuis, line 588: > It seems to be working, as this code was here for a half-year (or more).

"seems to be working" is not the same thing as "correct".  I guess it's possible there is simply no one still running OS X 10.5 except my dad-in-law.

bnoordhuis:
Last round of comments. If you fix those then it LGTM.

indutny:
Thanks, landed with fixes in cd2794c01fc84a4118f79e31071cb1bca78918f5. Backporting to v0.10 now.

piscisaureus:
Nice work @indutny !

indutny:
Thanks :)

Waiting for merging backport before closing this issue: https://github.com/joyent/libuv/pull/896

siebeneicher:
Hey guys, I still get this error using grunt-watch on mac os. Nodejs is up to date (0.10.26). Does this fix practically fixed the mentioned error in the inital comment? indutny?
Short reply very much appreciated! THanks

PR #895, opened 2013-08-21
build: enable all compiler warnings

bnoordhuis:
Currently generates 315 warnings when building `make libuv.la` with gcc 4.8.

Jenkins, I beseech thee.


indutny:
@bnoordhuis sorry for a delay! I suspect that gyp counterpart is needed as well, isn't it?

indutny:
Though, this commit is LGTM

bnoordhuis:
I don't think there is a good way to check if the compiler supports a particular flag with gyp.

indutny:
Ok, whatever f75bb7c

indutny:
Thanks

saghul:
@indutny IMHO we need to look into the warnings before landing this. As of right now, this is what happens: https://gist.github.com/saghul/9993473

I suggest we revert it for now, fix the problems and land it afterwards.

indutny:
Ok, pushed revert.

saghul:
Thanks Fedor!

PR #897, opened 2013-08-22
windows: uv_spawn() fixes

piscisaureus:
@bnoordhuis or @sblom, care to review?

This is necessary to make execSync work on windows.


bnoordhuis:
What does Windows return in the high 32 bits?  ExitProcess() only takes an UINT, right?

PR #898, opened 2013-08-22
darwin: remove CoreFoundation dependency

bnoordhuis:
Load the required symbols at run-time rather than linking against the
CoreFoundation (and CoreServices and ApplicationServices) frameworks
at build time.

Should make integration easier for people that bundle libuv with their
own projects because they no longer have to replicate magic -framework
incantations in their top-level build system.

Suggested reviewer: @indutny


src/unix/fsevents.c
```diff
@@ -94,6 +95,47 @@ struct uv__fsevents_event_s {
 static void* uv__cf_loop_runner(void* arg);
 static int uv__cf_loop_signal(uv_loop_t* loop, uv_fs_event_t* handle);
 
+/* Lazy-loaded by uv__fsevents_global_init(). */
+static CFArrayRef (*pCFArrayCreate)(CFAllocatorRef,
```
indutny, line 99: I'd suggest moving this and other stuff to another file and including it in files that are using it. Though,  you'll need to make them `extern` in order to do this. Will visibility `hidden` work for this?

bnoordhuis, line 99: Yes, it works and yes, the thought had crossed my mind.  The thing is that darwin-proctitle.c also needs ApplicationServices.  I'll think about it.

bnoordhuis:
I've landed this as-is for now in d48168a.  I couldn't really come up with something that was significantly better than the approach in this PR, in particular because I don't want to load ApplicationServices unless absolutely necessary.  I'll entertain patches. :-)

PR #900, opened 2013-08-23
build: add --enable-Werror switch

hasufell:
There are too many cases where general -Werror causes compile
failures without good reason including:
- new warnings on version bumps of GCC/GLIBC the developer was
  not aware of at the point of coding
- some autoconf checks will fail badly
- libraries adding deprecated API warnings although that API
  is still working/supported
- on less known architectures we may get different/more
  warnings than on common ones
- random breakage depending on what distro/architecture/library
  version/kernel/userland the developer was testing "-Werror" on

So we add a switch and have it default to off, since it
is only interesting for development.

Specific flags such as -Werror=implicit-function-declaration
might be ok to force unconditionally for everyone depending
on what we expect from the code.


bnoordhuis:
I think you have the wrong idea about what that -Werror flag in configure.ac means.  It's passed to automake, not the compiler, and it does that because anything in the Makefile.am that generates a warning will very likely stop the build from working correctly.

Passing -Werror to the compiler is something you can do by passing CFLAGS to configure, that doesn't have to be its own configure switch.  E.g.:

```
$ ./configure CFLAGS="-Wall -Werror"
```

Thanks for sending the pull request but it doesn't seem like it's necessary.  But if I missed something, please let me know. :-)

hasufell:
oh, that is confusing indeed

Issue #901, opened 2013-08-23
ctime is "creation time" on windows, "ctime" on unix


isaacs:
It'd be better if "ctime" meant the same thing on Windows as it does on Unix.

Yes, of course, most platforms treat "ctime" as "creation time" on Windows, and in fact, even some Unix tutorials and blogs say that "ctime" is "creation time".  But we should not contribute to the confusion!

Libuv and Node aim at cross-platform portability.  The disparate semantics causes problems in some cases, especially since the "creation time" of a file is not updated if it is deleted and then re-created within 15 seconds.  Please make "ctime" on windows mean "changed time", as in, "the last time that any of the file's metadata was changed".

A cursory perusal of google shows that there's no way to know on NTFS when a file's metadata was changed.  Perhaps this can be closed as "can't fix".

If this is unpossible, then that's too bad.  However, it would still be nice to explicitly provide "creation_time" on the stat object.  Something like this maybe?

``` c
#ifdef not_sure_offhand_what_this_should_be_for_darwin_and_freebsd
#define CREATE_TIME(x) x.st_birthtime
#elif defined(WIN32)
#define CREATE_TIME(x) x.st_ctime
#else
#define CREATE_TIME(x) 0
#endif

#ifdef WIN32
#define CTIME(x) 0
#else
#define CTIME(x) x.st_ctime
#endif
```

bnoordhuis:
Windows issue. Assigning to you, Bertje.

jsstorm:
I agree that those file time stamps should be consistent for a low level cross platform API, or at least well documented with the difference for platforms.

I also confused with the ctime on unix, and treated them as creation time as windows. and later realized unlike windows, there is no creation time for unix systems.

For windows system, call GetFileInformationByHandleEx with FileBasicInfo can obtain the FILE_BASIC_INFO structure which contains file ChangeTime, which should be the same as ctime on unix. file writing operations and file attribs or ACL changing will result this time stamp updated.

But GetFileInformationByHandleEx is only supported for windows vista and later. also the ChangeTime only works on NTFS file system. FAT, FAT32, exFAT is not supported.

So maybe the libuv API can try to obtain what it can for the current os and file system, and fill zero values for those not supported, and high level applications can choose what to use depending on what they need and the system currently running.

piscisaureus:
I can do it for windows, for any version of windows. (Although, note that FAT doesn't track change time and I don't know offhand what windows would make of it.

[...]

PR #903, opened 2013-08-24
Change the meaning of ctime on windows

piscisaureus:
Plus additional niceties.

For #901.


piscisaureus:
Landed in https://github.com/joyent/libuv/compare/7f756955b643...20a8e58adb2a

PR #907, opened 2013-08-26
Fix build issues on mingw

alexcrichton:
At least in the mingw I'm using, there's no `cc`, but rather `gcc`, so default to `gcc` to avoid having to type `CC=gcc` all the time. Also append to CFLAGS instead of overriding them just in case some extra ones are supplied from the command line.

Finally, the real build issue was `DEVICE_TYPE` not being defined early enough, so define it a bit earlier to get it to build.


bnoordhuis:
Can you split this into two commits, one for the Makefile, the other for the header issue?  The usual prefix is 'windows:' and please be detailed in the commit log (see 'git log' for inspiration): explain what changed and why.  Don't worry about being wordy, I rather have commit logs with too much detail than too little.  Thanks.

bnoordhuis:
> Don't worry about being wordy, I rather have commit logs with too much detail than too little.

And to expand on that, the commit log should be so descriptive that I don't need to look at the patch anymore to find out what changed.

bnoordhuis:
Thanks Alex, landed in a9b3fe2 and b647c27.

PR #912, opened 2013-08-31
unix, windows: remove struct passing/returning from the libuv API

bnoordhuis:
Take it for a ride, Jenkins!

@piscisaureus You probably should take a look.


bnoordhuis:
#913 has a better fix for the swprintf() usage, by the way.

PR #921, opened 2013-09-05
build, windows: clarify instructions for Windows

bkize:
Clarified up front that Python must be installed for all Windows builds
and it must be in the path. Simplified the Windows instructions by
removing the redundant msbuild guidance.

Also added subheadings for each unique build environment.


saghul:
I only see the subheadings part, is the commit complete?

bkize:
@saghul Good catch, thx. Fixed.

README.md
```diff
@@ -62,31 +62,39 @@ To build with autotools:
...
-To build with Visual Studio run the vcbuild.bat file which will
-checkout the GYP code into build/gyp and generate the uv.sln and
-related files.
+### Windows
 
-Windows users can also build from the command line using msbuild.
-This is done by running vcbuild.bat from Visual Studio command prompt.
+First, Python 2.6 or 2.7 must be installed as it is required by [GYP][]. Also,
+the directory for the Python executable must be included in the Windows `Path`
+environment variable.
```
saghul: The PYTHON environment variable will also be checked. You could have Python 3 in the path and specify PYTHON pointing to a Python 2 executable not in the path.

README.md
```diff
@@ -62,31 +62,39 @@ To build with autotools:
...
-Unix users run:
+### Unix
+
+Run:
 
     $ ./gyp_uv -f make
     $ make -C out
 
-Macintosh users run:
+### Macintosh
```
saghul: Silly nit, but can we say "OSX" here? Not even Apple calls it "Macintosh" anymore :-P

bkize:
@saghul Clarified Python env var options and used an Apple OS name from the current millennium.

README.md
```diff
@@ -62,31 +62,40 @@ To build with autotools:
...
     $ make install
 
-To build with Visual Studio run the vcbuild.bat file which will
-checkout the GYP code into build/gyp and generate the uv.sln and
-related files.
+### Windows
 
-Windows users can also build from the command line using msbuild.
-This is done by running vcbuild.bat from Visual Studio command prompt.
+First, Python 2.6 or later must be installed as it is required by [GYP][].
```
saghul: Only 2.6 and 2.7 are supported, not python 3.

bkize:
@saghul Got it. I was thrown off by the earlier Python 3 env var comment. Reverted to the original "2.6 or 2.7" verbiage.

saghul:
LGTM but I'm just an spectator, @bnoordhuis and @piscisaureus are the ones in charge :-)

bnoordhuis:
Thanks Brian, landed in 8e23b04.  I took the liberty of landing it with the email address that you used to sign the CLA rather than the noreply.github.com address.  That way, you'll show up in the AUTHORS list correctly.

PR #922, opened 2013-09-06
windows: Run close callbacks after polling for i/o

saghul:
Master version of #814


bnoordhuis:
Landed in 2b9c374, thanks Saúl.

PR #928, opened 2013-09-12
unix: use a heap for timers

bnoordhuis:
Replace the red-black tree with a heap.  The most common operation that
libuv performs on timers is looking up the first timer to expire.  With
a red-black tree, that operation is O(log n).  With a heap, it's O(1).

### DIFF
--- a/include/uv-unix.h
+++ b/include/uv-unix.h
@@ -176,16 +176,16 @@ typedef struct {
   void* idle_handles[2];                                                      \
   void* async_handles[2];                                                     \
   struct uv__async async_watcher;                                             \
-  /* RB_HEAD(uv__timers, uv_timer_s) */                                       \
-  struct uv__timers {                                                         \
-    struct uv_timer_s* rbh_root;                                              \
-  } timer_handles;                                                            \
+  struct {                                                                    \
+    void* min;                                                                \
+    unsigned int nelts;                                                       \
+  } timer_heap;                                                               \
+  uint64_t timer_counter;                                                     \
   uint64_t time;                                                              \
   int signal_pipefd[2];                                                       \
   uv__io_t signal_io_watcher;                                                 \
   uv_signal_t child_watcher;                                                  \
   int emfile_fd;                                                              \
-  uint64_t timer_counter;                                                     \
   UV_PLATFORM_LOOP_FIELDS                                                     \
 
 #define UV_REQ_TYPE_PRIVATE /* empty */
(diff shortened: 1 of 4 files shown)


bnoordhuis:
I'll probably polish it some more before landing; gcc 4.7.2 generates some really inefficient code for some of the constructs:

```
       │         for (n = 0; n < k; n += 1) {
 28.17 │415acc:   add    $0x1,%ecx
```

Yes, that's 28.17% of an inner loop spent on a single instruction.

piscisaureus:
@bnoordhuis 

That’s very odd. Do you have a reasonable explanation for that 28.17%, this looks pretty extreme? Also why is it not an `inc` instruction?

It would be trivial btw to cache the “next” timer which would largely mitigate this problem.

bnoordhuis:
That inner loop exists of only a few instructions.  Interestingly enough gcc only generates the add in speed-over-size mode; with -Os, it generates an inc.  I wonder if there is some timing or dependency issue with inc on modern CPUs that I'm not aware of.

> It would be trivial btw to cache the “next” timer which would largely mitigate this problem.

Yeah, but that kind of blows when with a heap the optimal solution falls out naturally.  It also shaves off a few bytes of the uv_timer_t struct.  Nothing critical, just nice.

piscisaureus:
This is what msvc makes of it. Impeccable.

```
   196: DWORD uv_get_poll_timeout(uv_loop_t* loop) {
   197:   uv_timer_t* timer;
   198:   int64_t delta;
   199: 
   200:   /* Check if there are any running timers */
   201:   timer = RB_MIN(uv_timer_tree_s, &loop->timers);
01262DD0  mov         eax,dword ptr [esi+34h]  
01262DD3  sub         esp,0Ch  
01262DD6  push        ebx  
01262DD7  push        edi  
01262DD8  test        eax,eax  
01262DDA  je          uv_get_poll_timeout+7Ch (1262E4Ch)  
01262DDC  lea         esp,[esp]  
01262DE0  mov         ebx,eax  
01262DE2  mov         eax,dword ptr [eax+20h]  
01262DE5  test        eax,eax  
01262DE7  jne         uv_get_poll_timeout+10h (1262DE0h)  
   202:   if (timer) {
01262DE9  test        ebx,ebx  
01262DEB  je          uv_get_poll_timeout+7Ch (1262E4Ch)  
   203:     uv_update_time(loop);
```

saghul:
Landed in f17c535, thanks Ben!

PR #932, opened 2013-09-19
unix, windows: add uv_fs_truncate() function

bnoordhuis:
Refs joyent/node#6233.

/cc @piscisaureus

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -1621,6 +1621,7 @@ typedef enum {
   UV_FS_STAT,
   UV_FS_LSTAT,
   UV_FS_FSTAT,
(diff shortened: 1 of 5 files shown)


piscisaureus:
@bnoordhuis Do I need to review this?

bnoordhuis:
@piscisaureus Initially yes, but now I'm not sure whether we really need it.  fs.truncate() in node.js works after joyent/node@b011811 using open() + ftruncate().  It requires an extra thread pool round-trip but that should be okay, it's not a very performance critical - as evidenced by the fact that fs.truncate() in node.js was broken for years and no one noticed.  Closing.

PR #933, opened 2013-09-20
build: use -pthread on sunos

tjfontaine:
When building on sunos with autoconf make sure to specify -pthread
otherwise there will be race conditions with errno and etc

@bnoordhuis


bnoordhuis:
LGTM. I take it that was the source of the test failures?

tjfontaine:
@bnoordhuis ya, errno was getting stomped on by something else

PR #935, opened 2013-09-23
unix, windows: add uv_fs_event_start/stop functions

saghul:
Make uv_fs_event behave like other handles, that is, it's inactive after
init, and it's active between start and stop.

That.

I didn't add any new test (I plan to), but I'd like to get some feedback to see if I didn't fuck up anything too badly to begin with. Tests pass (I tested OSX, Linux and Windows).

The basic idea was to make init pretty dummy, put the meat of init now in start and what close did, do it in stop. But well, the devil is in the details :-)

Please review:

@bnoordhuis everything in general and nothing in particular :-)

@indutny please overview if the logic looks sound on OSX, where the fsevent thread is used

@piscisaureus my Windows-fu is not very strong, I'm trying! Please check if the stop process is done correctly, also the close. Basically I added some extra check not to fire any callback if there is an in-flight request when we called stop, in close use an endgame as usual. Also, I'm reusing handle->buffer until you close the handle, I hope that's ok.

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -534,9 +534,6 @@ UV_EXTERN size_t uv_req_size(uv_req_type type);
  *      function, then it's active from the moment that function is called.
  *      Likewise, uv_foo_stop() deactivates the handle again.
  *
- *  - A uv_fs_event_t handle is currently always active.  Future versions
- *    of libuv may add uv_fs_event_start() and uv_fs_event_stop() functions.
- *
  */
 UV_EXTERN int uv_is_active(const uv_handle_t* handle);
 
@@ -1850,7 +1847,7 @@ UV_EXTERN void uv_loadavg(double avg[3]);
 
 
 /*
- * Flags to be passed to uv_fs_event_init.
+ * Flags to be passed to uv_fs_event_start.
  */
 enum uv_fs_event_flags {
   /*
@@ -1880,8 +1877,15 @@ enum uv_fs_event_flags {
 };
 
 
-UV_EXTERN int uv_fs_event_init(uv_loop_t* loop, uv_fs_event_t* handle,
-    const char* filename, uv_fs_event_cb cb, int flags);
+UV_EXTERN int uv_fs_event_init(uv_loop_t* loop, uv_fs_event_t* handle);
+
+UV_EXTERN int uv_fs_event_start(uv_fs_event_t* handle,
+                                uv_fs_event_cb cb,
+                                const char* filename,
+                                unsigned int flags);
+
+UV_EXTERN int uv_fs_event_stop(uv_fs_event_t* handle);
+
 
 /* Utility */
 
--- a/src/unix/kqueue.c
+++ b/src/unix/kqueue.c
@@ -296,23 +296,30 @@ static void uv__fs_event(uv_loop_t* loop, uv__io_t* w, unsigned int fflags) {
 }
 
 
-int uv_fs_event_init(uv_loop_t* loop,
-                     uv_fs_event_t* handle,
-                     const char* filename,
-                     uv_fs_event_cb cb,
-                     int flags) {
+int uv_fs_event_init(uv_loop_t* loop, uv_fs_event_t* handle) {
+  uv__handle_init(loop, (uv_handle_t*)handle, UV_FS_EVENT);
+  return 0;
+}
+
+
+int uv_fs_event_start(uv_fs_event_t* handle,
+                      uv_fs_event_cb cb,
+                      const char* filename,
+                      unsigned int flags) {
 #if defined(__APPLE__)
   struct stat statbuf;
 #endif /* defined(__APPLE__) */
   int fd;
+
+  if (uv__is_active(handle))
+    return -EINVAL;
 
   /* TODO open asynchronously - but how do we report back errors? */
   fd = open(filename, O_RDONLY);
   if (fd == -1)
     return -errno;
 
-  uv__handle_init(loop, (uv_handle_t*)handle, UV_FS_EVENT);
-  uv__handle_start(handle); /* FIXME shouldn't start automatically */
+  uv__handle_start(handle);
   uv__io_init(&handle->event_watcher, uv__fs_event, fd);
   handle->filename = strdup(filename);
   handle->cb = cb;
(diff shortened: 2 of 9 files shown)


include/uv.h
```diff
@@ -1880,8 +1877,15 @@ enum uv_fs_event_flags {
...
 
 
-UV_EXTERN int uv_fs_event_init(uv_loop_t* loop, uv_fs_event_t* handle,
-    const char* filename, uv_fs_event_cb cb, int flags);
+UV_EXTERN int uv_fs_event_init(uv_loop_t* loop, uv_fs_event_t* handle);
+
+UV_EXTERN int uv_fs_event_start(uv_fs_event_t* handle,
+                                uv_fs_event_cb cb,
+                                const char* filename,
+                                int flags);
```
bnoordhuis: s/int/unsigned int/ while you're here.

src/unix/kqueue.c
```diff
@@ -296,23 +296,30 @@ static void uv__fs_event(uv_loop_t* loop, uv__io_t* w, unsigned int fflags) {
...
+                      uv_fs_event_cb cb,
+                      const char* filename,
+                      int flags) {
 #if defined(__APPLE__)
   struct stat statbuf;
 #endif /* defined(__APPLE__) */
   int fd;
 
+  if (uv__is_active(handle))
+    return 0;
```
bnoordhuis: Maybe -EBUSY or -EINVAL?  I could live with turning this into an assert().

src/unix/kqueue.c
```diff
@@ -335,13 +342,16 @@ int uv_fs_event_init(uv_loop_t* loop,
...
+  uv__io_start(handle->loop, &handle->event_watcher, UV__POLLIN);
 
   return 0;
 }
 
 
-void uv__fs_event_close(uv_fs_event_t* handle) {
+int uv_fs_event_stop(uv_fs_event_t* handle) {
+  if (!uv__is_active(handle))
+    return 0;
```
bnoordhuis: Ditto.

src/unix/kqueue.c
saghul: I borrowed inspiration from fs_poll. I was actually thinking EALREADY for start and EINVAL for stop. This probably would need to be applied to other handles, I can do that afterwards, if that's ok.

src/win/fs-event.c
```diff
@@ -131,14 +107,42 @@ static int uv_split_path(const WCHAR* filename, WCHAR** dir,
...
+  handle->dir_handle = INVALID_HANDLE_VALUE;
+  handle->buffer = NULL;
+  handle->req_pending = 0;
+  handle->filew = NULL;
+  handle->short_filew = NULL;
+  handle->dirw = NULL;
+
+  uv_req_init(loop, (uv_req_t*)&handle->req);
+  handle->req.type = UV_FS_EVENT_REQ;
+  handle->req.data = (void*)handle;
```
bnoordhuis: Superfluous cast.

src/win/fs-event.c
```diff
@@ -213,15 +217,17 @@ int uv_fs_event_init(uv_loop_t* loop, uv_fs_event_t* handle,
...
                              0) == NULL) {
     last_error = GetLastError();
     goto error;
   }
 
-  handle->buffer = (char*)_aligned_malloc(uv_directory_watcher_buffer_size,
-    sizeof(DWORD));
+  if (!handle->buffer) {
+    handle->buffer = (char*)_aligned_malloc(uv_directory_watcher_buffer_size,
+                                            sizeof(DWORD));
```
bnoordhuis, line 229: Just curious, is the call to `_aligned_malloc()` necessary?  Put another way, can the normal malloc() return a pointer that's not at least sizeof(DWORD) aligned?

src/win/fs-event.c
```diff
@@ -456,22 +501,19 @@ void uv_process_fs_event_req(uv_loop_t* loop, uv_req_t* req,
...
   }
 
-  uv__handle_closing(handle);
 }
 
 
 void uv_fs_event_endgame(uv_loop_t* loop, uv_fs_event_t* handle) {
-  if (handle->flags & UV__HANDLE_CLOSING &&
-      !handle->req_pending) {
+  if (handle->flags & UV__HANDLE_CLOSING && !handle->req_pending) {
```
bnoordhuis: Parentheses around first expression.

src/win/fs-event.c
saghul, line 229: @piscisaureus can tell you better :-)

src/unix/kqueue.c
bnoordhuis: EBUSY or EINVAL is more POSIX-y.  I guess EINVAL for both start and stop is the least amount of mental overhead.

saghul: This expands to all handles, so I'd leave fs_events like fs_poll is and then make a change adjusting the behavior globally, whatever we decide is best :-)

saghul:
Force pushed addressing Ben's review. I made start and stop return UV_EINVAL if called when the handle was already started / stopped respectively, I'll also add a test for that.

bnoordhuis:
LGTM from a code POV but it seems to break fs_event_watch_dir on OS X.  @piscisaureus should sign off on the Windows part, it probably wouldn't hurt for @indutny to take a peek at the FSEvents changes.

saghul:
Thanks for the review! The OSX test failed for me almost every time (on 10.6), FWIW. I'll run the tests (before and after) on 10.8 to see if there is any difference.

PS: What's the OSX version of the build bot?

tjfontaine:
It's on 10.8 atm

saghul:
Thanks @tjfontaine.

saghul:
Ok, forget what I said, looks like I was remembering wrong. I could indeed reproduce the OSX issue reliably on my 10.8 machine. I'll look into it.

saghul:
Ok, rebased, updated and fixed the OSX test failure, should be good to go now :-) @indutny can you please take a peak uv_fs_event_stop? I modified how the fsevents thread is stopped since now the handle is not closed, so I changed it to check if the handle is not active.

indutny:
@saghul OSX changes seems to be very minor and are generally looking good to me. LGTM

bnoordhuis:
Another LGTM.

saghul:
Thanks guys, landed on https://github.com/joyent/libuv/commit/9d44d786ada6cf94e1bdcee7d777c790ca712a78 :-)

PR #939, opened 2013-09-27
win: fix immediate re-assignment

mscdex:


bnoordhuis:
Thanks Brian, but that's not a bug.  Zeroing locals at the top is something of a best practice in many cases, avoids uninitialized variable bugs when refactoring code.

PR #942, opened 2013-09-29
Test: relax some implicit assumptions

lucab:
The test-suite makes several assumptions on the build environment (mostly about tty availability and timers deadline) which may not be true under heavy-loaded autobuilders or slower/virtualized architectures.
This PR is target to the v0.10 stable branch and aims at fixing several occasional false-positive test-suite failures as experienced on Debian autobuilders, whose logs are available at https://buildd.debian.org/status/logs.php?pkg=libuv


lucab:
So, well, I didn't like to, but I did as suggested:
- reworded commit log to avoid tripping the 50-char check (this means dropping the "test: " prefix, meh)
- URL-shortened the build-log URL (hoping that the shortening service will keep it long enough, double-meh)

Properly signed the CLA now, I thought it was not needed for trivial changes.

test/test-loop-stop.c
```diff
@@ -62,7 +62,7 @@ static void timer_cb(uv_timer_t* handle, int status) {
 
   r = uv_run(uv_default_loop(), UV_RUN_NOWAIT);
   ASSERT(r != 0);
-  ASSERT(prepare_called == 3);
+  ASSERT(prepare_called > 1);
```
bnoordhuis, line 65: Isn't this an issue with the 100 ms expiry time of the timer a few lines up?

(EDIT: Granted, this test is somewhat timing sensitive by nature.)

test/test-timer-again.c
```diff
@@ -57,7 +57,7 @@ static void repeat_1_cb(uv_timer_t* handle, int status) {
   r = uv_timer_again(&repeat_2);
   ASSERT(r == 0);
 
-  if (uv_now(uv_default_loop()) >= start_time + 500) {
+  if (uv_now(uv_default_loop()) >= start_time + 500 && repeat_1_cb_called == 10) {
```
bnoordhuis: I think the wonky `uv_now(uv_default_loop()) >= start_time + 500` check can be dropped altogether.  Same for the 700 ms check at the bottom of the file, you'll probably hit that sooner or later too.

bnoordhuis:
Couple of comments but I'm pretty much okay with the changes.  That said, can you reword the commit logs to look like `test: <summary>`?  Thanks.

lucab:
Thanks for the quick comments, rebased.

bnoordhuis:
Thanks Luca, landed in master.

PR #943, opened 2013-10-01
unix: sanity-check fds before closing

bnoordhuis:
RFC from @piscisaureus and @saghul.

The idea is to catch a pretty common type of error that we have in node.js, namely closing stdio file descriptors.

It also ties in with joyent/node#6271 where libuv triggers a Linux kernel bug that results in a busy loop.


test/test-stdio-over-pipes.c
```diff
@@ -43,7 +43,6 @@
 
 
 static void close_cb(uv_handle_t* handle) {
-  printf("close_cb\n");
```
bnoordhuis, line 46: This was one of those bugs: calling printf() after fd 1 is closed.

src/unix/udp.c
```diff
@@ -40,8 +40,11 @@
 void uv__udp_close(uv_udp_t* handle) {
   uv__io_close(handle->loop, &handle->io_watcher);
   uv__handle_stop(handle);
-  close(handle->io_watcher.fd);
-  handle->io_watcher.fd = -1;
+
+  if (handle->io_watcher.fd != -1) {
```
saghul, line 44: I guess that before the change if you just did uv_udp_init and then close, fd stayed at -1 but we didn't care about the error, right?

bnoordhuis, line 44: Yep.  It did a `close(-1)` which isn't harmful but kind of sloppy (shows up in strace as failing with EBADF.)

src/unix/core.c
```diff
@@ -406,6 +406,26 @@ int uv__accept(int sockfd) {
...
 
 
+int uv__close(int fd) {
+  int saved_errno;
+  int rc;
+
+  assert(fd > -1);  /* Catch uninitialized io_watcher.fd bugs. */
+  assert(fd > STDERR_FILENO);  /* Catch stdio close bugs. */
+
+  saved_errno = errno;
```
saghul, line 416: Before this we didn't save the errno (only in some cases) so is this now a Good Thing (TM) ?

saghul:
LGTM :+1:

src/unix/core.c
bnoordhuis, line 416: Early versions of uv-unix saved errno in a lot of places but I kind of scaled back on that.  For uv__close(), it's mostly a convenience thing - it often gets called from error paths where the original errno needs to be preserved.  Dealing with that in uv__close() makes that easier.

saghul, line 416: I see. Thanks for explaining!

bnoordhuis:
Thanks @saghul, landed in 359d667.

PR #946, opened 2013-10-02
fsevents: fix clever rescheduling

indutny:
There're could be a situation, where one fsevents handle gets created
and another one is destroyed simultaneously. In such cases
`fsevent_need_reschedule` will be set to 1 twice and reset only once,
leaving handle destructor hanging in uv_sem_wait().

se https://github.com/joyent/node/issues/6296

/cc @bnoordhuis


indutny:
Added test.

test/test-fs-event.c
```diff
@@ -502,3 +502,31 @@ static void fs_event_cb_close(uv_fs_event_t* handle, const char* filename,
 }
 
 #endif /* HAVE_KQUEUE */
+
+TEST_IMPL(fs_event_start_and_close) {
+  uv_loop_t* loop;
+  static uv_fs_event_t fs_event1;
+  static uv_fs_event_t fs_event2;
```
bnoordhuis: Why static?

src/unix/fsevents.c
```diff
@@ -317,9 +317,13 @@ static void uv__fsevents_reschedule(uv_fs_event_t* handle) {
...
    */
-  if (!state->fsevent_need_reschedule)
+  uv_mutex_lock(&state->fsevent_mutex);
+  if (state->fsevent_need_reschedule == 0) {
+    uv_mutex_unlock(&state->fsevent_mutex);
     return;
-  state->fsevent_need_reschedule = 0;
+  }
+  state->fsevent_need_reschedule--;
+  uv_mutex_unlock(&state->fsevent_mutex);
```
bnoordhuis, line 326: Are similar issues possible with e.g. the call to uv__fsevents_create_stream() a few lines down?  It's also not guarded by the mutex.

indutny, line 326: Not really, but I just realized that this fix breaks this optimization.

bnoordhuis, line 326: Another thing, if all accesses to state->fsevent_need_reschedule are now guarded by the mutex, then it doesn't need to be volatile anymore.

indutny, line 326: Ok, figured it out. Going to push update in a bit.

indutny:
Fixed all nits, and improved patch to keep clever hack working.

src/unix/fsevents.c
```diff
@@ -317,9 +317,13 @@ static void uv__fsevents_reschedule(uv_fs_event_t* handle) {
   /* Optimization to prevent O(n^2) time spent when starting to watch
    * many files simultaneously
    */
-  if (!state->fsevent_need_reschedule)
-    return;
+  uv_mutex_lock(&state->fsevent_mutex);
+  if (state->fsevent_need_reschedule == 0) {
```
bnoordhuis, line 321: My comment about fsevent_need_reschedule still stands.  It doesn't need to be volatile anymore, right?

bnoordhuis:
One comment but otherwise LGTM.

indutny:
Landed in master in 429bb80.

PR #950, opened 2013-10-09
.gitignore: fix .deps/, add .dirstamp

rurban:


bnoordhuis:
Thanks Reini, landed in b0bb8de.

PR #956, opened 2013-10-18
unix: make uv_tty_reset_mode() async signal-safe

bnoordhuis:
Make uv_tty_reset_mode() async signal-safe and add some atomic operations and a spinlock type while we're here.

Suggested reviewer: @sam-github?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -92,6 +92,7 @@ libuv_la_SOURCES += src/unix/async.c \
                    src/unix/poll.c \
                    src/unix/process.c \
                    src/unix/signal.c \
+                   src/unix/spinlock.h \
                    src/unix/stream.c \
                    src/unix/tcp.c \
                    src/unix/thread.c \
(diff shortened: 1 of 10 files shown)


sam-github:
@bnoordhuis, sorry, way too late, but this should allow node.js to reset tty state in signal hander, rather than notifying main loop, resetting tty there, and exiting, right?

But node has not yet taken advantage of the signal safety, it still exits when signalled:

```
% ../../unstable-node/node exit-with-signal.js
worker pid 538
worker on exit 143 null process.exitCode 143
```

https://gist.github.com/sam-github/6236471

And other than re-raising the signal in the handler, I can't think of a way to both handle a signal, and then allow the default behaviour to occur (Terminate in this case).

I don't care that node exits on some signals. Its OK that SIGTERM and SIGINT are used internally, as is SIGUSR1, I just think it should be documented, the note [here](http://nodejs.org/api/child_process.html#child_process_event_exit) really needs to be in [process](http://nodejs.org/api/process.html). Want a PR?

bnoordhuis:
> sorry, way too late, but this should allow node.js to reset tty state in signal hander, rather than notifying main loop, resetting tty there, and exiting, right?

Correct.  It's not said that node is actually switching over to that but now at least it can.

> I don't care that node exits on some signals. Its OK that SIGTERM and SIGINT are used internally, as is SIGUSR1, I just think it should be documented, the note here really needs to be in process. Want a PR?

Sure, thanks.

PR #958, opened 2013-10-20
fsevents: poll when FSEvents fail

indutny:
As Apple documentation for FSEvents suggests, we should fall-back to
polling directory for changes when `FSEventStreamStart()` fails and
returns `false`.

The fall-back has one drawback though - it doesn't work for
subdirectories.

/cc @bnoordhuis


src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+
+#include "uv.h"
+#include "internal.h"
+
+#define POLL_INTERVAL 50  /* 50 ms */
+#define EMIT_ERROR(handle, code)                                              \
+    do {                                                                      \
+      (handle)->cb((handle), NULL, 0, (code));                                \
+      (handle)->cb = uv__fsevents_poller_nop;                                 \
+    } while (0)
```
bnoordhuis: This could be a static function, right?

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+                                           const uv_stat_t* prev,
+                                           const uv_stat_t* curr);
+static void uv__fsevents_poller_nop(uv_fs_event_t* handle,
+                                    const char* filename,
+                                    int events,
+                                    int status);
+
+
+int uv__fsevents_poller_init(uv_loop_t* loop, uv__fsevents_poller_t* poller) {
+  int r;
```
bnoordhuis: s/r/err/? That's what's used most everywhere else.

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+
+  poller->signal = malloc(sizeof(*poller->signal));
+  if (poller->signal == NULL)
+    goto signal_alloc_failed;
+
+  r = uv_async_init(loop, poller->signal, uv__fsevents_poller_signal_cb);
+  if (r != 0)
+    goto signal_init_failed;
+
+  uv_unref((uv_handle_t*) poller->signal);
```
bnoordhuis: uv__handle_unref()?  Doesn't need a cast.

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+  return r;
+}
+
+
+void uv__fsevents_poller_start(uv__fsevents_poller_t* poller) {
+  poller->active = 1;
+}
+
+
+int uv__fsevents_poller_is_active(uv__fsevents_poller_t* poller) {
```
bnoordhuis: const

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+  state = handle->loop->cf_state;
+  poller = handle->data;
+
+  /* Destroy previous items */
+  item = poller->items;
+  while (item != NULL) {
+    next = item->next;
+    uv__fsevents_poller_item_destroy(item);
+    item = next;
+  }
```
bnoordhuis: Same logic as uv__fsevents_poller_destroy().  Could be moved into a separate function.

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+
+readdir_failed:
+  uv_close((uv_handle_t*) item->poll, (uv_close_cb) free);
+
+alloc_failed:
+  free(item->poll);
+  free(item);
+
+item_alloc_failed:
+  if (r != 0)
```
bnoordhuis: Always true, right?

indutny: oh, yes.

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+  size_t i;
+  int j;
+  uv__fsevents_poller_file_t* file;
+  uv__fsevents_poller_file_t** files[2];
+  int count[2];
+
+  uv_close((uv_handle_t*) item->poll, (uv_close_cb) free);
+  if (item->readdir_req != NULL) {
+    item->readdir_req->data = NULL;
+    uv_cancel((uv_req_t*) item->readdir_req);
```
bnoordhuis, line 214: The NULL assignment is not thread-safe.

Note on uv_cancel(): it's not guaranteed to cancel the request, only when it's still in the 'pending work requests' queue.

indutny, line 214: I'm not checking it from different threads.

indutny, line 214: About, `uv_cancel()` that's exactly the reason, why I'm erasing `req->data` here.

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+
+  r = uv_fs_readdir(item->handle->loop,
+                    req,
+                    item->handle->filename,
+                    0,
+                    uv__fsevents_poller_readdir_cb);
+  if (r != 0)
+    free(req);
+
+  req->data = item;
```
bnoordhuis, line 273: Use after free.

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+  uv__fsevents_poller_item_t* item;
+  const char* dirname;
+  char* name;
+  char* path;
+  char* names;
+  int i;
+  int dirlen;
+  int len;
+
+  item = req->data;
```
bnoordhuis, line 290: Manipulated by another thread in a way that's not reorder safe.

EDIT: Please ignore, see comment some hundred lines up.

indutny, line 290: Well, I'm checking it here and skipping body and going string to fatal. So, in any order it should work, right?

If you have counterexample - please describe it.

src/unix/internal.h
```diff
@@ -229,10 +229,84 @@ void uv__work_submit(uv_loop_t* loop,
...
 int uv__fsevents_init(uv_fs_event_t* handle);
 int uv__fsevents_close(uv_fs_event_t* handle);
 void uv__fsevents_loop_delete(uv_loop_t* loop);
 
+int uv__fsevents_poller_init(uv_loop_t* loop, uv__fsevents_poller_t* poller);
+int uv__fsevents_poller_is_active(uv__fsevents_poller_t* poller);
+void uv__fsevents_poller_start(uv__fsevents_poller_t* poller);
+int uv__fsevents_poller_signal(uv__fsevents_poller_t* poller);
+void uv__fsevents_poller_destroy(uv__fsevents_poller_t* poller);
+
```
bnoordhuis: I'd prefer it if the AAPL cruft gets moved into its own header.  internal.h is intended for definitions that are shared across platforms.

src/unix/internal.h
```diff
@@ -229,10 +229,84 @@ void uv__work_submit(uv_loop_t* loop,
...
+struct uv__fsevents_poller_item_s {
+  /* Singly-linked list of items */
+  uv__fsevents_poller_item_t* next;
+
+  uv_fs_event_t* handle;
+  uv_fs_t* readdir_req;
+  uv_fs_poll_t* poll;
+
+  /* Current list of files */
+  int count;
```
bnoordhuis: Is there a reason for this field to be signed?

src/unix/internal.h
```diff
@@ -229,10 +229,84 @@ void uv__work_submit(uv_loop_t* loop,
...
+  uv__fsevents_poller_file_t** files;
+
+  /* List of previously-observed files */
+  int prev_count;
```
bnoordhuis: Ditto.

src/unix/internal.h
```diff
@@ -229,10 +229,84 @@ void uv__work_submit(uv_loop_t* loop,
...
+  uv__fsevents_poller_file_t** prev_files;
+
+  /* Number of files, waiting for uv_fs_stat() completion */
+  int stats_waiting;
```
bnoordhuis: (And ditto.)

src/unix/fsevents-poll.c
bnoordhuis, line 214: Oh, I guess that's not true - this runs on the libuv thread, right?  Maybe some comments that explain what functions runs in what thread are in order, it's not immediately obvious now.

indutny, line 214: Sure, but in fact `_signal()` is the only one, that runs off-thread here.

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+  count[1] = item->prev_count;
+
+  for (i = 0; i < ARRAY_SIZE(files); i++) {
+    if (files[i] == NULL)
+      continue;
+
+    for (j = 0; j < count[i]; j++) {
+      file = files[i][j];
+      if (file->stat_waiting) {
+        uv_cancel((uv_req_t*) &file->stat_req);
```
bnoordhuis: How many fs scandir and stat requests will there be in flight at any given time?  The thread pool is not infinitely scalable.

indutny: D - number of dirs, F - number of files. Answer: `D * F`.

Won't thread pool queue all this stuff and run in fixed amount of threads?

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+
+  /* Copy all file names */
+  names = req->ptr;
+  dirname = req->path;
+  dirlen = strlen(dirname);
+  for (i = 0; i < item->count; i++) {
+    item->files[i] = calloc(1, sizeof(*item->files[i]));
+    len = strlen(names);
+    path = malloc(len + 1 + dirlen + 1);
+    name = malloc(len + 1);
```
bnoordhuis: Can you malloc this once?

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+    uv__fsevents_poller_stats_done(item);
+
+fatal:
+  if (item != NULL)
+    item->readdir_req = NULL;
+
+  /* Emit error */
+  if (req->result != UV_ECANCELED && req->result < 0) {
+    EMIT_ERROR(item->handle, req->result);
+    return;
```
bnoordhuis: This leaks req, doesn't it?

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+  }
+
+  free(req);
+  return;
```
bnoordhuis: Superfluous statement.

src/unix/fsevents-poll.c
indutny: exactly. I'll remove `return here`.

src/unix/fsevents-poll.c
indutny: With `MAX_PATH` size?

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+
+  file = req->data;
+  file->stat_waiting = 0;
+  item = file->item;
+
+  /* Was cancelled - dealloc */
+  if (req->result == UV_ECANCELED) {
+    free(file->name);
+    free(file->path);
+    free(file);
```
bnoordhuis, line 375: Cleanup code for `uv__fsevents_poller_file_t` instances should probably be generalized into a function.  I see a few places where cleanup happens in an ad hoc way.

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+  /* Emit errors */
+  if (req->result != 0) {
+    EMIT_ERROR(item->handle, req->result);
+    return;
+  }
+
+  /* Save stats */
+  memcpy(&file->stat, req->ptr, sizeof(file->stat));
+
+  if (--item->stats_waiting == 0)
```
bnoordhuis, line 387: stats_waiting should be unsigned.  Maybe add an `assert(item->stats_waiting > 0)` just before to guard against underflow?

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+  }
+
+  /* Emit errors */
+  if (req->result != 0) {
+    EMIT_ERROR(item->handle, req->result);
+    return;
+  }
+
+  /* Save stats */
+  memcpy(&file->stat, req->ptr, sizeof(file->stat));
```
bnoordhuis: You can use `&req->statbuf` here and maybe use assignment instead of an unchecked memcpy().

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+    j = 0;
+    while (i < item->prev_count && j < item->count) {
+      prev = item->prev_files[i];
+      next = item->files[j];
+      r = strcmp(prev->path, next->path);
+
+      if (r < 0) {
+        /* File was removed */
+        i++;
+        item->handle->cb(item->handle, prev->name, UV_RENAME, 0);
```
bnoordhuis: Seem unsafe.  What happens when the callback calls uv_close()?

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,484 @@
...
+  item = handle->data;
+
+  /* Emit errors */
+  if (status != 0) {
+    EMIT_ERROR(item->handle, status);
+    return;
+  }
+
+  /* Directory was removed */
+  if (curr == NULL)
```
bnoordhuis: curr should never be NULL when status == 0.

src/unix/fsevents-poll.c
bnoordhuis: Yes, but that means the queue could back up indefinitely - and will if D or F are large.

src/unix/fsevents-poll.c
bnoordhuis: I don't understand the question.  What I mean is, instead of doing `p = malloc(a); q = malloc(b);`, do `p = malloc(a + b); q = p + a;`

bnoordhuis:
Left some comments.  General observations:
- It looks like the line count could be reduced by refactoring some common code.
- The patch seems pretty malloc heavy.  I'm counting 22 calls to free().
- I have some doubts about the scalability aspect: it seems you could easily saturate the thread pool this way.

src/unix/fsevents-poll.c
indutny: That's the price to pay, unfortunately.

src/unix/fsevents-poll.c
indutny: Added handling.

indutny:
@bnoordhuis please check it again, I think I fixed most of the nits.

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,497 @@
...
+
+int uv__fsevents_poller_init(uv_loop_t* loop, uv__fsevents_poller_t* poller) {
+  int err;
+
+  poller->loop = loop;
+  poller->items = NULL;
+  poller->active = 0;
+  err = -ENOMEM;
+
+  poller->signal = malloc(sizeof(*poller->signal));
```
saghul: You could store the async handle itself instead of a pointer, thus avoiding this malloc, right?

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,497 @@
...
+  if (poller->signal == NULL)
+    goto signal_alloc_failed;
+
+  err = uv_async_init(loop, poller->signal, uv__fsevents_poller_signal_cb);
+  if (err != 0)
+    goto signal_init_failed;
+
+  uv__handle_unref(poller->signal);
+  poller->signal->data = poller;
```
saghul: If the handle itself was used you could use container_of instead of using the data pointer for the back reference.

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,497 @@
...
+
+int uv__fsevents_poller_signal(uv__fsevents_poller_t* poller) {
+  if (poller->active == 0)
+    return -1;
+
+  return uv_async_send(poller->signal);
+}
+
+
+void uv__fsevents_poller_destroy(uv__fsevents_poller_t* poller) {
```
saghul: Set active = 0 here?

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,497 @@
...
+
+void uv__fsevents_poller_signal_cb(uv_async_t* handle, int status) {
+  uv__cf_loop_state_t* state;
+  uv__fsevents_poller_t* poller;
+  uv__fsevents_poller_item_t* item;
+  QUEUE* q;
+  uv_fs_event_t* curr;
+
+  state = handle->loop->cf_state;
+  poller = handle->data;
```
saghul: As @bnoordhuis once told me, real men use container_of :-)

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,497 @@
...
+
+  item = calloc(1, sizeof(*item));
+  if (item == NULL) {
+    err = -ENOMEM;
+    goto item_alloc_failed;
+  }
+
+  item->handle = handle;
+
+  item->poll = malloc(sizeof(*item->poll));
```
saghul: The fs_poll handle could also be embedded in uv__fsevents_poller_item_t instead of using a pointer, then you can use container_of in he callback and avoid malloc / free.

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,497 @@
...
+  if (err != 0)
+    goto readdir_failed;
+
+  return item;
+
+readdir_failed:
+  uv_close((uv_handle_t*) item->poll, (uv_close_cb) free);
+
+alloc_failed:
+  free(item->poll);
```
saghul: In case you continue doing malloc / free: this seems to be NULL always when reaching here.

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,497 @@
...
+  int len;
+
+  item = req->data;
+  if (req->result < 0 || item == NULL)
+    goto fatal;
+
+  /* Stop poller until all stats will be loaded */
+  err = uv_fs_poll_stop(item->poll);
+  if (err != 0) {
+    req->result = err;
```
saghul: Why override the req result here?

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,497 @@
...
+
+  /* Copy all file names */
+  names = req->ptr;
+  dirname = req->path;
+  dirlen = strlen(dirname);
+  for (i = 0; i < item->count; i++) {
+    item->files[i] = calloc(1, sizeof(*item->files[i]));
+    len = strlen(names);
+    path = malloc(len + 1 + dirlen + 1);
+    if (item->files[i] == NULL || path == NULL) {
```
saghul, line 319: Where is item->files[i] freed if the calloc failed?

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,497 @@
...
+fatal:
+  if (item != NULL)
+    item->readdir_req = NULL;
+
+  /* Emit error */
+  if (req->result != UV_ECANCELED && req->result < 0)
+    uv__fsevents_poller_emit_err(item->handle, req->result);
+
+  free(req);
+  return;
```
saghul: Unneeded return?

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,497 @@
...
+      if (cmp < 0) {
+        /* File was removed */
+        i++;
+        uv__fsevents_poller_emit_change(item->handle, prev->name, UV_RENAME);
+      } else if (cmp > 0) {
+        /* File was created */
+        j++;
+        uv__fsevents_poller_emit_change(item->handle, next->name, UV_RENAME);
+      } else {
+        if (prev->stat.st_mtim.tv_sec != next->stat.st_mtim.tv_sec ||
```
saghul: Maybe you could use this one: https://github.com/joyent/libuv/blob/master/src/fs-poll.c#L194 it takes more things into account.

src/unix/fsevents.h
```diff
@@ -0,0 +1,126 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ */
+
+#ifndef UV_UNIX_FSEVENTS_H_
+#define UV_UNIX_FSEVENTS_H_
+
+#if !defined(__APPLE__)
+# error "fsevents.h should be included only on OSX"
```
saghul, line 25: Maybe rename the file to darwin-fsevents.h?

saghul:
@indutny I left a few suggestions, hope it helps.

src/unix/fsevents-poll.c
indutny: To use it below.

src/unix/fsevents-poll.c
indutny, line 319: In `item_destroy()`

src/unix/fsevents-poll.c
indutny: Oh, I forgot to remove it.

src/unix/fsevents-poll.c
saghul: But can't we use err directly? It felt a bit odd when I was reading it.

src/unix/fsevents.h
indutny, line 25: I think it could be done after merging this into.

indutny:
@saghul thank you for comments! May I ask you to take a look at it again. I've embedded some of the handles into existing structures. I'd be really glad if you'll help me in verifying that there're no races here.

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,489 @@
...
+  err = -ENOMEM;
+
+  err = uv_async_init(loop, &poller->signal, uv__fsevents_poller_signal_cb);
+  if (err != 0)
+    goto signal_init_failed;
+
+  uv__handle_unref(&poller->signal);
+  return 0;
+
+signal_init_failed:
```
saghul: You can return error from above directly now.

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,489 @@
...
+                                    int status);
+
+
+int uv__fsevents_poller_init(uv_loop_t* loop, uv__fsevents_poller_t* poller) {
+  int err;
+
+  poller->loop = loop;
+  poller->items = NULL;
+  poller->active = 0;
+  err = -ENOMEM;
```
saghul: No need to initialize it. What's the policy here? (init even if not needed?)

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,489 @@
...
+
+
+int uv__fsevents_poller_is_active(const uv__fsevents_poller_t* poller) {
+  return poller->active;
+}
+
+
+int uv__fsevents_poller_signal(uv__fsevents_poller_t* poller) {
+  if (poller->active == 0)
+    return -1;
```
saghul: When / how can this happen? I see the only call to this function is done after checking with is_active. Race?

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,489 @@
...
+void uv__fsevents_poller_on_signal_close(uv_handle_t* handle) {
+  uv__fsevents_poller_t* poller;
+
+  poller = container_of(handle, uv__fsevents_poller_t, signal);
+  poller->destroy_cb(poller);
+}
+
+
+void uv__fsevents_poller_destroy(uv__fsevents_poller_t* poller,
+                                 uv__fsevents_poller_destroy_cb cb) {
```
saghul, line 97: Unused argument.

indutny, line 97: Oh, you're right.

A bit surprising thing here, `uv_close()` callback is not invoked... Even if I'll add `uv__handle_ref()` call right before it.

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,489 @@
...
+    goto readdir_failed;
+
+  return item;
+
+readdir_failed:
+  uv_close((uv_handle_t*) &item->poll, uv__fsevents_on_poll_close);
+  item = NULL;
+
+poll_init_failed:
+  free(item);
```
saghul, line 198: Does free(NULL) complain on vlagrind?

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,489 @@
...
+  item->files = NULL;
+  item->prev_files = NULL;
+  item->count = 0;
+  item->prev_count = 0;
+  item->stats_waiting = 0;
+}
+
+
+void uv__fsevents_poller_emit_err(uv_fs_event_t* handle, int code) {
+  if (handle->event_watcher.fd != -1) {
```
saghul: uv__is_active instead?

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,489 @@
...
+    handle->cb(handle, NULL, 0, code);
+    handle->cb = uv__fsevents_poller_nop;
+  }
+}
+
+
+void uv__fsevents_poller_emit_change(uv_fs_event_t* handle,
+                                     const char* filename,
+                                     int events) {
+  if (handle->event_watcher.fd != -1)
```
saghul: Ditto.

src/unix/fsevents.h
saghul, line 25: Ok.

src/unix/fsevents.c
```diff
@@ -312,12 +314,20 @@ static void uv__fsevents_create_stream(uv_loop_t* loop, CFArrayRef paths) {
                              flags);
   assert(ref != NULL);
 
-  state = loop->cf_state;
   pFSEventStreamScheduleWithRunLoop(ref,
                                     state->loop,
                                     *pkCFRunLoopDefaultMode);
-  if (!pFSEventStreamStart(ref))
-    abort();
+  if (!pFSEventStreamStart(ref)) {
```
saghul, line 323: Perhaps a short comment on why we need to use the poller in this case would be nice :-) (I actually don't know the case in which it's needed)

src/unix/fsevents-poll.c
```diff
@@ -0,0 +1,489 @@
...
+  if (item->stats_waiting == 0)
+    uv__fsevents_poller_stats_done(item);
+
+fatal:
+  if (item != NULL)
+    item->readdir_req = NULL;
+
+  /* Emit error */
+  if (req->result != UV_ECANCELED && err < 0)
+    uv__fsevents_poller_emit_err(item->handle, err);
```
saghul, line 358: What if item is NULL here? Looks possible, because you set the data field to NULL and call uv_cancel after, so if the request couldn't be canceled you'd reach here.

saghul:
Did a second round. It's hard to tell if there are any races (check that uv_cancel thing I mention somewhere though) because it's not very clear on what thread the stuff runs, maybe a short comment would help :-)

Also, when does this poller kick in? Can it be simulated with a test?

src/unix/fsevents-poll.c
indutny, line 358: Good catch!

indutny:
Added a comment about threads, there're no way to stimulate it right now. Which is really miserable. You can only replace `!...Start()` call with `1` to see if it works for you.

indutny:
Also, I've made some changes to loop_delete code, may I ask you to take a look at it? Basically, I added code to invoke all close callbacks.

src/unix/fsevents-poll.c
indutny: That's what I do in `fsevents.c`

indutny:
Anyway, I fixed all mentioned nits.

src/unix/fsevents-poll.c
bnoordhuis, line 198: No. (Or if it does, it's a bug in valgrind.  NULL is a valid argument to free().)

src/unix/loop.c
```diff
@@ -132,7 +132,15 @@ static int uv__loop_init(uv_loop_t* loop, int default_loop) {
...
 static void uv__loop_delete(uv_loop_t* loop) {
   uv__signal_loop_cleanup(loop);
+
+  /* Platform loop may contain internal handles, ensure that close
+   * callbacks will be invoked for each of them
+   */
+  assert(loop->closing_handles == NULL);
   uv__platform_loop_delete(loop);
+  uv__run_closing_handles(loop);
+  assert(loop->closing_handles == NULL);
```
bnoordhuis, line 142: As discussed on IRC, this is not the right approach.  I'm assuming you do this to clean up the poller and the readdir req, right?

IMO, the best way to address this is by postponing closing the uv_fs_event_t handle until the poller and the readdir req have closed and finished respectively.

Right now, that's not possible but it should be reasonably straightforward to implement.  When you do, you can fix up src/fs-poll.c while you're at it.  :-)

indutny, line 142: Wait a moment, I'm doing it because of `poller->signal` handle, not `readdir`. And I can't really postpone closing uv_fs_event_t handle, because it isn't related in any way to destroying loop.

I think I just didn't get your message here :) So if what I say doesn't feel right to you - please try to describe it in details.

bnoordhuis, line 142: Okay, let me put it another way: what handles are you closing here?  If they're inferiors of uv_fs_event_t handles, then this is the wrong approach.  If they're loop-local handles, then it _might_ be acceptable (but probably not and it doesn't look like they're loop-local anyway.)

If they're inferiors, then the right approach is to postpone closing the parent uv_fs_event_t handle until the inferiors have closed.

For the record, it is the libuv user's responsibility to close all handles before calling uv_loop_delete().  There should be no need to run close callbacks here because there should be no open handles at this point.

indutny, line 142: They're not inferiors and are loop-local unreffed handles.

bnoordhuis, line 142: What about item->poll handles.  They get unref'd, right?  How do you ensure that they have have closed before uv_run() returns?

indutny, line 142: Oh, I think its here mostly by mistake.

indutny, line 142: Ah, no...

indutny, line 142: Well, anyway this code is not for it.

indutny:
Closing this PR as approach looks unviable. Decided that after conversation with @bnoordhuis on IRC.

Better way would be to let users handle it, going to open separate PR soon.

PR #960, opened 2013-10-21
windows: Removed uneeded check

saghul:
Like we do in unix. Why? I just noticed :-)


bnoordhuis:
LGTM but there's a typo in the commit log and it should be present tense (and we don't usually capitalize; not a big deal but it stands out in `git log --oneline`.)

saghul:
Ups, fixed that

Issue #446, opened 2012-06-04
uv_cwd and uv_exepath should have similar interface


piscisaureus:
Right now it's a mess.

mmalecki:
The problem with `getcwd` used by `uv_cwd` is that it returns `ERANGE` when buffer is too small, so buffer can't be smaller than `MAXPATHLEN`. We'd have to truncate it ourselves to limit string length to size supplied by user.

saghul:
How about something like this?

```
diff --git a/src/unix/core.c b/src/unix/core.c
index a843044..936d895 100644
--- a/src/unix/core.c
+++ b/src/unix/core.c
@@ -589,16 +589,14 @@ ssize_t uv__recvmsg(int fd, struct msghdr* msg, int flags) {
 }


-int uv_cwd(char* buffer, size_t size) {
-  if (buffer == NULL)
+int uv_cwd(char* buffer, size_t* size) {
+  if (buffer == NULL || size == NULL)
     return -EINVAL;

-  if (size == 0)
-    return -EINVAL;
-
-  if (getcwd(buffer, size) == NULL)
+  if (getcwd(buffer, *size) == NULL)
     return -errno;

+  *size = strlen(buffer);
   return 0;
 }
```

That would make uv_cwd have the same API as uv_exepath and others. If buffer is not big enough, well, too bad, an error is returned. Of course the Windows version also needs adjusting, but you get the idea :-)

PR #961, opened 2013-10-21
unix, windows: Expose bound pipe name

saghul:
Q: Should we also expose the name when we connect?

If so, is there any way for knowing if a pipe is a server or not in unix without adding an extra attribute? We need to know not to unlink it on close.

/cc @bnoordhuis


bnoordhuis:
I'm not sure why you would want to do that (expose the pipe name, that is.)

Also, on Windows you're now storing the pipe name twice, once as an ASCII string and once as a wide character string.

saghul:
> I'm not sure why you would want to do that (expose the pipe name, that is.)

If I have multiple pipe servers I'd like to know which is which, a la getsockname for TCP.

> Also, on Windows you're now storing the pipe name twice, once as an ASCII string and once as a wide character string.

Yep, that's intentional, since apparently the Windows code needs the wide version.

bnoordhuis:
> If I have multiple pipe servers I'd like to know which is which, a la getsockname for TCP.

Right.  So what you really want is a uv_pipe_getsockname()?

> Yep, that's intentional, since apparently the Windows code needs the wide version.

I figured that much but 'intentional' doesn't automatically imply 'preferable'. :-)

saghul:
> Right. So what you really want is a uv_pipe_getsockname()?

Yes, but since it doesn't need to be dynamically calculated I thought storing the name and accessing it directly would be ok.

I'm open to suggestions though :-)

bnoordhuis:
uv_pipe_getsockname() on UNIX could just copy out handle->path.  (Though that may be inaccurate for long paths because those get truncated to UNIX_PATH_NAME characters.  Calling getsockname() will give you the kernel's idea of the socket path.)

An important reason to prefer a function over exposing the property directly is that once a property becomes API, it also becomes ABI - its position in the struct must stay fixed for the entire lifetime of the stable branch.  Needless to say that from a maintenance perspective that can get mighty inconvenient.

saghul:
> uv_pipe_getsockname() on UNIX could just copy out handle->path. (Though that may be inaccurate for long paths because those get truncated to UNIX_PATH_NAME characters. Calling getsockname() will give you the kernel's idea of the socket path.)

I see. Considering Windows, how about this interface?

```
int uv_pipe_getsockname(uv_pipe_t* handle, char* name, size_t len)
```

On unix we can extract it with getsockname and copy it to the given buffer, and on Windows we can save the original in a private struct member and copy it.

> An important reason to prefer a function over exposing the property directly is that once a property becomes API, it also becomes ABI - its position in the struct must stay fixed for the entire lifetime of the stable branch. Needless to say that from a maintenance perspective that can get mighty inconvenient.

Got it, thanks!

bnoordhuis:
Yep, sounds good.  The handle argument to uv_pipe_getsockname() can probably be const.

saghul:
Ok, thanks for confirming. I'll work on it.

saghul:
Ok, I'm back at this. I implemented the Unix version using getsockname and the Windows verison by keeping a copy (as non-wide char).

@indutny, @piscisaureus can you guys have a look? I'm specially interested in knowing if there is a better way for doing this on Windows.

Also, in the current version the len parameter gets the written bytes as output,  but I also considered just using it as the return value. Thoughts?

piscisaureus:
@saghul I think the better way would be to use `NtQueryInformationFile` with `FileNameInformation` which stores data in a `FILE_NAME_INFORMATION` struct. See http://msdn.microsoft.com/en-us/library/windows/hardware/ff545817%28v=vs.85%29.aspx.

saghul:
I have no idea what I'm doing, but I think I did what you suggested, @piscisaureus :-) I couldn't test it, lets see if Travis brings some good news!

saghul:
Doh, it failed :-( Do you see any obvious mistakes @piscisaureus? I'll run it on my Windows machine later today.

src/win/pipe.c
```diff
@@ -1738,3 +1739,39 @@ int uv_pipe_open(uv_pipe_t* pipe, uv_file file) {
...
+  IO_STATUS_BLOCK io_status;
+  FILE_NAME_INFORMATION name_info;
+  int addrlen;
+
+  if (handle->handle == INVALID_HANDLE_VALUE) {
+    *len = 0;
+    return UV_EINVAL;
+  }
+
+  nt_status = pNtQueryInformationFile(handle->handle,
```
saghul, line 1774: This one seems to fail with code -2147483643. Any obvious reason why that might be @piscisaureus ?

saghul, line 1774: The translated error is 234, ERROR_MORE_DATA according to http://msdn.microsoft.com/en-us/library/windows/hardware/ms681382(v=vs.85).aspx

I couldn't find a reason why this happens. I tried to use an idle handle and call uv_pipe_getsockname in the callback (I thought maybe binding took a tick on Windows) but got the same result.

I'm kind of lost at this point, Windows doesn't love me :-(

piscisaureus, line 1774: The FILE_NAME_INFORMATION struct fits only one character.

So you need to allocate more space as you need it. To quote:

> If ZwQueryInformationFile fails because of a buffer overflow, drivers that implement FileNameInformation should return as many WCHAR characters of the file name as will fit in the buffer and specify the full length that is required in the FileNameLength parameter of the FILE_NAME_INFORMATION structure. You should reissue the query by using the file name length so that you can retrieve the full file name. Drivers that do not follow this pattern might require a gradual increase in length until they retrieve the full file name. For more information about working with files, see Using Files in a Driver.

piscisaureus, line 1774: I think the named pipe driver does this so you don't have to worry about the "gradual increase"  scenario.

saghul:
Ok, almost there! Thanks for the pointers @piscisaureus :-) I need one last push though. The returned filename is "\uv-test" in WCHAR form, NULL terminated. If I use wcslen I get back 17, I'm assuming that's correct. Now, when I use `uv_utf16_to_utf8` it returns 33, so as a result the buffer contains the name plus some garbage. What am I doing wrong here?

Side question: to create the pipe we had to use that \pipe... thing, but that doesn't seem to be returned by this function, should we prepend that prefix?

Thanks!

piscisaureus:
@saghul 

Yes, you'll need to prepend with "\\.\pipe". I think "\\?\pipe" would be even more correct.
The name is not null-terminated, but windows tells you the exact length of the name in the FileNameLength field. 
I would suggest to use WideCharToMultiByte directly - I tried in the past to eradicate all uses of uv_ut16_to_utf8 because it has issues. Then add one byte to the buffer length and null-terminate it yourself

sam-github:
Unless there is some compelling reason, I'd like it if the output of this could be passed in to the init function, for unix and for windows, so I'd vote for prepending `\\pipe\`.

piscisaureus:
@sam-github you mean `\\?\pipe` I suppose ;)

saghul:
Hi @sam-github thanks for the feedback!

> You don't document the behaviour when the name is too long to fit in buf, which is to write a NON null terminated string into buf, which is a fairly cruddy API, its nice to have string output actually be a valid string, always, when return is success. I know some ancient unix system calls don't guarantee nul terminated strings, but libuv should be better than that.

I'll make sure they are null terminated.

> Also, I don't see any way with this API to determine that the provided buffer wasn't long enough for the path name. You can tell this with getsockname(), if it returns an addrlen longer than the address struct you passed in, the addr was truncated.
> 
> If there is an existing libuv API convention for string output buffers, using that convention here would be good. If not, I like returning EINVAL or some-such, and have *len be the required size.

Right now I copy as much as I can, but indeed, returning an error would make sense here, I'll go with ENOBUFS.

saghul:
> For non-abstract sockets, you end up memcpy()ing strlen() of data out... so output name is not a null terminated string? That seems bad. probably you need a +1 on the strlen() to get the nul.

I'll null it myself anyway.

> Except, wouldn't the first branch work for non-abstract sockets? And doesn't it include the nul since the addrlen returned by getsockname() probably does?

You mean here? https://github.com/joyent/libuv/pull/961/files#diff-1dcc314fa3f88181e7dcad56df613e2bR228 Non-abstract sockets don't start with a null byte.

Also note that for abstract sockets we cannot null-terminate the string, since null is allowed as part of the name.

saghul:
> Unless there is some compelling reason, I'd like it if the output of this could be passed in to the init function, for unix and for windows, so I'd vote for prepending \pipe. 

Yes, that's the idea. Won't work for abstract sockets though.

sam-github:
>  Won't work for abstract sockets though.
> 
> Also note that for abstract sockets we cannot null-terminate the string, since null is allowed as part of the name.

That's OK, the addrlen calculation you use will include a terminating NUL if the path has one. It probably always will. All the abstracts I've heard of are all normal, nul-terminated path strings, with a NUL prepended. Either way, your API does the right thing for abstracts, IMO.

> Won't work for abstract sockets though

You mean because uv_pipe_bind and uv_pipe_connect take only null terminated strings? That's unfortunate, it would be nice if they accepted `const char* name, size_t len` (where if len == 0, it defaults to `strlen(name)+1`. Maybe a new API is needed? It seems that this would prevent node.js from connecting to an abstract unix socket without constructing the socket yourself and passing the descriptor to uv_pipe_init (but I haven't tried, I might be missing something).

I'm about to make some fifo/unix socket-using node code I have portable to windows by writing code in javascript to "fixup" the name on windows by prepending the magic windows pipe signifier... It's a good thing @piscisaureus reminded me what it is! ;-)

Why doesn't libuv handle this? I'd rather the libuv APIs "auto prepended" the pipe thingy on windows during connect/bind, than that the windows getsockname for pipes auto-prepended it on return.

bnoordhuis:
> You mean because uv_pipe_bind and uv_pipe_connect take only null terminated strings?

Abstract sockets are a Linux-ism and libuv is not big on exposing platform-specific functionality in the API.  If you want that functionality, the normal approach is to create the socket yourself and pass that to uv_pipe_open().

saghul:
Agreed with @bnoordhuis.

I talked with @piscisaureus about the prepending thing, I'll handle that as a separated PR. This one will return the prefix, so you can pass the result of `uv_pipe_getsockname` to connect or bind.

saghul:
Ok, I think I got all the bases covered now. Test pass on my Windows VM and Linux. Give me some good news Jenkins!

sam-github:
Docs, please:

https://github.com/sam-github/libuv/commit/7850acc5090d309b2fce089866fecb8ce04ed473

txdv:
Why not a callback based allocation?
So the caller knows exactly how much to allocate. It is easier for bindings that way while the C API will be a little bit harder to use.

saghul:
> Docs, please:
> 
> sam-github@7850acc

I applied your patch, thanks! I still left the manual null termination there, just in case a 108 byte long name was used and a longer buffer is supplied, in which case I'm not sure what would happen, so I'd rather "manually" null-terminate it, just in case.

saghul:
Rebased and squashed all commits into one.

saghul:
> Why not a callback based allocation?
> So the caller knows exactly how much to allocate. It is easier for bindings that way while the C API will be a little bit harder to use.

Other similar APIs (uv_tcp_getsockname / peername) already require you to preallocate the space, I don't think we should do it any different here.

What problems do you foresee with this approach? I don't think we can please all binding writers, myself included.

saghul:
Reviewers (@indutny, @piscisaureus and @sam-github) please have one (hopefully) last look, thanks!

txdv:
I am saying it again, please do not ignore me, why not, like the people did in libgit2 to provide better interoperability with environments which have their own resource management, add an allocation callback to the function so the user doesn't have to allocate an arbitrary big buffer.

``` c
char *(*uv_pipe_getsockname_callback)(int size);
int uv_pipe_getsockname(const uv_pipe_t* handle, uv_pipe_getsockname_callback callback, char **buf)
```

libuv calls that callback with the exact size of the name, the user provides a method to allocate it (you still can just provide malloc in C) and you go girl!

It would even look ok in C:

``` c
char *buf;
uv_pipe_getsockname(handle, (uv_pipe_getsockname_callback)malloc, &buf);
// do something with buf
free(buf)
```

No need for allocating 1024 or 512 or whatever arbitrary power of 2 the user prefers. 

You want a 0 terminated string? create an allocating function which creates a buffer of the requested size + 1 and put a 0 in the end.

``` c
char *malloc_terminate(int size)
{
    char *buf = malloc(size + 1);
    buf[size] = 0;
    return buf;
}
```

No need to copy stuff around, immediate allocation of a buffer which just fits right

saghul:
> I am saying it again, please do not ignore me, why not, like the people did in libgit2 to provide better interoperability with environments which have their own resource management, add an allocation callback to the function so the user doesn't have to allocate an arbitrary big buffer.

Hi Andrius, don't get me wrong, I'm definitely open to suggestions! :-) This was my first thought when it came to implement it, but I'm open to change it. It'd be nice if other more experienced devs can comment as well. @piscisaureus @indutny, @bnoordhuis thoughts?

sam-github:
@txdv Why would you force the buffer to be allocated? Its not arbitrarily large (the sockaddr structure is fixed size on unix, UNIX_PATH_MAX is 108, see `unix(7)`), and it isn't necessarily true that the caller wants to manage storage for it. Stack space is free and cleans up after itself.

My API preference would be to allow name to be null, in which case `*len` would have the required size written to it (or some such variant, like returning the size that would be needed). This would allow you to poll for required size, allocate, and copy in, and also allow stack allocation for those who want a temporary.

txdv:
It wouldn't be impossible to pass a reference of the stack, it would just require more code.

``` c
char *buffer;
int* size;
char *allocator(int s)
{
  if (size != 0) {
    *size = s;
  }
  return buffer;
}

void set_allocator(char *buf, int *s)
{
  size = s;
  buffer = buf;
}

...
int main() {
  set_allocator(&b, &s);
  char b[UNIX_PATH_MAX];
  int s;
  uv_pipe_getsockname(handle, (uv_pipe_getsockname_callback)allocator, &s);
  // s contains length
  // b contains path
}
```

While having the current API would make it shorter, it would disallow to create buffers with the specific size requested. One could create a helper function which does nothing but wrap my proposed API and expose the already existing API.

saghul:
> My API preference would be to allow name to be null, in which case *len would have the required size written to it (or some such variant, like returning the size that would be needed). This would allow you to poll for required size, allocate, and copy in, and also allow stack allocation for those who want a temporary.

I like this idea. Sounds similar to how `WideCharToMultiByte` behaves: https://github.com/joyent/libuv/pull/961/files#diff-7be27b71e7f7a6d2c40c447747db1231R1813

txdv:
Wouldn't that result in calling the same functionality twice?

sam-github:
I just reviewed uv.h. I can find functions that return strings such as
this (uv_get_process_title, uv_inet_ntop, uv_cwd, uv_exepath, ...),
but no examples of them using a malloc callback to allocate the output
space, they all require the caller to make sufficient space available.

Of these, `uv_exepath(char* buffer, size_t* size)` has an API and
result almost identical to this new API (in that the output is also a
path).

I suggest that for this function to behave identically to uv_exepath
would be least surprising.

piscisaureus:
See also https://github.com/joyent/libuv/issues/446.

IMO, there are 2 reasonable interfaces:
1. Libuv malloc's memory for the user. This doesn't make custom allocator geeks happy but it works fine for most of us. Optionally we could have a specific free function, like `uv_free_addrinfo`.
2. The user provides a buffer. However it must be possible to retrieve the required buffer size, I would suggest by setting the buffer argument to null. So that'd become:
   `ssize_t uv_pipe_name(uv_pipe_t* pipe, void* buf, size_t length);`
   - if `buf == NULL`, return the length of the pipe name including terminating zero
   - if `buf != NULL`, copy the pipe name into the buffer + zero-terminate
     - return the length of the pipe name including final '\0;
     - but if the buffer is too small, return -ENOBUFS
   - if any error occurrs, return -errno.

But in any case, libuv should move to be consistent. I don't really care which one it's going to be.

Also I think that if the pipe has no name we should return the empty string.

saghul:
Thanks for your comments @sam-github and @piscisaureus. Here is my proposal then:
- We go with what is consistent _now_ and merge this PR. Personally, I like how `uv_exepath` does it, that is, returning 0 but setting the input size argument to the amount of written bytes. If there is no space, `ENOBUFS`. That's how it's currently implemented.
- We discuss how to make this consistent across libuv in its entirety (the mailing list is probably a better place for it). FWIW, I like the idea of setting the required size if NULL is passed as a buffer.

@piscisaureus I think I took care of the empty pipe name.

saghul:
Rebased and force pushed. @indutny can I ask you to please review this as well? I'd like to move this forward to apply the same idea in a couple if places. I'll also check uv_exepath and uv_cwd.

src/unix/pipe.c
```diff
@@ -212,5 +212,40 @@ void uv_pipe_connect(uv_connect_t* req,
...
 
 
+int uv_pipe_getsockname(const uv_pipe_t* handle, char* buf, size_t* len) {
+  struct sockaddr_un sa;
+  socklen_t addrlen;
+  int err;
+
+  addrlen = sizeof sa;
+  memset(&sa, 0, addrlen);
+  err = getsockname(uv__stream_fd(handle), (struct sockaddr*)&sa, &addrlen);
```
indutny: `(struct sockaddr*) &sa`

src/unix/pipe.c
```diff
@@ -212,5 +212,40 @@ void uv_pipe_connect(uv_connect_t* req,
...
+  if (err < 0) {
+    *len = 0;
+    return -errno;
+  }
+
+  if (sa.sun_path[0] == 0) {
```
indutny: Let's remove braces here.

src/unix/pipe.c
```diff
@@ -212,5 +212,40 @@ void uv_pipe_connect(uv_connect_t* req,
 }
 
 
+int uv_pipe_getsockname(const uv_pipe_t* handle, char* buf, size_t* len) {
+  struct sockaddr_un sa;
+  socklen_t addrlen;
+  int err;
+
+  addrlen = sizeof sa;
```
indutny: `sizeof(sa)`

src/unix/pipe.c
```diff
@@ -212,5 +212,40 @@ void uv_pipe_connect(uv_connect_t* req,
...
+  }
+
+  if (sa.sun_path[0] == 0) {
+    /* Linux abstract namespace */
+    addrlen -= offsetof(struct sockaddr_un, sun_path);
+  } else {
+    addrlen = strlen(sa.sun_path);
+  }
+
+  if (addrlen+1 > *len) {
```
indutny: `addrlen + 1`

src/unix/pipe.c
```diff
@@ -212,5 +212,40 @@ void uv_pipe_connect(uv_connect_t* req,
...
+  memset(&sa, 0, addrlen);
+  err = getsockname(uv__stream_fd(handle), (struct sockaddr*)&sa, &addrlen);
+  if (err < 0) {
+    *len = 0;
+    return -errno;
+  }
+
+  if (sa.sun_path[0] == 0) {
+    /* Linux abstract namespace */
+    addrlen -= offsetof(struct sockaddr_un, sun_path);
```
indutny, line 230: I think you wanted to do `sizeof(*sa) - offsetof(...)`

src/win/pipe.c
```diff
@@ -44,6 +44,10 @@
 
 static const int default_pending_pipe_instances = 4;
 
+/* Pipe prefix */
+static char pipe_prefix[] = "\\\\.\\pipe";
+static const int pipe_prefix_len = sizeof(pipe_prefix)-1;
```
indutny: `sizeof(pipe_prefix) - 1`

src/unix/pipe.c
indutny, line 230: Nvm

indutny:
Seems to be looking good, if tests are passing ;) But please fix style issues first.

saghul:
Fixed style issues, @indutny. Could you please reply to this comment? https://github.com/joyent/libuv/pull/961#issuecomment-31509287

saghul:
Foce pushed, all style nits should be fixed now.

saghul:
This is ready to land IMHO. Can I get a LGTM @piscisaureus, @indutny ? :-)

saghul:
@sam-github I just force-pushed addressing the issues you rightly pointed out. Care to take a (hopefully final) look?

src/win/pipe.c
```diff
@@ -44,6 +44,10 @@
 
 static const int default_pending_pipe_instances = 4;
 
+/* Pipe prefix */
+static char pipe_prefix[] = "\\\\.\\pipe";
```
piscisaureus: The prefix should really be "\\?\". Trust me... >:-)

saghul: Only that ? So a valid pipe name would be "\?\uberpipe" ?

saghul:
```
\\?\uberpipe
```

, that is.

retep998: See http://msdn.microsoft.com/en-us/library/gg465305

retep998: As well as http://msdn.microsoft.com/en-us/library/aa365247(VS.85).aspx

saghul: I trust you, @piscisaureus :-) I'll update this PR and change the test pipe name since I'm at it. Also thanks @retep998 for the links!

saghul:
Force pushed with the new pipe prefix. Should I also change it here: https://github.com/joyent/libuv/blob/master/src/win/pipe.c#L73 @piscisaureus?

saghul:
Pushed another update, making the change I suggested in my last comment.

sam-github:
LGTM

saghul:
Thanks for the help everyone, landed in master!

PR #964, opened 2013-10-23
fsevents: increase stack size for OSX 10.9

indutny:
Otherwise it fails with `EXC_BAD_ACCESS`.

/cc @bnoordhuis @saghul.


saghul:
No objection here. I guess fsevents has more stuff in it on 10.9. I have 10.9 around but haven't tested with it yet. Either way, LGTM.

indutny:
Thanks, landed in 0fdd99f.

PR #965, opened 2013-10-23
fsevents: report errors to user

indutny:
/cc @bnoordhuis


src/unix/fsevents.c
```diff
@@ -203,8 +233,10 @@ static void uv__fsevents_event_cb(ConstFSEventStreamRef streamRef,
   paths = eventPaths;
 
   /* For each handle */
+  uv_mutex_lock(&state->fsevent_mutex);
```
indutny, line 238: This probably should go in separate commit and backported to v0.10.

indutny, line 238: Ah, no it shouldn't.

src/unix/fsevents.c
```diff
@@ -169,15 +174,39 @@ static void uv__fsevents_cb(uv_async_t* cb, int status) {
...
+
+  uv_mutex_lock(&handle->cf_mutex);
+  curr = handle->cf_event;
+  if (curr == NULL) {
+    handle->cf_event = event;
+  } else {
+    while (curr->next != NULL)
+      curr = curr->next;
+    event->next = event;
+  }
```
bnoordhuis: You're not a two star programmer, are you, Fedor? :-)

indutny: What do you mean by that?

src/unix/fsevents.c
```diff
@@ -169,15 +174,39 @@ static void uv__fsevents_cb(uv_async_t* cb, int status) {
...
+
+  uv_mutex_lock(&handle->cf_mutex);
+  curr = handle->cf_event;
+  if (curr == NULL) {
+    handle->cf_event = event;
+  } else {
+    while (curr->next != NULL)
+      curr = curr->next;
+    curr->next = event;
+  }
```
bnoordhuis: You could rewrite this as:

``` c
uv__fsevents_event_t** next = &handle->cf_event;
while (*next != NULL) next = &(*next)->next;
*next = event;
```

src/unix/fsevents.c
```diff
@@ -262,23 +290,20 @@ static void uv__fsevents_event_cb(ConstFSEventStreamRef streamRef,
 
       if (tail != NULL)
         tail->next = event;
+      else
+        head = event;
```
bnoordhuis: Maybe swap the clauses around (and change the test from != to ==, of course.)

src/unix/fsevents.c
```diff
@@ -398,12 +436,36 @@ static void uv__fsevents_reschedule(uv_fs_event_t* handle) {
...
+      goto final;
+    }
+    err = uv__fsevents_create_stream(handle->loop, cf_paths);
   }
 
 final:
+  /* Deallocate all paths in case of failure */
+  if (err != 0) {
+    /* CFArray takes ownership of both strings and original C-array */
+    if (cf_paths != NULL) {
```
bnoordhuis: Ditto.  In general, if you have an if expression with two clauses, try to write it in a way that doesn't use negation.  It's easier on the human brain.

(EDIT: It's not the worst thing if you do but it makes it easier on your fellow maintainers.)

src/unix/fsevents.c
```diff
@@ -169,15 +174,35 @@ static void uv__fsevents_cb(uv_async_t* cb, int status) {
   handle = cb->data;
 
   UV__FSEVENTS_PROCESS(handle, {
-    if (handle->event_watcher.fd != -1)
-      handle->cb(handle, event->path[0] ? event->path : NULL, event->events, 0);
+    if (uv__is_active(handle)) {
```
bnoordhuis: This could probably use a comment why it's checking whether the handle is active.

src/unix/fsevents.c
```diff
@@ -169,15 +174,35 @@ static void uv__fsevents_cb(uv_async_t* cb, int status) {
...
 
+/* Runs in CF thread, pushed event into handle's event list */
+static void uv__fsevents_push_event(uv_fs_event_t* handle,
+                                    uv__fsevents_event_t* event) {
+  uv__fsevents_event_t** next;
+
+  uv_mutex_lock(&handle->cf_mutex);
+  next = (uv__fsevents_event_t**) &handle->cf_event;
+  while (*next != NULL)
+    next = (uv__fsevents_event_t**) &(*next)->next;
```
bnoordhuis: Why don't you change the type of uv__fsevents_event_t.next?

bnoordhuis: Love the two star programming, by the way.

bnoordhuis: Looking at this again, wouldn't it make sense to use a queue here rather than a linked list?  That way insertion is O(1) rather than O(n).

indutny: ABI compatibility.

src/unix/fsevents.c
```diff
@@ -260,25 +288,22 @@ static void uv__fsevents_event_cb(ConstFSEventStreamRef streamRef,
...
-    if (tail != NULL) {
-      uv_mutex_lock(&handle->cf_mutex);
-      tail->next = handle->cf_event;
-      handle->cf_event = tail;
-      uv_mutex_unlock(&handle->cf_mutex);
-
-      uv_async_send(handle->cf_cb);
-    }
+    if (head != NULL)
+      uv__fsevents_push_event(handle, head);
```
bnoordhuis: On the subject of two star programming...

``` diff
diff --git a/src/unix/fsevents.c b/src/unix/fsevents.c
index a42cef4..971da64 100644
--- a/src/unix/fsevents.c
+++ b/src/unix/fsevents.c
@@ -92,7 +92,7 @@ struct uv__cf_loop_signal_s {
 struct uv__fsevents_event_s {
   int events;
   int status;
-  void* next;
+  struct uv__fsevents_event_s* next;
   char path[1];
 };

@@ -220,8 +220,8 @@ static void uv__fsevents_event_cb(ConstFSEventStreamRef streamRef,
   uv_loop_t* loop;
   uv__cf_loop_state_t* state;
   uv__fsevents_event_t* event;
+  uv__fsevents_event_t** next;
   uv__fsevents_event_t* head;
-  uv__fsevents_event_t* tail;

   loop = info;
   state = loop->cf_state;
@@ -232,8 +232,7 @@ static void uv__fsevents_event_cb(ConstFSEventStreamRef streamRef,
   uv_mutex_lock(&state->fsevent_mutex);
   QUEUE_FOREACH(q, &state->fsevent_handles) {
     handle = QUEUE_DATA(q, uv_fs_event_t, cf_member);
-    head = NULL;
-    tail = NULL;
+    next = &head;

     /* Process and filter out events */
     for (i = 0; i < numEvents; i++) {
@@ -288,14 +287,11 @@ static void uv__fsevents_event_cb(ConstFSEventStreamRef streamRef,
       else
         event->events = UV_RENAME;

-      if (tail == NULL)
-        head = event;
-      else
-        tail->next = event;
-      tail = event;
+      *next = event;
+      next = &(*next)->next;
     }

-    if (head != NULL)
+    if (*next != head)
       uv__fsevents_push_event(handle, head);
   }
   uv_mutex_unlock(&state->fsevent_mutex);
```

I didn't test it deeply but that should work, right?

src/unix/fsevents.c
bnoordhuis: You're targeting master, right?  No need to worry about ABI compatibility, we don't make any guarantees for dev releases, only for the stable branches.

src/unix/fsevents.c
```diff
@@ -389,21 +429,48 @@ static void uv__fsevents_reschedule(uv_fs_event_t* handle) {
...
+      pCFRelease(cf_paths);
+    }
+
+    /* Broadcast error to all handles */
+    error_event.status = err;
+
+    uv_mutex_lock(&state->fsevent_mutex);
+    QUEUE_FOREACH(q, &state->fsevent_handles) {
+      curr = QUEUE_DATA(q, uv_fs_event_t, cf_member);
+      uv__fsevents_push_event(curr, &error_event);
```
bnoordhuis: error_event is static.  What if multiple threads hit on an error?  Is this thread-safe?  It doesn't look like it.

src/unix/fsevents.c
indutny: Won't it be backported to v0.10?

indutny:
Fixed almost all nits (except ABI), thanks for two-star tips!

src/unix/fsevents.c
bnoordhuis: It's a change in behavior so I'd say that's unlikely.

indutny: Lets keep it ABI compatible anyway, I like that it eats less memory for an event ;)

src/unix/fsevents.c
```diff
@@ -168,16 +175,40 @@ static void uv__fsevents_cb(uv_async_t* cb, int status) {
...
+                 event->path[0] ? event->path : NULL,
+                 event->events,
+                 event->status);
+    }
   });
 
-  if (!uv__is_closing(handle) && handle->event_watcher.fd == -1)
+  /* If, after callbacks, the handle is closing - let it close */
+  if (!uv__is_closing(handle) && !uv__is_active(handle))
     uv__fsevents_close(handle);
```
bnoordhuis: I'm not sure why it's necessary to call `uv__fsevents_close()` here.  Is this is a roundabout way of stopping handle->event_watcher?

src/unix/fsevents.c
bnoordhuis: I still think it should be a queue.  It's unlikely to land in v0.10 so ABI compatibility is not an issue and I don't want to have to worry about potential denial-of-services because of the O(n) insertion behavior.

By the way, you can drop the cast.

bnoordhuis:
LGTM if you fix the list insertion.

indutny:
@bnoordhuis Check it out for the last time, I've started using queues.

src/unix/fsevents.c
```diff
@@ -143,22 +143,36 @@ static void (*pFSEventStreamScheduleWithRunLoop)(FSEventStreamRef,
 
 #define UV__FSEVENTS_PROCESS(handle, block)                                   \
     do {                                                                      \
+      QUEUE events;                                                           \
+      QUEUE* q;                                                               \
       uv__fsevents_event_t* event;                                            \
-      uv__fsevents_event_t* next;                                             \
+      uv__cf_loop_state_t* state;                                             \
+      int err;                                                                \
+      state = (handle)->loop->cf_state;                                       \
```
bnoordhuis, line 151: Assigned but unused.

src/unix/fsevents.c
```diff
@@ -168,13 +182,31 @@ static void uv__fsevents_cb(uv_async_t* cb, int status) {
 
   handle = cb->data;
 
+  /* NOTE: Checking uv__is_active() is required here, because handle
+   * callback may close handle and invoking it after it will lead to
+   * incorrect behaviour */
```
bnoordhuis, line 187: Right comment, wrong place?

src/unix/fsevents.c
```diff
@@ -168,13 +182,31 @@ static void uv__fsevents_cb(uv_async_t* cb, int status) {
...
-  if (!uv__is_closing(handle) && handle->event_watcher.fd == -1)
-    uv__fsevents_close(handle);
+  /* Concatenate two queues */
+  if (events != NULL)
+    QUEUE_ADD(&handle->cf_events, events);
+
+  /* Propagate error */
+  if (err != 0)
+    handle->cf_error = err;
+  uv_mutex_unlock(&handle->cf_mutex);
```
bnoordhuis, line 207: Maybe add an assert at the top that verifies that `!(events == NULL && err == 0)`?

src/unix/fsevents.c
```diff
@@ -353,9 +383,15 @@ static void uv__fsevents_reschedule(uv_fs_event_t* handle) {
...
   int i;
+  int err;
   int path_count;
 
   state = handle->loop->cf_state;
+  paths = NULL;
+  cf_paths = NULL;
+  err = 0;
+  /* NOTE: `i` is used in deallocation loop below */
+  i = 0;
```
bnoordhuis, line 394: Apropos nothing, but ISTM that i, path_count and state->fsevent_handle_count should be unsigned rather than signed.  There is no legitimate reason for those last two to ever be < 0, right?

bnoordhuis:
Some final comments but nothing substantial.  It would be nice to have a test case that starts file watchers from multiple threads and event loops.

indutny:
Thanks, landed in 43bef41. I'll work on test soon.

PR #967, opened 2013-10-23
unix: close signal pipe fds when last loop is done

bnoordhuis:
Fixes #875.

Adds a leaked file descriptor check to the test runner that scans for open file descriptors when the test is done.  Most of the other commits are fix-ups that plug fd leaks in the tests.  I just noticed that the commit log that adds the check is a bit terse, I'll fix that up before landing this.

@indutny or @saghul, feel free to review it if you like.

### DIFF
--- a/test/test-tcp-close.c
+++ b/test/test-tcp-close.c
@@ -131,6 +131,9 @@ TEST_IMPL(tcp_close) {
   ASSERT(write_cb_called == NUM_WRITE_REQS);
   ASSERT(close_cb_called == 1);
 
+  uv_close((uv_handle_t*) &tcp_server, NULL);
+  ASSERT(0 == uv_run(loop, UV_RUN_DEFAULT));
+
   MAKE_VALGRIND_HAPPY();
   return 0;
 }
--- a/src/unix/internal.h
+++ b/src/unix/internal.h
@@ -179,11 +179,13 @@ void uv__run_timers(uv_loop_t* loop);
 int uv__next_timeout(const uv_loop_t* loop);
 
 /* signal */
+int uv__signal_global_init(void);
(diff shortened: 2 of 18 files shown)


test/test-tcp-close.c
```diff
@@ -131,6 +131,9 @@ static void start_server(uv_loop_t* loop, uv_tcp_t* handle) {
   ASSERT(write_cb_called == NUM_WRITE_REQS);
   ASSERT(close_cb_called == 1);
 
+  uv_close((uv_handle_t*) &tcp_server, NULL);
+  ASSERT(0 == uv_run(loop, UV_RUN_NOWAIT));
```
saghul: Any specific reason to use noawit here instead of default like in the other cases?

saghul:
LGTM. I guess those test failures are just spurious, right?

test/test-tcp-close.c
bnoordhuis: Hum, that's probably auto-complete fail.  Thanks for pointing that out.

bnoordhuis:
> I guess those test failures are just spurious, right?

They're probably legitimate insofar that I suspect that Jenkins is doing something to the test environment.  Guess I'll have to dig into that.

indutny:
Few nits, but generally LGTM.

indutny:
@bnoordhuis sorry for a delay, may I ask you to rebase it off the current master? I tried doing it, but got 17 test failures on osx...

bnoordhuis:
I don't really have an interest in libuv anymore.  I'll close the PR but feel free to pick up where I left off.

indutny:
@bnoordhuis sure, though, it is a bit sad, you know.

PR #968, opened 2013-10-23
unix: fix reopened fd bug

### LINKED ISSUE #826: unix: Assertion `events == 1' failed
node: ../deps/uv/src/unix/stream.c:494: uv__server_io: Assertion `events == 1' failed.

Im on node 0.10.10 and therefore livuv 0.10.10.
This happens when node-phantom tries to spawn phantom's process.
On Debian 7.0

indutny:
When fd is closed and new one (with the same number) is opened inside
kqueue/epoll/port loop's callback - stale events might invoke callbacks
on wrong watchers.

Check if watcher was changed after invocation and invalidate all events
with the same fd.

fix #826

/cc @bnoordhuis


indutny:
Wrong branch, reopening.

PR #969, opened 2013-10-24
windows: fix access violation on file_info->FileName

huxingyi:
file_info->FileName is not null terminated. 
this will crash the application(access violation) in _snwprintf.


piscisaureus:
Landed in da32344

PR #970, opened 2013-10-25
unix: update events from pevents between polls

indutny:
Watchers could be stopped between two `kevent()`/`epoll_wait()` calls
(which may happen in the same loop in `uv__io_poll()`), in such cases
`watcher->events` could be stale and won't be updated to
`watcher->pevents`.

Additionally, since epoll has only one entry per fd, whole update loop
should be invoked to keep events up-to-date.

/cc @bnoordhuis


src/unix/kqueue.c
```diff
@@ -205,7 +205,7 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
       }
 
       if (ev->filter == EVFILT_WRITE) {
-        if (w->events & UV__POLLOUT) {
+        if (w->pevents & UV__POLLOUT) {
```
bnoordhuis, line 208: Ai... that seems like a pretty obvious bug in hindsight.

src/unix/linux-core.c
```diff
@@ -132,8 +118,9 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
 
     if (w->events == 0)
       op = UV__EPOLL_CTL_ADD;
-    else
+    else if (w->pevents != w->events)
       op = UV__EPOLL_CTL_MOD;
+    continue;
```
bnoordhuis: Forgive me the lack of eloquence but... wut?

indutny: forgot `else`

indutny: Just added `else`.

src/unix/linux-core.c
```diff
@@ -132,8 +118,10 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
 
     if (w->events == 0)
       op = UV__EPOLL_CTL_ADD;
-    else
+    else if (w->pevents != w->events)
```
bnoordhuis: If w->events != 0, then this check must always be true, else the watcher wouldn't be in queue in the first place.

bnoordhuis:
Could really do with a regression test and "but it's hard!" it's not a valid excuse.

I imagine this change could introduce a performance regression because you're calling epoll_ctl() more often now, potentially a lot more often.  What do the before/after benchmark numbers look like?

indutny:
I agree, there would be a regression. Let me rethink the approach.

src/unix/linux-core.c
```diff
@@ -208,7 +209,14 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
         continue;
       }
 
-      w->cb(loop, w, pe->events);
+      /*
+       * Give users only events they're interested in. Prevents spurious
+       * callbacks when previous callback invokation in this loop has stopped
```
bnoordhuis, line 214: s/invokation/invocation/

src/unix/linux-core.c
```diff
@@ -208,7 +209,14 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
...
+       * current watcher.
```
bnoordhuis, line 215: 'the current watcher'

bnoordhuis, line 215: Also, that comment isn't quite correct, is it?  It filters out events that the watcher has stopped watching.  The watcher itself is still active.

bnoordhuis:
LGTM apart from the comment.  A regression test would be good to have though.

indutny:
Thanks, landed in 29fdb34.

bnoordhuis:
Next up the is the regression test, right?

indutny:
Right, working on it.

PR #971, opened 2013-10-25
unix: fix reopened fd bug

### LINKED ISSUE #826: unix: Assertion `events == 1' failed

indutny:
When fd is closed and new one (with the same number) is opened inside
kqueue/epoll/port loop's callback - stale events might invoke callbacks
on wrong watchers.

Check if watcher was changed after invocation and invalidate all events
with the same fd.

fix #826

/cc @bnoordhuis


bnoordhuis:
Here's a comment for you: where's the test case? :-)

indutny:
Haha, actually this one would be quite complicated. But I'll do my best.

indutny:
@bnoordhuis added test.

src/unix/core.c
```diff
@@ -595,20 +595,33 @@ static unsigned int next_power_of_two(unsigned int val) {
...
+  void* fake_watcher_count;
   unsigned int nwatchers;
   unsigned int i;
 
   if (len <= loop->nwatchers)
     return;
 
-  nwatchers = next_power_of_two(len);
-  watchers = realloc(loop->watchers, nwatchers * sizeof(loop->watchers[0]));
+  nwatchers = next_power_of_two(len + 2) - 2;
```
bnoordhuis, line 606: This logic is unsound.  Say that len == 14.  In that case, next_power_of_two(14 + 2) - 2 is still 14.

src/unix/darwin.c
```diff
@@ -102,6 +102,27 @@ void uv__platform_loop_delete(uv_loop_t* loop) {
 }
 
 
+void uv__platform_invalidate_fd(uv_loop_t* loop, int fd) {
+  struct kevent* events;
+  intptr_t i;
+  intptr_t nfds;
+
+  /* No watchers - no fake watcher */
+  if (loop->watchers == NULL)
```
bnoordhuis: Is that actually possible?  If not, you could turn this into an assert.

src/unix/darwin.c
```diff
@@ -102,6 +102,27 @@ void uv__platform_loop_delete(uv_loop_t* loop) {
...
+    return;
+
+  events = (struct kevent*) loop->watchers[loop->nwatchers];
+  nfds = (intptr_t) loop->watchers[loop->nwatchers + 1];
```
bnoordhuis: Could be uintptr_t, it's never < 0.

src/unix/darwin.c
```diff
@@ -102,6 +102,27 @@ void uv__platform_loop_delete(uv_loop_t* loop) {
...
+  if (events == NULL)
+    return;
+
+  /* Invalidate events with same file descriptor */
+  for (i = 0; i < nfds; i++)
+    if ((int) events[i].ident == fd)
```
bnoordhuis, line 119: I'd cast fd to uintptr_t here.  Casts from unsigned to signed types of equal or smaller width are implementation defined.

src/unix/linux-core.c
```diff
@@ -97,6 +97,27 @@ void uv__platform_loop_delete(uv_loop_t* loop) {
...
+void uv__platform_invalidate_fd(uv_loop_t* loop, int fd) {
+  struct uv__epoll_event* events;
+  intptr_t i;
+  intptr_t nfds;
+
+  /* No watchers - no fake watcher */
+  if (loop->watchers == NULL)
+    return;
+
+  events = (struct kevent*) loop->watchers[loop->nwatchers];
```
bnoordhuis: I know for a fact that you didn't test this code...

test/test-tcp-close-accept.c
```diff
@@ -0,0 +1,175 @@
...
+    ASSERT(status != 0);
+
+    /* Close check and incoming[0], time to finish test */
+    uv_close((uv_handle_t*) &tcp_incoming[0], close_cb);
+    uv_close((uv_handle_t*) &tcp_check, close_cb);
+    return;
+  }
+
+  ASSERT(status == 0);
+  i = (unsigned int) ((intptr_t) req - (intptr_t) connect_reqs) /  sizeof(*req);
```
bnoordhuis: Um, how about just `i = req - connect_reqs`?

test/test-tcp-close-accept.c
```diff
@@ -0,0 +1,175 @@
...
+}
+
+static void write_cb(uv_write_t* req, int status) {
+  ASSERT(status == 0);
+  write_cb_called++;
+}
+
+static void connect_cb(uv_connect_t* req, int status) {
+  unsigned int i;
+  uv_buf_t buf;
```
bnoordhuis, line 52: Shadows another variable.  Not terrible, just something that you should try to avoid if possible.

test/test-tcp-close-accept.c
```diff
@@ -0,0 +1,175 @@
...
+  i = (unsigned int) ((intptr_t) req - (intptr_t) connect_reqs) /  sizeof(*req);
+  ASSERT(i < ARRAY_SIZE(connect_reqs));
+
+  buf = uv_buf_init("x", 1);
+  outgoing = (uv_stream_t*) &tcp_outgoing[i];
+  ASSERT(0 == uv_write(&write_reqs[i], outgoing, &buf, 1, write_cb));
+}
+
+static uv_buf_t alloc_cb(uv_handle_t* handle, size_t suggested_size) {
+  return uv_buf_init(buf, sizeof(buf));
```
bnoordhuis, line 77: You could move buf into this function.

test/test-tcp-close-accept.c
```diff
@@ -0,0 +1,175 @@
...
+
+    /* Close check and incoming[0], time to finish test */
+    uv_close((uv_handle_t*) &tcp_incoming[0], close_cb);
+    uv_close((uv_handle_t*) &tcp_check, close_cb);
+    return;
+  }
+
+  ASSERT(status == 0);
+  i = (unsigned int) ((intptr_t) req - (intptr_t) connect_reqs) /  sizeof(*req);
+  ASSERT(i < ARRAY_SIZE(connect_reqs));
```
bnoordhuis: Undefined behavior when i >= ARRAY_SIZE(connect_reqs).  A pedantic optimizing compiler could optimize this away entirely.  A better check would be something like `ASSERT(req >= connect_reqs && req < connect_reqs + ARRAY_SIZE(connect_reqs))` (and move it to the top.)

EDIT: I guess a really pedantic compiler might consider that UB as well.  Sometimes you just can't win.

test/test-tcp-close-accept.c
```diff
@@ -0,0 +1,175 @@
...
+  /* Close all active incomings, except current one */
+  for (i = 1; i < got_connections; i++)
+    uv_close((uv_handle_t*) &tcp_incoming[i], close_cb);
+
+  /* Create new fd that should be one of the closed incomings */
+  ASSERT(0 == uv_tcp_init(loop, &tcp_check));
+  ASSERT(0 == uv_tcp_connect(&tcp_check_req, &tcp_check, addr, connect_cb));
+  ASSERT(0 == uv_read_start((uv_stream_t*) &tcp_check, alloc_cb, read_cb));
+
+  /* Close server, so noone will connect to it */
```
bnoordhuis: s/noone/no one/

test/test-tcp-close-accept.c
```diff
@@ -0,0 +1,175 @@
...
+  }
+}
+
+TEST_IMPL(tcp_close_accept) {
+  unsigned int i;
+  uv_loop_t* loop;
+  uv_tcp_t* client;
+
+  /*
+   * A little explanation of what'll go below:
```
bnoordhuis: s/what'll go/what goes on/

Appreciate the thorough comment though. :-)

src/unix/core.c
indutny, line 606: Yep, and that's ok. Since I'm allocating `nwatchers + 2` below.

src/unix/darwin.c
indutny: Could handle be stopped and closed outside `uv__io_poll()` callback? I think - yes, if `uv_close()` was called between two `uv_run()` calls.

src/unix/darwin.c
indutny, line 119: erm... we are doing the same thing in `kqueue.c`.

src/unix/linux-core.c
indutny: hahahaha, sorry. Actually, I did change things in a loop, but forgot to change types :P

test/test-tcp-close-accept.c
indutny: Isn't it a pointer arithmetic?

indutny: Ah, I guess it is ok, since they're in the same array.

indutny:
@bnoordhuis fixed all mentioned nits, except conversion to `(int)`. Waiting for your reply.

indutny:
And thank you :)

src/unix/darwin.c
bnoordhuis: On Sun, Nov 3, 2013 at 11:45 AM, Fedor Indutny notifications@github.com wrote:

> Could handle be stopped and closed outside uv__io_poll() callback? I think - yes, if uv_close() was called between two uv_run() calls.

loop->watchers never shrinks.  If it's non-NULL once, it stays non-NULL.

src/unix/darwin.c
bnoordhuis, line 119: > erm... we are doing the same thing in kqueue.c.

That's something we should fix then.

indutny, line 119: In another PR, right?

indutny:
Fixed all nits, again :)

indutny:
@bnoordhuis time to review it? :)

src/unix/core.c
```diff
@@ -595,20 +595,33 @@ static unsigned int next_power_of_two(unsigned int val) {
...
+  watchers = realloc(loop->watchers,
+                     (nwatchers + 2) * sizeof(loop->watchers[0]));
 
   if (watchers == NULL)
     abort();
 
+  /* Copy watchers, preserving fake one in the end */
+  if (loop->watchers == NULL) {
+    fake_watcher_list = watchers[loop->nwatchers];
+    fake_watcher_count = watchers[loop->nwatchers + 1];
```
bnoordhuis, line 616: Using loop->watchers here would be a little clearer.

src/unix/darwin.c
bnoordhuis, line 119: Yes.

src/unix/kqueue.c
```diff
@@ -162,11 +162,18 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
 
     nevents = 0;
 
+    assert(loop->watchers != NULL);
+    loop->watchers[loop->nwatchers] = (void*) events;
+    loop->watchers[loop->nwatchers + 1] = (void*) ((uintptr_t) nfds);
```
bnoordhuis, line 167: Superfluous parentheses.

src/unix/kqueue.c
```diff
@@ -162,11 +162,18 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
...
     for (i = 0; i < nfds; i++) {
       ev = events + i;
       fd = ev->ident;
       w = loop->watchers[fd];
 
+      /* Skip invalidated events */
```
bnoordhuis, line 173: Maybe point to uv__platform_invalidate_fd() here, else future readers will wonder how the fd became -1 all of a sudden.

test/test-tcp-close-accept.c
```diff
@@ -0,0 +1,177 @@
...
+
+    /* Close check and incoming[0], time to finish test */
+    uv_close((uv_handle_t*) &tcp_incoming[0], close_cb);
+    uv_close((uv_handle_t*) &tcp_check, close_cb);
+    return;
+  }
+
+  ASSERT(status == 0);
+  ASSERT(connect_reqs <= req &&
+         req <= (connect_reqs + ARRAY_SIZE(connect_reqs)));
```
bnoordhuis, line 66: Split these into two asserts.  The extra parentheses aren't necessary, by the way.

test/test-tcp-close-accept.c
```diff
@@ -0,0 +1,177 @@
...
+  i = (unsigned int) (req - connect_reqs) /  sizeof(*req);
```
bnoordhuis, line 67: Just req - connect_reqs?

bnoordhuis:
Few comments but mostly LGTM.  A possible future optimization is to store the current index in loop->watchers so you only have to scan (nfds - i) entries instead of nfds entries.

indutny:
Landed with fixes in bbccafb and f50ccd5. Merged into the master.

PR #972, opened 2013-10-26
unix: Move uv_shutdown assertion

Keno:
Allows uv_shutdown to return an error as intended if the stream has already been closed.

Judging from the fact that the `if` statement checks for `UV_CLOSING`, I believe this to be the correct behavior. 

Refs https://github.com/JuliaLang/julia/issues/4229


bnoordhuis:
Cheers Keno, landed in fb36c09.

PR #973, opened 2013-10-28
darwin-proctitle: avoid calling pGetCurrentProcess

### LINKED ISSUE #966: Setting `process.title` causes the process to show up as "Not Responding" in OS X's Activity Monitor, in Mavericks
It takes a few minutes to trigger, but eventually the process shows up red and is marked as "Not Responding", even though the process itself is running normal.

[](https://cloudup.com/cIKf2lX9DNQ)

My initial guess is that the trick we do in order to set the process title ends up making Activity Monitor think that the process is a GUI app, rather than a command line app, and then it's expecting some kind of ACK from the Cocoa event loop at that point which never happens.

indutny:
Use some black-magic from Apple to change process name without getting
a "Not responding" tag from Activity Manager.

fix #966

/cc @bnoordhuis

Also, it might be a good idea to backport it to v0.10.


bnoordhuis:
LGTM but are the MAC_OS_X_VERSION_10_9 checks necessary?  It looks to me like it should also work with older versions.

By the way, I'd replace 'darwin-proctitle:' with just 'darwin:' in the commit log.

indutny:
Yes, they seems to be necessary. I'm going to land it to v0.10 and then do a merge.

TooTallNate:
Thanks for looking into this @indutny!

indutny:
Landed in 0.10: 08e0e63

indutny:
And merged into master.

PR #975, opened 2013-10-29
linux: use CLOCK_MONOTONIC_COARSE if available

bnoordhuis:
clock_gettime(CLOCK_MONOTONIC_COARSE) eliminates the SYS_clock_gettime
system call entirely.  It just reads from the vDSO timestamp, which is
the kernel's idea of "now" and is good enough for our purposes.

Suggested reviewer: @indutny


src/unix/linux-core.c
```diff
@@ -245,10 +247,68 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
...
+/* Try clock_gettime() with clock |fst| but fall back to |snd| if the first
+ * one isn't supported.  Cache the availability in |*skip_fst|.
+ */
+static int uv__try_clocks(clock_t fst,
+                          clock_t snd,
+                          struct timespec* pt,
+                          int* skip_fst) {
+  int rc;
+
+  if (*skip_fst == 0) {
```
indutny: fst stands for first or fast?

bnoordhuis: Whatever works for you. :-)  It stands for 'first' in my head but that's because I'm a closet ML programmer.

src/unix/linux-core.c
```diff
@@ -245,10 +247,68 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
...
-  return (((uint64_t) ts.tv_sec) * NANOSEC + ts.tv_nsec);
+  static int no_clock_monotonic_coarse;
+  struct timespec t;
+  int err;
+
+  /* Prefer CLOCK_MONOTONIC_COARSE if available because it can be fully served
+   * from the vDSO, no system call necessary.  The supposedly lower resolution
+   * is not an issue for us, it still has sub-microsecond precision.
+   */
+  err = uv__try_clocks(CLOCK_MONOTONIC_COARSE,
```
indutny: Is it safe to try it out anyway? Is there any way to do a runtime check of it?

bnoordhuis: `uv__try_clocks()` _is_ doing a run-time check, that's the whole point of that function. :-)

src/unix/linux-core.c
```diff
@@ -245,10 +247,68 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
...
+  rc = clock_gettime(snd, pt);
+  if (rc == -1)
+    return -errno;
+
+  return 0;
+}
+
+
+static uint64_t timespec_to_u64(const struct timespec* pt) {
+  return pt->tv_sec * (uint64_t) 1e9 + pt->tv_nsec;
```
indutny: Magic number?

indutny: Hehe... just kidding, nvm

indutny:
LGTM

bnoordhuis:
Landed in 21c37a7 for great justice.

PR #977, opened 2013-10-30
test: add regression test for 29fdb3471

indutny:
Requires `ulimit -n 3000` on osx...

/cc @bnoordhuis


test/test-watcher-cross-stop.c
```diff
@@ -0,0 +1,95 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ */
+
+#include "uv.h"
+#include "task.h"
+#include "internal.h"
```
bnoordhuis, line 24:
```
../../test/test-watcher-cross-stop.c:24:22: fatal error: internal.h: No such file or directory
```

Not sure why you're including it in the first place...

test/test-watcher-cross-stop.c
```diff
@@ -0,0 +1,95 @@
...
+
+#include <string.h>
+#include <errno.h>
+#include <unistd.h>
```
bnoordhuis, line 28: Won't compile on Windows, missing header.

test/test-watcher-cross-stop.c
indutny, line 24: Ooops, this is a left-over.

bnoordhuis:
Here, have a patch.  :-)

``` diff
diff --git a/test/test-watcher-cross-stop.c b/test/test-watcher-cross-stop.c
index fdba3ca..5b53320 100644
--- a/test/test-watcher-cross-stop.c
+++ b/test/test-watcher-cross-stop.c
@@ -21,11 +21,14 @@

 #include "uv.h"
 #include "task.h"
-#include "internal.h"

 #include <string.h>
 #include <errno.h>
-#include <unistd.h>
+
+#if !defined(_WIN32)
+#include <sys/time.h>
+#include <sys/resource.h>  /* setrlimit() */
+#endif

 /* NOTE: Number should be big enough to trigger this problem */
 static uv_udp_t sockets[2500];
@@ -66,6 +69,16 @@ TEST_IMPL(watcher_cross_stop) {
   uv_buf_t buf;
   char big_string[1024];

+#if !defined(_WIN32)
+  {
+    struct rlimit lim;
+    lim.rlim_cur = ARRAY_SIZE(sockets) + 32;
+    lim.rlim_max = ARRAY_SIZE(sockets) + 32;
+    if (setrlimit(RLIMIT_NOFILE, &lim))
+      RETURN_SKIP("File descriptor limit too low.");
+  }
+#endif
+
   addr = uv_ip4_addr("127.0.0.1", TEST_PORT);
   memset(big_string, 'A', sizeof(big_string));
   buf = uv_buf_init(big_string, sizeof(big_string));
```

indutny:
Yay, landed in f996018.

indutny:
thank you!

PR #979, opened 2013-10-31
unix: return exec errors from uv_spawn, not async

### LINKED ISSUE #978: unix: Could posix_spawn be used to improve uv_spawn?
I've been using `uv_spawn` recently, and one of the weird things which I've found from using it is that both `fork` can fail, and then the child's `exec` can also fail, but one is reported as the return of `uv_spawn` and the other is reported via the `exit_cb`. This seems a little odd, as in theory both are a "spawn failure" and would both be reported as a return value of `uv_spawn`.

I'm curious if libuv would consider using the `posix_spawn` function instead of fork/exec? I certainly don't claim to be an expert on the posix_spawn, but it appears to combine the fork/exec into one, and I think that using it would solve this problem of reporting an exec failure immediately instead of delaying it for a callback. One thing that I would be concerned about is whether `posix_spawn` is supported everywhere, which I'm not sure of.

I briefly glanced at the windows code, and it looked like the "exec error" is reported asynchronously to behave the same as the unix implementation, but it could easily be modified to be returned immediately instead.

I've also heard of the suggestion of using a pipe with `O_CLOEXEC` set. That way you can know if the process successfully invoked `exec` or not from the caller side of things. I'm not sure this is completely feasible, as I'd want the calling thread to block on this read (waiting for the child to finish `exec`), and the block could possibly take awhile.

alexcrichton:
If spawning a process fails due to an exec() failure (but it succeeded in
forking), then this should be considered a spawn failure instead of an
asynchronous termination of the process. This allows to check for common exec()
failure conditions such as a bad path quickly instead of having to rely on
keeping track of the async callback.

Additionally, the meaning of the two fields returned in the callback are now
exactly what they advertise to be. The process exit argument is not one of two
values depending on what happened to the child.

Closes #978

### DIFF
--- a/src/unix/process.c
+++ b/src/unix/process.c
@@ -109,9 +109,6 @@ static void uv__chld(uv_signal_t* handle, int signum) {
       if (WIFSIGNALED(process->status))
         term_signal = WTERMSIG(process->status);
 
-      if (process->errorno != 0)
-        exit_status = process->errorno;  /* execve() failed */
-
       process->exit_cb(process, exit_status, term_signal);
     }
   }
@@ -359,6 +356,7 @@ int uv_spawn(uv_loop_t* loop,
   ssize_t r;
   pid_t pid;
   int err;
+  int exec_errorno;
   int i;
 
   assert(options->file != NULL);
@@ -434,14 +432,14 @@ int uv_spawn(uv_loop_t* loop,
   uv__close(signal_pipe[1]);
 
   process->status = 0;
-  process->errorno = 0;
+  exec_errorno = 0;
   do
-    r = read(signal_pipe[0], &process->errorno, sizeof(process->errorno));
+    r = read(signal_pipe[0], &exec_errorno, sizeof(exec_errorno));
   while (r == -1 && errno == EINTR);
 
   if (r == 0)
     ; /* okay, EOF */
-  else if (r == sizeof(process->errorno))
+  else if (r == sizeof(exec_errorno))
     ; /* okay, read errorno */
   else if (r == -1 && errno == EPIPE)
     ; /* okay, got EPIPE */
@@ -461,15 +459,18 @@ int uv_spawn(uv_loop_t* loop,
     goto error;
   }
 
-  q = uv__process_queue(loop, pid);
-  QUEUE_INSERT_TAIL(q, &process->queue);
+  /* Only activate this handle if exec() happened successfully */
+  if (exec_errorno == 0) {
+    q = uv__process_queue(loop, pid);
+    QUEUE_INSERT_TAIL(q, &process->queue);
+    uv__handle_start(process);
+  }
 
   process->pid = pid;
   process->exit_cb = options->exit_cb;
-  uv__handle_start(process);
 
   free(pipes);
-  return 0;
+  return exec_errorno;
 
 error:
   if (pipes != NULL) {
--- a/test/test-spawn.c
+++ b/test/test-spawn.c
@@ -51,7 +51,6 @@ static void close_cb(uv_handle_t* handle) {
   close_cb_called++;
 }
 
-
 static void exit_cb(uv_process_t* process,
                     int64_t exit_status,
                     int term_signal) {
@@ -60,6 +59,13 @@ static void exit_cb(uv_process_t* process,
   ASSERT(exit_status == 1);
   ASSERT(term_signal == 0);
   uv_close((uv_handle_t*)process, close_cb);
+}
+
+
+static void fail_cb(uv_process_t* process,
+                    int64_t exit_status,
+                    int term_signal) {
+  ASSERT(0 && "fail_cb called");
 }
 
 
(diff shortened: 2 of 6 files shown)


src/unix/process.c
```diff
@@ -434,14 +431,14 @@ int uv_spawn(uv_loop_t* loop,
   uv__close(signal_pipe[1]);
 
   process->status = 0;
-  process->errorno = 0;
+  int exec_errorno = 0;
```
bnoordhuis: Declarations go at the top (of the function.)

test/test-spawn.c
```diff
@@ -168,10 +168,11 @@ static void timer_cb(uv_timer_t* handle, int status) {
 TEST_IMPL(spawn_fails) {
   init_process_options("", exit_cb_expect_enoent);
```
bnoordhuis: The callback is no longer expected to run so it should be replaced with a fail_cb() that asserts unconditionally.

test/test-spawn.c
```diff
@@ -857,13 +858,13 @@ static void timer_cb(uv_timer_t* handle, int status) {
   options.uid = 0;
 
   r = uv_spawn(uv_default_loop(), &process, &options);
```
bnoordhuis, line 851: Ditto, I think?

test/test-spawn.c
```diff
@@ -889,13 +890,13 @@ static void timer_cb(uv_timer_t* handle, int status) {
   options.gid = 0;
 
   r = uv_spawn(uv_default_loop(), &process, &options);
-  ASSERT(r == 0);
+  ASSERT(r == UV_EPERM);
```
bnoordhuis, line 883: And ditto. :-)

piscisaureus:
We originally chose not to do this because exec() might block the event loop while scanning $PATH etc. By redefining the api we can now never make uv_spawn() truly asynchronous.

bnoordhuis:
I briefly touched on that in #978.  Reporting all errors to uv_spawn() is more in line with other API functions, returning errors is convenient and faster in some cases (say when walking directories in search of a binary to execute.)

I guess it's a tradeoff; I don't have a strong preference one way or the other but I agree with @alexcrichton that the current duality is kind of confusing and jarring.

piscisaureus:
Yeah whatever land it then. Are you going to change windows too?

alexcrichton:
Updated with test fixes (plus removed the unused static functions now), and I didn't personally test on windows (don't have a setup locally), but I did update the code paths (it was pretty much already structured to do this).

I think another possible option would be to have two callbacks. One could be invoked when the process has finished spawning, and it contains the result of the spawn (the exec failed, etc), and then the second could be invoked when the process actually exits. This would be a little complicated (another layer of async), but it would prevent the event loop from blocking waiting to see if the `exec` succeeded or failed.

sam-github:
I was surprised to see uv blocking on the child-pipe, waiting for exec to finish, uv works so hard to make things async, usually.

Fwiw. I'd mildly prefer the passing of both errors to the callback, or just documenting that errors can appear as return value OR in callback, so its clear. I'd guess exec would fail more often than fork. I don't understand why passing error to callback requires two callbacks.

Either way, since this changes behaviour, shouldn't it touch uv.h with a doc comment? Please.

alexcrichton:
I wasn't able to find much existing documentation about uv_spawn in uv.h, but I tried to add some in the relevant location.

I suppose I should mention my use case as well. In Rust we're using these bindings, but we transform the async callbacks into blocking our green threads appropriately. For us, we want `Process::new(...)` to return either a new process, or the error it had spawning if it failed to spawn.

With this change, we can immediately know that information (without having to block our green task internally at all). Without this change (or with only one callback), we in theory want to block the task waiting for the answer of `exec` to return to the caller whether the child has started or not, but no action is taken on a successful `exec` so we won't exactly know when to wake the task.

That may be a little too much of a rust-specific explanation, but I don't think that we can expose the right interface to process spawning if there's only one callback which reports any flavor of error, but no callback reporting a successful spawn.

sam-github:
On Thu, Oct 31, 2013 at 2:37 PM, Alex Crichton notifications@github.com wrote:

> I wasn't able to find much existing documentation about uv_spawn in uv.h, but I tried to add some in the relevant location.

As someone new to uv.... thanks!

> With this change, we can immediately know that information (without having to block our green task internally at all). Without this change (or with only one callback), we in theory want to block the task waiting for the answer of exec to return to the caller whether the child has started or not, but no action is taken on a successful exec so we won't exactly know when to wake the task.

That totally makes sense, its  not just about consistency, its about
knowing when this important event (exec) has occurred. So yeah, what
you have here is definitely an improvement.

bnoordhuis:
> By redefining the api we can now never make uv_spawn() truly asynchronous.

After thinking about it some more, I don't think we'll ever be able to really do that.

Node's child_process module and everything that's built on top of that is predicated on the assumption that e.g. `spawn('whatever').kill('SIGUSR1')` works; the API is effectively synchronous.

Retrofitting an asynchronous uv_spawn() probably isn't possible without breaking some edge cases. 

> I was surprised to see uv blocking on the child-pipe, waiting for exec to finish, uv works so hard to make things async, usually.

See above, that's why. :-)  Because the parent process gets parked until the child calls execve(), performance is generally a little better, too.

bnoordhuis:
As discussed on IRC, Bert and I are both okay with this change.  Landed in f764bff.  Thanks, Alex!

PR #980, opened 2013-11-01
fsevents: use FlagNoDefer for FSEventStreamCreate

indutny:
Otherwise `FSEventStreamCreate()` will coalesce events, even if they're
happening in the interval, bigger than supplied `latency`. In other
words, if this flag is not set events will happen in separate callback
only if there was a delay bigger than `latency` between two consequntive
events.


src/unix/fsevents.c
```diff
@@ -316,10 +316,10 @@ static int uv__fsevents_create_stream(uv_loop_t* loop, CFArrayRef paths) {
   ctx.release = NULL;
   ctx.copyDescription = NULL;
 
-  latency = 0.15;
+  latency = 0.05;
 
   /* Set appropriate flags */
```
bnoordhuis: Update the comment to explain why you picked these particular flags?

test/test-fs-event.c
```diff
@@ -36,11 +36,21 @@
 #endif
 
 static uv_fs_event_t fs_event;
+static const char* file_prefix = "fsevent-";
```
bnoordhuis: Make this const char[] and use sizeof below.

indutny:
Is that all I should fix there?

bnoordhuis:
Yes.  Were you expecting more?

indutny:
All fixed, please review grammar! ;)

bnoordhuis:
LGTM.  Apropos the commit log, it's 'consecutive.'

indutny:
Thanks, landed in 0f5c28b.

PR #981, opened 2013-11-01
test: test fsevents error reporting

indutny:
Create a lot of `FSEventStream`s to make `FSEventStreamStart()` fail.
Fetch the error code and verify that it is correct.

/cc @bnoordhuis


test/task.h
```diff
@@ -153,6 +158,24 @@ enum test_status {
     return TEST_SKIP;                                                         \
   } while (0)
 
+#if !defined(_WIN32)
+
+# define TEST_FILE_LIMIT(num)                                                 \
```
bnoordhuis, line 163: I don't think this should be in the header, it's only used in one place.

test/test-fs-event.c
```diff
@@ -556,3 +556,110 @@ static void fs_event_cb_close(uv_fs_event_t* handle, const char* filename,
...
+  ASSERT(handle != NULL);
+  close_cb_called++;
+  free(handle);
+}
+
+
+TEST_IMPL(fs_event_error_reporting) {
+  unsigned int i;
+  uv_loop_t* loops[1024];
+  uv_fs_event_t* events[ARRAY_SIZE(loops)];
```
bnoordhuis: You can just allocate these on the stack and not bother with malloc/free, right?

test/test-fs-event.c
```diff
@@ -556,3 +556,110 @@ static void fs_event_cb_close(uv_fs_event_t* handle, const char* filename,
...
+    ASSERT(0 == uv_timer_init(loop, &timer));
+    ASSERT(0 == uv_timer_start(&timer, timer_cb_nop, 2, 0));
+    uv_run(loop, UV_RUN_DEFAULT);
+    ASSERT(1 == timer_cb_called && 1 == close_cb_called);
+    if (fs_event_error_reported != 0)
+      break;
+  }
+
+  /* At least one loop should fail */
+  ASSERT(fs_event_error_reported == -EMFILE);
```
bnoordhuis: UV_EMFILE?

test/test-fs-event.c
```diff
@@ -556,3 +556,110 @@ static void fs_event_cb_close(uv_fs_event_t* handle, const char* filename,
...
+
+  /* Stop and close all events, and destroy loops */
+  while (1) {
```
bnoordhuis: do/while loop?

test/test-fs-event.c
```diff
@@ -556,3 +556,110 @@ static void fs_event_cb_close(uv_fs_event_t* handle, const char* filename,
...
+
+    ASSERT(0 == uv_fs_event_stop(event));
+    uv_ref((uv_handle_t*) event);
+    uv_close((uv_handle_t*) event, fs_event_error_report_close_cb);
+
+    close_cb_called = 0;
+    uv_run(loop, UV_RUN_DEFAULT);
+    ASSERT(close_cb_called == 1);
+
+    uv_loop_delete(loops[i]);
```
bnoordhuis: loop

test/test-fs-event.c
```diff
@@ -556,3 +556,110 @@ static void fs_event_cb_close(uv_fs_event_t* handle, const char* filename,
...
+                                  fs_event_error_report_cb,
+                                  "watch_dir",
+                                  0));
+    uv_unref((uv_handle_t*) event);
+
+    /* Let loop run for some time */
+    ASSERT(0 == uv_timer_init(loop, &timer));
+    ASSERT(0 == uv_timer_start(&timer, timer_cb_nop, 2, 0));
+    uv_run(loop, UV_RUN_DEFAULT);
+    ASSERT(1 == timer_cb_called && 1 == close_cb_called);
```
bnoordhuis: Split these into two asserts.

test/test-fs-event.c
indutny: Won't this be quite a lot for a stack?

bnoordhuis: How big is one element?  Even if it's 1 kB, that's still only 1 MB of stack space.

test/task.h
indutny, line 163: In two places, actually. I was planning to replace it in another test later on.

test/test-fs-event.c
indutny: Ok, if you wish so.

indutny:
@bnoordhuis pushed fixes.

test/test-fs-event.c
```diff
@@ -556,3 +556,110 @@ static void fs_event_cb_close(uv_fs_event_t* handle, const char* filename,
...
+    ASSERT(0 == uv_timer_start(&timer, timer_cb_nop, 2, 0));
+    uv_run(loop, UV_RUN_DEFAULT);
+    ASSERT(1 == timer_cb_called);
+    ASSERT(1 == close_cb_called);
+    if (fs_event_error_reported != 0)
+      break;
+  }
+
+  /* At least one loop should fail */
+  ASSERT(fs_event_error_reported == -UV_EMFILE);
```
bnoordhuis: This can't possibly work...

test/test-fs-event.c
bnoordhuis: Not fixed yet.

test/test-fs-event.c
```diff
@@ -556,3 +556,110 @@ static void fs_event_cb_close(uv_fs_event_t* handle, const char* filename,
...
+    ASSERT(close_cb_called == 1);
+
+    uv_loop_delete(loops[i]);
+
+    loops[i] = NULL;
+    if (i == 0)
+      break;
+    else
+      i--;
+  } while(1);
```
bnoordhuis: This is not quite what I mean.  E.g. `} while (i-- != 0);` was what I had in mind.

indutny:
@bnoordhuis pushed fixes.

test/test-fs-event.c
```diff
@@ -556,3 +556,106 @@ static void fs_event_cb_close(uv_fs_event_t* handle, const char* filename,
...
+    uv_close((uv_handle_t*) event, fs_event_error_report_close_cb);
+
+    close_cb_called = 0;
+    uv_run(loop, UV_RUN_DEFAULT);
+    ASSERT(close_cb_called == 1);
+
+    uv_loop_delete(loop);
+
+    loops[i] = NULL;
+  } while(i-- != 0);
```
bnoordhuis, line 645: Style: missing space

test/task.h
```diff
@@ -153,6 +158,24 @@ enum test_status {
...
+    do {                                                                      \
+      struct rlimit lim;                                                      \
+      lim.rlim_cur = (num);                                                   \
+      lim.rlim_max = (num);                                                   \
```
bnoordhuis, line 167: You're evaluating num twice here.

indutny:
Landed with fixes in 86b5c1e. Thank you!

Issue #984, opened 2013-11-04
Windows build broken on SMP


geertj:
I'm trying to compile the current master on Windows. I'm using the following configuration:
- Windows 7, 64-bit.
- Python 2.7.6rc1 32-bit is installed, but not in the PATH. I use `set PYTHON=C:\Python27\python.exe` to allow the build to find it.
- Microsoft Visual C++ Express 2010.

When running "vcbuild.bat" from the source directory, I get the following error:

https://gist.github.com/anonymous/7299594

For some reason, build command never exits and the error message keeps repeating itself.

I can fix the build by doing a `copy gyp_uv gyp_uv.py`.

Since Python 2.7.6 is currently a pre-release, I also tried 2.7.5, but I got the same issue.

@saghul, this also breaks the pyuv build on WIndows.

saghul:
Hum, this looks really weird. Nothing special changed in gyp_uv which would make it not work anymore. Also, I don't get why something from multiprocessing is used.

Unfortunately I won't have access to my Windows test machine for a few weeks in order to debug this.

bnoordhuis:
That looks like a python issue more than anything else.  Like @saghul said, nothing really changed in gyp_uv.

bnoordhuis:
Forgot to mention the obvious: if rolling back to 2.7.5 fixes it, then you know for sure it's an issue with python and not libuv.

saghul:
Apparently rolling back the Python version doesn't fix it. I know I tested this (using the Python env variable), it was added because pyuv needed it, so I'm not sure what's going on.

geertj:
OK, I found the problem. My build VM has 2 virtual CPUs, and it appears that gyp is trying to do a parallel build. This requires the multiprocessing module. Because Windows doesn't have fork(), multiprocessing needs to start up a new Python and re-import "gyp_uv" as a module, which fails because it is not a python module.

One solution is to add the option "--no-parallel" when calling "gyp_uv". The other is to rename gyp_uv to gyp_uv.py. However in the latter case see http://docs.python.org/2/library/multiprocessing.html#windows for some more requirements on the main module (parallel build appears to work though).

saghul:
I'm undecided here, should we go for the --no-parallel just to play safe?

bnoordhuis:
That's probably the path of least resistance unless someone feels like patching up gyp_uv to work with multiprocessing on Windows.

tjfontaine:
[...]

PR #987, opened 2013-11-05
build: fix windows smp build with gyp

geertj:
Gyp will try a parallel build if it detect the system has >1 processor.
This functionality depends on the Python "multiprocessing" package. The
multiprocessing package on Windows requires that the top-level module is
importable as a module, see:

  http://docs.python.org/2/library/multiprocessing.html#windows

This fixes issue #984.


bnoordhuis:
LGTM if it works.  First line of the commit log should be something like 'build: fix windows smp build with gyp' (50 columns max.)  Other lines should wrap at 72 columns.

geertj:
I tested this patch on Linux and Windows.

I updated the message on the pull request, but any pointers on how to update the commit message of the commit that I already pushed?

Also signed the CLA..

saghul:
LGTM. @geertj you can amend your commit (commit --amend) or do an interactive rebase (rebase -i HEAD~1) and then force push (push -f) to this branch. That should do.

geertj:
@saghul, thanks for the tip. Updated and re-pushed

bnoordhuis:
Cheers Geert, landed in 2445467 and 991409e (master and v0.10.)

PR #993, opened 2013-11-15
unix: set close-on-exec flag on received fds

bnoordhuis:
Set the close-on-exec flag on file descriptors that we've received with
recvmsg() so we don't leak them when calling fork() afterwards.

On Linux, we use the MSG_CMSG_CLOEXEC flag when supported (2.6.23 and
up.)

On older Linux versions and other platforms, we walk the received file
descriptors and set the close-on-exec flag for each fd manually.  That
won't entirely avoid race conditions when other threads call fork() or
clone() but at least we're less likely to leak file descriptors now.

### DIFF
--- a/src/unix/core.c
+++ b/src/unix/core.c
@@ -542,6 +542,44 @@ int uv__dup(int fd) {
 }
 
 
+ssize_t uv__recvmsg(int fd, struct msghdr* msg, int flags) {
+  struct cmsghdr* cmsg;
+  ssize_t rc;
+  int* pfd;
+  int* end;
+#if defined(__linux__)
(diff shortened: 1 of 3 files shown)


piscisaureus:
lgtm but a comment explaining "try with flag, if that fails with EINVAL, retry without flag and if that works, never use flag" would be helpful.

bnoordhuis:
I landed it as-is.  I initially had a comment like you suggested but I later decided to remove it.  A comment should tell _why_ code does what it does, merely telling _what_ it does is kind of superfluous (unless it's really complicated code but that's not the case here.)

At any rate, thanks for the review.  :-)

PR #999, opened 2013-11-21
fsevents: fix subfolder check

indutny:
First of all, a bit of explanation of what happens there:
1. FSEvents emits absolute paths to changed files or directories
2. We cut off the first part of such paths, which is equal to handle's
   real path ('/dir/subdir/subsubdir`, without trailing slash)
3. Then, if we are running in non-recursive mode, we discard paths that
   have slashes ('/') as a non-first character in them

/cc @bnoordhuis


indutny:
I'm not sure if its worth writing separate test for it, but I tested it in node.js master and its working.

indutny:
Added follow-up commit to fix watching '/' folder.

src/unix/fsevents.c
```diff
@@ -251,13 +251,15 @@ static void uv__fsevents_event_cb(ConstFSEventStreamRef streamRef,
...
         continue;
 
-      path += handle->realpath_len;
-      len -= handle->realpath_len;
-
-      /* Skip back slash */
-      if (*path != 0) {
-        path++;
-        len--;
+      if (handle->realpath_len > 1 || *handle->realpath != '/') {
```
bnoordhuis, line 254: Maybe `if (hande->realpath[0] != '\0' && handle->realpath[0] != '/') {`?  I'll leave that to your good judgment.

src/unix/fsevents.c
```diff
@@ -251,13 +251,15 @@ static void uv__fsevents_event_cb(ConstFSEventStreamRef streamRef,
...
+        path += handle->realpath_len;
+        len -= handle->realpath_len;
+
+        /* Skip back slash */
+        if (*path != 0) {
```
bnoordhuis, line 259: It's a forward slash.  Also, preferably use `'\0'` - that makes it obvious at first glance that you're comparing a character.

bnoordhuis:
Minor nits but LGTM.  A regression test would be nice though.

src/unix/fsevents.c
indutny, line 254: The way it is now, it should be ok for `realpath` to be NULL, I like it more (even if not all places are accepting `realpath` = NULL now).

indutny:
Landed in bf5038d, thank you!

src/unix/fsevents.c
bnoordhuis, line 254: > even if not all places are accepting realpath = NULL now

Yeah, like two lines up...

PR #1005, opened 2013-11-23
libuv: Forward additional getaddrinfo errors

skabbes:
These are defined on android and on mac (and possibly other platforms)


src/uv-common.c
```diff
@@ -433,6 +433,18 @@ int uv__getaddrinfo_translate_error(int sys_err) {
...
+  case EAI_BADHINTS: return UV_EAI_BADHINTS;
+#endif
+#if defined(EAI_PROTOCOL)
+  case EAI_PROTOCOL: return UV_EAI_PROTOCOL;
+#endif
+#if defined(EAI_OVERFLOW)
+  case EAI_OVERFLOW: return UV_EAI_OVERFLOW;
+#endif
+#if defined(EAI_MAX)
+  case EAI_MAX: return UV_EAI_MAX;
```
bnoordhuis: Is EAI_MAX actually returned or is it a 'highest error code plus one' marker?

skabbes:
I'm pretty sure that its only 'highest' plus one.  I didn't know what the libuv philosophy was for that kind of thing.  I also wanted to allow the user to be able to avoid [this abort](https://github.com/joyent/libuv/blob/master/src/uv-common.c#L437-L440)

I can remove it if you like

bnoordhuis:
> I can remove it if you like

Yes, please.

skabbes:
fixed, but i have no idea why the smartos and windows tests are failing now

bnoordhuis:
Looks unrelated.  Landed in dbe9905.  Thanks, Steven.

PR #1007, opened 2013-11-24
stream: fix emfile trick

indutny:
No comments.


bnoordhuis:
Wrong target branch?  I'm willing to believe I broke it in master after the errno rework but in v0.10, uv__accept() still returns either a file descriptor or -1.

indutny:
Yeah, you broke it in master then...

indutny:
mind pulling, anyway? :)

indutny:
@bnoordhuis btw, why are you returning `errno` anyway?

PR #1008, opened 2013-11-25
unix: fix accept() EMFILE error handling

bnoordhuis:
Fix a bug that was introduced in commit 3ee4d3f ("unix, windows:
return error codes directly") and add a regression test for good
measure.

Hat tip to Fedor for pointing out the issue.

Fixes #1007.

Suggested reviewer: @indutny


indutny:
LGTM.

PR #1010, opened 2013-11-25
stream: fix writing to connecting stream

indutny:
If stream is connecting, writing to it was never actually happening,
because we was returning in `uv__stream_io` right after handling it.

/cc @bnoordhuis


bnoordhuis:
Writing before having connected is currently only supported on Unices and IIRC is hard to implement on Windows (@piscisaureus knows the details, I remember having discussed that with him.)

If the current behavior is problematic, then maybe we should - for the sake of feature parity - make uv_write() simply return ENOTCONN.

@piscisaureus Thoughts?

test/test-tcp-write-and-connect.c
```diff
@@ -0,0 +1,142 @@
...
+
+static void write_cb(uv_write_t* req, int status) {
+  ASSERT(status == 0);
+  write_cb_called++;
+}
+
+
+static uv_buf_t alloc_cb(uv_handle_t* handle, size_t size) {
+  uv_buf_t buf;
+  buf.base = (char*)malloc(size);
```
bnoordhuis: Superfluous cast.

bnoordhuis: Also, is it necessary to malloc here?

test/test-tcp-write-and-connect.c
```diff
@@ -0,0 +1,142 @@
...
+}
+
+
+static void connection_cb(uv_stream_t* tcp, int status) {
+  int r;
+
+  ASSERT(status == 0);
+
+  r = uv_tcp_init(tcp->loop, &incoming);
+  ASSERT(r == 0);
```
bnoordhuis: You're being somewhat inconsistent with `ASSERT(0 == command)` vs `r = command;  ASSERT(r == 0)` in this file.  I personally prefer the former, it's a bit shorter.

test/test-tcp-write-and-connect.c
```diff
@@ -0,0 +1,142 @@
...
+
+  connection_cb_called++;
+  r = uv_read_start((uv_stream_t*) &incoming, alloc_cb, read_cb);
+  ASSERT(r == 0);
+}
+
+
+static void start_server(void) {
+  struct sockaddr_in addr = uv_ip4_addr("0.0.0.0", TEST_PORT);
```
bnoordhuis: You might as well put the assignment on a new line.  When you merge it into master, you'll have to update it anyway.

src/unix/stream.c
```diff
@@ -1106,7 +1106,8 @@ static void uv__stream_io(uv_loop_t* loop, uv__io_t* w, unsigned int events) {
 
   if (stream->connect_req) {
     uv__stream_connect(stream);
-    return;
+    if (uv__stream_fd(stream) == -1)
+      return; /* connect error */
```
bnoordhuis: I don't think this is necessarily always correct.  For one, you're assuming that the connect callback immediately closes the connection on error.

indutny: Well, I can rephrase comment, but the logic looks sounds to me here.

test/test-tcp-write-and-connect.c
indutny: Nope, I just copied it from other test.

indutny:
All fixed, please review again.

src/unix/stream.c
bnoordhuis: "The logic looks sound to me" doesn't really address the issue I raised... at the very least explain _why_ you think that's okay.

indutny: I don't follow, sorry.

bnoordhuis: Okay, so the issue that I see here is that you expect the user to close the handle immediately on error, else handle->io_watcher.fd wouldn't be -1.

It's not a terrible assumption but I don't think it's always going to be true.  If the handle is still open, execution falls through and what happens next depends entirely on the value of the events argument.

I think the way to address that is by having uv__stream_connect() return a status code and act on that.

indutny:
Pushed follow-up, please take a look.

src/unix/stream.c
```diff
@@ -1105,8 +1105,13 @@ static void uv__stream_io(uv_loop_t* loop, uv__io_t* w, unsigned int events) {
...
 
   if (stream->connect_req) {
-    uv__stream_connect(stream);
-    return;
+    /*
+     * There was either connect error, or user's cb that has
+     * closed the socket. Anyway, no reason to continue here without
+     * fd.
+     */
+    if (uv__stream_connect(stream) != 0 || uv__stream_fd(stream) < 0)
```
saghul, line 1113: What if uv__stream_connect returns EINPROGRESS?

saghul:
Also, as Ben pointed out, looks like this would fail in Windows: https://github.com/joyent/libuv/blob/master/src/win/stream.c#L150 because UV_HANDLE_WRITABLE is set after connecting: https://github.com/joyent/libuv/blob/master/src/win/tcp.c#L1062

So, shouldn't we try and do the same on both platforms? :-)

indutny:
Well... it seems that windows is not capable of buffering writes in stream.c, perhaps, I can just tweak my app to deal with it :)

indutny:
On other hand it may be a good idea to error on unix as well.

saghul:
Yep, that would be nice. Windows seems to give EPIPE, but I think Ben's
suggestion of using ENOTCON sounds better.

PR #1011, opened 2013-11-25
Added *.pyc to gitignore.

schlamar:
A *.pyc file is generated in the Windows build step with gyp since 2445467 / 991409e461e.


bnoordhuis:
Cheers, landed in c3e05ba.

PR #1006, opened 2013-11-23
Support VS2013 build

juanper:

PR #1003, opened 2013-11-22
windows: simplify vcbuild.bat

piscisaureus:
This patch is not ready for landing. It hasn't been fully polished!

cc @sblom for testing

PR #1012, opened 2013-11-25
build: add support for Windows SDK

schlamar:
With this patch vcbuild.bat is able to detect if it is run within a Windows SKD Command Prompt. In this case it does not try to configure variables because they are already set.


schlamar:
As motivation, Windows SDK allows 64bit builds for free.

tjfontaine:
can you try with #1006 instead, and see if relying on gyp's detection will work for you?

schlamar:
Do you mean #1003?

schlamar:
Closing in favor of #1107 and #1108. :)

Issue #444, opened 2012-06-04
unix, windows: queue write reqs when connect req is in progress


bnoordhuis:
See `test/test-tcp-dont-connect-after-write.c`. Make sure that write reqs are queued when the connection hasn't been established yet.

bnoordhuis:
Fixed (on Unices) in cb9ca14.

PR #1018, opened 2013-12-02
stream: return ENOTCONN on write before connect cb

indutny:
Be consistent with windows and error on `uv_write()` if called before
`uv_tcp_connect()`'s callback. We are not unqueuing data anyway.

/cc @saghul


saghul:
The Windows error code would need to be adjusted, currently it returns UV_EPIPE. Also, the failing test on Linux would need to be removed since it would no longer apply. Lastly, check this: https://github.com/joyent/libuv/issues/444

indutny:
Oh, so there is an issue about it. @piscisaureus do you think you may have time for that?

saghul:
Closing, Windows needs to eventually support queuing write reqs.

PR #1019, opened 2013-12-03
build: ignore generated makefiles on android

skabbes:
gyp currently doesn't support the --generator-output option for android,
this is the best way to ignore generated files for projects using libuv
as a dependency.


saghul:
@indutny any objections to this?

indutny:
Nope, if it works :)

indutny:
Thank you, landed in 4a7a72f

PR #1024, opened 2013-12-05
doc: document how to run tests

mmalecki:
Thanks for instructions @indutny.


indutny:
ninja is not really required, I'd rather add example with make as a gyp's output.

indutny:
LGTM, thank you

indutny:
Thank you, landed in 8094ac2.

PR #1027, opened 2013-12-06
unix: fix various memory leaks and undef behavior

indutny:
Kindly suggested by `cppcheck`.

/cc @saghul


src/unix/signal.c
```diff
@@ -348,6 +348,7 @@ static void uv__signal_event(uv_loop_t* loop,
   int r;
 
   bytes = 0;
+  end = 0;
```
saghul, line 351: Is this one needed? How can end be accessed without being initialized in this function?

saghul:
Minor comment about the 'end' variable, otherwise LGTM.

src/unix/signal.c
indutny, line 351: From there: https://github.com/indutny/libuv/blob/6b1227638db48236b7aaf45777826f43c231ae41/src/unix/signal.c#L357

saghul, line 351: Ah, right! It's good to go then :-)

indutny:
Thanks, landed in 646de34

Issue #1025, opened 2013-12-05
stream: create API for immediate writes


indutny:
`uv_write()`/`uv_write2()` may write data immediately on unixes, but it'll still queue up `uv_write_t` request and execute callback later. Its good for the most of the uses, but I'd like to propose API for handling immediate writes in user code and canceling queuing of the callback.

The few reasons for it:
- https://github.com/indutny/bud/commit/71ce128ac2a463c5e3334b27ce43dad3b05d4da8
- https://github.com/joyent/node/blob/master/lib/net.js#L659-L661

So having this API will allow us to eliminate allocations in node and other projects.

It could be something like this:

``` C
int uv_write_is_complete(uv_write_t* req);
int uv_write_finish(uv_write_t* req);
```

First one will return `1` if data was written to the socket and `0` otherwise.
Second one will execute request callback and remove request from the internal queue, so it could be reused by app.

indutny:
@saghul @piscisaureus thoughts?

indutny:
I know that it is most likely not case for windows, so we can make `uv_write_finish` return error on it unconditionally.

saghul:
Hum, I tried to understand why you would need this, but it's still not very clear to me :-) Even if the write happens immediately you already allocated a uv_write_t struct which was passed to uv_write, right? On Unixes, if the request wasn't sent on the spot, no more malloc-ing is necessary if less than 4 buffers were used (https://github.com/joyent/libuv/blob/master/src/unix/stream.c#L1251-L1253) so the difference is that the callback is called on the next tick, right?

I guess I'm missing something obvious here, can you elaborate a bit please? :-)

indutny:
Well, I'm mallocing it now. But it could be embedded in other struct if I'd be able to ask libuv to not call write_cb and dequeue `uv_write_t`.

saghul:
Ah, I get you now. What about

```
int uv_try_write(uv_write_t* req);
```

or

```
int uv_try_write(uv_stream_t* stream, uv_buf_t* bufs, int bufcnt);
```

The idea being that a write is attempted and if it fails the request is dequeued. In the second form no callback would be called, it would be up to you. Internally, a stack allocated request could be used.

indutny:
Hm... sounds like it would duplicate a lot of logic. I'm not totally against allocation, just want the `uv_write_t` to be reused. Any objections about API mentioned above?

[...]

PR #1028, opened 2013-12-06
stream: introduce uv_try_write(...)

indutny:
`uv_try_write(req, handle, bufs, nbufs)` acts like `uv_write()`,
but without queueing actual write until UV_POLLOUT (or IOCP completion).
This is useful for doing writes using on-stack `uv_write_t` requests.

see #1025 for details.


indutny:
@saghul I'm not really confident about windows, hopefully CI will let me know if I had wrong assumptions about it. @piscisaureus want to weigh in?

indutny:
Looks like its totally borked on windows, and won't work ever. Thanks @piscisaureus for explaining it on IRC. After all, it wouldn't be possible to have `uv_write_t` on-stack, and thus no point in additional API for this.

indutny:
@saghul looks like it may give us 30% improvement for some cases. Let's return `ENOTSUP` on windows, LGTY?

src/unix/stream.c
```diff
@@ -1299,6 +1299,46 @@ int uv_write(uv_write_t* req,
...
+
+  immediate = stream->write_queue_size == 0;
+
+  /* Unqueue request, regardless of immediateness */
+  QUEUE_REMOVE(&req->queue);
+  uv__req_unregister(stream->loop, req);
+  if (req->bufs != req->bufsml)
+    free(req->bufs);
+  req->bufs = NULL;
+  stream->write_queue_size = 0;
```
saghul: You should do something like this here instead: https://github.com/joyent/libuv/blob/master/src/unix/stream.c#L430

src/unix/stream.c
```diff
@@ -1299,6 +1299,46 @@ int uv_write(uv_write_t* req,
...
+int uv_try_write(uv_write_t* req,
+                 uv_stream_t* stream,
+                 const uv_buf_t bufs[],
+                 unsigned int nbufs) {
+  int r;
+  int has_pollout;
+  int immediate;
+
+  /* Already writing some data */
+  if (stream->write_queue_size != 0)
```
saghul, line 1321: Also check if we are connecting (stream->connect_req).

saghul:
> @saghul looks like it may give us 30% improvement for some cases. Let's return ENOTSUP on windows, LGTY?

Left a couple of comments on the Unix side. For Windows, why can't it succeed? I see this in the code: https://github.com/joyent/libuv/blob/master/src/win/tcp.c#L806

@piscisaureus can you shed some light? :-)

indutny:
@saghul I can't unqueue write.. that's why :)

saghul:
@indutny You mean the overlapped handle can't be cancelled?

saghul:
This just crossed my mind: since the req is kind of a throwaway in this function, what if we dump it from the function signature and use a stack allocated one internal to it instead?

```
int uv_try_write(uv_stream_t* stream,
                        const uv_buf_t bufs[],
                        unsigned int nbufs) {
   int r;
   int has_pollout;
   int immediate;
   uv_write_t req;

   ....
```

indutny:
Oh, that is a good idea, actually!

saghul:
Ah, looks like @piscisaureus already explained it to you over IRC, disregard my previous comment for Windows. I guess the appropriate errno would be ENOSYS.

bnoordhuis:
What about partial writes?  Don't take this the wrong way but the current approach seems pretty brittle.

saghul:
On 12/13/13 1:22 PM, Ben Noordhuis wrote:

> What about partial writes? Don't take this the wrong way but the current
> approach seems pretty brittle.

Thanks for chiming in Ben! I guess we should return the number of bytes 
we where able to write... Then I'm not sure if we can deal with trying 
to write multiple buffers.

Maybe we need to rethink this and do:

int uv_try_write(uv_stream_t *handle, const uv_buf_t *buf);

And return error or the number of bytes out of that buffer we were able 
to write.

## 

Saúl Ibarra Corretgé
http://bettercallsaghul.com

indutny:
@bnoordhuis thanks for a tip, didn't think about that.

indutny:
@saghul updated, please take a look.

include/uv.h
```diff
@@ -673,6 +673,16 @@ UV_EXTERN int uv_write2(uv_write_t* req,
...
 
+/*
+ * Same as `uv_write()`, but won't queue write request if it can't be completed
+ * immediately.
+ * Will return either:
+ * - positive number of bytes written
+ * - zero - if queued write is needed
+ * - negative error code
+ */
+UV_EXTERN ssize_t uv_try_write(uv_stream_t* handle, const uv_buf_t buf);
```
saghul: uv_fs_write (which is kind of similar) returns int and gets a void\* and a size_t instead of a uv_buf_t: https://github.com/joyent/libuv/blob/master/include/uv.h#L1681

Should we make it similar?

src/unix/stream.c
```diff
@@ -1299,6 +1299,48 @@ int uv_write(uv_write_t* req,
...
+
+
+ssize_t uv_try_write(uv_stream_t* stream, const uv_buf_t buf) {
+  int r;
+  int has_pollout;
+  ssize_t written;
+  uv_write_t req;
+
+  /* Already writing some data */
+  if (stream->write_queue_size != 0)
```
saghul, line 1321: Check if we are connecting here? That way we avoid trying to write altogether.

src/unix/stream.c
```diff
@@ -1299,6 +1299,48 @@ int uv_write(uv_write_t* req,
...
+    return 0;
+
+  has_pollout = uv__io_active(&stream->io_watcher, UV__POLLOUT);
+  r = uv_write(&req, stream, &buf, 1, uv_try_write_cb);
+  if (r != 0)
+    return r;
+
+  /* Remove not written bytes from write queue size */
+  written = buf.len;
+  if (req.bufs != NULL) {
```
saghul: Is this really possible?

src/unix/stream.c
```diff
@@ -1299,6 +1299,48 @@ int uv_write(uv_write_t* req,
...
+    written -= uv__write_req_size(&req);
+    stream->write_queue_size -= uv__write_req_size(&req);
```
saghul: Avoid calling uv__write_req_size twice here?

include/uv.h
saghul: If we still keep the buf, would it be better to pass a pointer instead?

test/test-tcp-try-write.c
```diff
@@ -0,0 +1,144 @@
...
+static void connect_cb(uv_connect_t* req, int status) {
+  static char rnd[1024];
+  ssize_t r;
+  uv_buf_t buf;
+  ASSERT(status == 0);
+  connect_cb_called++;
+
+  buf = uv_buf_init(rnd, sizeof(rnd));
+  do {
+    r = uv_try_write((uv_stream_t*) &client, buf);
```
saghul: Are we just sending null bytes here?

indutny: Yep.

src/unix/stream.c
indutny: Yep, if fully written.

include/uv.h
indutny: Please take another look, updated.

saghul:
LGTM.

src/unix/stream.c
```diff
@@ -1299,6 +1299,59 @@ int uv_write(uv_write_t* req,
...
+int uv_try_write(uv_stream_t* stream, const char* buf, size_t size) {
+  int r;
+  int has_pollout;
+  size_t written;
+  size_t req_size;
+  uv_write_t req;
+  uv_buf_t bufstruct;
+
+  /* Connecting */
+  if (stream->connect_req != NULL)
```
saghul, line 1317: Maybe combine these 2?

indutny, line 1317: Ok.

indutny:
Thanks, landed in b5e7798

PR #1029, opened 2013-12-07
implement UV_EXTERN_VISIBILITY_DEFAULT for GCC and Clang

wrl:
Previously, the `visibility` variable in common.gypi only controlled whether GCC or Clang builds would build with -fvisibility=hidden, but unfortunately this effect was negated by a define in uv.h setting UV_EXTERN to mark symbols with `default` visibility. This change sets -fvisibility=hidden as default and adds a define for setting UV_EXTERN to `default` visibility, which also enables building with `hidden` visibility, for example as a static library.

### DIFF
--- a/common.gypi
+++ b/common.gypi
@@ -154,7 +154,7 @@
             'cflags': [ '-pthread' ],
             'ldflags': [ '-pthread' ],
           }],
-          [ 'visibility=="hidden" and (clang==1 or gcc_version >= 40)', {
+          [ '(clang==1 or gcc_version >= 40)', {
             'cflags': [ '-fvisibility=hidden' ],
           }],
         ],
@@ -202,6 +202,11 @@
        'cflags': [ '-fno-omit-frame-pointer' ],
        # pull in V8's postmortem metadata
        'ldflags': [ '-Wl,-z,allextract' ]
+     }],
+     ['visibility=="default"', {
+       'defines': [
+         'UV_EXTERN_VISIBILITY_DEFAULT',
+       ]
      }],
     ],
   },
--- a/include/uv.h
+++ b/include/uv.h
@@ -39,7 +39,7 @@ extern "C" {
     /* Building static library. */
 #   define UV_EXTERN /* nothing */
 # endif
-#elif __GNUC__ >= 4
+#elif __GNUC__ >= 4 && defined(UV_EXTERN_VISIBILITY_DEFAULT)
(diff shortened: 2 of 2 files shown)


wrl:
i'm in the CLA now. not sure what subsystem this falls under, but if that and the commit line length are huge issues for you, let me know and i'll amend the commit.

indutny:
Thank you! I would place it under `gyp` subsystem, and yeah we care about commit line length. Perhaps you may omit `GCC and Clang` in it?

indutny:
@wrl ping ;)

wrl:
@indutny how's this?

common.gypi
```diff
@@ -154,7 +154,7 @@
             'cflags': [ '-pthread' ],
             'ldflags': [ '-pthread' ],
           }],
-          [ 'visibility=="hidden" and (clang==1 or gcc_version >= 40)', {
+          [ '(clang==1 or gcc_version >= 40)', {
```
indutny, line 157: Parens are not required.

include/uv.h
```diff
@@ -39,7 +39,7 @@
     /* Building static library. */
 #   define UV_EXTERN /* nothing */
 # endif
-#elif __GNUC__ >= 4
+#elif __GNUC__ >= 4 && defined(UV_EXTERN_VISIBILITY_DEFAULT)
```
bnoordhuis, line 42: What about the autotools and Makefile.mingw builds?

indutny, line 42: @wrl that is a good point too ^

wrl, line 42: @bnoordhuis yeah, very good point. currently they don't do anything about visibility (no `-fvisibility=hidden` anywhere).
i could either flip this conditional so `default` is...well, the default, and switch it to `UV_EXTERN_VISIBILITY_HIDDEN`, or define `UV_EXTERN_VISIBILITY_DEFAULT` in the autotools and mingw builds. which would you prefer?

win32 has its own separate issues, however (uses shared vs static builds to determine symbol visibility), and i was directed by @indutny to not worry about it initially.

indutny, line 42: @wrl it'd be good to have some sort of switch for `UV_EXTERN_VISIBILITY` in autotools too.

About mingw... I'm not sure about how it actually works, but I think that it is still creating dlls for shared libraries. Which means that the config we already have should work in that case too.

indutny:
@wrl ping

saghul:
No follow-up, closing.

Issue #1030, opened 2013-12-08
build fail OS 10.6.8 in ./configure


skaller:
checking for pthread_mutex_init in -lpthread... yes
checking for clock_gettime in -lrt... no
checking for sendfile in -lsendfile... no
checking for socket in -lsocket... no
checking for special C compiler options needed for large files... no
checking for _FILE_OFFSET_BITS value needed for large files... no
./configure: line 11346: syntax error near unexpected token `0.20'
./configure: line 11346:`PKG_PROG_PKG_CONFIG(0.20)'

indutny:
What version of autoconf are you using?

saghul:
Works for me on OSX 10.6.8, but using autoconf 2.69, since the system one is too old: https://github.com/joyent/libuv/issues/851

tjfontaine:
fwiw it's broken for all unix build slaves as well

http://jenkins.nodejs.org/job/libuv-master/374/

indutny:
@tjfontaine I'm quite sure it is autoconf dependent thing, @saghul perhaps we should revert that change?

saghul:
@tjfontaine I just tested it on my Linux machine and works here as well. Do the build slaves have the `pkg-config` package installed?

tjfontaine:
it's possible they don't, but that's a fairly obtuse error if so, we should do a feature test for pkg-config first

saghul:
Lets see if https://github.com/joyent/libuv/pull/1034 fixes it

saghul:
This was "fixed" already. pkg-config is no longer an issue, but a new version of autoconf is required, the one on 10.6 is ancient.

PR #1034, opened 2013-12-08
build: check if pkg-config is installed

saghul:
My autotools-fu is not very strong,but lets see if the build slaves turn green.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -45,12 +45,11 @@ AM_CONDITIONAL([NETBSD],  [AS_CASE([$host_os], [netbsd*],  [true], [false])])
 AM_CONDITIONAL([OPENBSD], [AS_CASE([$host_os], [openbsd*], [true], [false])])
 AM_CONDITIONAL([SUNOS],   [AS_CASE([$host_os], [solaris*], [true], [false])])
 AM_CONDITIONAL([WINNT],   [AS_CASE([$host_os], [mingw*],   [true], [false])])
-PKG_PROG_PKG_CONFIG([0.20])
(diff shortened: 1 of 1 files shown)


tjfontaine:
```
checking for pkg-config... no
./configure: line 11647: syntax error near unexpected token `0.20'
./configure: line 11647: `    PKG_PROG_PKG_CONFIG(0.20)'
```

it's checking for pkg-config but not bailing after that

tjfontaine:
I'll leave osx unfixed so we have something to test against

saghul:
Fun, even if I protect it, the mere occurrence of that undefined macro makes autoconf cry.

saghul:
Ok, I managed to avoid the problem. I removed that macro because libuv doesn't itself use pkg-config, it just generates the .pc file, so I don't think we need to check for a specific pkg-config version.

@tjfontaine @indutny thoughts?

indutny:
I'm not really good at autoconf too :) But if it works - I'm ok with it.

skaller:
I also fix that problem by deleting the macro in the configure file. I actually do have pkg-config installed. (but libuv didn't build right anyhow).

tjfontaine:
@saghul I second the notion that there's no need to dep on pkgconfig if you're only generating it

saghul:
Landed it on https://github.com/joyent/libuv/commit/69d2b69647564657625a6118ba911490d0cedcca

saghul:
@skaller what version of autoconf are you using? As I mentioned above the default one in the system won't work because it's too old and contains a bug. Please paste your `./configure` output with current master.

indutny:
@saghul +1

saghul:
I meant to type this on #1030, lets continue there, since the pkg-config thing is fixed now :-)

Issue #963, opened 2013-10-22
build: Upgrade to OS X Mavericks broke --disable-dtrace


rurban:
Without `--disable-dtrace` it compiles and tests okay.
Since my upgrade to teh latest OSX, with `--disable-dtrace` the command DTRACE is empty and executing -h -xnolibs ... is causing an error.

jneu:
I was able to reproduce this on OS X Mavericks with the following steps:
1. ./configure
2. make
3. make clean
4. ./configure --disable-dtrace
5. make
I don't think this issue is specific to any particular version of OS X, though.

Essentially the issue is this:
1. The initial make creates dependency files in the src/unix/.deps directory. These are not cleaned up by the make clean.
2. The second configure adds rules to make uv-dtrace.h, even though it's not added as a BUILT_SOURCES.
3. Since uv-dtrace.h is found in the dependency files, the second make causes it's recipe to run.
4. The recipe for uv-dtrace.h fails because it's incomplete (since --disable-dtrace was added).

To fix it, I added conditionals around the uv-dtrace.h rules in Makefile.am so they are not added to the Makefile if dtrace is not configured.

PR #1037, opened 2013-12-09
build: hide dtrace rules unless needed (issue 963)

jneu:
Fixes issue #963.

When using configure, there are situations where libuv will attempt
to build uv-dtrace.h, even if it is configured with --disable-dtrace.
For example, if libuv is first configured with dtrace enabled, then
built, the .deps files will contain references to include/uv-dtrace.h.
After a make clean and configure --disable-dtrace, the build will still
attempt to create include/uv-dtrace.h and fail. make will see the
dependency reference (which survives the make clean), use the rule
(which is always added to the Makefile), and fail since DTRACE is not
defined.

This commit protects the rules to make uv-dtrace.h with the proper
conditionals to ensure the rules are not written if --disable-dtrace
is chosen.


indutny:
LGTM

indutny:
Landed in f44f812 .

PR #1039, opened 2013-12-10
unix, windows: always update loop time

### LINKED ISSUE #846: uv_run doesn't update the time for uv_now if there isn't anything to do in the loop
Title speaks for itself. In https://github.com/joyent/libuv/blob/master/src/win/core.c#L282 (uv_run), `uv_update_time` should also be called when the loop doesn't run so the time is always updated. So that line should be copied to the line just before the while loop.

saghul:
Fixes #846


indutny:
LGTM

saghul:
Thanks Fedor, merged.

PR #1040, opened 2013-12-12
uv-common: add error detection in inet_addr()

yorkie:
Check this url: http://linux.die.net/man/3/inet_addr, please. And I noticed this point:

> Use of this function is problematic because -1 is a valid address (255.255.255.255)

Then If we except the `255.255.255.255` case, the `inet_addr()` function could throw the invalid ip error and work well.


saghul:
On that same page you can see that the use of inet_addr is discouraged. We should probably use inet_pton instead. Check here for an example: http://www.beej.us/guide/bgnet/output/html/multipage/inet_ntopman.html

yorkie:
Thanks, I see, then has this problem been fixed by `inet_pton()`?

saghul:
AFAIK yes, because in case of error it returns NULL and sets errno accordingly.

yorkie:
I have updated this file, using `inet_pton()`.

\cc @saghul

src/uv-common.c
```diff
@@ -102,9 +102,14 @@ int uv_ip4_addr(const char* ip, int port, struct sockaddr_in* addr) {
   memset(addr, 0, sizeof(*addr));
   addr->sin_family = AF_INET;
   addr->sin_port = htons(port);
-  /* TODO(bnoordhuis) Don't use inet_addr(), no good way to detect errors. */
-  addr->sin_addr.s_addr = inet_addr(ip);
-  return 0;
+  /* [Fixed] TODO(bnoordhuis) Don't use inet_addr(), no good way to detect errors. */
```
saghul: No need to keep the comment around.

src/uv-common.c
```diff
@@ -102,9 +102,14 @@ int uv_ip4_addr(const char* ip, int port, struct sockaddr_in* addr) {
...
+  int err = inet_pton(AF_INET, ip, &(addr->sin_addr.s_addr));
```
saghul: Define the variable at the top of the function.

src/uv-common.c
```diff
@@ -102,9 +102,14 @@ int uv_ip4_addr(const char* ip, int port, struct sockaddr_in* addr) {
...
+  if (err == 0) 
+    return errno = EADDRNOTAVAIL;
```
saghul: I'd just return EINVAL here, no need to assign it.

src/uv-common.c
```diff
@@ -102,9 +102,14 @@ int uv_ip4_addr(const char* ip, int port, struct sockaddr_in* addr) {
...
+  else if (err == -1 )
```
saghul: Extraneous space on closing parentheses.

src/uv-common.c
saghul: Also, check uv_inet_pton, you should use that, like uv_ip6_addr does. (not sure if it returns errors, please check that)

src/uv-common.c
saghul: Make that UV_EINVAL.

src/uv-common.c
```diff
@@ -102,9 +102,14 @@ int uv_ip4_addr(const char* ip, int port, struct sockaddr_in* addr) {
...
+    return errno;
```
saghul: You can't return errno here, this function is also executed on Windows, so the error would need to be translated.

saghul:
If you want go forward with this we'd need some tests as well :-)

yorkie:
Yep, updated these tips, check them again, thanks.

\cc @saghul

src/uv-common.c
```diff
@@ -99,12 +99,16 @@ const char* uv_strerror(int err) {
...
+  int err;
   memset(addr, 0, sizeof(*addr));
   addr->sin_family = AF_INET;
   addr->sin_port = htons(port);
-  /* TODO(bnoordhuis) Don't use inet_addr(), no good way to detect errors. */
-  addr->sin_addr.s_addr = inet_addr(ip);
-  return 0;
+  err = uv_inet_pton(AF_INET, ip, &(addr->sin_addr.s_addr));
+  if (err == 0 || err == -1) {
+    return errno;
```
saghul: This won't work on Windows.

src/uv-common.c
```diff
@@ -99,12 +99,16 @@ const char* uv_strerror(int err) {
...
+  } else {
```
saghul: If the returned value is UV_EAFNOSUPPORT (https://github.com/joyent/libuv/blob/master/src/inet.c#L165) you'd return 0, which is not right.

src/uv-common.c
saghul: Ah, wait a sec! The code in src/inet.c doesn't set the errno! If the returned value is != 0 just return err.

yorkie: Could you give more information about this? 
cc @saghul

saghul: Basically, uv_inet_pton doesn't set errno. It's cross platform, you can check the code in src/inet.c, so all you need to do is:

```
return uv_inet_pton(AF_INET, ip, &(addr->sin_addr.s_addr));
```

Because uv_inet_pton already returns 0 in case everything went OK and a UV_EXXX error otherwise.

yorkie: Thanks, fixed them, but sorry about I'm a new guy in windows programming.

saghul: On 12/12/13 8:08 PM, Yorkie wrote:

> In src/uv-common.c:
> 
> >    memset(addr, 0, sizeof(*addr));
> >    addr->sin_family = AF_INET;
> >    addr->sin_port = htons(port);
> > -  /\* TODO(bnoordhuis) Don't use inet_addr(), no good way to detect errors. */
> > -  addr->sin_addr.s_addr = inet_addr(ip);
> > -  return 0;
> > -  err = uv_inet_pton(AF_INET, ip, &(addr->sin_addr.s_addr));
> > -  if (err == 0 || err == -1) {
> > -    return errno;
> 
> Thanks, fixed them, but sorry about I'm a new guy in windows programming.

No worries, I think I also confused you a bit :-/

I think you can apply the same fix to uv_ip6_addr. Then we'd need to 
tests with a few valid and invalid IP addresses :-)

Thanks a lot for working on this btw!

## 

Saúl Ibarra Corretgé
http://bettercallsaghul.com

yorkie: `uv_ip6_addr()` is fixed with https://github.com/yorkie/libuv/commit/23be6d2ba49c67140868970b2307fc54f61bd85e, let me write a test for them later.

yorkie:
New test file `test-ip4-addr.c` has been added: https://github.com/yorkie/libuv/commit/dc7ca396449e82aae9679e7f49a8548d0b3c8784

And for do a correspondence with `test-ip6-addr.c`, also I updated the below in `include/uv-darwin.h`:

``` diff
+#define UV_PLATFORM_HAS_IP4_LINK_LOCAL_ADDRESS
 #define UV_PLATFORM_HAS_IP6_LINK_LOCAL_ADDRESS
```

Hope that it's no side effects!

cc @saghul

saghul:
On 12/12/13 9:02 PM, Yorkie wrote:

> New test file |test-ip4-addr.c| has been added: yorkie@dc7ca39
> https://github.com/yorkie/libuv/commit/dc7ca396449e82aae9679e7f49a8548d0b3c8784
> 
> And for do a correspondence with |test-ip6-addr.c|, I update the below
> in |include/uv-darwin.h|:
> 
> +#define UV_PLATFORM_HAS_IP4_LINK_LOCAL_ADDRESS
>    #define UV_PLATFORM_HAS_IP6_LINK_LOCAL_ADDRESS
> 
> Hope that it's no side effects!
> 
> cc @saghul https://github.com/saghul

That is not needed on IPv4. You can remove that part :-)

## 

Saúl Ibarra Corretgé
http://bettercallsaghul.com

saghul:
The test doesn't really blend well here. IPv4 address don't have a scope id. Instead, just test a few IPv4 addresses with some valid and invalid ones, and also try with a broken address family.

Thanks a lot for your effort!

yorkie:
New `test-ip4-addr.c`: https://github.com/yorkie/libuv/blob/d5003d37d8c1274fd337900b89e4dac2187f7f14/test/test-ip4-addr.c

@saghul , you can check it again, thanks! :)

test/test-ip4-addr.c
```diff
@@ -0,0 +1,51 @@
...
+
+  ASSERT(0 == uv_ip4_addr("127.0.0.1", TEST_PORT, &addr));
+  ASSERT(0 == uv_ip4_addr("255.255.255.255", TEST_PORT, &addr));
+  ASSERT(UV_EINVAL == uv_ip4_addr("255.255.255*000", TEST_PORT, &addr));
+  ASSERT(UV_EINVAL == uv_ip4_addr("255.255.255.256", TEST_PORT, &addr));
+  ASSERT(UV_EINVAL == uv_ip4_addr("2555.0.0.0", TEST_PORT, &addr));
+  ASSERT(UV_EINVAL == uv_ip4_addr("255", TEST_PORT, &addr));
+
+  // for broken address family
+  ASSERT(UV_EAFNOSUPPORT == uv_inet_pton(AF_UNIX, "127.0.0.1",
```
saghul: AF_UNIX won't be available on Windows, lets use an invented number like 42 :-)

test/test-ip4-addr.c
```diff
@@ -0,0 +1,51 @@
...
+ * IN THE SOFTWARE.
+ */
+
+#include "uv.h"
+#include "task.h"
+
+#include <stdio.h>
+#include <string.h>
+
+#ifdef __linux__
```
saghul: What do you need from these includes and why only in Linux?

yorkie: No, I forgot clean them up. one commit again.

saghul:
You need to add the test to the build system and the test list. Here is a very small commit that adds a new test, you can use is as an example: https://github.com/joyent/libuv/commit/15af49a7a0caa37516aa05c31fd84b0c1e952623

yorkie:
Check them again, please!
^.^

saghul:
Almost there! Now you can see the failures by clicking on the CI status. Right now:

```
test/test-ip4-addr.c: In function ‘run_test_ip4_addr’:
test/test-ip4-addr.c:41:3: error: invalid type argument of ‘->’ (have ‘struct sockaddr_in’)
```

yorkie:
Hi @saghul 
`test-ip4-addr.c` is ok, could you check them again? Thanks...!

saghul:
LGTM, I'll have another look tomorrow. Can you squash all commits into a single one with a nice commit message? :-)

yorkie:
Yes, check https://github.com/yorkie/libuv/commit/309e72d81e284366883d12af4f9fb7da6982561e again, please. :)

saghul:
Thanks @yorkie! I reworded the commit log a bit and landed it as https://github.com/joyent/libuv/commit/4ed6496b9ce416ac4bf1257b1eae06e0d4eb479c :sparkles:

yorkie:
:) cheers!

PR #1043, opened 2013-12-13
build: adds support for VS2013.

rioki:


saghul:
Related: https://github.com/joyent/libuv/pull/1012 https://github.com/joyent/libuv/pull/1006 https://github.com/joyent/libuv/pull/1003

IIRC the plan is to get rid of the VS version detection completely.

/cc @piscisaureus

saghul:
@piscisaureus can this be merged until we figure out if we can completely get rid of version detection?

tjfontaine:
@saghul I will spend some time with the build refactory today, but removing all detection should be pretty straight forward, we'll get either @piscisaureus or @sblom for confirmation

saghul:
Great, thanks for confirming @tjfontaine

piscisaureus:
@saghul I don't have 2013 either, but this looks "safe and obvious" so you're welcome to land it.

saghul:
Ok, I'll land it tomorrow if @tjfontaine hasn't found the magic solution thing :-)

rioki:
Sounds good, but unless the changes in GYP are made to use msbuild instead of devenv, the actual build should still be a call to msbuild, like the current version.

saghul:
Thanks @rioki landed as https://github.com/joyent/libuv/commit/a762934aa8c0947ed4d4e1dcbceaeaeb7e765329

PR #1044, opened 2013-12-13
unix: utime -> utimes in uv__fs_utime()

yorkie:


src/unix/fs.c
```diff
@@ -484,10 +484,12 @@ static ssize_t uv__fs_sendfile(uv_fs_t* req) {
 
 
 static ssize_t uv__fs_utime(uv_fs_t* req) {
-  struct utimbuf buf;
-  buf.actime = req->atime;
-  buf.modtime = req->mtime;
-  return utime(req->path, &buf); /* TODO use utimes() where available */
```
indutny, line 490: I think key point was _where available_, is it available on every supported platform? Have you  verified that?

yorkie, line 490: Hi, @indutny, I had done the test on OSX/linux, but not windows. then I think it's a downward compatibility, then I'd verify the other platform except window.

saghul, line 490: Given that this code is located in the unix part, windows doesn't apply. However, utimes might not be available on every unix we support. Unfortunately we don't have a jenkins build slave for every BSD we support.

yorkie, line 490: Could you look this man page: http://unixhelp.ed.ac.uk/CGI/man-cgi?utime+2 ? I'm confused that why unix could not support this?

\cc @indutny @saghul

yorkie, line 490: it's my fault, it's a linux man page, closing it. sorry!

\cc @indutny @saghul

Issue #1045, opened 2013-12-16
spawn_fails test fails for some PATHs


jneu:
On Mac OS X 10.9 (and likely most other unixes using glibc), `make check` fails for certain PATH values with the following error:

```
`spawn_fails` failed: exit code 6
Output from process `spawn_fails`:
Assertion failed in test/test-spawn.c on line 152: UV_ENOENT == uv_spawn(uv_default_loop(), &process, &options)
```

Inside uv_spawn, the child process uses execvp which will often return EACCES instead of ENOENT. One way to reproduce this is as follows:

```
0:$ BAD_DIR=/tmp/REMOVEME/bad_path_dir
0:$ mkdir -p $BAD_DIR
0:$ PATH=$PATH:$BAD_DIR
0:$ echo $PATH
/usr/bin:/bin:/usr/sbin:/sbin:/tmp/REMOVEME/bad_path_dir
0:$ chmod 000 $BAD_DIR
0:$ ls -l $BAD_DIR
ls: bad_path_dir: Permission denied
1:$ make check
...
`spawn_fails` failed: exit code 6
Output from process `spawn_fails`:
Assertion failed in test/test-spawn.c on line 152: UV_ENOENT == uv_spawn(uv_default_loop(), &process, &options)
...
```

While this example is contrived, this happened for me initially when a system directory was added to my PATH which I did not have access to.

PR #1046, opened 2013-12-16
test: allow UV_EACCES in spawn_fails (#1045)

jneu:
Addresses issue #1045.

Allow UV_EACCES as a successful return value when uv_spawn fails.
When the PATH environment variable contains a directory which the
user cannot access, execvp may return EACCES instead of ENOENT.


saghul:
Any way EACCES can be simulated in the test case?

jneu:
I don't think we can reliably. It's OS dependent on when `execvp` will return ENOENT or EACCES.

For example, on my linux system (glibc 2.5), `execvp` will return EACCES if any of the `execve` calls it makes on PATH elements return it. For reference, take a look at the `got_eacces` variable in the glibc source code here:
https://sourceware.org/git/?p=glibc.git;a=blob;f=posix/execvpe.c;h=588e7a71b1a5b507c7628b4ec7a467f4ba78b1d3;hb=HEAD

When testing with my Mac OS X 10.9 system, it appears `execvp` will only return EACCES when `execve` returns EACCES on the _last_ PATH element. The man page is particularly unhelpful:
https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man3/execvp.3.html
Reference the line containing: "... although only the critical error EACCES is really ambiguous ...".

saghul:
Thoughts, @indutny?

indutny:
Have no objections to it. LGTM.

indutny:
Thank you, landed in b44edf3

PR #1047, opened 2013-12-17
windows: improved handling of invalid FDs

orangemocha:
If passed and invalid FD, _get_osfhandle() sets an error code
through errno, not _doserrno. Hence we need to use
SET_REQ_WIN32_ERROR insted of SET_REQ_RESULT.

In debug builds, _get_osfhandle() also raises a superfluous
assert. I implemented a wrapper that disables all asserts
around the call to _get_osfhandle().

This fixes node.js unit tests test-fs-read-stream.js and
test-listen-fd-ebadf.js.

### DIFF
--- a/src/win/core.c
+++ b/src/win/core.c
@@ -26,6 +26,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <crtdbg.h>
 
 #include "uv.h"
 #include "internal.h"
@@ -39,6 +40,32 @@ static uv_loop_t uv_default_loop_;
 /* uv_once intialization guards */
 static uv_once_t uv_init_guard_ = UV_ONCE_INIT;
 static uv_once_t uv_default_loop_init_guard_ = UV_ONCE_INIT;
+
+
+#ifdef _DEBUG
+/* Our crt debug report handler allows us to temporarily disable asserts */
+/* just for the current thread. */
+
+__declspec( thread ) int uv__crt_assert_enabled = TRUE;
+
+static int uv__crt_dbg_report_handler(int report_type, char *message, int *ret_val) {
+  if (uv__crt_assert_enabled || report_type != _CRT_ASSERT)
+    return FALSE;
+  
+  if (ret_val) {
+    /* Set ret_val to 0 to continue with normal execution. */
+    /* Set ret_val to 1 to trigger a breakpoint. */
+
+    if(IsDebuggerPresent())     
+      *ret_val = 1;  
+    else
+      *ret_val = 0;  
+  }
+
+  /* Don't call _CrtDbgReport. */
+  return TRUE;
+}
+#endif
 
 
 static void uv__crt_invalid_parameter_handler(const wchar_t* expression,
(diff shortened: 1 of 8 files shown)


src/win/handle-inl.h
```diff
@@ -161,4 +162,18 @@ INLINE static void uv_process_endgames(uv_loop_t* loop) {
   }
 }
 
+INLINE static HANDLE uv_get_osfhandle(int fd)
```
tjfontaine: If this is purely internal I think the convention is `uv__`

orangemocha:
@tjfontaine Updated to uv__get_osfhandle()

src/win/core.c
```diff
@@ -41,6 +42,33 @@
 static uv_once_t uv_default_loop_init_guard_ = UV_ONCE_INIT;
 
 
+#ifdef _DEBUG
+/* Our crt debug report handler allows us to temporarily disable asserts */
+/* just for the current thread. */
+
+__declspec( thread ) BOOL uv__crt_assert_enabled = TRUE;
```
piscisaureus: prefer to use standard c types and not windows types if you can help it. I think in this case using an int and setting it to 1 would suffice?

src/win/core.c
```diff
@@ -41,6 +42,33 @@
...
+
+static int uv__crt_dbg_report_handler(int report_type, char *message, int *ret_val) {
+  if (uv__crt_assert_enabled || report_type != _CRT_ASSERT)
+    return FALSE;
+  
+  if (ret_val)
+  {
```
piscisaureus: nit: style, put the { on the same line as the if statement.

src/win/internal.h
```diff
@@ -35,6 +35,25 @@
 # define INLINE inline
 #endif
 
+
+#ifdef _DEBUG
+extern __declspec( thread ) BOOL uv__crt_assert_enabled;
+
+#define UV_BEGIN_DISABLE_CRT_ASSERT()                           \
+  {                                                             \
+    BOOL uv__saved_crt_assert_enabled = uv__crt_assert_enabled;  \
```
piscisaureus: backslash alignment looks off

src/win/internal.h
```diff
@@ -35,6 +35,25 @@
...
+    uv__crt_assert_enabled = FALSE;
+  
+
+#define UV_END_DISABLE_CRT_ASSERT()                             \
+    uv__crt_assert_enabled = uv__saved_crt_assert_enabled;       \
```
piscisaureus: ditto

piscisaureus:
Excellent, we were in dire need for a good solution for this problem. Thanks!

A couple of small nits. 
Also I think this may break the mingw build, but that can be fixed later when/if people complain about it.

src/win/core.c
orangemocha: no problem. I hope the TRUE/FALSE defines are still acceptable

orangemocha:
I believe all your suggested corrections have been incorporated. Thank you!

piscisaureus:
Thanks, landed in c0716b3.

PR #1048, opened 2013-12-17
osx: Fix a memory leak in uv_fs_readdir()

alexcrichton:
On OSX, apparently directories with 0 files in them still allocate memory on
the invocation of `scandir`, so if the function returns early it ends up
leaking memory allocated in `scandir`. I have altered uv__fs_readdir to
instead go to the "deallocation exit area" when 0 files are found in the
directory, and continue to return early on a return value of -1.

I went to add a test for this functionality, but it appears that one already
exists (reading an empty directory), so I imagine that the valgrind builds
must only be happening on linux instead of OSX as well? I have confirmed
manually that a program without this fix will infinitely leak memory, and with
this fix the memory usage stays constant.


src/unix/fs.c
```diff
@@ -205,7 +205,7 @@ static int uv__fs_readdir_filter(const struct dirent* dent) {
 
 /* This should have been called uv__fs_scandir(). */
 static ssize_t uv__fs_readdir(uv_fs_t* req) {
-  struct dirent **dents;
+  struct dirent **dents = NULL;
```
saghul: Please put the assignment on a different line.

src/unix/fs.c
```diff
@@ -215,7 +215,9 @@ static ssize_t uv__fs_readdir(uv_fs_t* req) {
 
   n = scandir(req->path, &dents, uv__fs_readdir_filter, alphasort);
 
-  if (n == -1 || n == 0)
+  if (n == 0)
+    goto out; // osx still needs to deallocate some memory
```
saghul: Use C style comments.

saghul:
If n is 0, then we won't free any extra stuff in the out label because the for loop won't be entered, right? What is freed exactly? I seem to be missing something :-)

alexcrichton:
Yes, the `dents` array won't get traversed because `n == 0`, and your guess is as good as mine as to what needs to get freed. Apparently OSX must reserve space in the returned `dents` array for 256 things and them I guess reallocs if there's more? The source of the leak is internal to the `scandir` implementation on OSX (which I have no idea where to find), and the only possible pointer it can return is `dents`, and then free'ing `dents` fixed the leak so I presume that it's the thing that `scandir` is allocating.

Also updated with comments!

saghul:
I think I found it :-) I downloaded the source for Libc from opensource.apple.com, here is the scandir implementation: https://gist.github.com/saghul/8013376

The dirent is allocated here: https://gist.github.com/saghul/8013376#file-scandir-c-L89

And it's returned even if the directory is empty, since this while loop won't run: https://gist.github.com/saghul/8013376#file-scandir-c-L93

So, you seem to be right :-)

@indutny, does it also LGTY?

saghul:
FWIW, I had a look at glibc and though the implementation is different, if the returned dirent is not NULL it needs to be freed as well.

Maybe we should also reword the commit log and remove that comment about OSX. FreeBSD probably has the same implementation. We could say that "Some scandir implementations allocate the dirent struct even if the directory is empty, ...".

alexcrichton:
Amended the message to be a be more general.

saghul:
Ok, I'll wait for @indutny's thumbs up, just in case my reasoning is totally broken (let's hope not ;-) ) but LGTM.

alexcrichton:
Thanks for looking into this and the other libc implementations!

indutny:
LGTM!

indutny:
Landed in v0.10 in 7c6bddb

indutny:
Thank you!

saghul:
Thanks @alexcrichton! :-)

PR #1050, opened 2013-12-20
stream: fix uv__stream_osx_select

indutny:
Relying on `readable`/`writable` when polling will lead to the select
thread spinning in the loop and calling `uv_async_send()`, because
stream may never become not `readable`/`writable`.

We should rely on `uv__io_active()` instead and interruprt select thread
every time it changes.

/cc @saghul see https://github.com/joyent/node/issues/6733#issuecomment-30916143


saghul:
How is it possible that it never becomes "readable"? The stream flags are set in `uv__stream_open`, which is called right after `uv__stream_try_select`.

indutny:
It never becomes not "readable", because we're no checking UV__POLLIN presence.

saghul:
Le me rephrase :-) Before we used to check for the intention, no we check if the stream is _actually_ readable or writable. If that is what we really want, then LGTM.

indutny:
Yeah, that's what we really need :) Otherwise it just loops there and sends `async`s to the main thread.

saghul:
I see. Go for it then! :+1:

indutny:
Thank you, landed in 08cafd0

PR #1052, opened 2013-12-22
osx: Fix a possible segfault in uv__io_poll

alexcrichton:
In our build infrastructure, I've seen a lot of segfaults recently that
were all only happening on OSX. Upon inspecting the coredumps, it
appearded that all segfaults happened at the same instruction, and upon
translating the assembly back to the source, I found that an array could
be indexed with a -1 index before the index was checked to be not -1.

As concrete evidence, here is the situation that I found caused the
segfault.  The instruction in question along with the relevant register
values was:

```
mov    (%r8,%r15,8),%r12

r8  = 0x7fb0ba800000
r15 = 0xffffffffffffffff

r8 + r15 * 8 == 0x7fb0ba7ffff8
```

It appears that the base of loop->watchers was page aligned, and by
going back one word I guess that the page wasn't mapped, causing our
segfaults.


txdv:
in other words, -1 index access on an array

indutny:
Thank you, LGTM!

indutny:
Landed in f166d6d and f6422af

stakach:
@alexcrichton good catch! I was having this issue too, all tests in that particular project now run without segfaulting. Thanks!

PR #1053, opened 2013-12-22
pipe: allow queueing pending handles

indutny:
Call read2_cb multiple times for queued fds.

/cc @saghul @piscisaureus

### DIFF
--- a/include/uv-unix.h
+++ b/include/uv-unix.h
@@ -229,6 +229,7 @@ typedef struct {
   uv_connection_cb connection_cb;                                             \
   int delayed_error;                                                          \
   int accepted_fd;                                                            \
+  int* queued_fds;                                                            \
   UV_STREAM_PRIVATE_PLATFORM_FIELDS                                           \
 
 #define UV_TCP_PRIVATE_FIELDS /* empty */
--- a/src/unix/stream.c
+++ b/src/unix/stream.c
@@ -916,7 +916,7 @@ static void uv__write_callbacks(uv_stream_t* stream) {
 }
 
 
-static uv_handle_type uv__handle_type(int fd) {
+uv_handle_type uv__handle_type(int fd) {
   struct sockaddr_storage ss;
   socklen_t len;
   int type;
--- a/src/unix/stream.c
+++ b/src/unix/stream.c
@@ -919,7 +919,7 @@ static void uv__write_callbacks(uv_stream_t* stream) {
 }
 
 
-uv_handle_type uv__handle_type(int fd) {
+static uv_handle_type uv__handle_type(int fd) {
   struct sockaddr_storage ss;
   socklen_t len;
   int type;
--- a/src/unix/stream.c
+++ b/src/unix/stream.c
@@ -602,6 +602,9 @@ done:
               server->queued_fds + 3,
               server->queued_fds[0]);
     }
+
+    /* Invoke read_cb one more time */
+    uv__io_feed(server->loop, &server->io_watcher);
   } else {
     server->accepted_fd = -1;
     if (err == 0)
@@ -1012,6 +1015,16 @@ static void uv__read(uv_stream_t* stream) {
   char cmsg_space[64];
   int count;
   int err;
+
+  /* Has queued fds */
+  if (stream->accepted_fd >= 0) {
+    uv_buf_t buf = { NULL, 0 };
+    stream->read2_cb((uv_pipe_t*) stream,
+                     0,
+                     &buf,
+                     uv__handle_type(stream->accepted_fd));
+    return;
+  }
 
   stream->flags &= ~UV_STREAM_READ_PARTIAL;
 
--- a/src/unix/stream.c
+++ b/src/unix/stream.c
@@ -119,6 +119,7 @@ void uv__stream_init(uv_loop_t* loop,
   stream->connect_req = NULL;
   stream->shutdown_req = NULL;
   stream->accepted_fd = -1;
+  stream->queued_fds = NULL;
   stream->delayed_error = 0;
   QUEUE_INIT(&stream->write_queue);
   QUEUE_INIT(&stream->write_completed_queue);
@@ -559,6 +560,7 @@ int uv_accept(uv_stream_t* server, uv_stream_t* client) {
   if (server->accepted_fd == -1)
     return -EAGAIN;
 
+  err = 0;
   switch (client->type) {
     case UV_NAMED_PIPE:
     case UV_TCP:
@@ -568,8 +570,7 @@ int uv_accept(uv_stream_t* server, uv_stream_t* client) {
       if (err) {
         /* TODO handle error */
         uv__close(server->accepted_fd);
-        server->accepted_fd = -1;
-        return err;
+        goto done;
       }
       break;
 
@@ -577,8 +578,7 @@ int uv_accept(uv_stream_t* server, uv_stream_t* client) {
       err = uv_udp_open((uv_udp_t*) client, server->accepted_fd);
       if (err) {
         uv__close(server->accepted_fd);
-        server->accepted_fd = -1;
-        return err;
+        goto done;
       }
       break;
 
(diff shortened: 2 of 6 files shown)


indutny:
I'll work on test soon.

indutny:
Nvm, I don't think that it is really possible...

src/unix/stream.c
```diff
@@ -951,6 +973,40 @@ static void uv__stream_eof(uv_stream_t* stream, const uv_buf_t* buf) {
...
+                                       sizeof(*stream->queued_fds));
+      if (stream->queued_fds == NULL)
+        return UV_ENOMEM;
+    }
+  }
+
+  /* Put fd in a queue */
+  stream->queued_fds[2 + queue_offset++] = fd;
+  stream->queued_fds[0] = queue_offset;
+  stream->queued_fds[1] = queue_len;
```
indutny, line 1004: Hm... this write isn't necessary, I'll move it above.

src/unix/stream.c
```diff
@@ -958,6 +1014,17 @@ static void uv__read(uv_stream_t* stream) {
   struct cmsghdr* cmsg;
   char cmsg_space[64];
   int count;
+  int err;
+
+  /* Has queued fds */
+  if (stream->accepted_fd >= 0) {
+    uv_buf_t buf = { NULL, 0 };
```
saghul, line 1021: Hum, I thought now we could call write2 and send some data and a handle. Wouldn't you be ignoring the received data here?

saghul:
The approach LGTM. Does Windows support any of this?

src/unix/stream.c
indutny, line 1021: Hm... right I would, perhaps, I just need to remove return.

txdv:
So does it work or not?

indutny:
@saghul I don't think that windows needs it, but it does support receiving handles on IPC pipe.

src/unix/stream.c
saghul, line 1021: Could be yeah.

saghul, line 1021: Perhaps the buf can be defined as static?

src/unix/stream.c
```diff
@@ -958,6 +1014,17 @@ static void uv__read(uv_stream_t* stream) {
   struct cmsghdr* cmsg;
   char cmsg_space[64];
   int count;
+  int err;
+
+  /* Has queued fds */
+  if (stream->accepted_fd >= 0) {
```
saghul, line 1020: I like stream->accepted_fd != -1 better :-)

saghul:
> @saghul I don't think that windows needs it, but it does support receiving handles on IPC pipe.

I mean, if there is any way to get more than one at the same time.

indutny:
@saghul I have no idea, but I think no. Otherwise @piscisaureus would had told me about it yesterday.

indutny:
Thank you, landed in 08aeaf6

saghul:
No test?

indutny:
@saghul yep, it's quite hard to get it queuing handles...

saghul:
Yeah. I believe you then :-)

src/unix/stream.c
```diff
@@ -951,6 +973,40 @@ static void uv__stream_eof(uv_stream_t* stream, const uv_buf_t* buf) {
...
+    queue_offset = stream->queued_fds[0];
+    queue_len = stream->queued_fds[1];
+
+    /* Grow */
+    if (queue_offset == queue_len) {
+      queue_len += 8;
+      stream->queued_fds = realloc(stream->queued_fds,
+                                   (queue_len + 2) *
+                                       sizeof(*stream->queued_fds));
+      if (stream->queued_fds == NULL)
```
bnoordhuis, line 996: Leaks memory.

src/unix/stream.c
```diff
@@ -1493,6 +1563,8 @@ void uv__stream_close(uv_stream_t* handle) {
     handle->accepted_fd = -1;
   }
 
+  free(handle->queued_fds);
```
bnoordhuis, line 1566: Why don't you NULL it?  Helps catch use-after-free bugs.

include/uv-unix.h
```diff
@@ -229,6 +229,7 @@ struct uv__work {
   uv_connection_cb connection_cb;                                             \
   int delayed_error;                                                          \
   int accepted_fd;                                                            \
+  int* queued_fds;                                                            \
```
bnoordhuis, line 232: I don't really understand why you didn't use a (possibly private) struct for this, the queue's logic is awfully implicit now.

src/unix/stream.c
```diff
@@ -586,9 +586,31 @@ int uv_accept(uv_stream_t* server, uv_stream_t* client) {
...
+
+    /* All read, free */
+    if (--server->queued_fds[0] == 0) {
+      free(server->queued_fds);
+      server->queued_fds = NULL;
+    } else {
+      /* Shift rest */
+      memmove(server->queued_fds + 2,
+              server->queued_fds + 3,
+              server->queued_fds[0]);
```
bnoordhuis, line 603: This memmove() doesn't do what you think it does.

indutny, line 603: Oh gosh, I forgot `sizeof()`

bnoordhuis:
I gather that the purpose of this change is to handle recvmsg()s that return multiple file descriptors?  How about changing the cmsg_space buffer so it only has room for a single file descriptor?

indutny:
Can't say I wasn't thinking about it, but what if some other message will sneak in?

indutny:
Temporarly reverted.

indutny:
@bnoordhuis @saghul pushed fixes. let's review and land it!

indutny:
@bnoordhuis additional comment about cmsg_space, what if not a libuv is writing to the other side of the pipe? I think we would support it this way, but definitely won't your way.

bnoordhuis:
You mean what if some non-libuv process is sending multiple file descriptors in a message but libuv's cmsg receive buffer is only large enough for one?  It depends on the platform:
- Linux returns the first file descriptor and closes the others.
- OS X returns the first file descriptor and closes the others.  cmsg->cmsg_len is the number of bytes that the kernel would have needed to return all file descriptors, not the actual number of bytes written (in other words, you have to be careful not to overrun the receive buffer.)
- OpenBSD fails with EMSGSIZE and closes all file descriptors.

I don't know what FreeBSD, NetBSD and Solaris do.  I guess the point is that you can't win: if the receive buffer is too small, you lose some or all file descriptors.  If you make it too big, you become vulnerable to (possibly unwitting) EMFILE denial-of-service attacks.

indutny:
Well, it couldn't be a true denial-of-service since it's inter-process communication anyway, and we're in trustful environment. I don't think that we should make the buffer significantly larger or even making it bigger at all. I just think that we should handle such cases and be able to queue file descriptors for a bit. This makes clustering in C on solaris much more simpler and faster.

include/uv-unix.h
jclulow, line 232: Yeah, I totally agree with @bnoordhuis -- this should be a `struct`.  Doing it with the int array requires surprising code, and is not self-describing or easily debuggable.

indutny, line 232: @jclulow please take a look at https://github.com/joyent/libuv/pull/1056/files

bnoordhuis:
> I just think that we should handle such cases and be able to queue file descriptors for a bit.

Yeah, but it will only ever a so-so solution because it breaks down the moment you receive more file descriptors than anticipated (in platform-specific ways no less, as noted above.)

There's also the issue that your PR doesn't quite work as advertised because libuv still only pulls the first file descriptor from the SCM_RIGHTS message, the others are leaked.  (Long-standing issue that I never got around to fixing.)

indutny:
@bnoordhuis may I ask you to continue on https://github.com/joyent/libuv/pull/1056/files ? Also, I don't see why it leaks excessive fds right now... may you shed some light on it, please?

bnoordhuis:
> Also, I don't see why it leaks excessive fds right now... may you shed some light on it, please?

The struct msghdr that's passed to recvmsg() has msg_controllen set to 64.  That's big enough to receive 11 or 12 file descriptors but libuv only reads the first one.

Put another way, it scans for the SCM_RIGHTS message but when it finds it, it only considers the first data element.  It needs to check cmsg->cmsg_len.

PR #1055, opened 2013-12-24
Add readme link to all projects that use libuv.

txdv:
Gives people who read the readme a link to all projects that use libuv.

Furthermore allows people to add their own projects(In the page).


txdv:
The subsystem is a readme file, I don't think this is a subsystem.

I already signed the CLA, I already contributed to this project.

saghul:
Subsystem is set to doc in these cases IIRC.

txdv:
Damn it, I tried to rename it...

Issue #461, opened 2012-06-11
read_start and read2_start should be mutually exclusive


piscisaureus:

saghul:
Closing, we no longer have read2_sart after #1056 landed :-)

PR #1056, opened 2013-12-24
Feature/pipe pending handles 2

indutny:
Github isn't updating previous PR because it was closed.

/cc @jclulow @bnoordhuis @saghul

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -376,6 +376,9 @@ typedef void (*uv_read_cb)(uv_stream_t* stream,
  * Just like the uv_read_cb except that if the pending parameter is true
  * then you can use uv_accept() to pull the new handle into the process.
  * If no handle is pending then pending will be UV_UNKNOWN_HANDLE.
+ *
+ * NOTE: The buffer may be a null buffer if multiple fds were accepted and
+ * read2_cb is called for pending ones.
  */
 typedef void (*uv_read2_cb)(uv_pipe_t* pipe,
                             ssize_t nread,
@@ -1147,16 +1150,6 @@ UV_EXTERN void uv_pipe_connect(uv_connect_t* req, uv_pipe_t* handle,
  * is waiting for connections.
  */
 UV_EXTERN void uv_pipe_pending_instances(uv_pipe_t* handle, int count);
-
-/*
- * This call could be used to determine if IPC pipe has any pending handles to
- * accept.
- *
- * NOTE: You should call `uv_accept()` from read2_cb anyway, if this function
- * returns 1 - `uv_accept()` may be called for another time.
- */
-UV_EXTERN int uv_pipe_has_pending_handles(uv_pipe_t* handle,
-                                          uv_handle_type* type);
 
 
 /*
--- a/include/uv.h
+++ b/include/uv.h
@@ -1148,6 +1148,15 @@ UV_EXTERN void uv_pipe_connect(uv_connect_t* req, uv_pipe_t* handle,
  */
 UV_EXTERN void uv_pipe_pending_instances(uv_pipe_t* handle, int count);
 
+/*
+ * This call could be used to determine if IPC pipe has any pending handles to
+ * accept.
+ *
+ * NOTE: You should call `uv_accept()` from read2_cb anyway, if this function
+ * returns 1 - `uv_accept()` may be called for another time.
+ */
+UV_EXTERN int uv_pipe_has_pending_handles(uv_pipe_t* handle);
+
 
 /*
  * uv_poll_t is a subclass of uv_handle_t.
--- a/include/uv.h
+++ b/include/uv.h
@@ -1155,7 +1155,8 @@ UV_EXTERN void uv_pipe_pending_instances(uv_pipe_t* handle, int count);
  * NOTE: You should call `uv_accept()` from read2_cb anyway, if this function
  * returns 1 - `uv_accept()` may be called for another time.
  */
-UV_EXTERN int uv_pipe_has_pending_handles(uv_pipe_t* handle);
+UV_EXTERN int uv_pipe_has_pending_handles(uv_pipe_t* handle,
+                                          uv_handle_type* type);
 
 
 /*
--- a/src/unix/pipe.c
+++ b/src/unix/pipe.c
@@ -216,6 +216,10 @@ void uv_pipe_pending_instances(uv_pipe_t* handle, int count) {
 }
 
 
-int uv_pipe_has_pending_handles(uv_pipe_t* handle) {
-  return handle->accepted_fd != -1;
+int uv_pipe_has_pending_handles(uv_pipe_t* handle, uv_handle_type* type) {
+  if (handle->accepted_fd == -1)
+    return 0;
+
+  *type = uv__handle_type(handle->accepted_fd);
+  return 1;
 }
--- a/src/unix/pipe.c
+++ b/src/unix/pipe.c
@@ -214,3 +214,8 @@ out:
 
 void uv_pipe_pending_instances(uv_pipe_t* handle, int count) {
 }
+
+
+int uv_pipe_has_pending_handles(uv_pipe_t* handle) {
+  return handle->accepted_fd != -1;
+}
--- a/src/unix/pipe.c
+++ b/src/unix/pipe.c
@@ -214,12 +214,3 @@ out:
 
 void uv_pipe_pending_instances(uv_pipe_t* handle, int count) {
 }
-
-
-int uv_pipe_has_pending_handles(uv_pipe_t* handle, uv_handle_type* type) {
-  if (handle->accepted_fd == -1)
-    return 0;
-
-  *type = uv__handle_type(handle->accepted_fd);
-  return 1;
-}
--- a/src/unix/stream.c
+++ b/src/unix/stream.c
@@ -56,6 +56,14 @@ struct uv__stream_select_s {
 };
 #endif /* defined(__APPLE__) */
 
+typedef struct uv__stream_queued_fds_s uv__stream_queued_fds_t;
+
+struct uv__stream_queued_fds_s {
+  int size;
+  int offset;
+  int fds[1];
+};
+
 static void uv__stream_connect(uv_stream_t*);
 static void uv__write(uv_stream_t* stream);
 static void uv__read(uv_stream_t* stream);
@@ -553,6 +561,7 @@ void uv__server_io(uv_loop_t* loop, uv__io_t* w, unsigned int events) {
 
 int uv_accept(uv_stream_t* server, uv_stream_t* client) {
   int err;
+  uv__stream_queued_fds_t* queued_fds;
 
   /* TODO document this */
   assert(server->loop == client->loop);
@@ -589,18 +598,20 @@ int uv_accept(uv_stream_t* server, uv_stream_t* client) {
 done:
   /* Process queued fds */
   if (server->queued_fds != NULL) {
+    queued_fds = server->queued_fds;
+
     /* Read first */
-    server->accepted_fd = server->queued_fds[2];
+    server->accepted_fd = queued_fds->fds[0];
 
     /* All read, free */
-    if (--server->queued_fds[0] == 0) {
-      free(server->queued_fds);
+    if (--queued_fds->offset == 0) {
+      free(queued_fds);
       server->queued_fds = NULL;
     } else {
       /* Shift rest */
-      memmove(server->queued_fds + 2,
-              server->queued_fds + 3,
-              server->queued_fds[0]);
+      memmove(queued_fds->fds,
+              queued_fds->fds + 1,
+              queued_fds->offset * sizeof(*queued_fds->fds));
     }
 
     /* Invoke read_cb one more time */
@@ -974,34 +985,33 @@ static void uv__stream_eof(uv_stream_t* stream, const uv_buf_t* buf) {
 
 
 static int uv__stream_queue_fd(uv_stream_t* stream, int fd) {
-  int queue_offset;
-  int queue_len;
+  uv__stream_queued_fds_t* queued_fds;
+  int queue_size;
 
-  if (stream->queued_fds == NULL) {
-    queue_offset = 0;
-    queue_len = 8;
-    stream->queued_fds = malloc((queue_len + 2) * sizeof(*stream->queued_fds));
-    if (stream->queued_fds == NULL)
+  queued_fds = stream->queued_fds;
+  if (queued_fds == NULL) {
+    queue_size = 8;
+    queued_fds = malloc((queue_size - 1) * sizeof(*queued_fds->fds) +
+                        sizeof(*queued_fds));
+    if (queued_fds == NULL)
       return UV_ENOMEM;
-  } else {
-    queue_offset = stream->queued_fds[0];
-    queue_len = stream->queued_fds[1];
+    queued_fds->size = queue_size;
+    queued_fds->offset = 0;
+    stream->queued_fds = queued_fds;
 
     /* Grow */
-    if (queue_offset == queue_len) {
-      queue_len += 8;
-      stream->queued_fds = realloc(stream->queued_fds,
-                                   (queue_len + 2) *
-                                       sizeof(*stream->queued_fds));
-      if (stream->queued_fds == NULL)
-        return UV_ENOMEM;
-    }
+  } else if (queued_fds->size == queued_fds->offset) {
+    queued_fds->size += 8;
+    queued_fds = realloc(queued_fds,
+                         (queued_fds->size - 1) * sizeof(*queued_fds->fds) +
+                             sizeof(*queued_fds));
+    if (queued_fds == NULL)
+      return UV_ENOMEM;
+    stream->queued_fds = queued_fds;
   }
 
   /* Put fd in a queue */
-  stream->queued_fds[2 + queue_offset++] = fd;
-  stream->queued_fds[0] = queue_offset;
-  stream->queued_fds[1] = queue_len;
+  queued_fds->fds[queued_fds->offset++] = fd;
 
   return 0;
 }
@@ -1526,6 +1536,9 @@ int uv___stream_fd(uv_stream_t* handle) {
 
 
 void uv__stream_close(uv_stream_t* handle) {
+  int i;
+  uv__stream_queued_fds_t* queued_fds;
+
 #if defined(__APPLE__)
   /* Terminate select loop first */
   if (handle->select != NULL) {
@@ -1563,7 +1576,14 @@ void uv__stream_close(uv_stream_t* handle) {
     handle->accepted_fd = -1;
   }
 
-  free(handle->queued_fds);
+  /* Close all queued fds */
+  if (handle->queued_fds != NULL) {
+    queued_fds = handle->queued_fds;
+    for (i = 0; i < queued_fds->offset; i++)
+      close(queued_fds->fds[i]);
+    free(handle->queued_fds);
+    handle->queued_fds = NULL;
+  }
 
   assert(!uv__io_active(&handle->io_watcher, UV__POLLIN | UV__POLLOUT));
 }
--- a/src/unix/stream.c
+++ b/src/unix/stream.c
@@ -916,7 +916,7 @@ static void uv__write_callbacks(uv_stream_t* stream) {
 }
 
 
-static uv_handle_type uv__handle_type(int fd) {
+uv_handle_type uv__handle_type(int fd) {
   struct sockaddr_storage ss;
   socklen_t len;
   int type;
--- a/src/unix/stream.c
+++ b/src/unix/stream.c
@@ -1005,8 +1005,15 @@ static int uv__stream_queue_fd(uv_stream_t* stream, int fd) {
     queued_fds = realloc(queued_fds,
                          (queued_fds->size - 1) * sizeof(*queued_fds->fds) +
                              sizeof(*queued_fds));
-    if (queued_fds == NULL)
+
+    /*
+     * Allocation failure, report back.
+     * NOTE: if it is fatal - sockets will be closed in uv__stream_close
+     */
+    if (queued_fds == NULL) {
+      queued_fds->size -= 8;
       return UV_ENOMEM;
+    }
     stream->queued_fds = queued_fds;
   }
 
--- a/src/unix/stream.c
+++ b/src/unix/stream.c
@@ -919,7 +919,7 @@ static void uv__write_callbacks(uv_stream_t* stream) {
 }
 
 
-uv_handle_type uv__handle_type(int fd) {
+static uv_handle_type uv__handle_type(int fd) {
   struct sockaddr_storage ss;
   socklen_t len;
   int type;
--- a/src/unix/stream.c
+++ b/src/unix/stream.c
@@ -602,6 +602,9 @@ done:
               server->queued_fds + 3,
               server->queued_fds[0]);
     }
+
+    /* Invoke read_cb one more time */
+    uv__io_feed(server->loop, &server->io_watcher);
   } else {
     server->accepted_fd = -1;
     if (err == 0)
@@ -1012,6 +1015,16 @@ static void uv__read(uv_stream_t* stream) {
   char cmsg_space[64];
   int count;
   int err;
+
+  /* Has queued fds */
+  if (stream->accepted_fd >= 0) {
+    uv_buf_t buf = { NULL, 0 };
+    stream->read2_cb((uv_pipe_t*) stream,
+                     0,
+                     &buf,
+                     uv__handle_type(stream->accepted_fd));
+    return;
+  }
 
   stream->flags &= ~UV_STREAM_READ_PARTIAL;
 
--- a/src/unix/stream.c
+++ b/src/unix/stream.c
@@ -119,6 +119,7 @@ void uv__stream_init(uv_loop_t* loop,
   stream->connect_req = NULL;
   stream->shutdown_req = NULL;
   stream->accepted_fd = -1;
+  stream->queued_fds = NULL;
   stream->delayed_error = 0;
   QUEUE_INIT(&stream->write_queue);
   QUEUE_INIT(&stream->write_completed_queue);
@@ -559,6 +560,7 @@ int uv_accept(uv_stream_t* server, uv_stream_t* client) {
   if (server->accepted_fd == -1)
     return -EAGAIN;
 
+  err = 0;
   switch (client->type) {
     case UV_NAMED_PIPE:
     case UV_TCP:
@@ -568,8 +570,7 @@ int uv_accept(uv_stream_t* server, uv_stream_t* client) {
       if (err) {
         /* TODO handle error */
         uv__close(server->accepted_fd);
-        server->accepted_fd = -1;
-        return err;
+        goto done;
       }
       break;
 
@@ -577,8 +578,7 @@ int uv_accept(uv_stream_t* server, uv_stream_t* client) {
       err = uv_udp_open((uv_udp_t*) client, server->accepted_fd);
       if (err) {
         uv__close(server->accepted_fd);
-        server->accepted_fd = -1;
-        return err;
+        goto done;
       }
       break;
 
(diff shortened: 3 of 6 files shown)


indutny:
<blockquote>
Put another way, it scans for the SCM_RIGHTS message but when it finds it, it only considers the first data element. It needs to check cmsg->cmsg_len
</blockquote>

@bnoordhuis - good point, I'll make walk all fds in cmsg tomorrow.

indutny:
@bnoordhuis addressed that thing.

src/unix/stream.c
```diff
@@ -56,6 +56,14 @@ struct uv__stream_select_s {
 };
 #endif /* defined(__APPLE__) */
 
+typedef struct uv__stream_queued_fds_s uv__stream_queued_fds_t;
+
+struct uv__stream_queued_fds_s {
+  int size;
+  int offset;
```
bnoordhuis: Don't use signed types for fields that always contain values >= 0.

src/unix/stream.c
```diff
@@ -586,9 +595,33 @@ int uv_accept(uv_stream_t* server, uv_stream_t* client) {
...
+done:
+  /* Process queued fds */
+  if (server->queued_fds != NULL) {
+    queued_fds = server->queued_fds;
+
+    /* Read first */
+    server->accepted_fd = queued_fds->fds[0];
+
+    /* All read, free */
+    if (--queued_fds->offset == 0) {
```
bnoordhuis, line 600: Should probably be preceded by an assert that checks that queued_fds->offset > 0.

src/unix/stream.c
```diff
@@ -586,9 +595,33 @@ int uv_accept(uv_stream_t* server, uv_stream_t* client) {
...
+      server->queued_fds = NULL;
+    } else {
+      /* Shift rest */
+      memmove(queued_fds->fds,
+              queued_fds->fds + 1,
+              queued_fds->offset * sizeof(*queued_fds->fds));
+    }
+
+    /* Invoke read_cb one more time */
+    uv__io_feed(server->loop, &server->io_watcher);
```
bnoordhuis: Borderline incorrect: uv__io_feed() queues a UV__POLLOUT event but a server handle is only primed for UV__POLLIN.

indutny: Oh gosh, you're right.

src/unix/stream.c
```diff
@@ -951,6 +984,46 @@ static void uv__stream_eof(uv_stream_t* stream, const uv_buf_t* buf) {
...
+  uv__stream_queued_fds_t* queued_fds;
+  int queue_size;
+
+  queued_fds = stream->queued_fds;
+  if (queued_fds == NULL) {
+    queue_size = 8;
+    queued_fds = malloc((queue_size - 1) * sizeof(*queued_fds->fds) +
+                        sizeof(*queued_fds));
+    if (queued_fds == NULL)
+      return UV_ENOMEM;
```
bnoordhuis: Convention is to return -ENOMEM.

src/unix/stream.c
```diff
@@ -951,6 +984,46 @@ static void uv__stream_eof(uv_stream_t* stream, const uv_buf_t* buf) {
...
+    queued_fds->size += 8;
+    queued_fds = realloc(queued_fds,
+                         (queued_fds->size - 1) * sizeof(*queued_fds->fds) +
+                             sizeof(*queued_fds));
+
+    /*
+     * Allocation failure, report back.
+     * NOTE: if it is fatal - sockets will be closed in uv__stream_close
+     */
+    if (queued_fds == NULL) {
```
bnoordhuis: Leaks memory.

indutny: No, it doesn't

src/unix/stream.c
```diff
@@ -958,6 +1031,17 @@ static void uv__read(uv_stream_t* stream) {
...
   int count;
+  int err;
+
+  /* Has queued fds */
+  if (stream->accepted_fd >= 0) {
+    uv_buf_t buf = { NULL, 0 };
+    stream->read2_cb((uv_pipe_t*) stream,
+                     0,
+                     &buf,
+                     uv__handle_type(stream->accepted_fd));
```
bnoordhuis: Seems like a terrible change because you're putting the burden on the user to always check if buf.base == NULL.  It's unlikely that existing code does that, so you'll be breaking pretty much everything that's already out there.

It's made worse because receiving multiple file descriptors is an unlikely event so any bugs will be hidden for a long time and difficult to reproduce.  If you put together this PR for your own convenience, then I suggest you drop it.

src/unix/stream.c
```diff
@@ -1046,17 +1130,24 @@ static void uv__read(uv_stream_t* stream) {
...
-
             /* silence aliasing warning */
-            {
-              void* pv = CMSG_DATA(cmsg);
-              int* pi = pv;
-              stream->accepted_fd = *pi;
+            void* pv = CMSG_DATA(cmsg);
+            int* pi = pv;
+            unsigned int i;
+            unsigned int count = (cmsg->cmsg_len - CMSG_LEN(0)) / sizeof(*pi);
```
bnoordhuis: You cannot depend on count actually being the number of file descriptors in the message.  As I mentioned in the other PR, it doesn't work that way on OS X (and it's probably the same on FreeBSD) - it reports the _original_ number of file descriptors, which may be more than what fits in the receive buffer.

src/unix/stream.c
indutny: It was reproducible every time when I did C clustering benchmarks on a smartos server. The only alternative to that is to ACK all incoming fds in a child processes, which is a bit of burden and waste of resources.

src/unix/stream.c
```diff
@@ -1046,17 +1130,24 @@ static void uv__read(uv_stream_t* stream) {
...
+
+            for (i = 0; i < count; i++) {
+              /* Already has accepted fd, queue now */
+              if (stream->accepted_fd != -1) {
+                err = uv__stream_queue_fd(stream, pi[i]);
+                if (err != 0) {
+                  uv__stream_read_cb(stream, err, NULL, UV_UNKNOWN_HANDLE);
+                  return;
```
bnoordhuis: Leaks the remaining file descriptors.

You should probably set the close-on-exec flag on the file descriptors first.  Linux has a MSG_CMSG_CLOEXEC flag that you can pass to recvmsg() which does that atomically but it wasn't added until 2.6.22 or 2.6.23.  You'll need a fallback path for when recvmsg() returns EINVAL.

src/unix/stream.c
```diff
@@ -1493,6 +1587,15 @@ void uv__stream_close(uv_stream_t* handle) {
     handle->accepted_fd = -1;
   }
 
+  /* Close all queued fds */
+  if (handle->queued_fds != NULL) {
+    queued_fds = handle->queued_fds;
+    for (i = 0; i < queued_fds->offset; i++)
+      close(queued_fds->fds[i]);
```
bnoordhuis: uv__close()

src/unix/stream.c
bnoordhuis: Oh, you're right.  I can see it now, I just discovered GH's new expand functionality.

src/unix/stream.c
bnoordhuis: I don't understand what you mean.  My point is that you're changing the API in a subtle way that's bound to break existing programs.  That's not a good thing.

bnoordhuis:
If you go through with this, it _really_ needs some regression tests.  Should be easy: create file descriptors with socketpair(), feed one to uv_pipe_open() and call sendmsg() on the other.  You'll want to cover several cases: one file descriptor, several and more than fit in the receive buffer.  Keep in mind that OpenBSD will fail with EMSGSIZE if number_of_fds > sizeof_receive_buffer.

src/unix/stream.c
txdv: Size is always a positive too

src/unix/stream.c
indutny: It can already be a null-buffer, when nread is UV_EOF

src/unix/stream.c
indutny: I think I can just close them.

src/unix/stream.c
indutny: Would using `msg_controllen` fix it?

indutny:
@bnoordhuis all fixed, working on a test.

indutny:
@bnoordhuis added test. @saghul please review it too!

src/unix/stream.c
```diff
@@ -951,13 +985,112 @@ static void uv__stream_eof(uv_stream_t* stream, const uv_buf_t* buf) {
...
+static int uv__stream_recv_cmsg(uv_stream_t* stream, struct msghdr* msg) {
+  struct cmsghdr *cmsg;
+  void* pv;
+  int* pi;
+  unsigned int i;
+  unsigned int count;
+  int err;
+
+  /*
+   * Some implementations can send multiple file descriptors in a
```
saghul: I see you use CMSG_NXTHDR, is this comment still relevant?

indutny: No, I'll remove it.

saghul:
I had a look at it, but TBH I don't think I know enough about cmsg\* to make any meaningful comment.

re: passing a NULL buf: in pyuv I assume the buffer base is not NULL if nread is >= 0, so this code would break my assumption. Not sure about what others do. Maybe the buffer should contain the empty string instead?

indutny:
@saghul the zero-length buffer won't work and may break existing things too, considering that users may think that the buffer has came from `alloc_cb`.

saghul:
Ah, indeed. Hum I don't see how we can get away with this without "breaking" something :-S

indutny:
Yeah... Let's try discussing it this weekend or next week on irc. I'll write a message on the mailing list to get some feedback from users.

indutny:
@jclulow @bnoordhuis @saghul done

include/uv-win.h
```diff
@@ -424,10 +424,11 @@ typedef int (WSAAPI* LPFN_WSARECVFROM)
   uv_write_t ipc_header_write_req;                                            \
   int ipc_pid;                                                                \
   uint64_t remaining_ipc_rawdata_bytes;                                       \
-  unsigned char reserved[sizeof(void*)];                                      \
```
jclulow, line 427: What was this reserved space for?  Is removing it OK?

indutny, line 427: It is safe to remove it in unstable version. According to b496c1221c3816f06a3fcd3e8c3efedcf6eb8445 it was introduced for a future feature implementation in a stable branch.

@piscisaureus wanna weigh in?

piscisaureus, line 427: That used to be another field that we removed on a stable branch. So we added a reserved spot to not break the ABI
(I see now that replacing `void*` by `char[sizeof void]` only works accidentally but was not right.)

src/win/pipe.c
```diff
@@ -31,6 +31,17 @@
 #include "stream-inl.h"
 #include "req-inl.h"
 
+typedef struct uv__ipc_queue_item_s uv__ipc_queue_item_t;
+
+struct uv__ipc_queue_item_s {
+  /*
+   * NOTE: It is important for socket_info to be the first field,
```
piscisaureus, line 38: I don't get this. You mean you're going to assign a `struct uv__ipc_queue_item_s*` to a `WSAPROTOCOL_INFOW*` ?
Why not use container_of (or what have you on windows) then?

src/win/pipe.c
```diff
@@ -358,11 +373,22 @@ void uv_pipe_endgame(uv_loop_t* loop, uv_pipe_t* handle) {
     assert(!(handle->flags & UV_HANDLE_CLOSED));
 
     if (handle->flags & UV_HANDLE_CONNECTION) {
+      /* Free pending socket */
       if (handle->pending_ipc_info.socket_info) {
         free(handle->pending_ipc_info.socket_info);
         handle->pending_ipc_info.socket_info = NULL;
       }
 
+      /* Free rest pending sockets */
```
piscisaureus: Grammar...

src/win/pipe.c
```diff
@@ -358,11 +373,22 @@ void uv_pipe_endgame(uv_loop_t* loop, uv_pipe_t* handle) {
...
+      /* XXX: Need to free sockets themselves? */
```
piscisaureus: Yes you need to "materialize" the socket with `WSASocket`, then close it with `CloseSocket`.

src/win/pipe.c
```diff
@@ -1345,6 +1372,34 @@ static void uv_pipe_read_error_or_eof(uv_loop_t* loop, uv_pipe_t* handle,
 }
 
 
+void uv__pipe_insert_pending_socket(uv_pipe_t* handle,
+                                    WSAPROTOCOL_INFOW* info,
+                                    int tcp_connection) {
+  WSAPROTOCOL_INFOW* socket_info;
+  uv__ipc_queue_item_t* item;
+
+  if (handle->pending_ipc_info.socket_info == NULL) {
```
piscisaureus: I don't agree with having both a queue _and_ having the queue 'head' data stored in separate fields.
Please drop the `handle->pending_ipc_info` field and replace it with a queue head.

src/win/pipe.c
```diff
@@ -1745,3 +1778,21 @@ int uv_pipe_open(uv_pipe_t* pipe, uv_file file) {
...
   return 0;
 }
+
+
+int uv_pipe_pending_count(uv_pipe_t* handle) {
+  assert(handle->ipc);
+  if (handle->pending_ipc_info.socket_info == NULL)
+    return 0;
+
+  return 1 + handle->pending_ipc_info.queue_len;
```
piscisaureus: And with `handle->pending_ipc_info` being replaced by a queue, you don't have to add 1 here.

src/win/pipe.c
indutny, line 38: No, because it'll be passed to `free()`.

src/win/pipe.c
indutny: It is somewhat complicated if we don't want to export the `uv__ipc_queue_item_t` declaration. Also this was done to make it work without allocations in the most of the cases.

src/unix/stream.c
```diff
@@ -552,6 +552,7 @@ void uv__server_io(uv_loop_t* loop, uv__io_t* w, unsigned int events) {
 
 int uv_accept(uv_stream_t* server, uv_stream_t* client) {
   int err;
+  uv__stream_queued_fds_t* queued_fds;
```
jclulow: This variable doesn't seem to be used outside of the `if` block starting at L591.  It would be clearer if it were declared down there.

src/unix/stream.c
```diff
@@ -931,23 +953,102 @@ static uv_handle_type uv__handle_type(int fd) {
...
   if (!uv__io_active(&stream->io_watcher, UV__POLLOUT))
     uv__handle_stop(stream);
-  uv__stream_read_cb(stream, UV_EOF, buf, UV_UNKNOWN_HANDLE);
+  stream->read_cb(stream, UV_EOF, buf);
+}
+
+
+static int uv__stream_queue_fd(uv_stream_t* stream, int fd) {
+  uv__stream_queued_fds_t* queued_fds;
+  unsigned int queue_size;
```
jclulow, line 968: It's confusing to use `queue_size` for the `malloc` case, but not the `realloc` case.  Either move it inside the `if`, or use it consistently for both branches.

src/unix/stream.c
```diff
@@ -931,23 +953,102 @@ static uv_handle_type uv__handle_type(int fd) {
...
+    queued_fds = realloc(queued_fds,
+                         (queued_fds->size - 1) * sizeof(*queued_fds->fds) +
+                             sizeof(*queued_fds));
+
+    /*
+     * Allocation failure, report back.
+     * NOTE: if it is fatal - sockets will be closed in uv__stream_close
+     */
+    if (queued_fds == NULL) {
+      queued_fds->size -= 8;
```
jclulow: If `queue_fds == NULL` here, surely this will segfault.

src/unix/stream.c
indutny, line 968: Decided to use it in both branches

src/win/pipe.c
indutny: Please let me know you have any suggestions for embedding internal struct in a public one.

indutny:
@jclulow I think I have fixed all nits, thank you! Could you please take a look?

indutny:
@piscisaureus pushed windows update

src/win/pipe.c
```diff
@@ -358,10 +370,30 @@ void uv_pipe_endgame(uv_loop_t* loop, uv_pipe_t* handle) {
...
+      while (!QUEUE_EMPTY(&handle->pending_ipc_info.queue)) {
+        QUEUE* q;
+        SOCKET socket;
+
+        q = QUEUE_HEAD(&handle->pending_ipc_info.queue);
+        QUEUE_REMOVE(q);
+        item = QUEUE_DATA(q, uv__ipc_queue_item_t, member);
+
+        /* Materialize socket and close it */
+        socket = WSASocketW(AF_INET,
```
piscisaureus: That should be `FROM_PROTOCOL_INFO`

src/win/pipe.c
```diff
@@ -358,10 +370,30 @@ void uv_pipe_endgame(uv_loop_t* loop, uv_pipe_t* handle) {
...
+                            SOCK_STREAM,
```
piscisaureus: ditto

src/win/pipe.c
```diff
@@ -358,10 +370,30 @@ void uv_pipe_endgame(uv_loop_t* loop, uv_pipe_t* handle) {
...
+                            IPPROTO_IP,
```
piscisaureus: ditto

src/win/pipe.c
```diff
@@ -358,10 +370,30 @@ void uv_pipe_endgame(uv_loop_t* loop, uv_pipe_t* handle) {
     assert(!(handle->flags & UV_HANDLE_CLOSED));
 
     if (handle->flags & UV_HANDLE_CONNECTION) {
-      if (handle->pending_ipc_info.socket_info) {
-        free(handle->pending_ipc_info.socket_info);
-        handle->pending_ipc_info.socket_info = NULL;
+      /* Free rest of the pending sockets */
+      /* XXX: Need to free sockets themselves? */
```
piscisaureus: Obsolete comment

src/win/pipe.c
```diff
@@ -358,10 +370,30 @@ void uv_pipe_endgame(uv_loop_t* loop, uv_pipe_t* handle) {
     assert(!(handle->flags & UV_HANDLE_CLOSED));
 
     if (handle->flags & UV_HANDLE_CONNECTION) {
-      if (handle->pending_ipc_info.socket_info) {
-        free(handle->pending_ipc_info.socket_info);
-        handle->pending_ipc_info.socket_info = NULL;
+      /* Free rest of the pending sockets */
```
piscisaureus: s/rest of//

src/win/pipe.c
```diff
@@ -358,10 +370,30 @@ void uv_pipe_endgame(uv_loop_t* loop, uv_pipe_t* handle) {
...
+        /* Materialize socket and close it */
+        socket = WSASocketW(AF_INET,
+                            SOCK_STREAM,
+                            IPPROTO_IP,
+                            &item->socket_info,
+                            0,
+                            WSA_FLAG_OVERLAPPED);
+        free(item);
+
+        /* XXX: Handle error? */
```
piscisaureus: I don't think handling the error here is necessary.

indutny:
@piscisaureus thank you, all fixed.

indutny:
@jclulow LGTY?

piscisaureus:
LGTM

src/unix/stream.c
trevnorris: Doesn't libuv follow the C89 spec? Part of which requires all variables to
be declared at the top of the function.

src/unix/stream.c
```diff
@@ -586,9 +585,33 @@ int uv_accept(uv_stream_t* server, uv_stream_t* client) {
       assert(0);
   }
 
-  uv__io_start(server->loop, &server->io_watcher, UV__POLLIN);
-  server->accepted_fd = -1;
-  return 0;
+done:
+  /* Process queued fds */
+  if (server->queued_fds != NULL) {
+    uv__stream_queued_fds_t* queued_fds;
```
trevnorris, line 591: @piscisaureus Doesn't libuv follow the C89 spec and require all variables to be declared at the top of the function?

bnoordhuis, line 591: That's correct C89.  Variable declarations need to go at the top of a block which can be a function but doesn't have to be.

trevnorris, line 591: Ah OK. Thanks. For some reason though it must be the function block. Guess
that's what I get for being a JS dev. :-P

src/unix/stream.c
```diff
@@ -931,23 +954,101 @@ static uv_handle_type uv__handle_type(int fd) {
...
+
+  /*
+   * Some implementations can send multiple file descriptors in a
+   * single message. We should be using CMSG_NXTHDR() to walk the
+   * chain to get at them all. This would require changing the API to
+   * hand these back up the caller, is a pain.
+   */
+
+  for (cmsg = CMSG_FIRSTHDR(msg);
+       msg->msg_controllen > 0 && cmsg != NULL;
```
jclulow: The check for `msg->msg_controllen > 0` is not necessary, as I understand -- it is included in `CMSG_FIRSTHDR` .  In fact, it's a more complete check in there, anyway: that `msg_controllen` is not just non-zero, but long enough to contain a valid `struct cmsghdr`.

src/unix/stream.c
```diff
@@ -931,23 +954,101 @@ static uv_handle_type uv__handle_type(int fd) {
...
+       cmsg = CMSG_NXTHDR(msg, cmsg)) {
+
+    if (cmsg->cmsg_type != SCM_RIGHTS) {
+      fprintf(stderr, "ignoring non-SCM_RIGHTS ancillary data: %d\n",
+          cmsg->cmsg_type);
+      continue;
+    }
+
+    /* silence aliasing warning */
```
jclulow, line 1027: What warning are you actually getting, here?  Can you not just `pi = (int *)CMSG_DATA(cmsg);` ?

indutny, line 1027: Actually, I copied it from another place :) Decided to keep it as it is.

I think it is done this way to remove `-fstrict-aliasing` warnings.

src/unix/stream.c
```diff
@@ -977,21 +1081,20 @@ static void uv__read(uv_stream_t* stream) {
...
-    if (stream->read_cb) {
+    if (!is_ipc) {
       do {
         nread = read(uv__stream_fd(stream), buf.base, buf.len);
       }
       while (nread < 0 && errno == EINTR);
     } else {
-      assert(stream->read2_cb);
-      /* read2_cb uses recvmsg */
+      /* ipc uses recvmsg */
```
jclulow, line 1101: So, looking at various things around the place, it feels like `msg_controllen` should be set with the `CMSG_LEN()` macro, having previously allocated memory of size returned by the `CMSG_SPACE()` macro.  i.e. depending on how many file descriptors we wish to be able to receive in a single message:

``` c
#define   MAX_RECV_FD_COUNT     16
#define   SCM_RIGHTS_LENGTH     (MAX_RECV_FD_COUNT * sizeof (int))

char cmsg_space[CMSG_SPACE(SCM_RIGHTS_LENGTH)];

cmsg.cmsg_len = msg.msg_controllen = CMSG_LEN(SCM_RIGHTS_LENGTH);
```

src/unix/stream.c
```diff
@@ -931,23 +954,101 @@ static uv_handle_type uv__handle_type(int fd) {
...
+    if (cmsg->cmsg_type != SCM_RIGHTS) {
+      fprintf(stderr, "ignoring non-SCM_RIGHTS ancillary data: %d\n",
+          cmsg->cmsg_type);
+      continue;
+    }
+
+    /* silence aliasing warning */
+    pv = CMSG_DATA(cmsg);
+    pi = pv;
+    count = (msg->msg_controllen - CMSG_LEN(0)) / sizeof(*pi);
```
jclulow: I believe you actually want to use `CMSG_DATA(cmsg)` here, rather than `CMSG_LEN(0)`.  These macros seem to do alignment stuff on different platforms, so it seems unwise to assume that the data region begins immediately following the end of the `struct cmsg_hdr` without padding.

indutny:
@jclulow fixed nits, please take another look.

indutny:
@jclulow also pushed a hack for linux, it seems that it is adding a lot of alignment to `CMSG_LEN()`.

indutny:
Please review ;)

indutny:
Seems to be failing on linux right now, will figure it out tomorrow.

src/unix/pipe.c
```diff
@@ -214,3 +214,27 @@ void uv_pipe_connect(uv_connect_t* req,
 
 void uv_pipe_pending_instances(uv_pipe_t* handle, int count) {
 }
+
+
+int uv_pipe_pending_count(uv_pipe_t* handle) {
+  uv__stream_queued_fds_t* queued_fds;
+
+  assert(handle->ipc);
```
saghul: Can we be a bit more gentle here and return 0 instead?

src/unix/pipe.c
```diff
@@ -214,3 +214,27 @@ void uv_pipe_connect(uv_connect_t* req,
...
+  if (handle->queued_fds == NULL)
+    return 1;
+
+  queued_fds = handle->queued_fds;
+  return 1 + queued_fds->offset;
+}
+
+
+uv_handle_type uv_pipe_pending_type(uv_pipe_t* handle) {
+  assert(handle->ipc);
```
saghul: Ditto, but with UV_UNKNOWN_HANDLE.

indutny:
@saghul done!

indutny:
Also fixed alignment issues, seems to be working fine on linux.

indutny:
@jclulow ^

saghul:
@indutny sweet! I can't comment much on the rest as it's beyond my knowledge, so I'll leave it up to you guys :-)

indutny:
@jclulow ping.

saghul:
@indutny, please close https://github.com/joyent/libuv/issues/461 when you land this.

indutny:
@jclulow reminding about it again :)

include/uv.h
```diff
@@ -1148,6 +1129,15 @@ UV_EXTERN void uv_pipe_connect(uv_connect_t* req, uv_pipe_t* handle,
  */
 UV_EXTERN void uv_pipe_pending_instances(uv_pipe_t* handle, int count);
 
+/*
+ * Should be used to receive handles on ipc pipe.
```
saghul, line 1133: "Used to receive handles over ipc pipes."

src/unix/pipe.c
```diff
@@ -214,3 +214,31 @@ void uv_pipe_connect(uv_connect_t* req,
...
+    return 0;
+
+  if (handle->accepted_fd == -1)
+    return 0;
+
+  if (handle->queued_fds == NULL)
+    return 1;
+
+  queued_fds = handle->queued_fds;
+  return 1 + queued_fds->offset;
```
saghul, line 232: return handle->queued_fds->offset + 1 ?

src/unix/stream.c
```diff
@@ -931,23 +954,105 @@ static uv_handle_type uv__handle_type(int fd) {
...
+    queued_fds->size = queue_size;
+    queued_fds->offset = 0;
+    stream->queued_fds = queued_fds;
+
+    /* Grow */
+  } else if (queued_fds->size == queued_fds->offset) {
+    queue_size = queued_fds->size + 8;
+    queued_fds = realloc(queued_fds,
+                         (queue_size - 1) * sizeof(*queued_fds->fds) +
+                             sizeof(*queued_fds));
```
saghul, line 986: Align with realloc open parens? (maybe GH shows it weird)

src/unix/stream.c
saghul, line 1101: Did you check this @indutny ?

indutny, line 1101: Yes, I think I fixed it. Tried it on Linux and Mac, was working fine.

src/unix/stream.c
indutny, line 986: Actually, this is intended 4 space indent here, as it is not a new argument, but a continuation of expression on the previous line.

indutny:
Landed in b05a3ee, thank you!

PR #1057, opened 2013-12-26
unix: `uv_pipe_init` should exports error

yorkie:
Related lines:
1. https://github.com/joyent/libuv/blob/7e68e6d895a9007736e53b592192fd9620a34ac6/src/unix/stream.c#L77
2. https://github.com/joyent/libuv/blob/master/src/win/pipe.c#L92


src/unix/pipe.c
```diff
@@ -36,7 +36,7 @@ int uv_pipe_init(uv_loop_t* loop, uv_pipe_t* handle, int ipc) {
   handle->connect_req = NULL;
   handle->pipe_fname = NULL;
   handle->ipc = ipc;
-  return 0;
+  return errno;
```
indutny, line 39: Hm... why?

indutny:
Sorry, it doesn't look like a right thing

src/unix/pipe.c
yorkie, line 39: Did you check this branch: https://github.com/joyent/libuv/blob/7e68e6d895a9007736e53b592192fd9620a34ac6/src/unix/stream.c#L77?

indutny, line 39: And? :) It is only for `emfile_fd` trick, if it fails - everything is still fine. See https://github.com/joyent/libuv/blob/7e68e6d895a9007736e53b592192fd9620a34ac6/src/unix/tcp.c#L31 for reference

yorkie, line 39: when `uv__stream_init` fails, the `errno` would be assigned with the specified non-zero value.
Actually, check this https://github.com/joyent/node/blob/master/src/pipe_wrap.cc#L134 in node, we should export an error for this.

indutny:
Sorry, you're doing wrong conclusions from the comments you see.

yorkie:
I see, it's my fault, sorry!

PR #1058, opened 2013-12-26
doc: add readme link to all projects that use libuv.

txdv:
All examples can be then dynamically added in the documentation without the need of changing the readme all the time.


txdv:
Already signed.

saghul:
Thanks @txdv, landed!

PR #1060, opened 2013-12-27
osx: fix -Wextra warnings

jneu:
On OS X 10.9.1, libuv does not compile cleanly with all warnings
enabled:

```
CFLAGS="-g -O2 -Wall -Wextra -Wno-unused-parameter -Werror"
CFLAGS="$CFLAGS --std=gnu89 -pedantic"
# The following is required for include/uv-dtrace.h
CFLAGS="$CFLAGS -Wno-dollar-in-identifier-extension"
./configure CFLAGS="$CFLAGS"
```

Most of the warnings are caused by function pointer casts in
`src/unix/darwin-proctitle.c` and `src/unix/fsevents.c`. In general I try
to avoid typedef-ing function prototypes, but in this case I use them
because it cleans up the code nicely.


indutny:
I think all those conversions are superfluous. We allow omitting casts when casting from `void*`. Also that code already looks quite clobbered and unreadable :) I don't want to make it even worse again.

Thank you for contributing anyway!

jneu:
No problem. Thanks @indutny for reading and commenting!

PR #1062, opened 2013-12-28
unix: fix non-OS X builds

unixwitch:
uv__stream_osx_interrupt_select was only defined on OS X, but is used
elsewhere on all platforms.  The intention was to implement it as a
no-op on other platforms, but the entire definition was inside
"#ifdef __APPLE__", so this didn't actually work.

Fix: move it above the #ifdef.

### DIFF
--- a/src/unix/stream.c
+++ b/src/unix/stream.c
@@ -138,6 +138,31 @@ void uv__stream_init(uv_loop_t* loop,
 }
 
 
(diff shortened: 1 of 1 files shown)


indutny:
Oh, good catch!

indutny:
Landed in 17648be, thank you!

PR #1063, opened 2013-12-29
unix: optimise UDP buffer allocation

unixwitch:
Instead of always allocating a 64KB buffer, use FIONREAD to check the size of the next incoming packet.

I intended to use FIONREAD here to check if any packet is pending, and avoid allocating a buffer (simply returning instead) if no packet is pending.  However, this isn't possible with FIONREAD: if a zero-byte packet is pending, we would spin forever as the socket will appear readable but we'll never read the packet.

Still, this is a minor worthwhile optimisation.

### DIFF
--- a/src/unix/udp.c
+++ b/src/unix/udp.c
@@ -22,6 +22,8 @@
 #include "uv.h"
 #include "internal.h"
 
+#include <sys/ioctl.h>
+
 #include <assert.h>
 #include <string.h>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


unixwitch:
NB: this actually ends up swapping a malloc() for a syscall (ioctl), so perhaps benchmarking would be worthwhile to see which is better.

unixwitch:
With these commits, we now avoid allocating a buffer entirely when there's no packet to be received.

saghul:
IIRC @bnoordhuis had some thoughts on this. Can you comment, Ben?

bnoordhuis:
I've experimented with FIONREAD / SIOCINQ it in the past.  The tl;dr is that the extra system call cuts throughput in half on most platforms (ioctls are expensive) and that it's almost always faster to malloc/realloc.  If you worry about malloc overhead, use some kind of slab allocation scheme.

A separate but related issue is that, on some platforms, the ioctl returns the size of the first pending datagram while on other platforms, it returns the size of all pending datagrams.

unixwitch:
Okay, that makes sense - shame though.

I'm not overly concerned about the speed of modern malloc()s, just thought I'd have a look as someone mentioned this behaviour on IRC.

Issue #869, opened 2013-08-05
freebsd: broken autotools + dtrace build 


bnoordhuis:
```
# uname -a
FreeBSD 15f3cf02-b72e-45c5-8c07-7a5119e20fcb 9.1-RELEASE FreeBSD 9.1-RELEASE #0: Sat Jun 15 00:11:29 UTC 2013     root@freebsd:/usr/obj/usr/src/sys/JOYENT  amd64

# ./configure
<snip>
checking for dtrace... dtrace
checking if dtrace works... yes
checking if dtrace should instrument object files... no
<snip>

# make
  GEN      include/uv-dtrace.h
dtrace: failed to open -o: No such file or directory
*** [include/uv-dtrace.h] Error code 1
```

jneu:
I was able to reproduce this issue on FreeBSD 9.2:

```
[jneuheisel@free /usr/home/jneuheisel]$ uname -a
FreeBSD free.jn 9.2-RELEASE FreeBSD 9.2-RELEASE #0 r255898: Thu Sep 26 22:50:31 UTC 2013     root@bake.isc.freebsd.org:/usr/obj/usr/src/sys/GENERIC  amd64
```

Looks like the FreeBSD version of make only sets `$<` on suffix-transformation rules - although this is not well documented anywhere I can find.

FWIW, building with dtrace support on FreeBSD is still not for the faint of heart. First, you need to add libelf manually (I did this as `./configure LDFLAGS="-lelf"`). This could be automated through autoconf. More problematic, though, is the failure of `dtrace` to correctly determine the 32/64-bitness of the build machine. I manually edited `m4/dtrace.m4` and added `-64` to each call to `dtrace` to force the correct sizes. I don't think this has anything to do with libuv but is probably a bug in the FreeBSD software.

PR #1064, opened 2013-12-29
build: fix dtrace in FreeBSD build (#869)

jneu:
Addresses joyent/libuv#869.

FreeBSD make only sets `$<` (aka `.IMPSRC`) in suffix-transformation rules.
Changed the syntax in the recipe for the rule to make
`include/uv-dtrace.h` to explicitly name `src/unix/uv-dtrace.d` instead of
using `$<`.


tjfontaine:
isn't it reasonable to just require gmake for the build?

jneu:
If libuv plans on maintain a FreeBSD port in the future, it might be easier if gmake is not made a requirement and the native make works correctly. If not, I'd be OK with requiring gmake.

indutny:
Yeah, I don't think that we are going to add hacks for bsd make. Node and libuv should be built using `gmake` or other build tool like `ninja`, BSD make is _not supported_.

PR #1067, opened 2014-01-02
windows: fix compatibility with cygwin pipes

vtjnash:
This makes libuv more tolerant of the properties of the pipes
that it can use. this is necessary because Cygwin (and hence Mintty)
opens `STDIN` without `FILE_WRITE_ATTRIBUTES`. Also, this always duplicate
handles from external sources, to ensure the original handle can remain
valid after uv_close (this can be important for stdio since the system
libraries do not expect it to be closed by the user).


txdv:
Everything after the : should go into a new line, because it is a small description of the headline.

vtjnash:
@txdv are you looking at my second referenced commit? It's not clear from GitHub's view, but it was in a working branch (before I cleaned up the commit description to keep Jenkins from complaining).

Keno:
bump

oberlies:
Would this be fixing https://github.com/joyent/node/issues/6459 ?

saghul:
Looks like it would. Unfortunately this is way out of my Windows knowledge so we'd need some imput on it. @piscisaureus, @orangemocha does this look like a good approach?

piscisaureus:
Hey @vtjnash, it's hard for me to review this code because I don't understand what the issues are with cygwin pipes and your code has no comments that explain it. 

Of course I can now bitch about how you're duplicating handles without ever closing them and code style etc. but before that I'd like to understand it better. 

Can you maybe explain the issue that this solves; if that's unpractical we can maybe make an appointment so I'll be on IRC (or some other medium) and we can chat about it.

src/win/pipe.c
```diff
@@ -230,7 +230,7 @@ static int uv_set_pipe_handle(uv_loop_t* loop,
   NTSTATUS nt_status;
   IO_STATUS_BLOCK io_status;
   FILE_MODE_INFORMATION mode_info;
-  DWORD mode = PIPE_TYPE_BYTE | PIPE_READMODE_BYTE | PIPE_WAIT;
+  DWORD mode = PIPE_READMODE_BYTE | PIPE_WAIT;
```
vtjnash, line 233: `PIPE_TYPE_BYTE` isn't a valid flag, although it didn't matter since it's value is 0.

src/win/pipe.c
```diff
@@ -246,11 +246,9 @@ static int uv_set_pipe_handle(uv_loop_t* loop,
       if (!GetNamedPipeHandleState(pipeHandle, &current_mode, NULL, NULL,
                                    NULL, NULL, 0)) {
         return -1;
-      } else if (current_mode != mode) {
+      } else if (current_mode&PIPE_NOWAIT) {
         SetLastError(ERROR_ACCESS_DENIED);
         return -1;
-      } else {
-        duplex_flags &= ~UV_HANDLE_WRITABLE;
```
vtjnash, line 253: this is now handled below, by testing for whether the pipe is writeable (whereas this previously checked that properties of the pipe could be modified -- which is controlled by a separate permissions bit)

vtjnash:
A portion of this patch was independently executed in ba47e68824dbd00c51d2c659ab05a83989283aab

I've reworked (shortened) this to incorporate those changes, and join IRC when I have availability (usually late in the evenings).

It looks like I had some rebase issues when I copied this from the Julia fork (we have other changes there that fix the read/write flags of pipes on other platforms and the behavior of uv_spawn for all platforms per #451. I removed the changes -- such as duplicating handles from the host process -- that aren't directly relevant)

src/win/pipe.c
```diff
@@ -246,11 +246,9 @@ static int uv_set_pipe_handle(uv_loop_t* loop,
       if (!GetNamedPipeHandleState(pipeHandle, &current_mode, NULL, NULL,
                                    NULL, NULL, 0)) {
         return -1;
-      } else if (current_mode != mode) {
+      } else if (current_mode&PIPE_NOWAIT) {
```
vtjnash, line 249: previously, this verified that the pipe is in `PIPE_READMODE_BYTE`, but cygwin puts it in `PIPE_READMODE_MESSAGE` (IIRC, this was changed because it gives better behavior when there are multiple writers) and doesn't give us the permissions to edit the mode. however this should be fine, since we can always just read bytes at a time.

Keno:
BUMP

Keno:
Bump^2 ;). If there's anything we can clarify, please let us know.

saghul:
The new cleaned-up code LGTM. It just takes care of settig the readable/writable flags approriately. Can you do a quick lookover, @piscisaureus?

PS: There are a few style issues (mostly space between operators), we can fix those during merge.

piscisaureus:
@Keno @vtjnash 

Ok, I understand what you're trying to do now and it seems good. 

I would really like it if you could add a comment _to the code_ stating that cygwin opens the pipe in message mode but we can support that just fine. That way the knowledge in our heads doesn't get lost - you've probably already noticed that I am not around a lot any more, but there will be maintainers in the future and it'd be great if they don't have to reverse engineer your intent here.

@saghul is right, there are a few style nits but nothing that takes more than 5 minutes to fix. 

so lgtm
@saghul feel free to land this when you think it's good.

src/win/pipe.c
saghul, line 249: Style: spaces around '&'

src/win/pipe.c
```diff
@@ -1771,8 +1769,33 @@ static void eof_timer_close_cb(uv_handle_t* handle) {
 
 
 int uv_pipe_open(uv_pipe_t* pipe, uv_file file) {
-  HANDLE os_handle = uv__get_osfhandle(file);
-  DWORD duplex_flags = UV_HANDLE_READABLE | UV_HANDLE_WRITABLE;
+  HANDLE os_handle = (HANDLE)uv__get_osfhandle(file);
```
saghul, line 1772: Unneeded cast?

src/win/pipe.c
```diff
@@ -1771,8 +1769,33 @@ static void eof_timer_close_cb(uv_handle_t* handle) {
...
+  }
+
+  if (pipe->ipc) {
+    if (!(access.AccessFlags&FILE_WRITE_DATA) ||
+        !(access.AccessFlags&FILE_READ_DATA)) {
+      return UV_EINVAL;
+    }
+  }
+
+  DWORD duplex_flags = 0;
```
saghul, line 1794: Declare at the top.

src/win/pipe.c
```diff
@@ -1771,8 +1769,33 @@ static void eof_timer_close_cb(uv_handle_t* handle) {
...
+  if (access.AccessFlags&FILE_WRITE_DATA)
```
saghul, line 1795: Spaces around '&'.

src/win/pipe.c
```diff
@@ -1771,8 +1769,33 @@ static void eof_timer_close_cb(uv_handle_t* handle) {
...
+    duplex_flags |= UV_HANDLE_WRITABLE;
+  if (access.AccessFlags&FILE_READ_DATA)
```
saghul, line 1797: Ditto.

saghul:
Thanks @piscisaureus! I made some comments about the style issues, @Keno @vtjnash please fix those and add a explanatory comment and this is good to go!

saghul:
Made the changes myself and landed it in https://github.com/joyent/libuv/commit/ebafb90acd6de8f2bf766ac58dc1091ddfacfad4, thanks!

PR #1068, opened 2014-01-04
`free(path)` is possible to cause abort

yorkie:
`realpath` returns NULL when it failed, not only in out of memory case.

> If there is no error, realpath() returns a pointer to the resolved_path.
> 
> Otherwise, it returns a NULL pointer, the contents of the array resolved_path are undefined, and errno is set to indicate the error.

See the refs from `libuv`:

``` c
path = malloc(2 * PATH_MAX);
fullpath = realpath(buffer, path);
if (fullpath == NULL) {
  SAVE_ERRNO(free(path));
  return -errno;
}
```

If `path` is NULL, `free(path)` will throw an error, even though the fullpath is equal to path.


txdv:
2 \* PATH_MAX

I don't even...

yorkie:
HI, @txdv, could you give me your thought more?
I checked some docs on `realpath()`, I know if `path` here is NULL, it will be allocated with `PATH_MAX`.

indutny:
`free(NULL)` is totally correct thing.

PR #1069, opened 2014-01-04
Hpux support

jachym1:
Hi,
comments included in commit messages, hpux.c has been modified from sunos.c source. 
We would like to use Avatar-js in our projects on all platforms including hpux and libuv is requiered by Avatar. To test libuv in real scenario we have to wait for Java 8 availability on hpux which is prereq for Avatar.
Regards
Jachym
Czech Republic

### DIFF
--- a/src/unix/hpux.c
+++ b/src/unix/hpux.c
@@ -35,9 +35,7 @@
 #include <fcntl.h>
 
 #include <sys/pstat.h>
-#include <sys/dk.h>
 #include <sys/devpoll.h>
-
 
 int uv__platform_loop_init(uv_loop_t* loop, int default_loop) {
    int err;
@@ -238,12 +236,9 @@ update_timeout:
   }
 }
 
-extern  hrtime_t        gethrtime __((void)); 
-
 uint64_t uv__hrtime(uv_clocktype_t type) {
   return gethrtime();
 }
-
 
 /*
  * We could use a static buffer for the path manipulations that we need outside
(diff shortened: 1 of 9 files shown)


indutny:
Wrong rebase?

src/unix/hpux.c
```diff
@@ -0,0 +1,643 @@
...
+
+  if (loop->backend_fd != -1) {
+    uv__close(loop->backend_fd);
+    loop->backend_fd = -1;
+  }
+}
+
+
+void uv__platform_invalidate_fd(uv_loop_t* loop, int fd) {
+struct pollfd* events;
```
indutny, line 75: Indent.

src/unix/hpux.c
```diff
@@ -0,0 +1,643 @@
...
+
+  while (!QUEUE_EMPTY(&loop->watcher_queue)) {
+    q = QUEUE_HEAD(&loop->watcher_queue);
+    QUEUE_REMOVE(q);
+    QUEUE_INIT(q);
+
+    w = QUEUE_DATA(q, uv__io_t, watcher_queue);
+    assert(w->pevents != 0);
+
+   pc.fd = w->fd;
```
indutny, line 123: Indent.

src/unix/hpux.c
```diff
@@ -0,0 +1,643 @@
...
+    }
+
+    /* Work around a kernel bug where nfds is not updated. */
+    events[0].revents = 0;
+
+    saved_errno = 0;
+
+    dvp.dp_fds = events;
+    dvp.dp_nfds = 1024;
+    dvp.dp_timeout = -1;
```
indutny, line 148: No timeout?

src/unix/hpux.c
```diff
@@ -0,0 +1,643 @@
...
+    return;
+
+  /* Invalidate events with same file descriptor */
+  for (i = 0; i < nfds; i++)
+    if ((int) events[i].fd == fd)
+      events[i].events = -1;
+}
+
+
+void uv__io_poll(uv_loop_t* loop, int timeout) {
```
indutny, line 93: I guess it should go to the `poll.c`.

src/unix/hpux.c
```diff
@@ -0,0 +1,643 @@
...
+int uv_get_process_title(char* buffer, size_t size) {
+  if (size > 0) {
+    buffer[0] = '\0';
+  }
+  return 0;
+}
+
+
+int uv_resident_set_memory(size_t* rss) {
+   return 0;
```
indutny, line 338: Returning `0` feels quite wrong here.

src/unix/hpux.c
```diff
@@ -0,0 +1,643 @@
...
+}
+
+
+int uv_uptime(double* uptime) {
+
+struct pst_status p;
+
+ if (pstat_getproc(&p, sizeof(p), 0, 1) == 1)
+  {
```
indutny, line 347: Remove braces, they are superfluous.

src/unix/hpux.c
```diff
@@ -0,0 +1,643 @@
...
+ else {
+   *uptime = -1;
+  }
+
+  return 0;
+}
+
+
+int uv_cpu_info(uv_cpu_info_t** cpu_infos, int* count) {
+uv_cpu_info_t* cpu_info;
```
indutny: Indent?

src/unix/hpux.c
```diff
@@ -0,0 +1,643 @@
...
+  if (!*cpu_infos) {
+    free(psp);
+    abort();
+  }
+
+  *count = nspu;
+
+  cpu_info = *cpu_infos;
+  while (i < nspu) {
+    cpu_info->cpu_times.user = (((uint64_t)psp[i].psp_usercycles.psc_hi << 32) + psp[i].psp_usercycles.psc_lo);
```
indutny: please follow 80 column limit.

indutny:
@jachym1 this seems to be generally a right thing. The main problem with it is that I don't really want to support/test it. @saghul what do you think?

bnoordhuis:
> The main problem with it is that I don't really want to support/test it.

This would be the same as AIX support: if it works, good for you - if it doesn't, fix it yourself and forward us the patch.

indutny:
@bnoordhuis that sounds easy :)

saghul:
What @bnoordhuis said :-)

indutny:
@jachym1 ping, do you still want this to be landed?

indutny:
No follow-up, closing.

PR #1076, opened 2014-01-15
doc: describe signal send emulation on windows

sam-github:
Signals don't exist on Windows, but libuv emulates the behaviour of
several Unix signals. This wasn't documented, and the existing
documentation for signal reception emulation on Windows was worded as if
it applied to signal sending, which it does not.

/to @piscisaureus


saghul:
Thanks @sam-github, landed it as https://github.com/joyent/libuv/commit/a3c0e48faa6018b1867f4897d1c74b78df6dd416.

PR #1081, opened 2014-01-20
linux: relax assumption on /proc/stat parsing

### LINKED ISSUE #1080: linux-core: too strict assumption on /proc/stat parsing
In the Debian build machine park we have an [UltraSparc II box (U60)](https://db.debian.org/machines.cgi?host=spontini) with the following `/proc/stat`:

```
$ cat /proc/stat
cpu  29733059 2644 2838265 32944962 2338973 15558 52150 0 0
cpu0 14908964 1765 1494903 15699901 1739345 15558 37979 0 0
cpu2 14824094 879 1343361 17245060 599627 0 14171 0 0
intr 147091428 134883696 0 0 0 0 0 0 901 14 6 0 2 0 0 7559720 2709164 1937925 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
ctxt 185951785
btime 1389833870
processes 2239487
procs_running 1
procs_blocked 0
softirq 279821254 0 122089522 1415859 6656126 4202387 0 465 20866767 4418 124585710
```

This is tripping the assert at https://github.com/joyent/libuv/blob/v0.10/src/unix/linux-core.c#L606 which is expecting the `cpuX` in linear progression, and barking due to the missing `cpu1`. An example of it is the [libuv-0.10.22 testsuite failing](https://buildd.debian.org/status/fetch.php?pkg=libuv&arch=sparc&ver=0.10.22-1&stamp=1390147318).

Not sure about your preferred solution, but we can either: 
1. skip /proc/stat when such weirdness appears (returning instead of asserting)
2. relax the assumption and just parse `numcpus` entries after the first one

lucab:
CPU entries in /proc/stat are not guaranteed to be monotonically
increasing, asserting on this assumption can break in cases such
as the UltraSparc II shown below. 

This fixes #1080.


indutny:
@lucab thanks for contributing this, but I don't think that this is a proper solution for the problem.

With your fix - cpu stats will be collected in a wrong order. It may be a better idea to parse num from `cpu%u` line and use it to insert into cpus array.

lucab:
Hmm, what you mean "in a wrong order"? I still believe they are collected in the same order, just that `cpu2` will go into `ci[1]` instead of `ci[2]`. But as you still have `numcpus` cpu, `cpu1` is not really there, and `cpu2` is really the second cpu, I don't get the issue‽

Are you instead suggesting to leave `ci[1]` empty and go for `ci[2]`? This will still give bad surprises, as we are only allocating up to `ci[1]` and reporting two cpus to the user.

indutny:
Hm... @lucab is it just skipping some cpus in output? Or could it also output it unordered?

If former is the case - then LGTM!

lucab:
In this case it is just missing some cpus (that were probably taken offline by software). I'm not sure if they can be unordered, will probably have to check kernel code to know.

Full output also here below as per #1080.

```
$ cat /proc/stat
cpu  29733059 2644 2838265 32944962 2338973 15558 52150 0 0
cpu0 14908964 1765 1494903 15699901 1739345 15558 37979 0 0
cpu2 14824094 879 1343361 17245060 599627 0 14171 0 0
intr 147091428 134883696 0 0 0 0 0 0 901 14 6 0 2 0 0 7559720 2709164 1937925 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
ctxt 185951785
btime 1389833870
processes 2239487
procs_running 1
procs_blocked 0
softirq 279821254 0 122089522 1415859 6656126 4202387 0 465 20866767 4418 124585710
```

src/unix/linux-core.c
```diff
@@ -603,7 +603,7 @@ static int read_times(unsigned int numcpus, uv_cpu_info_t* ci) {
     {
       unsigned int n = num;
       for (len = sizeof("cpu0"); n /= 10; len++);
```
indutny: I think you should calculate len after reading `n` now.

lucab: Ah right, for more than 9 cpus. Swapping the `for` and `assert` lines should be just enough, right? Also, initialising `n` to `num` is useless in that case...

indutny: Yep, swapping should be fine.

lucab:
Rebased as agreed.

src/unix/linux-core.c
```diff
@@ -601,9 +601,9 @@ static int read_times(unsigned int numcpus, uv_cpu_info_t* ci) {
...
     /* skip "cpu<num> " marker */
     {
-      unsigned int n = num;
+      unsigned int n;
+      assert(sscanf(buf, "cpu%u ", &n) == 1);
       for (len = sizeof("cpu0"); n /= 10; len++);
-      assert(sscanf(buf, "cpu%u ", &n) == 1 && n == num);
     }
 
     /* Line contains user, nice, system, idle, iowait, irq, softirq, steal,
```
indutny, line 609: Perhaps we should add assertion at https://github.com/joyent/libuv/pull/1081/files#diff-ef6a692d0a1d5779d26d4f57afcf3d42L632 or memset rest of the `ts`.

indutny, line 609: erm, not `ts`, `cpu_times`.

lucab, line 609: I'm more about explicit failures, so I'll assert `num == numcpus` before returning. Hopefully this won't trip too many times.

indutny, line 609: Ok, sounds good.

lucab:
Added final assertion as a safety net.

indutny:
Thank you, landed in 993151b

PR #1082, opened 2014-01-20
sunos: handle rearm errors

### LINKED ISSUE #1078: uv__fs_event_rearm return codes are ignored
I ran into a problem where fs.watch would pretend to be watching a file when the port_associate call was actually returning ENOENT. The reason appears to be:

https://github.com/joyent/libuv/blob/6149b66ccb9998f561a207162195bf5cba647280/src/unix/sunos.c#L415

is not checking the return value from uv__fs_event_rearm() even though that should have returned the errno from the failed port_associate().

indutny:
fix #1078


indutny:
@bcantrill May I ask you to review it?

src/unix/sunos.c
```diff
@@ -361,11 +363,13 @@ static void uv__fs_event_read(uv_loop_t* loop,
     assert(events != 0);
     handle->fd = PORT_FIRED;
     handle->cb(handle, NULL, events, 0);
+
+    if (handle != NULL && handle->fd != PORT_DELETED) {
```
indutny: It should be a better place to rearm handle if we are going to receive multiple events from port.

indutny:
@bcantrill or @jclulow any idea if I could land it?

src/unix/sunos.c
jclulow: Surely `handle` is _always_ `!= NULL` here, or we would have aborted while dereferencing on line 365?

indutny: Good catch.

indutny: @jclulow pushed fix.

src/unix/sunos.c
```diff
@@ -361,11 +363,12 @@ static void uv__fs_event_read(uv_loop_t* loop,
     assert(events != 0);
     handle->fd = PORT_FIRED;
     handle->cb(handle, NULL, events, 0);
+
+    if (handle->fd != PORT_DELETED)
```
saghul, line 367: handle->fd was just set to PORT_FIRED above, so I guess this if is redundant?

saghul, line 367: Ah, but the user maybe closed the handle in the callback! Forget that.

indutny, line 367: It could become `DELETED` inside callback.

src/unix/sunos.c
```diff
@@ -387,14 +390,16 @@ int uv_fs_event_init(uv_loop_t* loop,
...
   handle->fd = PORT_UNUSED;
   handle->cb = cb;
 
   memset(&handle->fo, 0, sizeof handle->fo);
   handle->fo.fo_name = handle->filename;
-  uv__fs_event_rearm(handle);
+  if (uv__fs_event_rearm(handle) != 0)
+    return -1;
+
+  uv__handle_start(handle); /* FIXME shouldn't start automatically */
```
saghul, line 402: I changed this a while ago so we now have uv_fsevent_start, I guess we should do that here?

saghul, line 402: Sorry, this is the v0.10 branch, disregard that as well.

src/unix/sunos.c
jclulow, line 367: I would brace the outer `if`, for readability -- but that's just a style nit.

jclulow:
I posted another style nit, but with or without that: LGTM!

indutny:
Thank you!

indutny:
Thank you, landed in 703a9e601e976b74ce459e4db9e23fca107d3f4b

PR #1083, opened 2014-01-20
process: close stdio after dup2'ing it

### LINKED ISSUE #1074: uv_spawn won't duplicate the same original fd to two fds in the child
Instead, it results in the child not having a valid fd at that position.

Reproduceable in node with https://gist.github.com/sam-github/8341916

strace shows what's happening in child: https://gist.github.com/sam-github/8341573

This results in fd 2 in child not being valid (an unusual condition, which libuv tries to avoid). The invalidity causes and write to fd 2 (console.error in node), to cause node to terminate (seemingly silently, its probably trying to write a stack trace, but since that would go to stderr, that doesn't work so well).

indutny:
Thus allow passing the same file descriptor as the source of multiple
stdios.

fix #1074


sam-github:
Maybe more like https://github.com/sam-github/libuv/commit/12a8f2c6808d45661d0b586b01d1e23df8e22317 ?

I think all three branches in the for loop up above need accounting for, including the temp fd from the open of /dev/null, so I moved the handling of close_fd to end of the loop block. This might work a bit better, at least allowing the possibility of the same parent fd to be mapped to two fds in the child, but...

@piscisaureus pointed out even more interesting ways to cause chaos, if the output fds overlap with the input fds. We are basically mapping internally to a set, so previous actions effect later.Example, if you map <parentfd> to <should be fd in child> like this:

```
0 -> 0
1 -> 1
2 -> 2
5 -> 3,4,5
4 -> 6,7,8
```

we won't get `[0,1,2,5,5,5,4,4,4]`, we'll get `[0,1,2,5,X,X,5,5,5]` where X is closed, or something like that.

Not too sure what to do about that, you could in theory map all the src fds to numbers greater than stdio_count, then do the mapping... but that sounds like lots of work. 

Maybe a better fix is checking the return values of dup2() and close(), and erroring if they fail, because its a fd setup that isn't achievable? That would probably fix the node bug, since the spawn would have failed explicitly.

indutny:
@sam-github added your fix, hope you don't mind. LGTY?

sam-github:
Looks OK, if it passes your tests! Thanks. Will try with node when I have a chance, but I'm buried in work at this moment.

indutny:
Yes, it does pass it. But the test isn't that complicated, though :)

indutny:
@piscisaureus may I ask you to review it? There are not much people who could do it right now, except you and me ;)

rmustacc:
@tjfontaine asked me to take a look, I'll have feedback later today.

piscisaureus:
@sam-github: Fedor and I just decided that merging stdout and stderr is out of scope for this patch. This just makes it possible to inherit the same fd multiple times from the parent process; it doesn't let you bind a child stdio fd to a newly created pipe created for another child fd.

indutny:
Thank you, everyone! Landed in e403a2c

PR #1085, opened 2014-01-22
linux: silence uninitialized variable warning

trevnorris:
When building using gyp and BUILDTYPE=Release using clang 3.4 received
this warning:

```
../src/unix/linux-core.c:640:34: warning: variable 'n' is uninitialized
      for (len = sizeof("cpu0"); n /= 10; len++);
                                 ^
```

Initializing n = 0 silences this build warning.


indutny:
Oh, wrong fix, actually. :) Could you just unwrap that line from `assert()`? And thanks for catching it!

PR #1086, opened 2014-01-22
core: uv__close() should accept low fds

indutny:
Low (`<= 2`) file descriptor may arrive to `uv__close()` from kqueue
hack and perhaps other places. The scenario is following: one of the
stdio fds is closed and new one with the same number is opened,
eventually leading to the `abort()` call because of that check.


bnoordhuis:
You know why I changed that in the first place, right?  If there is a legitimate case for calling uv__close() with a stdio file descriptor, you should call close() directly with a comment explaining why.

indutny:
Ok... I see, actually after discussing it with @piscisaureus I'm starting to think that there is another bug which causes it to fail...

indutny:
Ha! Found node.js bug that was causing it, it was calling uv_fs_close(0).

PR #1087, opened 2014-01-22
linux: move sscanf() out of the assert()

trevnorris:
If asserts are turned off then the sscanf() wouldn't have run, being
placed directly in the assert() itself.


indutny:
LGTM

indutny:
Landed in 74b2900

PR #1091, opened 2014-01-29
fs: `uv__cloexec()` opened fd

indutny:
Every file descriptor opened using libuv should be automatically marked
as CLOEXEC to prevent it from leaking to a child process. Note that
since we are opening fds in a thread pool, there is a possible race
condition between `uv_spawn()` and the `open()` + `uv__cloexec()`. The
rwlock was added to avoid it.

see https://github.com/joyent/node/issues/6905


include/uv-unix.h
```diff
@@ -169,6 +169,7 @@ struct uv__work {
   void* wq[2];                                                                \
   uv_mutex_t wq_mutex;                                                        \
   uv_async_t wq_async;                                                        \
+  uv_rwlock_t cloexec_lock;                                                   \
```
rmustacc, line 172: It's worth going through and making sure you understand what the length of your read hold times is. When you have very short hold times, you're actually better served by a mutex as the rwlock_t actually creates more cache contention when hold time are short. See http://queue.acm.org/detail.cfm?id=1454462 for more of the history/explanation of this.

src/unix/fs.c
```diff
@@ -620,6 +621,21 @@ static void uv__fs_work(struct uv__work* w) {
...
     break;
 
+#define RW_X(type, action)                                                    \
+  case UV_FS_ ## type:                                                        \
+    uv_rwlock_rdlock(&req->loop->cloexec_lock);                               \
+    r = action;                                                               \
+    if (r >= 0) {                                                             \
+      err = uv__cloexec(r, 1);                                                \
+      if (err != 0) {                                                         \
+        close(r);                                                             \
```
rmustacc: Is there a reason that you aren't checking the return value of close? Also, shouldn't you save the original errno here before calling close which may very well clobber errno?

src/unix/fs.c
```diff
@@ -620,6 +621,21 @@ static void uv__fs_work(struct uv__work* w) {
...
+        r = err;                                                              \
+      }                                                                       \
+      errno = 0;                                                              \
```
rmustacc: Is there a reason you clobber errno and don't save/propagate it back somehow?

src/unix/process.c
```diff
@@ -422,6 +422,8 @@ int uv_spawn(uv_loop_t* loop,
 
   uv_signal_start(&loop->child_watcher, uv__chld, SIGCHLD);
 
+  /* Acquire write lock to prevent opening new fds in worker threads */
+  uv_rwlock_wrlock(&loop->cloexec_lock);
```
rmustacc, line 426: In the case where pid == -1, you don't unlock the write lock and it appears that the error case doesn't handle that either.

indutny, line 426: Good catch!

src/unix/fs.c
indutny: Yeah, I better use `uv__close()` here, I think

include/uv-unix.h
indutny, line 172: You convinced me!

rmustacc, line 172: I haven't quite convinced myself. As this could be used to access network file systems which can have rather painful failure modes, eg. think NFS. So it's something to think through and consider. In this case it may not make sense to change.

indutny, line 172: Actually, since that comment I had some time to think too. Better leave rwlock here.

src/unix/fs.c
indutny: Because it could become "dirty" in `uv__cloexec`, and it is set to zero right before the switch, and the `open()` has already succeeded so it is zero.

indutny:
@rmustacc please take another look, should be fine now.

src/unix/fs.c
```diff
@@ -620,6 +621,23 @@ static void uv__fs_work(struct uv__work* w) {
...
+      uv_rwlock_rdlock(&req->loop->cloexec_lock);                             \
+    r = action;                                                               \
+    if (r >= 0) {                                                             \
+      err = uv__cloexec(r, 1);                                                \
+      if (err != 0) {                                                         \
+        uv__close(r);                                                         \
+        r = err;                                                              \
+      }                                                                       \
+      errno = 0;                                                              \
+    }                                                                         \
```
rmustacc: Looking at this in more detail, I feel like r is being conflated in its purpose. If we look at line 675, we check to see if r is negative one and set it to errno if that's the case. However, functions like uv__cloexec and uv__close return -errno. In this case it's conceivable that any of these operations could legitimately return -1 (based on whatever errno 1 is on a given system), and then we'll do the wrong thing, especially as we've just clobbered errno up above.

Also uv__close() can return an error. Is there some reason you're not checking it?

indutny: Oh, you are right! Thank you very much, I rewrote it again!

src/unix/fs.c
```diff
@@ -646,6 +645,23 @@ static void uv__fs_work(struct uv__work* w) {
...
+        uv_rwlock_rdlock(&req->loop->cloexec_lock);
+      r = open(req->path, req->flags, req->mode);
+
+      /*
+       * In case of failure `uv__cloexec` will leave error in `errorno`,
+       * so it is enough to just set `r` to `-1`.
+       */
+      if (r >= 0 && uv__cloexec(r, 1) != 0) {
+        if (uv__close(r) != 0)
+          abort();
```
rmustacc, line 674: It looks like uv__close here can return EINTR or EINPROGRESS. Don't we want to retry on the case of EINTR and not abort()?

src/unix/fs.c
```diff
@@ -646,6 +645,23 @@ static void uv__fs_work(struct uv__work* w) {
     X(UNLINK, unlink(req->path));
     X(UTIME, uv__fs_utime(req));
     X(WRITE, uv__fs_write(req));
+    case UV_FS_OPEN:
+      if (req->cb != NULL)
+        uv_rwlock_rdlock(&req->loop->cloexec_lock);
+      r = open(req->path, req->flags, req->mode);
+
+      /*
+       * In case of failure `uv__cloexec` will leave error in `errorno`,
```
rmustacc: I think you meant `errno` instead of `errorno`.

src/unix/fs.c
bnoordhuis, line 674: uv__close() translates EINTR to EINPROGRESS (for 'not completed yet.')

Retrying a close() after EINTR is unsafe on Linux and the BSDs and (IIRC) on Solaris as well because the file descriptor has indeed been closed.  If you call close() again, best case you get a EBADF, worst case you close a file descriptor that another thread has just opened.

(Edit: reworded for clarity.)

src/unix/fs.c
```diff
@@ -646,6 +645,23 @@ static void uv__fs_work(struct uv__work* w) {
     X(UNLINK, unlink(req->path));
     X(UTIME, uv__fs_utime(req));
     X(WRITE, uv__fs_write(req));
+    case UV_FS_OPEN:
+      if (req->cb != NULL)
+        uv_rwlock_rdlock(&req->loop->cloexec_lock);
+      r = open(req->path, req->flags, req->mode);
```
bnoordhuis, line 665: So... why don't you use O_CLOEXEC on platforms that support it instead of this mutex hackery?

indutny:
Fixed all nits. @bnoordhuis added support for `O_CLOEXEC`, from my googling I found that it is not supported only on the platforms where it is not defined, right?

bnoordhuis:
> added support for O_CLOEXEC, from my googling I found that it is not supported only on the platforms where it is not defined, right?

Up to a point.  See example usage [here](https://github.com/joyent/libuv/blob/b0ec132/src/unix/stream.c#L71-80).  The idea is to try the system call and keep a note of whether the kernel supports it in case libuv was compiled against system headers that are newer than the actual kernel.

By the way, the `#if defined(__linux__)` check that's in there is now somewhat wrong because FreeBSD 10 also supports O_CLOEXEC.  Any POSIX.1-2008 compliant system does but so far that's limited to Linux and FreeBSD, I think.

In case you're looking for it, O_CLOEXEC == 0x100000 on FreeBSD 10.  The #define itself is only visible when you compile with -D_POSIX_C_SOURCE=200809.

indutny:
And OSX > 10.6 . I'll see what I can do about it, perhaps will amend that function and use it there.

indutny:
@bnoordhuis please take another look

indutny:
@rmustacc do you have any other comments? @bnoordhuis I pushed a hack for FreeBSD.

indutny:
@rmustacc ping

src/unix/process.c
```diff
@@ -422,10 +422,17 @@ int uv_spawn(uv_loop_t* loop,
 
   uv_signal_start(&loop->child_watcher, uv__chld, SIGCHLD);
 
+#ifndef O_CLOEXEC
+  /* Acquire write lock to prevent opening new fds in worker threads */
+  uv_rwlock_wrlock(&loop->cloexec_lock);
+#endif  /* !O_CLOEXEC */
```
rmustacc: Is this actually safe to do? Earlier you used the presence of O_CLOEXEC as a hint that the feature may be present and falled back. Here you're assuming that it exists. So for example, on a FreeBSD system that doesn't support O_CLOEXEC, you'll do the wrong thing. I'm less concerned with bringing a binary built for a newer system to an older one, as that is fraught with perils, but if that's something you care about based on the other work that you have trying to fall back, then this won't work.

src/unix/internal.h
```diff
@@ -111,6 +112,14 @@
 # define UV__POLLHUP  8
 #endif
 
+#if !defined(O_CLOEXEC) && defined(__FreeBSD__)
+/*
+ * It may be that we are just missing `__POSIX_VISIBLE >= 200809`.
+ * Try using fixed value const and give up, if it doesn't work
+ */
+# define O_CLOEXEC 0x00100000
```
rmustacc, line 120: Why do this? Why not just actually set the __POSIX_VISIBLE macro in the appropriate builds to declare that you want that support? This hack seems well, rather hacky and not really worth it.

src/unix/fs.c
```diff
@@ -646,6 +648,35 @@ static void uv__fs_work(struct uv__work* w) {
...
     X(UTIME, uv__fs_utime(req));
     X(WRITE, uv__fs_write(req));
+    case UV_FS_OPEN:
+#ifdef O_CLOEXEC
+      /* Try O_CLOEXEC before entering locks */
+      if (!no_cloexec_support) {
+        r = open(req->path, req->flags | O_CLOEXEC, req->mode);
+        if (r >= 0)
+          break;
+        if (errno != EINVAL)
```
rmustacc, line 658: Have you verified that all the platforms will actually return EINVAL in this case? Also this is the other spot of inconsistent meaning of the O_CLOEXEC macro. This causes problems with the locking surrounding fork down below.

rmustacc:
@indutny Sorry for the delay, comments inline.

rmustacc:
@indutny Does libuv guarantee that all other operations that could manipulate file descriptors are guaranteed to happen serially with respect to the uv apis that call fork? eg. sockets, pipes, etc. Given that the current code base generally only uses the atomic methods under `ifdef __linux__`, most platforms would have problems if that wasn't the case.

indutny:
@rmustacc all other operations are happening in the same thread, where `fork()` happens. Or are not opening any fds at all, this is the only problematic case indeed.

src/unix/fs.c
indutny, line 658: Sorry, I don't seem to be getting it. What kind of problems does it cause below? If this line is reached and `errno` is `EINVAL` it'll just keep note that `O_CLOEXEC` does not work and will enter critical section.

src/unix/internal.h
indutny, line 120: Honestly saying I have no idea what consequences will it have, but let me check it in a virtualbox.

src/unix/process.c
indutny: This was a mistake, fixed it

src/unix/fs.c
rmustacc, line 658: There were a few issues. One was the ifdef logic down below. The thing that you want to verify is that all the platforms you care about will return EINVAL when they don't support O_CLOEXEC. If, for example, the kernel doesn't error out on things it doesn't recognize (which does happen in some interfaces), then uv will mistakenly thing it has opened the fd O_CLOEXEC when it hasn't.

src/unix/internal.h
rmustacc, line 120: I think you'll want to test just doing the proper build with __POSIX_VISIBLE rather than trying to hardcode a header constant to work around it. It also simplifies the question of what happens on older FreeBSD platforms that don't support O_CLOEXEC.

src/unix/fs.c
bnoordhuis, line 658: > If, for example, the kernel doesn't error out on things it doesn't recognize (which does happen in some interfaces), then uv will mistakenly thing it has opened the fd O_CLOEXEC when it hasn't.

That would make it a kernel bug.  Returning EINVAL for unrecognized flags is mandated by POSIX.  Doesn't mean it's a bad thing to check, of course, but I would be taken aback if one of our supported platforms manages to screw up such a basic thing.

rmustacc, line 658: I would be surprised as well. But generally most systems are backwards compatible and not forwards compatible, meaning you can bring a binary forward, but not backwards. While I don't suspect it to show up in as much in `open()` and `O_CLOEXEC`, in general, it's problematic and error prone because it's hard to anticipate how things will change arbitrarily in the future.

src/unix/internal.h
bnoordhuis, line 120: The reason for doing it like that is that it doesn't depend on build-time flags because that can break down when people integrate libuv into a larger project and it's not susceptible to the issue of compiling on $PLATFORM_VERSION but running it on $PLATFORM_VERSION-1.

Probably less of an issue on FreeBSD because the core system evolves as a whole (that is, there is no ABI stability between major releases) but you can imagine a statically linked binary that's compiled on FreeBSD 10 but run on FreeBSD 9.

indutny, line 120: Oh right, I forgot about it.

src/unix/fs.c
indutny, line 658: @rmustacc I think I fixed that `ifdef` thing, haven't I?

bnoordhuis, line 658: > I would be surprised as well. But generally most systems are backwards compatible and not forwards compatible, meaning you can bring a binary forward, but not backwards

Libuv (and by extension node.js) on Linux is designed to do exactly that.  You can compile it on a 3.13 system and run it on 2.6.9 and it will work.

indutny:
I agree with @bnoordhuis on the OS support, if you have no other objections - I'd like to land it. Thank you!

rmustacc:
In general, I don't think that forwards compatibility is something that makes sense, but given that's the current choice and design of libuv, what you have there seems fine.

indutny:
Ok, thanks for reviewing it. Going to land it.

indutny:
Thank you, landed in 6abe1e4

PR #1092, opened 2014-01-31
osx: thread starts before the assignments

obastemur:
Changed the order of the member assignments since the thread may start just before the parameters are assigned. 

This especially happens when the osx scheduler is very busy.


src/unix/stream.c
```diff
@@ -353,16 +353,20 @@ int uv__stream_try_select(uv_stream_t* stream, int* fd) {
   s->fake_fd = fds[0];
   s->int_fd = fds[1];
 
-  if (uv_thread_create(&s->thread, uv__stream_osx_select, stream))
-    goto fatal4;
-
   s->stream = stream;
   stream->select = s;
+  const int old_fd = *fd;
```
indutny: C89 style declarations should be at the top of the block.

indutny:
One small nit, otherwise LGTM. Thanks for catching it!

obastemur:
Done. Thanks.

indutny:
Thank you, landed in a6ff04d2c48be408cdcd76e5d78f4dc0a8fb9312

PR #1095, opened 2014-01-31
gitignore: don't use bare leading slash

calid:
The current .gitignore does not work correctly with my version of git (1.8.4), as some test paths are not getting filtered out:

```
# On branch gitignore
# Untracked files:
#   (use "git add <file>..." to include in what will be committed)
#
#   test/.libs/
#   test/run-tests
nothing added to commit but untracked files present (use "git add" to track)
```

Removing the leading slash seems to always filter correctly, regardless of version


indutny:
This is rather surprising, as I am using the same version too and it seems to be working fine for me.

There is a big semantical difference between `/file` and `file` in git ignore. First one matches exactly one filename and the latter one matches every file with name `file`.

Please update your git version, or just ignore them ;)

calid:
@indutny I upgraded to latest and greatest and still have the issue:

selection_004

That's on my arch linux box, but to prove the case I also installed the bleeding edge on my Fedora 19, and see the issue there as well:

selection_005

Reading the gitignore man page I think the problem is the test filters are not rooted at /test:

>  A leading slash matches the beginning of the pathname. For example, "/*.c" matches "cat-file.c" but not "mozilla-sha1/sha1.c".

I've modified my previous patch to leave the existing leading slashes, but to explicitly root test ignores at /test.  Let me know what you think (I'm not sure if I have to submit a new pull request for you to see the diffs).

Thanks!

.gitignore
```diff
@@ -11,44 +11,44 @@ vgcore.*
...
 # Generated by dtrace(1) when doing an in-tree build.
-/include/uv-dtrace.h
+include/uv-dtrace.h
 
 # Generated by gyp for android
 *.target.mk
 
-/out/
-/build/gyp
+out/
```
indutny, line 43: Please keep this as it is.

.gitignore
```diff
@@ -11,44 +11,44 @@ vgcore.*
...
+build/gyp
```
indutny, line 44: Ditto.

.gitignore
```diff
@@ -11,44 +11,44 @@ vgcore.*
...
+libuv.so
+ltmain.sh
+missing
+test-driver
 Makefile
 Makefile.in
 
 # Generated by dtrace(1) when doing an in-tree build.
-/include/uv-dtrace.h
+include/uv-dtrace.h
```
indutny, line 38: Are you sure about this?

indutny:
nvm, I'll push fixes myself

calid:
I think you're still referencing the old diff (it seems github doesn't update the branch if the issue is closed), on the current branch those lines are unchanged: calid@979444c

all I've added is:

``` C

/test/.libs/
/test/run-tests
/test/run-tests.exe
/test/run-tests.dSYM
/test/run-benchmarks
/test/run-benchmarks.exe
/test/run-benchmarks.dSYM
```

indutny:
Oh, I think github has stopped updating it after I closed it.

indutny:
LGTM

indutny:
Thank you, landed in 60d621c6072062cbb65458a1d9d721dac5d9b41f

PR #1097, opened 2014-02-04
gyp: qualify `library` variable

indutny:
Use prefixed variable name for better control over dependencies.


saghul:
LGTM, but why are there so many failing tests?

indutny:
I have no idea :) Looks good on my machine. Thank you, landed in 7301d21.

PR #1100, opened 2014-02-05
linux: always deregister closing fds from epoll

### LINKED ISSUE #1099: closing a shared socket can lead to 100% CPU usage
Another manifestation of #826. When a socket is closed in one process, if it is open in another process, it may stay registered in epoll and thus continously generate junk events that can't be silenced. Reproducible by the following node.js program:

``` js
var cluster = require('cluster');

if (cluster.isMaster) {
    console.log("%d: master", process.pid);
    cluster.fork();
} else {
    // Create a new fd. The fd will be shared with the master.
    var sock = require('dgram').createSocket('udp4');
    // Send a packet to get the socket registered in the epoll.
    sock.send(new Buffer(1000), 0, 1000, 3000, '127.0.0.1', end);
    function end() {
        // Close the socket.
        sock.close();
        console.log("%d: worker: done", process.pid);
        // At this point, the socket is not deregistered from the epoll fd,
        // since it is still open - but not in this process.
        // It is still in EPOLLIN|EPOLLOUT mode.
        // It also cannot be removed, since we no longer have an fd for it.
        // The process will continuously receive EPOLLOUT events, causing 100% CPU usage.
    }
}
```

This patch corrects this behaviour, although as I understand it, it is not acceptable as-is because of the performance of the extra system call.

``` patch
diff --git a/deps/uv/src/unix/linux-core.c b/deps/uv/src/unix/linux-core.c
index 74294eb..351fdd0 100644
--- a/deps/uv/src/unix/linux-core.c
+++ b/deps/uv/src/unix/linux-core.c
@@ -98,21 +98,11 @@ void uv__platform_loop_delete(uv_loop_t* loop) {


 void uv__platform_invalidate_fd(uv_loop_t* loop, int fd) {
-  struct uv__epoll_event* events;
-  uintptr_t i;
-  uintptr_t nfds;
-
-  assert(loop->watchers != NULL);
-
-  events = (struct uv__epoll_event*) loop->watchers[loop->nwatchers];
-  nfds = (uintptr_t) loop->watchers[loop->nwatchers + 1];
-  if (events == NULL)
-    return;
-
-  /* Invalidate events with same file descriptor */
-  for (i = 0; i < nfds; i++)
-    if ((int) events[i].data == fd)
-      events[i].data = -1;
+  /* Pass in a dummy epoll_event,
+   * to work around a bug in old kernels. */
+  struct uv__epoll_event e;
+  if (loop->backend_fd == -1) return;
+  uv__epoll_ctl(loop->backend_fd, UV__EPOLL_CTL_DEL, fd, &e);
 }
```

goffrie:
If the same file description is open in two different processes, then
closing the file descriptor is not sufficient to deregister it from the
epoll instance (as described in epoll(7)), resulting in spurious events
that cause the event loop to spin repeatedly. So always explicitly
deregister it.

Fixes #1099.

### DIFF
--- a/src/unix/linux-core.c
+++ b/src/unix/linux-core.c
@@ -107,6 +107,7 @@ void uv__platform_loop_delete(uv_loop_t* loop) {
 
 void uv__platform_invalidate_fd(uv_loop_t* loop, int fd) {
   struct uv__epoll_event* events;
+  struct uv__epoll_event dummy;
   uintptr_t i;
   uintptr_t nfds;
 
@@ -114,13 +115,20 @@ void uv__platform_invalidate_fd(uv_loop_t* loop, int fd) {
 
   events = (struct uv__epoll_event*) loop->watchers[loop->nwatchers];
   nfds = (uintptr_t) loop->watchers[loop->nwatchers + 1];
-  if (events == NULL)
-    return;
+  if (events != NULL)
+    /* Invalidate events with same file descriptor */
+    for (i = 0; i < nfds; i++)
+      if ((int) events[i].data == fd)
+        events[i].data = -1;
 
-  /* Invalidate events with same file descriptor */
-  for (i = 0; i < nfds; i++)
-    if ((int) events[i].data == fd)
-      events[i].data = -1;
+  /* Remove the file descriptor from the epoll.
+   * This avoids a problem where the same file description remains open
+   * in another process, causing repeated junk epoll events.
+   *
+   * We pass in a dummy epoll_event, to work around a bug in old kernels.
+   */
+  if (loop->backend_fd >= 0)
+    uv__epoll_ctl(loop->backend_fd, UV__EPOLL_CTL_DEL, fd, &dummy);
 }
 
 
--- a/test/test-list.h
+++ b/test/test-list.h
@@ -238,6 +238,7 @@ TEST_DECLARE   (spawn_setuid_setgid)
 TEST_DECLARE   (we_get_signal)
 TEST_DECLARE   (we_get_signals)
 TEST_DECLARE   (signal_multiple_loops)
+TEST_DECLARE   (closed_fd_events)
 #endif
 #ifdef __APPLE__
 TEST_DECLARE   (osx_select)
@@ -479,6 +480,7 @@ TASK_LIST_START
   TEST_ENTRY  (we_get_signal)
   TEST_ENTRY  (we_get_signals)
   TEST_ENTRY  (signal_multiple_loops)
+  TEST_ENTRY  (closed_fd_events)
 #endif
 
 #ifdef __APPLE__
--- a/test/test-spawn.c
+++ b/test/test-spawn.c
@@ -40,6 +40,7 @@ static char exepath[1024];
 static size_t exepath_size = 1024;
 static char* args[3];
 static int no_term_signal;
+static int timer_counter;
 
 #define OUTPUT_SIZE 1024
 static char output[OUTPUT_SIZE];
@@ -118,6 +119,12 @@ static void on_read(uv_stream_t* tcp, ssize_t nread, const uv_buf_t* buf) {
 }
 
 
+static void on_read_once(uv_stream_t* tcp, ssize_t nread, const uv_buf_t* buf) {
+  uv_read_stop(tcp);
+  on_read(tcp, nread, buf);
+}
+
+
 static void write_cb(uv_write_t* req, int status) {
   ASSERT(status == 0);
   uv_close((uv_handle_t*)req->handle, close_cb);
@@ -142,6 +149,11 @@ static void init_process_options(char* test, uv_exit_cb exit_cb) {
 static void timer_cb(uv_timer_t* handle, int status) {
   uv_process_kill(&process, /* SIGTERM */ 15);
   uv_close((uv_handle_t*)handle, close_cb);
+}
+
+
+static void timer_counter_cb(uv_timer_t* handle, int status) {
+  ++timer_counter;
 }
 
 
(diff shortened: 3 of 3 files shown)


src/unix/linux-core.c
```diff
@@ -121,6 +122,16 @@ void uv__platform_invalidate_fd(uv_loop_t* loop, int fd) {
...
     if ((int) events[i].data == fd)
       events[i].data = -1;
+
+  /* Remove the file descriptor from the epoll.
+   * This avoids a problem where the same file description remains open
+   * in another process, causing repeated junk epoll events.
+   *
+   * We pass in a dummy epoll_event, to work around a bug in old kernels.
+   */
+  if (loop->backend_fd >= 0) {
```
indutny: Please omit braces here.

indutny:
Thank you! LGTM, but it really needs to have a test.

Just FYI, the relevant files to introduce a new test:
- https://github.com/joyent/libuv/blob/master/test/test-list.h (to add new test declaration to the list)
- https://github.com/joyent/libuv/blob/master/test/test-spawn.c (test implementation)
- https://github.com/joyent/libuv/blob/master/test/run-tests.c#L156 (for writing a spawn helper)

indutny:
@goffrie just thought that you may find this useful for implementing test: https://gist.github.com/indutny/8842007 . Anyway, if you don't feel like you want or have time to do it, let me know.

goffrie:
Alright, I fixed a little issue (`return` early-exit was messing things up) and added a test.

test/test-closed-fd-events.c
```diff
@@ -0,0 +1,131 @@
...
+  uv_read_stop(handle);
+}
+
+static void timeout(uv_timer_t* handle, int status) {
+  timed_out = 1;
+}
+
+TEST_IMPL(closed_fd_events) {
+  /* spawn_helper4 blocks indefinitely. */
+  char *argv[] = { NULL, "spawn_helper4", NULL };
```
indutny: `char* argv`, not `char *argv`, please

test/test-closed-fd-events.c
```diff
@@ -0,0 +1,131 @@
...
+#include <unistd.h>
+
+static int timed_out;
+static int read_cb_called;
+
+static void alloc_cb(uv_handle_t *handle, size_t size, uv_buf_t *buf) {
+  static char slab[1];
+  buf->base = slab;
+  buf->len = sizeof(slab);
+}
```
indutny: two newlines between functions

test/test-closed-fd-events.c
```diff
@@ -0,0 +1,131 @@
...
+  r = pipe(fd);
+  ASSERT(r == 0);
+
+  r = fcntl(fd[0], F_SETFL, O_NONBLOCK);
+  ASSERT(r == 0);
+
+  memset(&options, 0, sizeof(options));
+  exepath_size = sizeof(exepath);
+
+  r = uv_exepath(exepath, &exepath_size);
```
indutny: I'd rather do `ASSERT(0 == uv_exepath(...))`.

test/test-closed-fd-events.c
```diff
@@ -0,0 +1,131 @@
...
+  size_t exepath_size;
+  char exepath[256];
+  uv_process_t process;
+  int r;
+  uv_pipe_t pipe_handle;
+  int fd[2];
+  uv_timer_t timer;
+
+  /* create a pipe and share it with a child process */
+  r = pipe(fd);
```
indutny: Ditto, and everywhere else.

test/test-closed-fd-events.c
```diff
@@ -0,0 +1,131 @@
...
+  ASSERT(r == 0);
+
+  /* read from the pipe with uv */
+  r = uv_pipe_init(uv_default_loop(), &pipe_handle, 0);
+  ASSERT(r == 0);
+  r = uv_pipe_open(&pipe_handle, fd[0]);
+  ASSERT(r == 0);
+  fd[0] = -1;
+
+  read_cb_called = 0;
```
indutny: No need to initialize it, all static variables are defaulting to `0`.

test/test-closed-fd-events.c
```diff
@@ -0,0 +1,131 @@
...
+  r = uv_read_start((uv_stream_t *) &pipe_handle, alloc_cb, read_cb);
+  ASSERT(r == 0);
+
+  r = write(fd[1], "", 1);
+  ASSERT(r == 1);
+
+  uv_run(uv_default_loop(), UV_RUN_ONCE);
```
indutny: I guess it should be `ASSERT(0 == uv_run(...))`

test/test-closed-fd-events.c
```diff
@@ -0,0 +1,131 @@
...
+
+  ASSERT(read_cb_called == 1);
+
+  /* close the pipe and see if we still get events */
+  uv_close((uv_handle_t *) &pipe_handle, 0);
```
indutny: Please use `close_cb` here, not `0`.

test/test-closed-fd-events.c
```diff
@@ -0,0 +1,131 @@
...
+
+  r = write(fd[1], "", 1);
+  ASSERT(r == 1);
+
+  uv_timer_init(uv_default_loop(), &timer);
+  timed_out = 0;
+  uv_timer_start(&timer, timeout, 10, 0);
+
+  /* see if any spurious events interrupt the timer */
+  uv_run(uv_default_loop(), UV_RUN_ONCE);
```
indutny: Why are you calling it two times? Needs are a bit more details comment, I think.

test/test-closed-fd-events.c
```diff
@@ -0,0 +1,131 @@
...
+  r = fcntl(fd[0], F_SETFL, O_NONBLOCK);
+  ASSERT(r == 0);
+
+  memset(&options, 0, sizeof(options));
+  exepath_size = sizeof(exepath);
+
+  r = uv_exepath(exepath, &exepath_size);
+  ASSERT(r == 0);
+
+  argv[0] = exepath;
```
indutny: I think you could move this test into `test-spawn.c` and use `init_process_options` function which does absolutely the same.

indutny:
Looks awesome! Thank you very much! Some style issue, but generally LGTM. I hope you'll agree with me that it would be simpler to have it in `test-spawn.c`, rather than in a separate file ;)

test/test-closed-fd-events.c
goffrie: For some reason the timer doesn't trigger with just one iteration of UV_RUN_ONCE, but the second is enough. I don't know why this is.

goffrie:
Alright, I've moved the test into `test-spawn.c` and made the style fixes you mentioned.

indutny:
@goffrie looks totally good now! ;) Will land it with very minor style fixes that I'll make myself.

Could you please sign [CLA](http://nodejs.org/cla.html) to make me be able to push it to github?

goffrie:
Sorry for the delay -- I'm waiting for my employer to get back to me.

indutny:
Hey man! Any news?

goffrie:
I've signed the CLA now. :)

indutny:
Thank you, landed in 780d8ad8e58c675dc60f5b73ea22a203760da2f2

Issue #839, opened 2013-06-24
Integrate FS operations into uv_write?


Keno:
The filesystem operations being so separate in the current implementation is a little unfortunate. It would be nice if writing to files could be integrated with the rest of the libuv I/O types to provide a more uniform interface. There is of course the issue that unlike the other types of I/O, files support positional writes whereas the other types of I/O do not. Discussing this with @bnoordhuis on IRC, one solution might be to have `uv_pwrite` for files that at least uses the same API and callback structure. 

/cc @vtjnash

vtjnash:
Is it crazy to think that the exact same uv_write call should be usable for file handles? The harder issue that I see is with synchronizing reads and writes. If I want file-like access then it makes sense to me to use the existing file API.

(edit: deleted text "It seems that you just fixed the julia issue with this anyways" since it only applies to loladiro and doesn't make sense in this context)

saghul:
I wouldn't object exporing the possibility of creating some "fs_stream" handle, which would use the majority of the stream functions, but as it stands we cannot merge the functionality, we need a new handle for that.

Please open an RFC/issue at libuv/libuv if this is still something you'd like to see / work on.

PS: I'm really sorry for the huge delay...

PR #1103, opened 2014-02-09
Vectored IO for uv_fs_read/write

Ralith:
I don't have a Windows development environment on hand, so the emulation of vectored IO under Windows is untested.

These changes should make #839 easier.

I had to fix test-udp-multicast-interface's use of the old `uv_ip4_addr` API to get the tests to build.


Ralith:
Test are failing to link `preadv` and `pwritev`. This is a bit perplexing, as those are defined in glibc >=2.10. They are, however, not POSIX, so perhaps there's some feature test gate hiding them? Would it be better to fall back on emulating vectored IO when an offset is used?

On further review, it appears that preadv and pwritev may also not be available on Darwin. An emulated approach along the lines of what I did for windows could be used there, though that's another platform I'm not able to test on.

bnoordhuis:
preadv() and pwritev() are BSD-isms.  Solaris and OS X don't have them and Linux only acquired them about four years ago.

It's okay to use newer system calls in libuv but because the baseline is kernel 2.6.9 and glibc 2.3.3 you need to make the system call directly and have a fallback for the ENOSYS case.  Most of the relevant code is in src/unix/linux-syscalls.[ch].

saghul:
> I had to fix test-udp-multicast-interface's use of the old uv_ip4_addr API to get the tests to build.

My fault, sorry about that. It's fixed in master now.

src/unix/fs.c
```diff
@@ -191,10 +192,39 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
...
+#else
+# if defined(__linux__)
+    result = uv__preadv(req->file, (struct iovec*)req->bufs, req->nbufs, req->off);
+    if(result == -1 && errno == ENOSYS)
+# endif
+    {
+      off_t cursor;
+      size_t index;
+      size_t bytes = 0;
+      for(index = 0, result = 0, cursor = req->off;
```
indutny: Perhaps, it'd be better to rewrite this as a while loop?

src/unix/fs.c
```diff
@@ -865,14 +923,23 @@ int uv_fs_open(uv_loop_t* loop,
...
+
+  req->nbufs = nbufs;
+  req->bufs = req->bufsml;
+  if (nbufs > ARRAY_SIZE(req->bufsml))
+    req->bufs = malloc(nbufs * sizeof(bufs[0]));
+
+  if (req->bufs == NULL)
+    return -ENOMEM;
+
+  memcpy(req->bufs, bufs, nbufs * sizeof(bufs[0]));
```
indutny: Style nit: `sizeof(*bufs)`

src/unix/linux-syscalls.c
```diff
@@ -386,3 +406,19 @@ int uv__utimesat(int dirfd,
   return errno = ENOSYS, -1;
 #endif
 }
+
+ssize_t uv__preadv(int fd, const struct iovec *iov, int iovcnt, off_t offset) {
+#if defined(__NR_preadv)
+  return syscall(__NR_preadv, fd, iov, iovcnt, offset);
+#else
+  return errno = ENOSYS, -1;
```
indutny, line 414: Rather complicated, no? Why not just split it into two separate lines?

bnoordhuis, line 414: Aligns with the style used throughout the rest of the file though.

src/unix/fs.c
Ralith: This as written to be consistent with the existing vectored IO code, e.g. src/unix/udp.c:411 . If *bufs is preferred, should that be changed too?

Ralith:
I've rewritten the large for loops as while loops, removed a redundant #include, and improved discrimination of native preadv/pwritev availability.

src/win/fs.c
```diff
@@ -573,7 +575,16 @@ void fs__read(uv_fs_t* req) {
     overlapped_ptr = NULL;
   }
 
-  if (ReadFile(handle, req->buf, req->length, &bytes, overlapped_ptr)) {
+  for(result = 1, index = 0, bytes = 0;
+      result && index < req->nbufs;
+      ++index) {
+    DWORD incremental_bytes;
+    result = ReadFile(handle, req->bufs[index].base, req->bufs[index].len,
```
saghul: Error is only checked after the for loop. What about checkint for it on each iteration and bailing early if so?

src/win/fs.c
```diff
@@ -620,7 +633,15 @@ void fs__write(uv_fs_t* req) {
     overlapped_ptr = NULL;
   }
 
-  if (WriteFile(handle, req->buf, length, &bytes, overlapped_ptr)) {
+  for(result = 1, index = 0, bytes = 0;
+      result && index < req->nbufs;
+      ++index) {
+    DWORD incremental_bytes;
+    result = WriteFile(handle, req->bufs[index].base, req->bufs[index].len, &incremental_bytes, overlapped_ptr);
```
saghul: Ditto.

src/win/fs.c
```diff
@@ -1569,13 +1590,27 @@ int uv_fs_close(uv_loop_t* loop, uv_fs_t* req, uv_file fd, uv_fs_cb cb) {
...
-  req->buf = buf;
-  req->length = length;
+
+  req->nbufs = nbufs;
+  req->bufs = req->bufsml;
+  if (nbufs > ARRAY_SIZE(req->bufsml))
+    req->bufs = malloc(nbufs * sizeof(bufs[0]));
+
+  if (req->bufs == NULL)
+    return -ENOMEM;
```
saghul: Use UV_ENOMEM, we only use -ENOMEM on unix.

src/win/fs.c
```diff
@@ -1593,8 +1628,17 @@ int uv_fs_write(uv_loop_t* loop, uv_fs_t* req, uv_file fd, const void* buf,
...
-  req->buf = (void*) buf;
-  req->length = length;
+
+  req->nbufs = nbufs;
+  req->bufs = req->bufsml;
+  if (nbufs > ARRAY_SIZE(req->bufsml))
+    req->bufs = malloc(nbufs * sizeof(bufs[0]));
+
+  if (req->bufs == NULL)
+    return -ENOMEM;
```
saghul: Ditto.

saghul:
Left a few comments, but looks like we are close!

Ralith:
I've replaced the win32 for loops with while loops for improved clarity and consistency w/ the unix implementation, and swapped in UV_ENOMEM.

src/unix/fs.c
```diff
@@ -191,10 +201,37 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
...
+  else {
+#if HAVE_PREADV
+    result = preadv(req->file, (struct iovec*)req->bufs, req->nbufs, req->off);
+#else
+# if defined(__linux__)
+    result = uv__preadv(req->file, (struct iovec*)req->bufs, req->nbufs, req->off);
+    if(result == -1 && errno == ENOSYS)
+# endif
+    {
+      off_t cursor = 0;
```
saghul: Declare and initialize these in different lines.

src/unix/fs.c
```diff
@@ -507,14 +544,40 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
...
+  else {
+#if HAVE_PREADV
+    r = pwritev(req->file, (struct iovec*)req->bufs, req->nbufs, req->off);
+#else
+# if defined(__linux__)
+    r = uv__pwritev(req->file, (struct iovec*)req->bufs, req->nbufs, req->off);
+    if(r == -1 && errno == ENOSYS)
+# endif
+    {
+      off_t cursor = 0;
```
saghul: Ditto.

src/unix/fs.c
```diff
@@ -507,14 +544,40 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
...
+      size_t index = 0;
+      r = 0;
+      while(index < req->nbufs && r > 0) {
+        r = pwrite(req->file, req->bufs[index].base, req->bufs[index].len, cursor);
```
saghul: What if we do 3 pwrites, and the one in the middle fails? This code continues writing, shouldn't it stop?

saghul:
LGTM, but I'm uncertain about what we should do in case of error when pread/pwrite emulations are used. Thoughts?

src/unix/fs.c
Ralith: If any write fails, `r > 0` will become false and the while loop will terminate, preventing any further writes.

Ralith:
Under emulated pread/pwrite, IO ceases if an error occurs. Only if an error occurs after some data has been read or written is it suppressed, and even then no additional IO is automatically performed. Is there any case in which this behavior might cause a problem? It seems like any error that occurs after some successful IO should recur on a subsequent IO operation, or if it fails to do so it is arguably not of consequence.

src/unix/fs.c
```diff
@@ -507,14 +547,43 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
...
+    if(r == -1 && errno == ENOSYS)
+# endif
+    {
+      off_t cursor;
+      size_t index;
+
+      cursor = 0;
+      index = 0;
+      r = 0;
+      while(index < req->nbufs && r > 0) {
```
saghul: r >= 0? Looks like a valid value according to the man page.

saghul:
> Under emulated pread/pwrite, IO ceases if an error occurs. Only if an error occurs after some data has been read or written is it suppressed, and even then no additional IO is automatically performed. Is there any case in which this behavior might cause a problem? It seems like any error that occurs after some successful IO should recur on a subsequent IO operation, or if it fails to do so it is arguably not of consequence.

Well, it might be a bit of a stretch, but since we are writing to files, disk might become full between writes. Now, I'm ok with swallowing the error, if we return the right amount of data we actually wrote.

src/unix/fs.c
```diff
@@ -191,10 +201,40 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
...
+      off_t cursor;
+      size_t index;
+
+      cursor = 0;
+      index = 0;
+      result = 0;
+      while(index < req->nbufs && result > 0) {
+        result = pread(req->file, req->bufs[index].base, req->bufs[index].len, cursor);
+        index++;
+        if(result > 0) {
```
indutny: Style nit: please remove braces.

src/unix/fs.c
```diff
@@ -191,10 +201,40 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
...
+          cursor += result;
+        }
+      }
+      if(cursor > 0) {
```
indutny: Ditto.

src/unix/fs.c
```diff
@@ -44,10 +44,20 @@
 #include <utime.h>
 #include <poll.h>
 
+#define HAVE_PREADV \
+  (defined(__GLIBC__) && __GLIBC__ >= 2 && __GLIBC_MINOR__ >= 10) ||    \
```
indutny: May be turn it the other way around?

``` C
#if (defined(__GLIBC__) ...
# define HAVE_PREADV 1
#else
# define HAVE_PREADV 0
#endif
```

src/unix/fs.c
```diff
@@ -191,10 +201,40 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
 
 
 static ssize_t uv__fs_read(uv_fs_t* req) {
+  ssize_t result;
   if (req->off < 0)
-    return read(req->file, req->buf, req->len);
-  else
-    return pread(req->file, req->buf, req->len, req->off);
+    result = readv(req->file, (struct iovec*)req->bufs, req->nbufs);
```
indutny: Style nit: `(struct iovec*) req->bufs`

src/win/fs.c
```diff
@@ -620,7 +635,17 @@ void fs__write(uv_fs_t* req) {
     overlapped_ptr = NULL;
   }
 
-  if (WriteFile(handle, req->buf, length, &bytes, overlapped_ptr)) {
+  result = 1;
+  index = 0;
+  bytes = 0;
+  while(result && index < req->nbufs) {
```
indutny: Style nit: `while (`.

src/unix/fs.c
```diff
@@ -191,10 +201,40 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
...
+    if(result == -1 && errno == ENOSYS)
+# endif
+    {
+      off_t cursor;
+      size_t index;
+
+      cursor = 0;
+      index = 0;
+      result = 0;
+      while(index < req->nbufs && result > 0) {
```
indutny: Ditto.

src/win/fs.c
```diff
@@ -573,7 +575,18 @@ void fs__read(uv_fs_t* req) {
     overlapped_ptr = NULL;
   }
 
-  if (ReadFile(handle, req->buf, req->length, &bytes, overlapped_ptr)) {
+  result = 1;
+  index = 0;
+  bytes = 0;
+  while(result && index < req->nbufs) {
```
indutny: Ditto.

src/unix/fs.c
```diff
@@ -191,10 +201,40 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
...
-  else
-    return pread(req->file, req->buf, req->len, req->off);
+    result = readv(req->file, (struct iovec*)req->bufs, req->nbufs);
+  else {
+#if HAVE_PREADV
+    result = preadv(req->file, (struct iovec*)req->bufs, req->nbufs, req->off);
+#else
+# if defined(__linux__)
+    result = uv__preadv(req->file, (struct iovec*)req->bufs, req->nbufs, req->off);
+    if(result == -1 && errno == ENOSYS)
```
indutny: Ditto.

src/unix/fs.c
indutny: Also, please add space between `if` and `(`.

src/unix/fs.c
```diff
@@ -191,10 +201,40 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
...
+          cursor += result;
+        }
+      }
+      if(cursor > 0) {
+        result = cursor;
+      }
+    }
+#endif
+  }
+  if(req->bufs != req->bufsml)
```
indutny: Ditto.

src/unix/fs.c
indutny: Ditto.

indutny: @saghul I think this is to prevent infinite looping here.

src/unix/fs.c
```diff
@@ -507,14 +547,43 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
...
+      off_t cursor;
+      size_t index;
+
+      cursor = 0;
+      index = 0;
+      r = 0;
+      while(index < req->nbufs && r > 0) {
+        r = pwrite(req->file, req->bufs[index].base, req->bufs[index].len, cursor);
+        index++;
+        if(r > 0) {
```
indutny: Remove braces.

src/unix/fs.c
```diff
@@ -507,14 +547,43 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
...
+          cursor += r;
+        }
+      }
+      if(cursor > 0) {
```
indutny: Ditto.

src/unix/fs.c
indutny: And add space.

src/unix/linux-syscalls.c
indutny, line 414: Ok, gotcha

src/unix/fs.c
bnoordhuis: Apropos the glibc check, it's the wrong thing to check for because libuv doesn't (or shouldn't) use the glibc wrappers.

src/unix/fs.c
bnoordhuis: Record the fact that the kernel doesn't support preadv() so you don't keep making the system call unnecessarily in the future.

src/unix/fs.c
bnoordhuis: r == 0 means EOF or that you passed zero-length input.  This check looks bogus though because at the first iteration r == 0, hence r > 0 is automatically false.

Ralith:
Applied style tweaks and fixes.

The purpose of unix/fs.c:230 and unix/fs.c:579 is to return the amount of data we read/wrote rather than an error if we do successfully perform some IO. If e.g. the filesystem becomes full, then that error should either recur on a subsequent call, or be resolved in the intervening time and not really matter.

src/unix/fs.c
```diff
@@ -507,14 +556,47 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
...
+      size_t index;
+
+    retry:
+      cursor = 0;
+      index = 0;
+      do {
+        r = pwrite(req->file, req->bufs[index].base, req->bufs[index].len, cursor);
+        index++;
+        if (r > 0) cursor += r;
+      } while (index < req->nbufs && r > 0);
```
saghul, line 600: This would only write 1 buffer if I call uv_fs_write with this 3 buffers: "foo", "", "bar", IMHO the while condition needs to be adjusted to r > = 0.

src/unix/fs.c
```diff
@@ -191,10 +203,46 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
...
+      off_t cursor;
+      size_t index;
+
+    retry:
+      cursor = 0;
+      index = 0;
+      do {
+        result = pread(req->file, req->bufs[index].base, req->bufs[index].len, cursor);
+        index++;
+        if (result > 0) cursor += result;
```
saghul: Put this if in 2 lines :-)

src/unix/fs.c
```diff
@@ -191,10 +203,46 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
...
+      } while (index < req->nbufs && result > 0);
+      if (cursor > 0) result = cursor;
```
saghul: Ditto.

src/unix/fs.c
```diff
@@ -507,14 +556,47 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
...
+      off_t cursor;
+      size_t index;
+
+    retry:
+      cursor = 0;
+      index = 0;
+      do {
+        r = pwrite(req->file, req->bufs[index].base, req->bufs[index].len, cursor);
+        index++;
+        if (r > 0) cursor += r;
```
saghul: Ditto.

src/unix/fs.c
```diff
@@ -507,14 +556,47 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
...
+      } while (index < req->nbufs && r > 0);
+      if (cursor > 0) r = cursor;
```
saghul: 2 lines.

Ralith:
Applied style fixes and adjusted the emulated preadv and pwritev loops to not abort early if zero-length buffers are passed.

I opted to skip writing buffers of zero length rather than to continue when `r == 0` to prevent a disk becoming full for one write but having room for the next resulting in dropped data.

saghul:
LGTM

indutny:
A lot of minor style fixes are required, by LGTM

Ralith:
Applied style fixes, adjusted emulation on windows to not report a false error if a zero length operation succeeds, squashed into a single commit, rebased onto latest master

src/unix/fs.c
```diff
@@ -191,10 +203,50 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
...
+#else
+# if defined(__linux__)
+    if (no_preadv)
+# endif
+    {
+      off_t cursor;
+      size_t index;
+
+    retry:
+      cursor = 0;
```
indutny: Why are we starting from `0` and not from `req->off` here?

src/unix/fs.c
```diff
@@ -191,10 +203,50 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
...
+      index = 0;
+      do {
+        if (req->bufs[index].len > 0) {
+          result = pread(req->file, req->bufs[index].base, req->bufs[index].len, cursor);
```
indutny: `cursor` should probably stay initialized to `0`, but I think that you should add `req->off` to it here.

src/unix/fs.c
```diff
@@ -507,14 +560,51 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
...
+    {
+      off_t cursor;
+      size_t index;
+
+    retry:
+      cursor = 0;
+      index = 0;
+      do {
+        if (req->bufs[index].len > 0) {
+          r = pwrite(req->file, req->bufs[index].base, req->bufs[index].len, cursor);
```
indutny: Same here, perhaps?

src/unix/fs.c
```diff
@@ -507,14 +560,51 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
...
+          if (r > 0)
+            cursor += r;
+        }
+        index++;
+      } while (index < req->nbufs && r > 0);
```
indutny, line 600: May be reached with unitialized `r`

src/unix/fs.c
```diff
@@ -191,10 +203,50 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
...
+      cursor = 0;
+      index = 0;
+      do {
+        if (req->bufs[index].len > 0) {
+          result = pread(req->file, req->bufs[index].base, req->bufs[index].len, cursor);
+          if (result > 0)
+            cursor += result;
+        }
+        index++;
+      } while (index < req->nbufs && result > 0);
```
indutny, line 238: May be reached with uninitalized `result`

src/win/fs.c
```diff
@@ -573,7 +575,20 @@ void fs__read(uv_fs_t* req) {
...
+    result = ReadFile(handle,
+                      req->bufs[index].base,
+                      req->bufs[index].len,
+                      &incremental_bytes,
+                      overlapped_ptr);
+    bytes += incremental_bytes;
+    ++index;
+  } while (result && index < req->nbufs);
+
+  if (result || bytes > 0) {
```
indutny, line 591: What would be a condition where `result` is negative, but bytes are `positive`? Why should we not fail at it?

src/win/fs.c
```diff
@@ -620,7 +637,16 @@ void fs__write(uv_fs_t* req) {
...
+  index = 0;
+  bytes = 0;
+  do {
+    DWORD incremental_bytes;
+    result = WriteFile(handle, req->bufs[index].base, req->bufs[index].len, &incremental_bytes, overlapped_ptr);
+    bytes += incremental_bytes;
+    ++index;
+  } while (result && index < req->nbufs);
+
+  if (result || bytes > 0) {
```
indutny, line 653: Same question here.

src/win/fs.c
```diff
@@ -620,7 +637,16 @@ void fs__write(uv_fs_t* req) {
     overlapped_ptr = NULL;
   }
 
-  if (WriteFile(handle, req->buf, length, &bytes, overlapped_ptr)) {
+  index = 0;
+  bytes = 0;
+  do {
+    DWORD incremental_bytes;
+    result = WriteFile(handle, req->bufs[index].base, req->bufs[index].len, &incremental_bytes, overlapped_ptr);
```
indutny: 80 column limit

test/test-spawn.c
```diff
@@ -301,8 +303,9 @@ static void timer_cb(uv_timer_t* handle, int status) {
   ASSERT(exit_cb_called == 1);
   ASSERT(close_cb_called == 1);
 
-  r = uv_fs_read(uv_default_loop(), &fs_req, file, output, sizeof(output),
-      0, NULL);
+  buf.base = output;
```
indutny: `uv_buf_init`, please

indutny:
Looks awesome now! A bit more questions and minor nits, hopefully a last round! ;) Thank you for your patience.

Ralith:
Applied fixes.

If `result` is false but `bytes` is positive, it means that one or more operations succeeded before a failure. I treat this as a success because it's important that calling code knows how much, if any, IO occurred so data is not lost or duplicated. I can't think of a case where suppressing errors for partially successful operations would cause problems: if the error is persistent, it will be accurately reported on a subsequent call, and if it is not then it might as well be spurious.

indutny:
Could it be that one of the buffers was written partially? Should we rather let user know that we failed?

Ralith:
If any data is written at all, `bytes` will be nonzero and success will be indicated to the user. If a partial write occurred (I'm not sure `WriteFile` is actually capable of that), that remains the case.

Reviewing libuv documentation, it does seem that this may change the semantics of the API slightly. Previously, filesystem IO operations appear to have been guaranteed to either completely succeed or completely fail, presuming `WriteFile` adheres to those semantics itself. With these changes, calling code will need to handle partial success. We could simply treat partial success as an error for reads, but I'm not sure there's any recourse for partial writes other than reporting them to the caller, and it's probably best to maintain consistent semantics between the two.

src/unix/fs.c
```diff
@@ -191,10 +203,59 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
...
+# if defined(__linux__)
+    if (no_preadv)
+# endif
+    {
+      off_t written, base;
+      size_t index;
+
+    retry:
+      written = 0;
+      base = req->off < 0 ? 0 : req->off;
```
indutny: I don't think that this check is required, this is a false branch of `if (req->off < 0)`.

src/unix/fs.c
```diff
@@ -191,10 +203,59 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
...
+      do {
+        if (req->bufs[index].len > 0) {
+          result = pread(req->file,
+                         req->bufs[index].base,
+                         req->bufs[index].len,
+                         base + written);
+          if (result > 0)
+            written += result;
+        } else {
+          result = 1;
```
indutny: Better initialize it before loop.

src/unix/fs.c
```diff
@@ -507,14 +569,60 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
...
+# if defined(__linux__)
+    if (no_pwritev)
+# endif
+    {
+      off_t written, base;
+      size_t index;
+
+    retry:
+      written = 0;
+      base = req->off < 0 ? 0 : req->off;
```
indutny: Same.

src/unix/fs.c
```diff
@@ -507,14 +569,60 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
...
+      do {
+        if (req->bufs[index].len > 0) {
+          r = pwrite(req->file,
+                     req->bufs[index].base,
+                     req->bufs[index].len,
+                     base + written);
+          if (r > 0)
+            written += r;
+        } else {
+          r = 1;
```
indutny: Ditto.

indutny:
Ok, 4 last things and LGTM

Ralith:
Applied cleanup.

src/unix/fs.c
```diff
@@ -191,10 +203,57 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
 
 
 static ssize_t uv__fs_read(uv_fs_t* req) {
+  static int no_preadv;
```
indutny: should be inside `# if defined(__linux__)`

src/unix/fs.c
```diff
@@ -191,10 +203,57 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
...
+    result = preadv(req->file, (struct iovec*) req->bufs, req->nbufs, req->off);
+#else
+# if defined(__linux__)
+    if (no_preadv)
+# endif
+    {
+      off_t nread;
+      size_t index;
+
+    retry:
```
indutny, line 223: Should be wrapped in `if defined(__linux__)`.

src/unix/fs.c
```diff
@@ -507,14 +567,58 @@ static ssize_t uv__fs_write(uv_fs_t* req) {
...
+    r = pwritev(req->file, (struct iovec*) req->bufs, req->nbufs, req->off);
+#else
+# if defined(__linux__)
+    if (no_pwritev)
+# endif
+    {
+      off_t written;
+      size_t index;
+
+    retry:
```
indutny, line 585: Should be wrapped in `if defined(__linux__)`.

src/unix/fs.c
```diff
@@ -495,6 +554,7 @@ static ssize_t uv__fs_utime(uv_fs_t* req) {
 
 
 static ssize_t uv__fs_write(uv_fs_t* req) {
+  static int no_pwritev;
```
indutny: Please move it below as with `no_preadv`.

Ralith:
Unused static variables and labels both have zero cost; are you sure this is necessary? To remain consistent with the style guide I'd also have to introduce an additional redundant scope to place no_preadv/no_pwritev inside the linux guards.

src/unix/fs.c
```diff
@@ -191,10 +203,57 @@ static ssize_t uv__fs_futime(uv_fs_t* req) {
...
   if (req->off < 0)
-    return read(req->file, req->buf, req->len);
-  else
-    return pread(req->file, req->buf, req->len, req->off);
+    result = readv(req->file, (struct iovec*) req->bufs, req->nbufs);
+  else {
+#if HAVE_PREADV
+    result = preadv(req->file, (struct iovec*) req->bufs, req->nbufs, req->off);
+#else
+# if defined(__linux__)
```
indutny, line 214: you could move `no_preadv` right after this statement, it'll be fine.

indutny:
@Ralith yes, I'm sure. Unused variable warnings is something that I don't really want to see at each build.

src/unix/fs.c
indutny, line 214: Nope, it is right after `else {` starts, `#if ....` doesn't count here as it is ignored by the compiler

indutny:
Thank you for your patience and effort, finally landed in 7df2458

PR #1104, opened 2014-02-10
build: skip vcvarsall if already set up

mscdex:
Without this, running vcbuild repeatedly for some time will cause errors with msbuild saying the command line is too long (due to concatenation happening in vcvarsall).


saghul:
LGTM, @piscisaureus any objection to this?

saghul:
Thanks @mscdex, landed as https://github.com/joyent/libuv/commit/b738d315fbf634f6c571234b8f028e6384eb42ec

Issue #1093, opened 2014-01-31
rpmlint warning: missing-call-to-setgroups-before-setuid


daxim:
When packaging libuv-0.11.19 on openSUSE 13.1, rpmlint reports the warning:

```
libuv11.x86_64: W: missing-call-to-setgroups-before-setuid /usr/lib64/libuv.so.11.0.0
This executable is calling setuid and setgid without setgroups or initgroups.
There is a high probability this mean it didn't relinquish all groups, and
this would be a potential security issue to be fixed. Seek POS36-C on the web
for details about the problem.
```

Is this warning legimitate or bogus? If the former, that should be fixed in the libuv source code. If the latter, I can suppress the warning in the specfile.

Sannis:
@daxim are you plan to deploy that rpm to devel:languages:nodejs or some other channel on openSUSE build services?

daxim:
libuv is one of the indirect dependencies of rakudo; I plan to submit it to devel:languages:parrot.

bnoordhuis:
I came here to report this as a bug.  Yes, this is a potential security issue that needs to be addressed.

(EDIT: Another thing that libuv should do on Linux is drop some or all capabilities.)

/cc @indutny

saghul:
Landed a fix for this warning in https://github.com/joyent/libuv/commit/66ab38918c911bcff025562cf06237d7fedaba0c. We should also drop some (or all) capabilities on Linux, I created https://github.com/joyent/libuv/issues/1106 for that.

lkundrak:
FWIW this patch is incorrect. With size of 0, it doesn't set anything:

>  If size is zero, list is not modified, but the total number of supplementary group IDs for the process is returned. This allows the caller to determine the size of a dynamically allocated list to be used in a further call to getgroups().

saghul:
On 08/18/2015 09:57, Lubomir Rintel wrote:

> FWIW this patch is incorrect. With size of 0, it doesn't set anything:
> 
> ```
> If size is zero, list is not modified, but the total number of
> supplementary group IDs for the process is returned. This allows the
> caller to determine the size of a dynamically allocated list to be
> used in a further call to getgroups().
> ```

My understanding is that what you quoted applies to getgroups, not
setgroups, which is what we are using here. Can you point me to another
resource stating this is incorrect? I'm surprised, because many eyes saw
this patch, as we had to issue a CVE, and I'd hate it if it was wrong
after all. Thanks!

lkundrak:
[...]

PR #1105, opened 2014-02-10
unix: call setgoups before calling setuid/setgid

saghul:
Partial fix for #1093

Shamelessly taken from: https://github.com/mozilla/rust/pull/12085

/cc @indutny

If you have a few milliseconds to spare, it'd be great to get your input, @bnoordhuis.

As for the dropping caps part, I'll need to look into that, doesn't look anywhere straightforward :-/

### DIFF
--- a/src/unix/process.c
+++ b/src/unix/process.c
@@ -330,6 +330,20 @@ static void uv__process_child_init(const uv_process_options_t* options,
     _exit(127);
   }
 
+  if (options->flags & (UV_PROCESS_SETUID | UV_PROCESS_SETGID)) {
+    /* When dropping privileges from root, the `setgroups` call will
+     * remove any extraneous groups. If we don't call this, then
+     * even though our uid has dropped, we may still have groups
+     * that enable us to do super-user things. This will fail if we
+     * aren't root, so don't bother checking the return value, this
+     * is just done as an optimistic privilege dropping function.
+     */
+    int saved_errno;
+    saved_errno = errno;
+    setgroups(0, NULL);
+    errno = saved_errno;
+  }
+
   if ((options->flags & UV_PROCESS_SETGID) && setgid(options->gid)) {
     uv__write_int(error_fd, -errno);
     perror("setgid()");


src/unix/process.c
```diff
@@ -330,6 +330,17 @@ static void uv__process_child_init(const uv_process_options_t* options,
     _exit(127);
   }
 
+  if (options->flags & (UV_PROCESS_SETUID|UV_PROCESS_SETGID)) {
```
indutny: Please add space between: `...SETUID | ...SETGID`

src/unix/process.c
```diff
@@ -330,6 +330,17 @@ static void uv__process_child_init(const uv_process_options_t* options,
...
+    /* When dropping privileges from root, the `setgroups` call will
+     * remove any extraneous groups. If we don't call this, then
+     * even though our uid has dropped, we may still have groups
+     * that enable us to do super-user things. This will fail if we
+     * aren't root, so don't bother checking the return value, this
+     * is just done as an optimistic privilege dropping function.
+     */
+    setgroups(0, NULL);
```
indutny, line 343: Perhaps, clear errno?

indutny, line 343: I mean restore.

indutny:
Minor style nit, otherwise LGTM. But yeah, having @bnoordhuis opinion on this would be totally cool!

src/unix/process.c
saghul, line 343: I thought about it, but we don't check for it unless any of the remaining function calls fail, so does it matter much? Or is it just good practice? I could just set errno to 0 if setgroups fails.

indutny, line 343: Better just restore it, yeah it is just for the sake of style.

saghul:
Force pushed addressing feedback, thanks @indutny!

src/unix/process.c
```diff
@@ -330,6 +330,20 @@ static void uv__process_child_init(const uv_process_options_t* options,
...
+     * remove any extraneous groups. If we don't call this, then
+     * even though our uid has dropped, we may still have groups
+     * that enable us to do super-user things. This will fail if we
+     * aren't root, so don't bother checking the return value, this
+     * is just done as an optimistic privilege dropping function.
+     */
+    int saved_errno;
+    saved_errno = errno;
+    setgroups(0, NULL);
+    errno = saved_errno;
```
bnoordhuis, line 344: Can be shortened to `SAVE_ERRNO(setgroups(0, NULL))` but apart from that LGTM.

bnoordhuis:
Change itself LGTM but as food for thought, you may want to consider adding an API that lets the user retain some or all of the ancillary groups.  Ditto for capabilities.

src/unix/process.c
saghul, line 344: Thanks Ben, will fix.

saghul:
Thanks guys! Landed as https://github.com/joyent/libuv/commit/66ab38918c911bcff025562cf06237d7fedaba0c. I also created https://github.com/joyent/libuv/issues/1106 to track the Linux capabilites part and the API suggesting by Ben.

Issue #711, opened 2013-02-15
Failing tests on Windows8.


txdv:
I have some tests failing on windows:

```
=============================================================
[%   8|+  15|-   0]: pipe_connect_to_file
`pipe_connect_to_file` failed: exit code 3
Output from process `pipe_connect_to_file`:
Assertion failed in test\test-pipe-connect-error.c on line 56: uv_last_error(uv_
default_loop()).code == UV_ENOTSOCK || uv_last_error(uv_default_loop()).code ==
UV_ECONNREFUSED
=============================================================
[%  76|+ 139|-   1]: fs_file_loop
`fs_file_loop` failed: exit code 3
Output from process `fs_file_loop`:
Assertion failed in test\test-fs.c on line 529: uv_last_error(loop).code == UV_E
LOOP
=============================================================
[%  86|+ 156|-   2]: fs_event_watch_file_twice
`fs_event_watch_file_twice` failed: exit code 3
Output from process `fs_event_watch_file_twice`:
Assertion failed in test\test-fs-event.c on line 255: 0 == uv_fs_event_init(loop
, watchers + 0, path, fail_cb, 0)
=============================================================
[%  91|+ 163|-   3]: fs_readdir_file
`fs_readdir_file` failed: exit code 3
Output from process `fs_readdir_file`:
Assertion failed in test\test-fs.c on line 1662: uv_last_error(loop).code == UV_
ENOTDIR
=============================================================
[%  94|+ 168|-   4]: threadpool_cancel_getaddrinfo
`threadpool_cancel_getaddrinfo` failed: exit code 3
Output from process `threadpool_cancel_getaddrinfo`:
Assertion failed in test\test-threadpool-cancel.c on line 90: 0 == uv_cancel((uv
_req_t*) req)
=============================================================
[%  95|+ 168|-   5]: threadpool_cancel_work
`threadpool_cancel_work` failed: exit code 3
Output from process `threadpool_cancel_work`:
Assertion failed in test\test-threadpool-cancel.c on line 90: 0 == uv_cancel((uv
_req_t*) req)
=============================================================
[%  95|+ 168|-   6]: threadpool_cancel_fs
`threadpool_cancel_fs` failed: exit code 3
Output from process `threadpool_cancel_fs`:
Assertion failed in test\test-threadpool-cancel.c on line 90: 0 == uv_cancel((uv
_req_t*) req)
=============================================================
[%  96|+ 168|-   7]: threadpool_cancel_single
[...]

PR #1107, opened 2014-02-11
build, windows: do not fail on Windows SDK Prompt

schlamar:


schlamar:
Hello bot, I have already signed the CLA :-)

vcbuild.bat
```diff
@@ -111,8 +111,9 @@ echo Project files generated.
 if defined nobuild goto run
 
 @rem Check if VS build env is available
-if not defined VCINSTALLDIR goto msbuild-not-found
-goto msbuild-found
+if defined VCINSTALLDIR goto msbuild-found
+if defined WindowsSDKDir goto msbuild-found
+goto msbuild-not-found
```
saghul: This label seems useless now since we only jump from the line above, right?

saghul:
I don't know what that "Windows SDK" is, does this LGTY @piscisaureus?

vcbuild.bat
schlamar: Ah right, will fix that.

schlamar:
> I don't know what that "Windows SDK" is, does this LGTY @piscisaureus?

This is a free tool from Microsoft which allows to compile Visual Studio projects on x64 (
http://www.microsoft.com/en-us/download/details.aspx?id=8279). So this is an alternative to VS Express as build tool because VS Express can only compile x86.

saghul:
Hey @schlamar, I installed the Windows SDK and I get the following error when running `vcbuild.bat test`:

```
  runner-win.c
LINK : fatal error LNK1123: failure during conversion to COFF: file invalid or
corrupt [c:\Users\saghul\src\libuv\run-tests.vcxproj]
```

I have Visual Studio Express 2008, 2013 and the Windows SDK 7.1 installed. Any ideas?

schlamar:
Are you in the Windows SDK prompt and are you targeting x86 debug (`setenv /x86 /debug`)?

Works for me with VS 2008, VS Express 2010 and Windows SDK 7.1 (except the test failures noted in #711).

schlamar:
Maybe you need to `set GYP_MSVS_VERSION=2010`. AFAIS Windows SDK 7.1 is targeting VS 2010 projects.

saghul:
I don't have VS 2010 :-/ I guess vcbuild.bat is detecting my 2013 version, should that work? Else, is there a way to for my 2008 express version to be forced?

schlamar:
> I don't have VS 2010

This should work anyway. I can create a VS 2013 project with `set GYP_MSVS_VERSION=2013` without having it installed. So with `GYP_MSVS_VERSION=2010` it should create a VS 2010 project which then could be compiled with the SDK (without having VS 2010 installed).

For 2008 you can either `set GYP_MSVS_VERSION=2008` and run it in the VS 2008 command prompt or prevent detection of 2013 by `set VS120COMNTOOLS=`.

saghul:
I tried setting it to use VS 2013 extress and this is what I get:

```
c:\Users\saghul\src\libuv>set GYP_MSVS_VERSION=2013e

c:\Users\saghul\src\libuv>vcbuild.bat test
['-Dtarget_arch=ia32', '-Duv_library=static_library', '.\\uv.gyp', '-I', '.\\com
mon.gypi', '--depth=.', '-Dhost_arch=x86', '-Dcomponent=static_library']
Project files generated.
Building the projects in this solution one at a time. To enable parallel build,
please add the "/m" switch.
c:\Users\saghul\src\libuv\libuv.vcxproj(18,3): error MSB4019: The imported proj
ect "C:\Program Files\MSBuild\Microsoft.Cpp\v4.0\V110\Microsoft.Cpp.Default.pro
ps" was not found. Confirm that the path in the <Import> declaration is correct
, and that the file exists on disk.
c:\Users\saghul\src\libuv\uv_dtrace_header.vcxproj(18,3): error MSB4019: The im
ported project "C:\Program Files\MSBuild\Microsoft.Cpp\v4.0\V110\Microsoft.Cpp.
Default.props" was not found. Confirm that the path in the <Import> declaration
 is correct, and that the file exists on disk.
c:\Users\saghul\src\libuv\uv_dtrace_provider.vcxproj(18,3): error MSB4019: The
imported project "C:\Program Files\MSBuild\Microsoft.Cpp\v4.0\V110\Microsoft.Cp
p.Default.props" was not found. Confirm that the path in the <Import> declarati
on is correct, and that the file exists on disk.

c:\Users\saghul\src\libuv>
```

Of course, I don't have that file. Should I install something else?

schlamar:
The error with 2013 should be not relevant to my patch. Might be an issue with gyp.

Looks like the linker error is a MS bug which can be resolved with a patch: http://www.microsoft.com/en-us/download/details.aspx?id=4422. You should be able to install the patch even if you have not installed VS 2010 (see http://www.chromium.org/developers/how-tos/build-instructions-windows#TOC-Setting-up-the-environment-for-building-with-Visual-C-2010-Express-or-Windows-7.1-SDK).

schlamar:
Rebased against master. 

BTW, I have seen the linker error on x64 test build, too. But it was successfully resolved with the patch from MS linked above.

saghul:
Thanks for the help! I'll try that later and let you know how it goes!

saghul:
Success! That patch made it! I'll land it in a bit.

saghul:
Thanks Marc, landed in https://github.com/joyent/libuv/commit/2f6d4b4b58ea290961dc983117d31b19e198403a

schlamar:
@saghul and this would be also required in v0.10 for proper pyuv builds.

saghul:
Ok, I'll backport them both. Thanks for letting us know!

saghul:
Backport landed in https://github.com/joyent/libuv/commit/cd6db8bbefb11a780977052f1a561920d6d60bbe

PR #1113, opened 2014-02-14
unix: Fail gracefully when calling uv_shutdown twice

Keno:
Before this, one of the requests was left as a ghost in the system. See https://github.com/JuliaLang/julia/issues/5793

### DIFF
--- a/test/test-shutdown-twice.c
+++ b/test/test-shutdown-twice.c
@@ -1,0 +1,85 @@
+/* Copyright Joyent, Inc. and other Node contributors. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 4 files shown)


indutny:
Wow, thanks for catching it! May I ask you to write a test for it? (and also fix all issues mentioned by @Nodejs-Jenkins )

Keno:
Updated.

test/test-shutdown-twice.c
```diff
@@ -0,0 +1,85 @@
...
+  ASSERT(0 == uv_ip4_addr("127.0.0.1", TEST_PORT, &addr));
+  loop = uv_default_loop();
+
+  r = uv_tcp_init(loop, &h);
+
+  r = uv_tcp_connect(&connect_req,
+                     &h,
+                     (const struct sockaddr*) &addr,
+                     connect_cb);
+  printf("%d\n", r);
```
saghul, line 75: Debug leftover?

saghul:
LGTM. If you don't mind please remove the printf, otherwise I can remove it when merging.

saghul:
Thanks @loladiro, landed in https://github.com/joyent/libuv/commit/a284b90c03735095763523a7c63f80140dc6a03a

saghul:
For the record, I made a backport for 0.10 and landed it in https://github.com/joyent/libuv/commit/6e2021ca111a44449932792e0cb578e63705c2a2

Issue #1111, opened 2014-02-14
export UV_VERSION macro 


sukinull:
If API changes so dramatically version by version, is it possible to move
- UV_VERSION_MAJOR 
- UV_VERSION_MINOR 
- UV_VERSION_PATCH 
- UV_VERSION_IS_RELEASE 
- UV_VERSION 

to uv.h?

saghul:
Nope, you should be using these: https://github.com/joyent/libuv/blob/master/include/uv.h#L231-L242

retep998:
Knowing the version at runtime doesn't help with compiling the program in the first place, where a macro would be much more helpful to conditionally compile the correct calls.

indutny:
@saghul indeed, we need to consider exporting them. Do you have any objections?

saghul:
Hum, I hadn't considered that. I would definitely not export UV_VERSION_IS_RELEASE.

indutny:
@saghul actually, it seems that this was intentionally removed from v0.11, because all patch versions have ABI compatibility.

tjfontaine:
Patch versions may have ABI compatibility, but minor have significant API/ABI differences, and thus it's difficult to know against whom you're building, so you can't write code that targets two stables easily.

In Node it's not a problem because we do keep the same minor, but we also export useful macros.

A well behaved C library should definitely include version macros.

indutny:
@saghul mind if I'll ask you help me with this? ;)

saghul:
No problem, I'll make a patch.

saghul:
Fixed in https://github.com/joyent/libuv/commit/d4c12adff648bc4ebd5038e8186069193b22a693

PR #1114, opened 2014-02-14
include: expose libuv version in header files

saghul:
Fix for #1111

/cc @indutny

Anyone feel free to provide feedback.

### DIFF
--- a/include/uv-version.h
+++ b/include/uv-version.h
@@ -19,31 +19,20 @@
  * IN THE SOFTWARE.
  */
 
-#include "uv.h"
+#ifndef UV_VERSION_H
+#define UV_VERSION_H
 
-#define UV_VERSION  ((UV_VERSION_MAJOR << 16) | \
-                     (UV_VERSION_MINOR <<  8) | \
-                     (UV_VERSION_PATCH))
+ /*
+ * Versions with an even minor version (e.g. 0.6.1 or 1.0.4) are API and ABI
+ * stable. When the minor version is odd, the API can change between patch
+ * releases. Make sure you update the -soname directives in configure.ac
+ * and uv.gyp whenever you bump UV_VERSION_MAJOR or UV_VERSION_MINOR (but
+ * not UV_VERSION_PATCH.)
+ */
 
-#define UV_STRINGIFY(v) UV_STRINGIFY_HELPER(v)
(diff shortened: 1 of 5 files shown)


sam-github:
Looks quite useful to me.

indutny:
LGTM, thank you!

include/uv-version.h
```diff
@@ -0,0 +1,39 @@
...
+ * not UV_VERSION_PATCH.)
+ */
+
+#define UV_VERSION_MAJOR 0
+#define UV_VERSION_MINOR 11
+#define UV_VERSION_PATCH 20
+#define UV_VERSION_IS_RELEASE 0
+
+#endif /* UV_VERSION_H */
+
```
bnoordhuis: Newline at end.

include/uv-version.h
```diff
@@ -0,0 +1,39 @@
...
+ * IN THE SOFTWARE.
+ */
+
+#ifndef UV_VERSION_H
+#define UV_VERSION_H
+
+ /*
+ * Versions with an even minor version (e.g. 0.6.1 or 1.0.4) are API and ABI
+ * stable. When the minor version is odd, the API can change between patch
+ * releases. Make sure you update the -soname directives in config-unix.mk
```
bnoordhuis: This comment is slightly outdated. :-)

bnoordhuis:
Is there a reason to put the defines in a separate header file?

src/version.c
```diff
@@ -19,21 +19,7 @@
  * IN THE SOFTWARE.
  */
 
-#include "uv.h"
```
bnoordhuis, line 22: You need to include uv.h here in order to pick up the right visibility for uv_version() and uv_version_string().  They won't be exported from the .so otherwise.

saghul, line 22: Ah, right. Thanks Ben!

saghul:
Addressed all feedback. About the separated file, @bnoordhuis: no specific reason, it felt right and I went for it, anything wrong with the approach?

Thanks!

saghul:
Thanks everyone, landed in https://github.com/saghul/libuv/commit/d4c12adff648bc4ebd5038e8186069193b22a693.

@piscisaureus, @tjfontaine you may need to update the release tool, since the version needs to be modified in a different file now.

PR #1117, opened 2014-02-16
windows: Always initialize the uv_process_t

alexcrichton:
The unix implementation of uv_spawn always starts out with a
uv__handle_init, but the windows implementation sometimes bails out early
before calling uv__handle_init. This means that uv_close on a failed
uv_spawn will always succeed on unix but sometimes fail on windows.

This commit lifts the initialization of the uv_process_t above all of the
error checking to ensure that uv_close will always work when uv_spawn
returns an error.


saghul:
Thanks! Landed in https://github.com/joyent/libuv/commit/6f62d62c905c2335232ec8e5697036426bd171d4.

PR #1118, opened 2014-02-16
windows: Fix building on MinGW

alexcrichton:
The crtdbg.h header was added in c0716b3d, but MinGW does not have this
header present on the system. This commit takes the same approach of
2684f876a and just ignores this header and functionality on MinGW


saghul:
Thanks @alexcrichton! Landed in https://github.com/joyent/libuv/commit/8692bbc254b89ef4ccd1a9bf7be9a1bc8bc6eb5f.

PR #1119, opened 2014-02-16
unix, windows: add uv_getrusage function

Sannis:
BTW, I have no idea about what is the best names for `uv_rusage` function and `getrusage.c` file.
Maybe the better name is `uv_getrusage_times`.


uv.gyp
```diff
@@ -139,6 +140,7 @@
             'src/unix/dl.c',
             'src/unix/fs.c',
             'src/unix/getaddrinfo.c',
+            'src/unit/getrusage.c',
```
saghul: Typo: unix.

include/uv.h
```diff
@@ -1646,6 +1646,8 @@ struct uv_interface_address_s {
 UV_EXTERN int uv_resident_set_memory(size_t* rss);
 UV_EXTERN int uv_uptime(double* uptime);
 
+UV_EXTERN int uv_rusage(double* user_time, double* system_time);
```
saghul: A comment here would be nice :-)

uv.gyp
Sannis: Fixed.

saghul:
Here are some comments about the general idea of uv_rusage after thinking a bit about it.

I think we should make this be `int uv_getrusage(uv_rusage_t *rusage)`, which would fill in a `uv_rusage_t` structure. Consider `uv_rusage_t` the cross-platform equivalent of `struct rusage`. I don't know if all values of it are available or make sense in all platforms, we should make sure that sane values are provided when we cannot them from the OS (in Windows, for example, if that is the case).

I'm not sure if we'd need/want the process specified or just use `RUSAGE_SELF`. I'd be fine that way, but I don't experience in using `getrusage` myself.

What's your take on this @indutny?

Sannis:
The only one reasone that I have to do it in different way is that I didn't found Windows equivalents for other `struct rusage` fields. But your proposal looks reasonable.

saghul:
@Sannis if there are no Windows equivalents just fill the structure with zeroes (or the appropriate value), we should try to expose as much of it as possible in platforms that do have the data available IMHO, similar to what we do with `uv_stat_t`.

Sannis:
@saghul ok, I'll do it.

saghul:
@Sannis sweet, thanks!

Sannis:
I've reused `struct uv_timespec_t` here instead of creating new struct that similar to `struct timeval`. Should I make it more similar to original Unix `struct rusage` or not?

saghul:
On 02/17/2014 06:40 PM, Oleg Efimov wrote:

> I've reused |struct uv_timespec_t| here instead of creating new struct
> that similar to |struct timeval|. Should I make it more similar to
> original Unix |struct rusage| or not?

I'd make it as similar as possible, so please define some uv_timeval_t.

Thanks for your work!

## 

Saúl Ibarra Corretgé
bettercallsaghul.com

Sannis:
I'll merge all these commit at the end.

include/uv.h
```diff
@@ -1646,6 +1646,37 @@ struct uv_interface_address_s {
...
+
+typedef struct {
+  long tv_sec;
+  long tv_usec;
+} uv_timeval_t;
+
+typedef struct {
+   uv_timeval_t ru_utime; /* user CPU time used */
+   uv_timeval_t ru_stime; /* system CPU time used */
+   /* Unix only: */
```
saghul: Remove this comment, but in uv.h mention that not all fields will be filled in Windows.

include/uv.h
```diff
@@ -1646,6 +1646,37 @@ struct uv_interface_address_s {
...
+   uint64_t ru_oublock;    /* block output operations */
+   uint64_t ru_msgsnd;     /* IPC messages sent */
+   uint64_t ru_msgrcv;     /* IPC messages received */
+   uint64_t ru_nsignals;   /* signals received */
+   uint64_t ru_nvcsw;      /* voluntary context switches */
+   uint64_t ru_nivcsw;     /* involuntary context switches */
+} uv_rusage_t;
+
+/*
+ * Get information about OS resource utilization for process
```
saghul: "for the current process"

src/unix/getrusage.c
```diff
@@ -0,0 +1,65 @@
+/* Copyright Joyent, Inc. and other Node contributors. All rights reserved.
```
saghul: This can be moved to src/unix/core.c, it doesn't really need a new file.

src/unix/getrusage.c
```diff
@@ -0,0 +1,65 @@
...
+int uv_getrusage(uv_rusage_t *uv_rusage) {
+  struct rusage usage;
+  int err;
+
+  err = getrusage(RUSAGE_SELF, &usage);
+  if (err < 0) {
+    return err;
+  }
+
+  uv__to_rusage(&usage, uv_rusage);
```
saghul: Since we only use it here there is no need for the extra function.

src/win/getrusage.c
```diff
@@ -0,0 +1,60 @@
+/* Copyright Joyent, Inc. and other Node contributors. All rights reserved.
```
saghul: You can put this in src/win/util.c, that's where similar functions are defined in Windows.

saghul:
I left some comments. Please leave a comment when you add new commits, GitHub doesn't send notifications for those... :-/

Sannis:
Rewritten acording to your comment.

> GitHub doesn't send notifications for those...

It also doesn't show your comments in onsite notifications :-(

saghul:
Looks great @Sannis! Lets see what Jenkins says :-)

Sannis:
Sure) I checked this on OS X and openSUSE, will try to build on Windows tomorrow.

test/test-platform-output.c
```diff
@@ -47,6 +48,20 @@
   ASSERT(uptime > 0);
   printf("uv_uptime: %f\n", uptime);
 
+  err = uv_getrusage(&rusage);
+  ASSERT(err == 0);
+  ASSERT(rusage.ru_utime.tv_sec >= 0);
+  ASSERT(rusage.ru_utime.tv_usec > 0);
```
saghul: Looks like this assert if blowing up in the tests. Maybe relax it with >= 0?

test/test-platform-output.c
```diff
@@ -47,6 +48,20 @@
...
+  ASSERT(rusage.ru_stime.tv_sec >= 0);
+  ASSERT(rusage.ru_stime.tv_usec > 0);
```
saghul: Ditto.

Sannis:
@saghul you are right, it is often system_time == 0 and it it ok. Fixed.

Sannis:
Compiles and works well on Windows 7 with VS 2012.

saghul:
Thanks @Sannis, LGTM. @indutny, can you take a quick look at this? Implementation looks sound to me.

indutny:
LGTM

saghul:
Thanks @Sannis! Landed in https://github.com/joyent/libuv/commit/da9a2b1d3d86eeeeaea6380620c7ab315e8c142d. I made a slight change to return the proper value on error, which slipped during review.

Sannis:
You are right. Thanks for merging.

PR #1123, opened 2014-02-17
unix: use a heap for timers

saghul:
> Replace the red-black tree with a heap.  The most common operation that
> libuv performs on timers is looking up the first timer to expire.  With
> a red-black tree, that operation is O(log n).  With a heap, it's O(1).

This work was done by @bnoordhuis a while ago. I talked with him and the code is ready to be merged. While there might be room for more optimizations when using millions of timers, the common use cases are already better with this.

This patch is a squash and rebase of https://github.com/joyent/libuv/pull/928

@bnoordhuis: does heap-inl.h need to be listed in Makefile.am and uv.gyp? It didn't complain here, but we seem to list all internal include files.

### DIFF
--- a/src/heap-inl.h
+++ b/src/heap-inl.h
@@ -1,0 +1,238 @@
+/* Copyright (c) 2013, Ben Noordhuis <info@bnoordhuis.nl>
+ *
+ * Permission to use, copy, modify, and/or distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#ifndef UV_SRC_HEAP_H_
+#define UV_SRC_HEAP_H_
+
+#include <stddef.h>  /* NULL */
+
+#if defined(__GNUC__)
+# define HEAP_EXPORT(declaration) __attribute__((unused)) static declaration
+#else
+# define HEAP_EXPORT(declaration) static declaration
+#endif
+
+struct heap_node {
+  struct heap_node* left;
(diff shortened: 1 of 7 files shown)


src/unix/timer.c
```diff
@@ -20,35 +20,41 @@
...
-    return -1;
-  if (a->timeout > b->timeout)
     return 1;
-  /*
-   *  compare start_id when both has the same timeout. start_id is
-   *  allocated with loop->timer_counter in uv_timer_start().
+  if (b->timeout < a->timeout)
+    return 0;
+
+  /* Ccompare start_id when both has the same timeout. start_id is
```
bnoordhuis: Typo. :-)

bnoordhuis:
> does heap-inl.h need to be listed in Makefile.am and uv.gyp?

Yes, it would be good to have it listed.

saghul:
Fixed the typo and added the file to Makefile.am and uv.gyp.

retep998:
Speak of the devil. I was just benchmarking running millions of timers, firing every second.
Using the latest official release, I can get to about 1.6 million timers before it can't keep up.
Using the version in this pull request lets me get to about 2.5 million before it slows down.

saghul:
@retep998 sweet, thanks for sharing the results!

src/heap-inl.h
```diff
@@ -0,0 +1,238 @@
...
+ * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#ifndef UV_SRC_HEAP_H_
+#define UV_SRC_HEAP_H_
+
+#include <stddef.h>  /* NULL */
+
+#if defined(__GNUC__)
+# define HEAP_EXPORT(declaration) __attribute__((unused)) static declaration
```
indutny, line 22: Is there any point for `declaration` arg to be here, except style? I'd rather not use it.

saghul, line 22: How would you put it?

```
# define HEAP_EXPORT __attribute__((unused)) static
```

indutny, line 22: Yep, I think we are doing `UV_EXTERN` the same way.

src/heap-inl.h
```diff
@@ -0,0 +1,238 @@
...
+  unsigned int k;
+
+  newnode->left = NULL;
+  newnode->right = NULL;
+  newnode->parent = NULL;
+
+  /* Calculate the path from the root to the insertion point.  This is a min
+   * heap so we always insert at the left-most free node of the bottom row.
+   */
+  path = 0;
```
indutny, line 122: Is this extra step really required?

src/heap-inl.h
```diff
@@ -0,0 +1,238 @@
...
+    if (path & 1)
+      child = &(*child)->right;
+    else
+      child = &(*child)->left;
+    path >>= 1;
+    k -= 1;
+  }
+
+  /* Insert the new node. */
+  newnode->parent = *parent;
```
indutny, line 139: Am I right that root will be it's own parent?

indutny, line 139: nvm, I was wrong.

src/heap-inl.h
indutny, line 122: Ah, I see...

src/heap-inl.h
saghul, line 22: I don't mind changing it, but I trust the original author's criteria ;-)

indutny:
LGTM

saghul:
Landed, thanks!

Issue #435, opened 2012-05-30
windows: mark internal handles as internal


bnoordhuis:
So `uv_walk()` filters them out. Here is an example: 71a66c1

saghul:
I looked over the Windows side of things and didn't spot any internal handle defined there, closing.

PR #1124, opened 2014-02-17
unix, windows: add uv_loop_init and uv_loop_close

saghul:
These functions supersede uv_loop_new and uv_loop_delete.
uv_loop_init initialized a user allocated loop and uv_loop_close
removes all associated resources a loop uses after it has finished
execution.

uv_loop_new and uv_loop_delete are now deprecated.

@indutny, @piscisaureus: please review :-)


include/uv.h
```diff
@@ -261,6 +257,32 @@
...
+ * Initializes a uv_loop_t structure.
+ */
+UV_EXTERN int uv_loop_init(uv_loop_t* loop);
+
+/*
+ * Closes all internal loop resources.  This function must only be called once
+ * the loop has finished it's execution or it will return UV_EBUSY.  After this
+ * function returns the user shall free the memory allocated for the loop.
+ */
+UV_EXTERN uv_loop_close(uv_loop_t* loop);
```
retep998: Shouldn't this have a return type?

saghul: Ups, why didn't you tell me Mr Compiler?!

indutny: Because this is C :) That's how people was writing stuff before return types.

saghul:
Note to self: look into https://github.com/joyent/libuv/issues/435, I might need to solve that one otherwise uv_loop_close will always fail on Windows.

src/unix/loop.c
```diff
@@ -45,6 +45,31 @@ uv_loop_t* uv_default_loop(void) {
...
+  if (!QUEUE_EMPTY(&(loop)->active_reqs))
+    return -EBUSY;
+  QUEUE_FOREACH(q, &loop->handle_queue) {
+    h = QUEUE_DATA(q, uv_handle_t, handle_queue);
+    if (!(h->flags & UV__HANDLE_INTERNAL))
+      return -EBUSY;
+  }
+  uv__loop_close(loop);
+#ifndef NDEBUG
+  memset(loop, -1, sizeof(*loop));
```
indutny, line 66: Huh? `-1` ?

indutny, line 66: Ah, that's how it was.

saghul:
Force pushed, fixing some nasty problem in Windows: initializing libuv is kind of important :-) Tests pass on all platforms now (on Windows there are some known failures, this PR doesn't introduce new failures).

Reviewers, please review!

indutny:
If it builds and runs on win and unixes, and if tests are passing - LGTM.

saghul:
Thanks @indutny, landed it!

yann2192:
@saghul uv_loop_close always returns -EBUSY when I use signals : http://pastebin.com/ZqA2hhZG

indutny:
@yann2192 perhaps you should call `uv_close` on `uv_signal_t` first?

saghul:
@yann2192 what @indutny said. Stopping is not enough. For the loop to properly close all handles need to be uv_close'd first.

saghul:
This works: https://gist.github.com/9504483

yann2192:
Oh, I didn't know that uv_signal_t requires uv_close, my bad.

saghul:
On 03/12/2014 11:41 AM, Yann GUIBET wrote:

> Oh, I didn't know that uv_signal_t requires uv_close, my bad.

No worries. All handles do, actually.

## 

Saúl Ibarra Corretgé
bettercallsaghul.com

PR #1125, opened 2014-02-18
build: update dtrace m4 check for sunos

tjfontaine:
The check to see if -G is needed for dtrace requires a bit more effort,
make a header, c file, and object file.

### DIFF
--- a/m4/dtrace.m4
+++ b/m4/dtrace.m4
@@ -39,10 +39,18 @@ provider Example {
   probe increment(int);
 };
 _ACEOF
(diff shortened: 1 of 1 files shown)


saghul:
I don't have a SunOS to test on, so if it works, LGTM.

indutny:
@tjfontaine why -G could not be needed?

tjfontaine:
@indutny -G is needed on platforms that post-process object files, compared to OSX's implementation where the ASM logic is in the generated header

indutny:
Ok, LGTM

tjfontaine:
Thanks landed in 3310ff4

PR #1127, opened 2014-02-18
tcp: freeze in uv_tcp_endgame

orangemocha:
The event_handle field of unused accept requests was not being
initialized properly. As a result, uv_tcp_endgame would try to close
an invalid handle and this could lead to unexpected behavior.

This fixes node.js test test-cluster-disconnect.js on Windows.


piscisaureus:
Thanks! Landed in 4f913b669adc1765829ce57b8a022394e363a757.

PR #1132, opened 2014-02-20
Fix setgroups() compile warning on Linux

wrl:
On Linux, `setgroups()` is in `grp.h`.

### DIFF
--- a/src/unix/process.c
+++ b/src/unix/process.c
@@ -40,6 +40,10 @@
 extern char **environ;
 #endif
 
+#ifdef __linux__
+# include <grp.h>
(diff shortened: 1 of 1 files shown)


saghul:
Thanks! I wonder how never noticed this :-/ Have you signed the CLA?

wrl:
for some reason, the warning didn't show up in the autoconf build, only the gyp one. weird!

and yeah, i've signed the CLA. should be under the same email my commit is from, let me know if i don't show up.

saghul:
Thanks @wrl!

PR #1133, opened 2014-02-21
test, build: fix Android NDK r9c compatibility

ghost:
r9c removed the toolchain currently used, and e403a2c4 introduced a test that fails when built against the NDK.

More info on both:
http://www.gnu.org/savannah-checkouts/gnu/libc/manual/html_node/Permission-Bits.html
http://developer.android.com/tools/sdk/ndk/index.html#Revisions


ghost:
I'm not sure about these failing tests, they don't seem related to my changes.

ghost:
I addressed Jenkins' comments

saghul:
LGTM, have you signed the CLA?

saghul:
Thanks @lgierth! Landed in https://github.com/joyent/libuv/commit/1ad61807ffa893f610f1cc87007e332582617dbf and https://github.com/joyent/libuv/commit/cd67e10eceba9c1e13fcd023e4c804714ec2b211.

ghost:
> and e403a2c introduced a test that fails when built against the NDK.

That should actually say "fails to build against the NDK". A failing test is ok, but the latest v0.11.x doesn't even build against the NDK.

PR #1134, opened 2014-02-21
fs: vectored IO API for filesystem read/write

Ralith:
#1103 with a windows error fixed

### DIFF
--- a/src/win/fs.c
+++ b/src/win/fs.c
@@ -539,13 +539,14 @@ void fs__close(uv_fs_t* req) {
 
 void fs__read(uv_fs_t* req) {
   int fd = req->fd;
-  size_t length = req->length;
   int64_t offset = req->offset;
   HANDLE handle;
   OVERLAPPED overlapped, *overlapped_ptr;
   LARGE_INTEGER offset_;
   DWORD bytes;
   DWORD error;
+  int result;
+  unsigned int index;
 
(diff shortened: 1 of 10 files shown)


saghul:
Jenkinds doesn't seem to have the Windows slave available, I'll run the test suite later today. Thanks!

src/win/fs.c
```diff
@@ -1588,13 +1632,27 @@ int uv_fs_read(uv_loop_t* loop, uv_fs_t* req, uv_file fd, void* buf,
 }
 
 
-int uv_fs_write(uv_loop_t* loop, uv_fs_t* req, uv_file fd, const void* buf,
-    size_t length, int64_t offset, uv_fs_cb cb) {
+int uv_fs_write(uv_loop_t* loop,
+                uv_fs_t* req,
+                uv_file file,
```
saghul: uv_file fd.

saghul:
Fails to compile on windows because there are 4 places where req->length is still used. Lines 542, 606, 1120 and 1983.

Ralith:
Applied style tweak and removed or replaced references to req->length as appropriate.

saghul:
Compiles and test suite passes on Windows now, thanks @Ralith! Landed in https://github.com/joyent/libuv/commit/13dd3502a7aedf4284559a88b3662e222723b7ce

PR #1135, opened 2014-02-21
Revert "stream: add an API to make streams do blocking writes"

saghul:
This reverts commit 92040eb7123e892261c8da71c3eedb97c00e4b88.

I haven't even attempted to compile this, I will do so later today. Just wanted to gather comments from you guys.

AFAICT this is not used anywhere, doesn't have a test and doesn't work as advertised.

Of course, I could be totally wrong here.

/cc @piscisaureus @indutny

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -733,31 +733,6 @@ UV_EXTERN int uv_is_writable(const uv_stream_t* handle);
 
 
 /*
- * Enable or disable blocking mode for a stream.
- *
- * When blocking mode is enabled all writes complete synchronously. The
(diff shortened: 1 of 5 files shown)


indutny:
+1

saghul:
I think I found where this comes from: https://github.com/joyent/node/issues/3584

saghul:
Closing, looks like we might actually need this and node 0.12 plans to use it. PR for clarifying its usage + some error checking here: https://github.com/joyent/libuv/pull/1152

PR #1137, opened 2014-02-21
win: minor error handle fix to uv_pipe_write_impl

zerhacken:
This commit ensures a proper win32 error code is returned
from uv_pipe_write_impl in case of failure.


indutny:
LGTM, thank you!

indutny:
Landed in 0943d3c

PR #1141, opened 2014-02-23
windows: fix warnings of MinGW -Wall -O3

starwing:
sweep MinGW warnings.

still two warnings remains:

```
src/win/fs.c: In function 'fs__utime_handle':
src/win/fs.c:1187:3: warning: dereferencing type-punned pointer will break stric
t-aliasing rules [-Wstrict-aliasing]
   TIME_T_TO_FILETIME((time_t) atime, &filetime_a);
   ^
src/win/fs.c:1188:3: warning: dereferencing type-punned pointer will break stric
t-aliasing rules [-Wstrict-aliasing]
   TIME_T_TO_FILETIME((time_t) mtime, &filetime_m);
   ^
```

warnings fixed:
- unsigned/signed unmatched, use int i to iteration array with ARRAY_SIZE(...) -- which is unsigned
- uv__crt_invalid_parameter_handler not used on MinGW
- several uninitialized warning for several places. but some are caused by assert(0); and abort(); (but MinGW doesn't know it will abort)
- some Windows API returns int, not DWORD.
- InterlockedExchange's first argument is volatile LONG_, not unsigned long_.

### DIFF
--- a/src/win/pipe.c
+++ b/src/win/pipe.c
@@ -1077,7 +1077,7 @@ static int uv_pipe_write_impl(uv_loop_t* loop,
   int err;
   int result;
   uv_tcp_t* tcp_send_handle;
-  uv_write_t* ipc_header_req;
+  uv_write_t* ipc_header_req = NULL;
   uv_ipc_frame_uv_stream ipc_frame;
 
   if (nbufs != 1 && (nbufs != 0 || !send_handle)) {
@@ -1256,6 +1256,7 @@ static int uv_pipe_write_impl(uv_loop_t* loop,
       /* Request completed immediately. */
       req->queued_bytes = 0;
     } else {
+      assert(ipc_header_req != NULL);
       /* Request queued by the kernel. */
       if (WaitForSingleObject(ipc_header_req->overlapped.hEvent, INFINITE) !=
           WAIT_OBJECT_0) {
--- a/src/win/poll.c
+++ b/src/win/poll.c
@@ -85,6 +85,7 @@ static void uv__fast_poll_submit_poll_req(uv_loop_t* loop, uv_poll_t* handle) {
     handle->mask_events_2 = 0;
   } else {
     assert(0);
+    return;
   }
 
   /* Setting Exclusive to TRUE makes the other poll request return if there */
@@ -119,7 +120,7 @@ static void uv__fast_poll_submit_poll_req(uv_loop_t* loop, uv_poll_t* handle) {
 
 static int uv__fast_poll_cancel_poll_req(uv_loop_t* loop, uv_poll_t* handle) {
   AFD_POLL_INFO afd_poll_info;
-  DWORD result;
+  int result;
 
   afd_poll_info.Exclusive = TRUE;
   afd_poll_info.NumberOfHandles = 1;
@@ -158,6 +159,7 @@ static void uv__fast_poll_process_poll_req(uv_loop_t* loop, uv_poll_t* handle,
     mask_events = handle->mask_events_2;
   } else {
     assert(0);
+    return;
   }
 
   /* Report an error unless the select was just interrupted. */
@@ -281,7 +283,7 @@ static SOCKET uv__fast_poll_get_peer_socket(uv_loop_t* loop,
   SOCKET peer_socket;
 
   index = -1;
-  for (i = 0; i < ARRAY_SIZE(uv_msafd_provider_ids); i++) {
+  for (i = 0; (size_t) i < ARRAY_SIZE(uv_msafd_provider_ids); i++) {
     if (memcmp((void*) &protocol_info->ProviderId,
                (void*) &uv_msafd_provider_ids[i],
                sizeof protocol_info->ProviderId) == 0) {
@@ -392,6 +394,7 @@ static void uv__slow_poll_submit_poll_req(uv_loop_t* loop, uv_poll_t* handle) {
     handle->mask_events_2 = 0;
   } else {
     assert(0);
+    return;
   }
 
   if (!QueueUserWorkItem(uv__slow_poll_thread_proc,
@@ -418,6 +421,7 @@ static void uv__slow_poll_process_poll_req(uv_loop_t* loop, uv_poll_t* handle,
     mask_events = handle->mask_events_2;
   } else {
     assert(0);
+    return;
   }
 
   if (!REQ_SUCCESS(req)) {
--- a/src/win/process-stdio.c
+++ b/src/win/process-stdio.c
@@ -327,7 +327,7 @@ int uv__stdio_create(uv_loop_t* loop,
         /* an uv_pipe_t for use by the parent. The other one is given to */
         /* the child. */
         uv_pipe_t* parent_pipe = (uv_pipe_t*) fdopt.data.stream;
-        HANDLE child_pipe;
+        HANDLE child_pipe = INVALID_HANDLE_VALUE;
 
         /* Create a new, connected pipe pair. stdio[i].stream should point */
         /* to an uninitialized, but not connected pipe handle. */
@@ -389,6 +389,7 @@ int uv__stdio_create(uv_loop_t* loop,
 
           default:
             assert(0);
+            return -1;
         }
 
         CHILD_STDIO_HANDLE(buffer, i) = child_handle;
@@ -423,11 +424,9 @@ int uv__stdio_create(uv_loop_t* loop,
         }
 
         /* Make an inheritable copy of the handle. */
-        if (uv__duplicate_handle(loop,
-                             stream_handle,
-                             &child_handle) < 0) {
+        err = uv__duplicate_handle(loop, stream_handle, &child_handle);
+        if (err)
           goto error;
-        }
 
         CHILD_STDIO_HANDLE(buffer, i) = child_handle;
         CHILD_STDIO_CRT_FLAGS(buffer, i) = crt_flags;
@@ -436,6 +435,7 @@ int uv__stdio_create(uv_loop_t* loop,
 
       default:
         assert(0);
+        return -1;
     }
   }
 
--- a/src/win/signal.c
+++ b/src/win/signal.c
@@ -78,7 +78,8 @@ int uv__signal_dispatch(int signum) {
   for (handle = RB_NFIND(uv_signal_tree_s, &uv__signal_tree, &lookup);
        handle != NULL && handle->signum == signum;
        handle = RB_NEXT(uv_signal_tree_s, &uv__signal_tree, handle)) {
-    unsigned long previous = InterlockedExchange(&handle->pending_signum, signum);
+    unsigned long previous = InterlockedExchange(
+            (volatile LONG*) &handle->pending_signum, signum);
 
     if (!previous) {
       POST_COMPLETION_FOR_REQ(handle->loop, &handle->signal_req);
@@ -307,12 +308,13 @@ int uv_signal_start(uv_signal_t* handle, uv_signal_cb signal_cb, int signum) {
 
 void uv_process_signal_req(uv_loop_t* loop, uv_signal_t* handle,
     uv_req_t* req) {
-  unsigned long dispatched_signum;
+  long dispatched_signum;
 
   assert(handle->type == UV_SIGNAL);
   assert(req->type == UV_SIGNAL_REQ);
 
-  dispatched_signum = InterlockedExchange(&handle->pending_signum, 0);
+  dispatched_signum = InterlockedExchange(
+          (volatile LONG*) &handle->pending_signum, 0);
   assert(dispatched_signum != 0);
 
   /* Check if the pending signal equals the signum that we are watching for. */
--- a/src/win/tty.c
+++ b/src/win/tty.c
@@ -698,7 +698,7 @@ void uv_process_tty_read_raw_req(uv_loop_t* loop, uv_tty_t* handle,
         buf.base[buf_used++] = handle->last_key[handle->last_key_offset++];
 
         /* If the buffer is full, emit it */
-        if (buf_used == buf.len) {
+        if ((size_t) buf_used == buf.len) {
           handle->read_cb((uv_stream_t*) handle, buf_used, &buf);
           buf = uv_null_buf_;
           buf_used = 0;
--- a/src/win/core.c
+++ b/src/win/core.c
@@ -70,11 +70,13 @@ static int uv__crt_dbg_report_handler(int report_type, char *message, int *ret_v
 #endif
 
 
+#if !defined(__MINGW32__) || __MSVCRT_VERSION__ >= 0x800
 static void uv__crt_invalid_parameter_handler(const wchar_t* expression,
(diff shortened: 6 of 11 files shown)


src/win/pipe.c
```diff
@@ -1040,7 +1040,7 @@ static int uv_pipe_write_impl(uv_loop_t* loop,
   int err;
   int result;
   uv_tcp_t* tcp_send_handle;
-  uv_write_t* ipc_header_req;
+  uv_write_t* ipc_header_req = NULL; /* fix MinGW uninitialized warning */
```
indutny: Please use two spaces betwen colon and comment. `NULL;  /*`

src/win/process-stdio.c
```diff
@@ -436,6 +437,7 @@ int uv__stdio_create(uv_loop_t* loop,
 
       default:
         assert(0);
+        return -1; /* avoid err not initialzed warning from MinGW */
```
indutny: Ditto.

src/win/signal.c
```diff
@@ -78,7 +78,7 @@ int uv__signal_dispatch(int signum) {
   for (handle = RB_NFIND(uv_signal_tree_s, &uv__signal_tree, &lookup);
        handle != NULL && handle->signum == signum;
        handle = RB_NEXT(uv_signal_tree_s, &uv__signal_tree, handle)) {
-    unsigned long previous = InterlockedExchange(&handle->pending_signum, signum);
+    unsigned long previous = InterlockedExchange((volatile LONG*)&handle->pending_signum, signum);
```
indutny: Does it fit into 80 column limit?

src/win/signal.c
```diff
@@ -307,12 +307,12 @@ int uv_signal_start(uv_signal_t* handle, uv_signal_cb signal_cb, int signum) {
...
 void uv_process_signal_req(uv_loop_t* loop, uv_signal_t* handle,
     uv_req_t* req) {
-  unsigned long dispatched_signum;
+  long dispatched_signum;
 
   assert(handle->type == UV_SIGNAL);
   assert(req->type == UV_SIGNAL_REQ);
 
-  dispatched_signum = InterlockedExchange(&handle->pending_signum, 0);
+  dispatched_signum = InterlockedExchange((volatile LONG*)&handle->pending_signum, 0);
```
indutny: Space between cast and expression `(volatile LONG*) &handle`

src/win/tty.c
```diff
@@ -698,7 +698,7 @@ void uv_process_tty_read_raw_req(uv_loop_t* loop, uv_tty_t* handle,
         buf.base[buf_used++] = handle->last_key[handle->last_key_offset++];
 
         /* If the buffer is full, emit it */
-        if (buf_used == buf.len) {
+        if ((size_t)buf_used == buf.len) {
```
indutny: ditto.

src/win/process-stdio.c
```diff
@@ -423,9 +424,9 @@ int uv__stdio_create(uv_loop_t* loop,
         }
 
         /* Make an inheritable copy of the handle. */
-        if (uv__duplicate_handle(loop,
+        if ((err = uv__duplicate_handle(loop,
```
indutny: Mind splitting into two statements: one assignment and one `if`?

src/win/poll.c
```diff
@@ -281,7 +281,7 @@ static SOCKET uv__fast_poll_get_peer_socket(uv_loop_t* loop,
   SOCKET peer_socket;
 
   index = -1;
-  for (i = 0; i < ARRAY_SIZE(uv_msafd_provider_ids); i++) {
+  for (i = 0; (size_t)i < ARRAY_SIZE(uv_msafd_provider_ids); i++) {
```
indutny: `(size_t) i`

indutny:
Generally, LGTM, but has some style issues. @saghul mind reviewing?

starwing:
@indutny rebased and force pushed, reviewing again? thank you very much!

src/win/poll.c
```diff
@@ -145,8 +145,8 @@ static int uv__fast_poll_cancel_poll_req(uv_loop_t* loop, uv_poll_t* handle) {
 
 static void uv__fast_poll_process_poll_req(uv_loop_t* loop, uv_poll_t* handle,
     uv_req_t* req) {
-  unsigned char mask_events;
-  AFD_POLL_INFO* afd_poll_info;
+  unsigned char mask_events = 0;
+  AFD_POLL_INFO* afd_poll_info = NULL;
```
indutny: No point in initializing it here, it is set below in both cases.

src/win/poll.c
```diff
@@ -145,8 +145,8 @@ static int uv__fast_poll_cancel_poll_req(uv_loop_t* loop, uv_poll_t* handle) {
 
 static void uv__fast_poll_process_poll_req(uv_loop_t* loop, uv_poll_t* handle,
     uv_req_t* req) {
-  unsigned char mask_events;
-  AFD_POLL_INFO* afd_poll_info;
+  unsigned char mask_events = 0;
```
indutny: Same thing.

src/win/pipe.c
```diff
@@ -1040,7 +1040,7 @@ static int uv_pipe_write_impl(uv_loop_t* loop,
   int err;
   int result;
   uv_tcp_t* tcp_send_handle;
-  uv_write_t* ipc_header_req;
+  uv_write_t* ipc_header_req = NULL;  /* fix MinGW uninitialized warning */
```
indutny: Could you please initialize it after all declarations? I think C89 prohibits initializing assignments.

src/win/poll.c
```diff
@@ -407,7 +407,7 @@ static void uv__slow_poll_submit_poll_req(uv_loop_t* loop, uv_poll_t* handle) {
 
 static void uv__slow_poll_process_poll_req(uv_loop_t* loop, uv_poll_t* handle,
     uv_req_t* req) {
-  unsigned char mask_events;
+  unsigned char mask_events = 0;
```
indutny: No need in initialization.

src/win/process-stdio.c
```diff
@@ -327,7 +327,7 @@ int uv__stdio_create(uv_loop_t* loop,
         /* an uv_pipe_t for use by the parent. The other one is given to */
         /* the child. */
         uv_pipe_t* parent_pipe = (uv_pipe_t*) fdopt.data.stream;
-        HANDLE child_pipe;
+        HANDLE child_pipe = INVALID_HANDLE_VALUE;
```
indutny, line 330: No need in initialization.

src/win/process-stdio.c
```diff
@@ -266,7 +266,7 @@ int uv__stdio_create(uv_loop_t* loop,
                      BYTE** buffer_ptr) {
   BYTE* buffer;
   int count, i;
-  int err;
+  int err = 0;
```
indutny: Are you sure this is needed? Most likely - no.

src/win/process-stdio.c
```diff
@@ -423,11 +423,10 @@ int uv__stdio_create(uv_loop_t* loop,
         }
 
         /* Make an inheritable copy of the handle. */
-        if (uv__duplicate_handle(loop,
-                             stream_handle,
-                             &child_handle) < 0) {
+        err = uv__duplicate_handle(loop, stream_handle, &child_handle);
+
```
indutny: Superfluos whitespace.

indutny:
Looks cool, though, I dislike initializing all stuff. Those `assert(0)`s are already a pretty fatal error case, perhaps they should be made an `abort()` s intead.

src/win/poll.c
starwing: it doesn't initialize on assert(0);/abort(); branch, which we know will abort execute, but gcc doesn't know :(

src/win/pipe.c
starwing: C89 can initialize assignment, but must be const expression, and NULL is really a const expression.

src/win/poll.c
starwing: maybe we should add a return; after assert(0); or abort();? or use a gcc no_return attribute (which will add a new function)

starwing:
okay, I will try using abort(); maybe this can avoid warnings.

starwing:
add __declspec(noreturn) and use abort(), clear all warnings without -O3, but still some warnings at -O3 level, how to solve these warnings?

```
gcc -O3 -Wall -Wextra -Wno-unused-parameter -Iinclude -Isrc -Isrc/win -DWIN32_LE
AN_AND_MEAN -D_WIN32_WINNT=0x0600 -c -o src/win/fs.o src/win/fs.c
src/win/fs.c: In function 'fs__utime_handle':
src/win/fs.c:1187:3: warning: dereferencing type-punned pointer will break stric
t-aliasing rules [-Wstrict-aliasing]
   TIME_T_TO_FILETIME((time_t) atime, &filetime_a);
   ^
src/win/fs.c:1188:3: warning: dereferencing type-punned pointer will break stric
t-aliasing rules [-Wstrict-aliasing]
   TIME_T_TO_FILETIME((time_t) mtime, &filetime_m);
   ^
gcc -O3 -Wall -Wextra -Wno-unused-parameter -Iinclude -Isrc -Isrc/win -DWIN32_LE
AN_AND_MEAN -D_WIN32_WINNT=0x0600 -c -o src/win/pipe.o src/win/pipe.c
src/win/pipe.c: In function 'uv_pipe_write_impl':
src/win/pipe.c:1226:20: warning: 'ipc_header_req' may be used uninitialized in t
his function [-Wmaybe-uninitialized]
         CloseHandle(ipc_header_req->overlapped.hEvent);
                    ^
gcc -O3 -Wall -Wextra -Wno-unused-parameter -Iinclude -Isrc -Isrc/win -DWIN32_LE
AN_AND_MEAN -D_WIN32_WINNT=0x0600 -c -o src/win/process-stdio.o src/win/process-
stdio.c
src/win/process-stdio.c: In function 'uv__stdio_create':
src/win/process-stdio.c:345:39: warning: 'child_pipe' may be used uninitialized
in this function [-Wmaybe-uninitialized]
         CHILD_STDIO_HANDLE(buffer, i) = child_pipe;
                                       ^
```

starwing:
There are still many assert(0) in codes, shall I change all of them to abort() for consistency?

starwing:
```
src/win/pipe.c:1226:20: warning: 'ipc_header_req' may be used uninitialized in t
his function [-Wmaybe-uninitialized]
         CloseHandle(ipc_header_req->overlapped.hEvent);
                    ^
```

this maybe a bug -- ipc_header_req will only initialized if handle->ipc, so maybe in `handle->flags & UV_HANDLE_BLOCKING_WRITES` branch ipc_header_req will not initialized, can you review this? @indutny

starwing:
fix remain 3 warnings, now there are only one warnings, about `ipc_header_req`, please review :)

saghul:
Still want me to review this @indutny?

src/win/error.c
```diff
@@ -33,6 +33,9 @@
 /*
  * Display an error message and abort the event loop.
  */
+#if defined(_MSC_VER) || defined(__MINGW32__)
+    __declspec(noreturn)
```
saghul: Since this is windows code, what compile would not need this? Can't we just do it always?

src/win/process-stdio.c
```diff
@@ -140,7 +140,8 @@ static int uv__create_stdio_pipe_pair(uv_loop_t* loop,
                            NULL);
   if (child_pipe == INVALID_HANDLE_VALUE) {
     err = GetLastError();
-    goto error;
+    if (err)
```
saghul: Why the if? We do know there was an error.

saghul:
I'm not sure about replacing assert(0) with abort(). I'm generally more inclined to fail gracefully.

src/win/error.c
starwing: okay... you are total right :)

src/win/process-stdio.c
starwing: because gcc doesn't know err is always non-zero, so it warnings about not initialized child_pipe, has any ideas about this?

starwing:
removed `#ifdef`s.

src/win/process-stdio.c
saghul: Uh, but child_pipe is actually initialized to INVALID_HANDLE. If the compiler is stupid, I'd rather leave the warning in, than have weird code I'll scratch my head about later. FTR, no warning is printed on VS.

starwing: but the variable gcc complain _is_ child_pipe_ptr, i.e. it complains the caller's child_pipe. we just doesn't initialize child_pipe_ptr if a error occurs.
so, add a child_pipe_ptr assignment on failure path?

saghul: You can assign it to INVALID_HANDLE on top, just like child_pipe IMHO.

starwing:
okay:) fix this, now this patch resolve all -Wall -O3 warnings from gcc, except this one:

```
src/win/pipe.c:1226:20: warning: 'ipc_header_req' may be used uninitialized in t
his function [-Wmaybe-uninitialized]
         CloseHandle(ipc_header_req->overlapped.hEvent);
                    ^
```

hope for review this warning, and tell us whether its a bug or how to fix this, thanks! @indutny @saghul

saghul:
I'll run the test suite on VS a bit later, to check that we didn't create warning there :-)

@piscisaureus, what's your stand on the assert/abort thing?

starwing:
if you use VS2013, you can open /sdl check, sdl stands for software development lifetime, it may give you some warnings about initializing. I tried last time and notice several warnings.

starwing:
I have tested on VS2013, no bug report, I will open /sdl to see further

starwing:
`ipc_header_req` is nearly a bug -- /sdl check also warning about that. /sdl check also report fs.c(789), tty(216), but I have checked these, they are safe.

starwing:
hi dear @saghul has some news on this topic?

saghul:
LGTM for the most part, but I still don't like the `abort()`s. I'd like a second opinion from @indutny or @piscisaureus on that.

starwing:
@saghul I don't like abort(), though. what about add a return; after assert(0)? notice that assert will disable in Relese mode, a return may insure the control flow is correct.
Another way is use uv_fatal_error, to fail gracefully in these case.
how some opinions?

starwing:
@saghul Hi, Sorry for the long time repy, I'm busy these days.

I have removed all of `abort()`, using `assert(0) + return;` instead. still hoping for a LGTM :)

several improvement: 
- removed `abort()`, as mentioned above. 
- fix another warning from pipe.c, introduced by previous commits.
- tested on VS2013 and run tests (5 test failure, unfortunately).
- fix all warnings include the `ipc_header_req` ones (add a assert for safe).

hopes for a review and merge, thank you :)

saghul:
@starwing thnaks! I'll have it one last look tonight.

saghul:
Thanks @starwing! Landed in https://github.com/joyent/libuv/commit/1ec4c234388c0548db149772b2b8fd510a6dd843.

PR #1142, opened 2014-02-24
add uv_fe_event_getpath

saghul:
The idea is to make the path field of the uv_fs_event_t structure private. After this one I plan to make the equivalent for uv_fspoll_t.

/cc @indutny


include/uv.h
```diff
@@ -2027,11 +2028,15 @@ enum uv_fs_event_flags {
 
 UV_EXTERN int uv_fs_event_start(uv_fs_event_t* handle,
                                 uv_fs_event_cb cb,
-                                const char* filename,
+                                const char* path,
                                 unsigned int flags);
 
 UV_EXTERN int uv_fs_event_stop(uv_fs_event_t* handle);
 
+UV_EXTERN int uv_fs_event_getpath(uv_fs_event_t* handle,
```
indutny, line 2043: Needs documentation, at least a bit of.

src/uv-common.c
```diff
@@ -444,3 +444,23 @@ int uv__getaddrinfo_translate_error(int sys_err) {
   abort();
   return 0;  /* Pacify compiler. */
 }
+
```
indutny, line 447: Two "LF"s?

indutny:
LGTM, I hope all tests are passing and it is building fine on windows ;)

include/uv.h
saghul, line 2043: Will add!

src/uv-common.c
saghul, line 447: Ups.

saghul:
Yep. I tried it on Linux and Windows and works fine. I'll fix the nits and add some docs later during the day. Thanks for the review!

saghul:
Force pushed. I don't see 2 LFs, I had to add one because the function I added goes at the bottom of the file.

saghul:
@indutny, I updated the PR and created a separated test, can you have a final look? (I just changed the test)

test/test-fs-event.c
```diff
@@ -628,6 +628,38 @@ static void fs_event_cb_close(uv_fs_event_t* handle, const char* filename,
...
+  ASSERT(r == 0);
+  len = sizeof buf;
+  r = uv_fs_event_getpath(&fs_event, buf, &len);
+  ASSERT(r == UV_EINVAL);
+  r = uv_fs_event_start(&fs_event, fail_cb, "watch_dir", 0);
+  ASSERT(r == 0);
+  len = sizeof buf;
+  r = uv_fs_event_getpath(&fs_event, buf, &len);
+  ASSERT(r == 0);
+  ASSERT(memcmp(buf, "watch_dir", len) == 0);
```
indutny, line 649: Define `watch_dir` somewhere above and reuse it?

test/test-fs-event.c
```diff
@@ -628,6 +628,38 @@ static void fs_event_cb_close(uv_fs_event_t* handle, const char* filename,
...
+  uv_loop_t* loop = uv_default_loop();
+  int r;
+  char buf[1024];
+  size_t len;
+
+  create_dir(loop, "watch_dir");
+
+  r = uv_fs_event_init(loop, &fs_event);
+  ASSERT(r == 0);
+  len = sizeof buf;
```
indutny, line 641: `sizeof(buf)` I guess.

indutny:
LGTM

saghul:
I merged this one as is, since all (nearby) tests seem to do the same.

PR #1144, opened 2014-02-24
unix, windows: map ENOPROTOOPT errno

### LINKED ISSUE #1143: Add ENOPROTOOPT error to the error map
Steps of the problem:
1. Call `uv_tcp_init`
2. Call `uv_tcp_bind` with an IPv4 address
3. Call `uv_tcp_bind` with an IPv6 address
4. The previous operation fails with resulting code -92 (-ENOPROTOOPT)
5. Call `uv_err_name` with that `err` value
6. End up in `assert(0)` (or segfault with NDEBUG)

Interestingly enough that if you swap steps 2 and 3, the error code is -22 (-EINVAL).

Also, please consider returning "UNKNOWN" from `uv_err_name` instead of asserting (symmetrical to `uv_strerror`'s behaviour).

Code snippet:

``` C
#include <assert.h>
#include <stdio.h>
#include <uv.h>

int main() {
    int err;
    struct sockaddr_in ba1;
    struct sockaddr_in6 ba2;
    uv_tcp_t tcp;

    uv_tcp_init(uv_default_loop(), &tcp);

    assert(uv_ip4_addr("0.0.0.0", 3333, &ba1) == 0);
    err = uv_tcp_bind(&tcp, (const struct sockaddr *)&ba1, 0);
    if (err) {
        printf("err: %d\n", err);
        printf("name: %s\n", uv_err_name(err));
        return 0;
    }
    printf("ipv4 bound ok\n");

    assert(uv_ip6_addr("::", 4444, &ba2) == 0);
    err = uv_tcp_bind(&tcp, (const struct sockaddr *)&ba2, UV_TCP_IPV6ONLY);
    if (err) {
        printf("err: %d\n", err);
        printf("name: %s\n", uv_err_name(err));
        return 0;
    }
    printf("ipv6 bound ok\n");
}
```

Output:

```
ipv4 bound ok
err: -92
test: src/uv-common.c:83: uv_err_name: Assertion `0' failed.
```

saghul:
Fixes #1143

### DIFF
--- a/include/uv-errno.h
+++ b/include/uv-errno.h
@@ -376,4 +376,10 @@
 # define UV__EFBIG (-4036)
 #endif
 
+#if defined(ENOPROTOOPT) && !defined(_WIN32)
+# define UV__ENOPROTOOPT (-ENOPROTOOPT)
+#else
+# define UV__ENOPROTOOPT (-4035)
+#endif
+
 #endif /* UV_ERRNO_H_ */
--- a/include/uv.h
+++ b/include/uv.h
@@ -114,6 +114,7 @@ extern "C" {
   XX(ENOENT, "no such file or directory")                                     \
   XX(ENOMEM, "not enough memory")                                             \
   XX(ENONET, "machine is not on the network")                                 \
+  XX(ENOPROTOOPT, "protocol not available")                                   \
   XX(ENOSPC, "no space left on device")                                       \
   XX(ENOSYS, "function not implemented")                                      \
   XX(ENOTCONN, "socket is not connected")                                     \


indutny:
Should it be backported to v0.10?

saghul:
On 02/24/2014 09:57 AM, Fedor Indutny wrote:

> Should it be backported to v0.10?
> 
> —
> https://github.com/joyent/libuv/pull/1144#issuecomment-35867577.

Hum, I guess so. I'll make a backport of this one together with 
https://github.com/joyent/libuv/commit/107be2bed38afa6279aa53b13b946eb60c204969 
a bit later today, ok?

## 

Saúl Ibarra Corretgé
bettercallsaghul.com

indutny:
Yep, thank you!

PR #1145, opened 2014-02-24
errno: add ETXTBSY

indutny:
Add `ETXTBSY` which could be returned by file system reads on some
unixes and emulate it with -148 (`ERROR_PATH_BUSY`) on windows.

/cc @saghul


include/uv-errno.h
```diff
@@ -364,6 +364,12 @@
 # define UV__ETIMEDOUT (-4039)
 #endif
 
+#if defined(ETXTBSY) && !defined(_WIN32)
+# define UV__ETXTBSY (-ETXTBSY)
+#else
+# define UV__ETXTBSY (-148)
```
saghul: Why -148? On Windows we remap all errors ourselfs, so shouldn't it be -4038?

indutny: Oh, didn't know about that. Perhaps, you are right.

saghul: You can add the windows mapping here: https://github.com/joyent/libuv/blob/master/src/win/error.c#L67

indutny:
@saghul force pushed.

include/uv-errno.h
indutny: I guess it is fine to leave it unmapped.

saghul:
LGTM.

indutny:
Oh, actually it should land in v0.10 first.

saghul:
Well, I can backport it with the other 2, no worries.

indutny:
Ok, thank you, landed in aaaefe3

saghul:
Landed also in v0.10

PR #1148, opened 2014-02-24
unix, windows: add uv_fs_poll_getpath

saghul:
Equivalent of https://github.com/joyent/libuv/pull/1142 for fspoll handles.

/cc @indutny.

### DIFF
--- a/test/test-fs-poll.c
+++ b/test/test-fs-poll.c
@@ -33,6 +33,11 @@ static void poll_cb(uv_fs_poll_t* handle,
                     const uv_stat_t* prev,
                     const uv_stat_t* curr);
 
+static void poll_cb_fail(uv_fs_poll_t* handle,
+                         int status,
+                         const uv_stat_t* prev,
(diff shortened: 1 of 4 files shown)


test/test-fs-poll.c
```diff
@@ -144,3 +157,31 @@ static void poll_cb(uv_fs_poll_t* handle,
...
   return 0;
 }
+
+
+TEST_IMPL(fs_poll_getpath) {
+  char buf[1024];
+  size_t len;
+  loop = uv_default_loop();
+
+
```
indutny: Superfluos newline.

indutny:
LGTM

saghul:
Landed with the whitespace fixes. Thanks for the help reviewing Fedor!

trevnorris:
I'm just curious about an architecture decision. Why not pass a uv_buf_t
instead of the char*?

saghul:
Size is passed as a pointer, a buf takes the value.

PR #1149, opened 2014-02-25
unix, windows: set required size on UV_ENOBUFS

saghul:
When the supplied buffer is not big enough and UV_ENOBUFS is
returned, hint the user about the required size by setting
the len paramemeter to the required value.

Applies to:
- uv_pipe_getsockname
- uv_fs_event_getpath
- uv_fs_poll_getpath


saghul:
I thought I had already done this, but I guess it only happened in my head :-/

/cc @indutny

indutny:
LGTM, but please verify that tests are passing.

saghul:
Sure thing.

saghul:
Landed. Tests run ok, I fixed a typo on Windows that I caught during testing. Thanks for reviewing!

PR #1152, opened 2014-02-25
unix, windows: clarify what uv_stream_set_blocking does

saghul:
Also replace the assert with an error on unix.

### DIFF
--- a/src/unix/stream.c
+++ b/src/unix/stream.c
@@ -1511,7 +1511,5 @@ void uv__stream_close(uv_stream_t* handle) {
 
 
 int uv_stream_set_blocking(uv_stream_t* handle, int blocking) {
-  assert(0 && "implement me");
-  abort();
-  return 0;
+  return UV_ENOSYS;
 }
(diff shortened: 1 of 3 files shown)


src/unix/stream.c
```diff
@@ -1511,7 +1511,5 @@ void uv__stream_close(uv_stream_t* handle) {
 
 
 int uv_stream_set_blocking(uv_stream_t* handle, int blocking) {
-  assert(0 && "implement me");
-  abort();
-  return 0;
+  return UV_ENOSYS;
```
tarruda, line 1514: @saghul , is there some reason for this not being implemented? If you want, I can send a PR, something along these lines:

``` c
int uv_stream_set_blocking(uv_stream_t* handle, int blocking) {
  int fd = handle->io_watcher.fd, rv;

  if (blocking) {
    rv = fcntl(fd, F_SETFL, fcntl(fd, F_GETFL, 0) & ~O_NONBLOCK);
  } else {
    rv = fcntl(fd, F_SETFL, fcntl(fd, F_GETFL, 0) | O_NONBLOCK);
  }

  return rv;
}
```

What do you say?

saghul, line 1514: I left it unimplemented on purpose :-) This function was added to "fix" a problem with Node, because pipes have non-blocking writes in certain situations, so output could be missed. On Windows this function just makes pipes use blocking writes, it doesn't even work with other types of streams. It's unclear if it will stay or we'll do something else to make the situation better, but that will happen after 0.12.

PR #1155, opened 2014-02-26
build: fix compilation warning if dtrace doesn't work

saghul:
Check if dtrace should instrument object files only if dtrace
support is detected to work


saghul:
This fixes the build error I'm seeing on FreeBSD 10, where dtrace is present, but configure says it's not working, so don't bother checking if object files should be instrumented.

/cc @indutny @tjfontaine

indutny:
LGTM, but needs @tjfontaine acception too

saghul:
Sure, thanks for having a look Fedor!

tjfontaine:
lgtm

saghul:
Thanks guys, landed.

PR #1156, opened 2014-02-26
fs: use preadv on Linux if available

mscdex:


saghul:
Looks like build is failing on non-Linux systems:

```
../src/unix/fs.c:56:42: error: token is not a valid binary operator in a preprocessor subexpression
     LINUX_VERSION_CODE >= KERNEL_VERSION(2,6,30) &&                      \
                           ~~~~~~~~~~~~~~^
```

Maybe we need another set of "if"s just for linux?

indutny:
I bet it should be inside separate `ifdef`.

mscdex:
changed :-)

saghul:
Looks a bit convoluted :-) How about setting it to 1 or 0 for the BSDs and then doing the Linux check so that we only set it to 1 if all conditions are met else it will remain as 0?

mscdex:
If we do that, gcc warns about redefining. Is that ok?

saghul:
Oh, forget it then :-) LGTM.

saghul:
Thanks a lot! Landed in https://github.com/joyent/libuv/commit/269ff0beaeb8560dc6156ef34aa5dcf9b574540d

PR #1157, opened 2014-02-26
process: remove debug perror() prints

### LINKED ISSUE #1128: libuv writing to child stderr during spawn when some system calls fail
execvp(): No such file or directory, should libuv really be calling perror()?                                                                                                         It's writing the error number to the parent, and direct writes to fd 2 makes it hard in node to cleanly handle non-existence of a command, try:

```
node -e 'require("child_process").spawn("no-such-command", {stdio:"inherit"}).on("error",function(){})'
```

output is

```
execvp(): No such file or directory
```

(expected is nothing, the error is handled).

I was hoping it was stray debug code, but https://github.com/joyent/libuv/blob/master/src/unix/process.c#L362 has been there since a30e45f1, last June. The setuid and setgid just above also perror().

With this, there is no way in node to spawn, catch ENOENT, and report it without unexpected output on stderr.

/to @indutny

indutny:
fix #1128


saghul:
LGTM. I wonder if we should still have them in debug builds?

indutny:
@saghul I think no, we are returning error as an exit code, so it should be fine anyway.

sam-github:
Thanks, Fedor.

indutny:
Landed in c0c9480, thank you!

PR #1159, opened 2014-02-27
unix, windows: add assertion in uv_loop_delete

saghul:


saghul:
We'll deprecate these, but just to be on the safe side in the meantime.

indutny:
Kewl.

PR #1162, opened 2014-02-27
process: fix bug that mistakenly closes child stdin in some cases

maxtaco:
If a parent process closes 0, and then calls spawn, the following will
happen:
- the first call to socketpair() will return a pair like [0,14]
  in the parent. The kernel allocates fd=0 since it's the lowest
  free file descriptor.
- the child, after fork, in uv__process_child_init,
  calls dup2(use_fd=14,fd=0), which is fine.
- the child then calls close(close_fd=0), which is wrong.

This problem is easily solved by closing the close_fd first, and then
calling dup2 as before.

### DIFF
--- a/src/unix/process.c
+++ b/src/unix/process.c
@@ -309,6 +309,11 @@ static void uv__process_child_init(const uv_process_options_t* options,
       }
     }
 
+    if (close_fd != -1) {
+      /* Can't call uv__close, since that might fail an assertion if close_fd=0 */
+      close(close_fd);
+    }
+
     if (fd == use_fd)
       uv__cloexec(use_fd, 0);
     else
@@ -316,9 +321,6 @@ static void uv__process_child_init(const uv_process_options_t* options,
 
     if (fd <= 2)
       uv__nonblock(fd, 0);
-
-    if (close_fd != -1)
-      uv__close(close_fd);
   }
 
   for (fd = 0; fd < stdio_count; fd++) {
--- a/test/run-tests.c
+++ b/test/run-tests.c
@@ -161,6 +161,14 @@ static int maybe_run_test(int argc, char **argv) {
 
     return 1;
   }
+
+  if (strcmp(argv[1], "spawn_helper9") == 0) {
+    struct stat statbuf;
+    int rc;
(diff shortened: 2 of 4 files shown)


saghul:
Can you provide a test?

maxtaco:
Sure, it's taken me a while to get my build env setup.  I assume you don't want to see my node app that was crashing after 1000 lines of debug output.  At the very least I did verify this fixes my node app.  I'll hopefully get something to you soon for test.

saghul:
Great, thank you!

maxtaco:
Ok, a test is provided.  It fails with the master code and passes with my fix.  But, it might be a bug in node, which shouldn't be closing 0 in the first place.

indutny:
@maxtaco what node version are we talking about? I remember fixing one such bug in v0.11, perhaps it wasn't backported to v0.10?

indutny:
cc @tjfontaine

maxtaco:
Correct, my application fails on v0.10 but works on v0.11.  But I didn't drill down too far on why it started working in v0.11.  Our application needs to work on all v0.10's, so we'll put in a workaround.

indutny:
@maxtaco I'm afraid this patch won't make into libuv as it isn't fixing a real problem, which is closing stdio. And the problem was fixed there: https://github.com/joyent/node/commit/fc26fd6b . @tjfontaine could we backport it to v0.10?

maxtaco:
This behavior might come as a surprise for those who use libuv in any application outside of node.  Those applications will see buggy `uv_spawn` behavior after closing stdin.

indutny:
In fact, its not only `uv_spawn()` it may happen in other places too. For me it was happening in a kqueue hack. Anyway, there is no surprise that closing stdio will break things, it will break stuff in many other programs even without libuv.

maxtaco:
For anyone who comes across this PR with similar problems in Node v0.10.x, just do this before spawn, which will open a dummy fd=0 and allow future spawns to work:

``` javascript
if (!process.stdin._handle) {
  fs.fstat(0, function (err) {
    if (err) {
      fs.open(process.execPath, "r", function (err, fd) {});
    }
  });
}
```

Issue #1161, opened 2014-02-27
Failing tests on compilation


dariocravero:
Hi all,

I'm getting an error while running tests on a libuv build in Ubuntu both the latest release and the git repo.

Here's the compilation process output: https://gist.github.com/dariocravero/9250901.

Platform: Linux uxtemple 3.8.0-29-generic #42~precise1-Ubuntu SMP Wed Aug 14 16:19:23 UTC 2013 x86_64 x86_64 x86_64 GNU/Linux.

Any clues?

Thanks,
Darío

saghul:
These are not "real" failures. Some handles are left unref's and the loop now asserts. I'll make a change to the test runner which closes all remaining handles and runs the loop one last time.

dariocravero:
Good stuff :) Thanks @saghul. Feel free to close the issue then.

saghul:
Thanks for the report Dario. Should be fixed in master after https://github.com/joyent/libuv/pull/1163.

dariocravero:
Thanks @saghul, that sorted it out!
However, one test is still failing from that set, see [this](https://gist.github.com/dariocravero/9250901#file-libuv-compilation-error-L638):

```
=============================================================
[%  74|+ 159|-   0|T   0|S   0]: spawn_setuid_setgid
`spawn_setuid_setgid` failed: exit code 134
Output from process `spawn_setuid_setgid`:
Assertion failed in test/test-spawn.c on line 925: r == 0
=============================================================
```

Is that of any importance too?

saghul:
On 02/28/2014 10:38 AM, Darío Javier Cravero wrote:

> Thanks @saghul https://github.com/saghul, that sorted it out!

Great, thanks for confirming!

> However, one test is still failing from that set, see this
> https://gist.github.com/dariocravero/9250901#file-libuv-compilation-error-L638:
> 
> |=============================================================
> 
> `spawn_setuid_setgid` failed: exit code 134
> Output from process `spawn_setuid_setgid`:
> 
> # Assertion failed in test/test-spawn.c on line 925: r == 0
> 
> |
> |
> 
> Is that of any importance too?

Maybe :-) Please open a new issue for that one. Could be platform 
related. I'mupgrading my BBB as I type :-P

> —
> Reply to this email directly or view it on GitHub
> https://github.com/joyent/libuv/issues/1161#issuecomment-36334761.

## 

Saúl Ibarra Corretgé
bettercallsaghul.com

dariocravero:
[...]

PR #1163, opened 2014-02-28
test: make sure all handles in the loop are closed

saghul:
After https://github.com/joyent/libuv/commit/2611ef43b15dc083ac2ffd5f16901a4ff076be03 hell broke loose because we don't clean up after ourselves in the test suite. I could reproduce the failures in https://github.com/joyent/libuv/issues/1161 when using autoconf but not when using gyp (need to investigate that). Either way, this patch makes sure all handles in the loop are closed before destroying it.

### DIFF
--- a/test/task.h
+++ b/test/task.h
@@ -22,6 +22,8 @@
 #ifndef TASK_H_
 #define TASK_H_
 
+#include "uv.h"
+
 #include <stdio.h>
 #include <stddef.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


indutny:
This sounds like a hack. Perhaps, we should find what we are not closing instead?

saghul:
Well, look at test-run-once for example. It's not necessarily wrong to bail early from uv_run if you unref all handles. MAKE_VALGRIND_HAPPY was only added, well, to make valgrind happy, it's not needed, but since we are here we can also make saghul happy ;-)

saghul:
Not to mention that MAKE_VALGRIND_HAPPY is the last thing we do in all tests, so all asserts have already been checked and you know the test passed.

indutny:
Ok, I guess those are not serious problems anyway. LGTM

saghul:
On 02/28/2014 09:53 AM, Fedor Indutny wrote:

> Ok, I guess those are not serious problems anyway. LGTM

Yep, they are not :-) Thanks, Fedor.

## 

Saúl Ibarra Corretgé
bettercallsaghul.com

PR #1167, opened 2014-02-28
build: Fix configure when PKG_CONFIG is set

thierry-FreeBSD:
If you define the environment variable PKG_CONFIG and assign it to
pkgconf (or pkg-config), configure will say it's OK, but it won't be
used.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -13,7 +13,7 @@
 # OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
 AC_PREREQ(2.57)
-AC_INIT([libuv], [0.11.21], [https://github.com/joyent/libuv/issues])
+AC_INIT([libuv], [0.11.22], [https://github.com/joyent/libuv/issues])
 AC_CONFIG_MACRO_DIR([m4])
 m4_include([m4/libuv-extra-automake-flags.m4])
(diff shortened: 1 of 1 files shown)


saghul:
@thierry-FreeBSD thanks! could you guide me on validating this? That is, what is the expected outcome after the patch?

thierry-FreeBSD:
After `./autogen.sh', run configure with this command:

PKG_CONFIG=pkgconf ./configure

Without the proposed patch, libuv.pc won't be generated, and the following lines from Makefile.in:

@HAVE_PKG_CONFIG_TRUE@pkgconfigdir = $(libdir)data/pkgconfig
@HAVE_PKG_CONFIG_TRUE@pkgconfig_DATA = @PACKAGE_NAME@.pc

will commented out in Makefile.

With the patch, it will be OK (of course, if pkgconf exists in your $PATH!).

saghul:
Thanks! I'll give it a shot!

saghul:
Thanks @thierry-FreeBSD! I squashed the 2 commits into 1 and landed it in https://github.com/joyent/libuv/commit/b870d914ebbcec2f0124dda6f96e47cdd4abd7d3. FTR, I tested this by setting the PKG_CONFIG env variable pointing to a pkg-config-ish thing written in Python I found in some project (https://gist.github.com/saghul/9287458)

PR #1170, opened 2014-03-02
unix: workaround broken pthread_sigmask on Android

### LINKED ISSUE #1169: uv_signal_start segfaults on android
```
#include <stdio.h>
#include "uv.h"

void signal_handler(uv_signal_t *handle, int signum)
{
    printf("Signal received: %d\n", signum);
}

int main()
{
    uv_loop_t *loop = uv_loop_new();
    uv_signal_t sig1;
    uv_signal_init(loop, &sig1);
    uv_signal_start(&sig1, signal_handler, SIGUSR1);
    uv_run(loop, UV_RUN_DEFAULT);
    return 0;
}
```

The above segfaults on android ARMv7 on uv_signal_start. Tested on b870d914ebbcec2f0124dda6f96e47cdd4abd7d3

pyokagan:
Fixes #1169

Redefine all references of `pthread_sigmask` to `sigprocmask` in order
to workaround broken behaviour of `pthread_sigmask` on Android
versions <4.1. See:
- https://android.googlesource.com/platform/bionic/+/9bf330b5
- https://code.google.com/p/android/issues/detail?id=15337


src/unix/signal.c
```diff
@@ -43,6 +43,19 @@
 static void uv__signal_stop(uv_signal_t* handle);
 
 
+#ifdef __ANDROID__
```
indutny: What do you think about making it a define and moving to `src/unix/internal.h` ?

Something like this:

``` C
#define uv__sigmask(how, set, oset) pthread_sigmask(how, set, oset)
```

pyokagan: sure thing. I'll modify the PR.

saghul: Another approach would be to add a working ptherad_sigmask in pthread-fixes.c, which we already use for Android IIRC.

indutny: Oooooh, that's right @saghul. @pyokagan sorry, but could you please do it that way?

pyokagan: Would redefining pthread_sigmask in pthread-fixes.h be okay?

saghul: On 03/02/2014 11:50 AM, Paul Tan wrote:

> In src/unix/signal.c:
> 
> > @@ -43,6 +43,19 @@
> >  static void uv__signal_stop(uv_signal_t\* handle);
> > 
> > +#ifdef **ANDROID**
> 
> Would redefining pthread_sigmask in pthread-fixes.h be okay?

Maybe you could first try to use pthread_sigmask and if it returns 
EINVAL switch to sigprocmask? Also, remember that with a static variable 
for next calls to pthread_sigmask.

## 

Saúl Ibarra Corretgé
bettercallsaghul.com

pyokagan: Okay, implemented in 072f76d7

include/pthread-fixes.h
```diff
@@ -56,4 +56,17 @@ int pthread_barrier_init(pthread_barrier_t* barrier,
 #endif  /* defined(PTHREAD_BARRIER_SERIAL_THREAD) */
 
 int pthread_yield(void);
+
+#if __ANDROID__
```
saghul: You souldn't need this, since pthread-fixes.h is only used in Android, at least for now.

src/unix/pthread-fixes.c
```diff
@@ -29,6 +29,31 @@
    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
+/* Android versions <4.1 have a broken pthread_sigmask
+ * Note that this block of code must come before any inclusion of
+ * pthread-fixes.h so that the real pthread_sigmask can be referenced.
+ * */
+#ifdef __ANDROID__
```
saghul: Ditto.

src/unix/pthread-fixes.c
```diff
@@ -29,6 +29,31 @@
...
+#include <errno.h>
+#include <pthread.h>
+
+int uv__pthread_sigmask(int how, const sigset_t* set, sigset_t* oset) {
+  static int workaround = 0;
```
saghul: NO need to initialize it to 0.

src/unix/pthread-fixes.c
```diff
@@ -29,6 +29,31 @@
    OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
+/* Android versions <4.1 have a broken pthread_sigmask
```
saghul: Space after <

pyokagan:
@saghul Thanks for reviewing! I have implemented your changes in fdab0c4.

saghul:
Left a comment, other than that, LGTM.

saghul:
Thanks! Landed in https://github.com/joyent/libuv/commit/347000929de775ae6b975a2959f5fabe81d6f1f0

PR #1173, opened 2014-03-02
windows: fall back for volume info query

ihnorton:
Julia's build system uses mingw and wine for windows packaging. We hit the following issue after rebasing to libuv 0.11.20: Wine [does not](https://github.com/mirrors/wine/blob/0e42fd97c05beb50c0788164647631c00485623b/dlls/ntdll/file.c#L2679) currently support `pNtQueryVolumeInformationFile` for `FileFsVolumeInformation` (implemented in https://github.com/joyent/libuv/commit/315d7001e8e0a86f30dda68b237d0b25dce48d73 and https://github.com/joyent/libuv/commit/20a8e58adb2a3fc2e92ceb6e1ceb6f05d7e0d439)

This patch checks `io_status` and falls back to previous behavior if the query is not implemented.

### DIFF
--- a/src/win/fs.c
+++ b/src/win/fs.c
@@ -861,9 +861,13 @@ INLINE static int fs__stat_handle(HANDLE handle, uv_stat_t* statbuf) {
                                             FileFsVolumeInformation);
 
   /* Buffer overflow (a warning status code) is expected here. */
-  if (NT_ERROR(nt_status)) {
+  if (io_status.Status == STATUS_NOT_IMPLEMENTED) {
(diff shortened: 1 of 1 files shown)


ihnorton:
I signed CLA and fixed commit message.

saghul:
LGTM. @piscisaureus ?

piscisaureus:
LGTM too

saghul:
Thanks @ihnorton! Landed in https://github.com/joyent/libuv/commit/2930d04e5b2d47f889290ed82bbb27c99f2dd430

PR #1179, opened 2014-03-06
freebsd: use accept4, introduced in version 10

saghul:


saghul:
Tested it, works, landed. You can blame me later :-)

bnoordhuis:
Belated LGTM FWIW.  You may get a compiler warning on other platforms because `__FreeBSD__` is not defined but only when using very strict settings.

saghul:
On 03/06/2014 06:20 PM, Ben Noordhuis wrote:

> Belated LGTM FWIW. You may get a compiler warning on other platforms
> because |**FreeBSD**| is not defined but only when using very strict
> settings.

Thanks Ben!

## 

Saúl Ibarra Corretgé
bettercallsaghul.com

PR #1180, opened 2014-03-06
error: add ENXIO for O_NONBLOCK FIFO open()

indutny:
When opening FIFO with `O_NONBLOCK` flag, `ENXIO` could be returned if
the readable side hasn't yet opened this FIFO.


saghul:
My Linux says: " ENXIO           No such device or address (POSIX.1)"

saghul:
I think we should use the "default" message.

saghul:
LGTM.

PR #1182, opened 2014-03-07
build: allow CC to be overridden on Makefile.mingw

saghul:
I'm testing cross compilation for MinGW (32 and 64) on a Debian system, using the Makefile.mingw file (I know autotools works, but I want to try this one out).

I'm using the following script:

```
#!/bin/sh

PREFIX=i686-w64-mingw32
export CC=$PREFIX-gcc
export CXX=$PREFIX-g++
export CPP=$PREFIX-cpp
export AR=$PREFIX-ar
export RANLIB=$PREFIX-ranlib
exec "$@"
```

so overridding CC is a needed.


saghul:
@indutny, quick review for this 1-char PR? :-)

indutny:
LGTM

saghul:
Thanks, Fedor :-)

PR #1183, opened 2014-03-07
dev/win32 minor fixes

zerhacken:
- fix uv_handle_size and uv_req_size when failing return 0
- fix VC++ warning C4244
- fix VC++ warning C4389


zerhacken:
@Nodejs-Jenkins signed CLA

src/uv-common.c
```diff
@@ -52,7 +52,7 @@ size_t uv_handle_size(uv_handle_type type) {
   switch (type) {
     UV_HANDLE_TYPE_MAP(XX)
     default:
-      return -1;
+      return 0;
```
retep998, line 55: Wouldn't it be better to return (size_t)-1 here?

zerhacken:
@retep998 thanks, you are right. Will fix it.

src/uv-common.c
```diff
@@ -52,7 +52,7 @@ size_t uv_handle_size(uv_handle_type type) {
   switch (type) {
     UV_HANDLE_TYPE_MAP(XX)
     default:
-      return -1;
+      return (size_t) - 1;
```
saghul: This is not a good idea, it will return the biggest possible integer, which is not what we want.

zerhacken:
Revert to my original fix returning 0?

saghul:
I guess so, yes. The size of a handle needs to be > 0.

zerhacken:
@saghul reverted to original fix, hope this is better?

src/uv-common.c
```diff
@@ -101,7 +101,7 @@ const char* uv_strerror(int err) {
 int uv_ip4_addr(const char* ip, int port, struct sockaddr_in* addr) {
   memset(addr, 0, sizeof(*addr));
   addr->sin_family = AF_INET;
-  addr->sin_port = htons(port);
+  addr->sin_port = htons((u_short) port);
```
saghul: I'm not seeing any warning here, and u_short is a Windows-ism. uint16_t would be fine but I've never seen a warning, what system are you using?

src/uv-common.c
```diff
@@ -115,7 +115,7 @@ int uv_ip6_addr(const char* ip, int port, struct sockaddr_in6* addr) {
 
   memset(addr, 0, sizeof(*addr));
   addr->sin6_family = AF_INET6;
-  addr->sin6_port = htons(port);
+  addr->sin6_port = htons((u_short) port);
```
saghul: Ditto.

src/win/poll.c
```diff
@@ -209,7 +209,7 @@ static int uv__fast_poll_set(uv_loop_t* loop, uv_poll_t* handle, int events) {
   assert(!(handle->flags & UV__HANDLE_CLOSING));
   assert((events & ~(UV_READABLE | UV_WRITABLE)) == 0);
 
-  handle->events = events;
+  handle->events = (unsigned char) events;
```
saghul: Maybe we should declare events as unsigned int instead.

src/win/poll.c
```diff
@@ -451,7 +451,7 @@ static int uv__slow_poll_set(uv_loop_t* loop, uv_poll_t* handle, int events) {
...
```
saghul, line 454: Ditto.

src/uv-common.c
zerhacken: VC++ (2012 + 2013) its a /L4 warning, you are right uint16_t would be better.

saghul: What is a /L4 warning?

Also, the function itself accepts an int, but ports cannot be negative, so maybe we need to adjust uv_ip4_add itself...

src/win/poll.c
zerhacken: That would be perfect.

src/uv-common.c
zerhacken: /L4 short hand for compiler warning level 4 (http://msdn.microsoft.com/en-us/library/thxezb7y.aspx).

Will look into adjusting uv_ip4_add to take an unsigned int if you are fine with that?
Actually "uv_ip4_add" might as well take an unsigned short?

zerhacken:
@saghul this I like better, should be warning-less and even a tad prettier

zerhacken:
The test failure on smartos build doesnt appear as related

PR #1184, opened 2014-03-07
unix: reopen tty as /dev/tty

saghul:
Reopen the file descriptor when it refers to a tty. This lets us put the tty in
non-blocking mode without affecting other processes that share it with
us.

This brings back commit 31f9fbc, which was reverted in 20bb1bf. The OSX
select trick is working now.

Original patch by @bnoordhuis

### DIFF
--- a/src/unix/core.c
+++ b/src/unix/core.c
@@ -59,12 +59,15 @@
 # include <sys/filio.h>
 # include <sys/ioctl.h>
 # include <sys/wait.h>
-#endif
-
-#if __FreeBSD__ >= 10
-# define uv__accept4 accept4
-# define UV__SOCK_NONBLOCK SOCK_NONBLOCK
-# define UV__SOCK_CLOEXEC  SOCK_CLOEXEC
+# define UV__O_CLOEXEC O_CLOEXEC
+# if __FreeBSD__ >= 10
+#  define uv__accept4 accept4
+#  define UV__SOCK_NONBLOCK SOCK_NONBLOCK
+#  define UV__SOCK_CLOEXEC  SOCK_CLOEXEC
+# endif
+# if !defined(F_DUP2FD_CLOEXEC) && defined(_F_DUP2FD_CLOEXEC)
+#  define F_DUP2FD_CLOEXEC  _F_DUP2FD_CLOEXEC
+# endif
 #endif
 
 static void uv__run_pending(uv_loop_t* loop);
@@ -820,4 +823,94 @@ int uv_getrusage(uv_rusage_t* rusage) {
   rusage->ru_nivcsw = usage.ru_nivcsw;
 
   return 0;
+}
+
+
+int uv__open_cloexec(const char* path, int flags) {
+  int err;
+  int fd;
+
+#if defined(__linux__) || (defined(__FreeBSD__) && __FreeBSD__ >= 9)
+  static int no_cloexec;
+
+  if (!no_cloexec) {
+    fd = open(path, flags | UV__O_CLOEXEC);
+    if (fd != -1)
+      return fd;
+
+    if (errno != EINVAL)
+      return -errno;
+
+    /* O_CLOEXEC not supported. */
+    no_cloexec = 1;
+  }
+#endif
+
+  fd = open(path, flags);
+  if (fd == -1)
+    return -errno;
+
+  err = uv__cloexec(fd, 1);
+  if (err) {
+    uv__close(fd);
+    return err;
+  }
+
+  return fd;
+}
+
+
+int uv__dup2_cloexec(int oldfd, int newfd) {
+  int r;
+#if defined(__FreeBSD__) && __FreeBSD__ >= 10
+  do
+    r = dup3(oldfd, newfd, O_CLOEXEC);
+  while (r == -1 && errno == EINTR);
+  if (r == -1)
+    return -errno;
+  return r;
+#elif defined(__FreeBSD__) && defined(F_DUP2FD_CLOEXEC)
+  do
+    r = fcntl(oldfd, F_DUP2FD_CLOEXEC, newfd);
+  while (r == -1 && errno == EINTR);
+  if (r != -1)
+    return r;
+  if (errno != EINVAL)
+    return -errno;
+  /* Fall through. */
+#elif defined(__linux__)
+  static int no_dup3;
+  if (!no_dup3) {
+    do
+      r = uv__dup3(oldfd, newfd, UV__O_CLOEXEC);
+    while (r == -1 && (errno == EINTR || errno == EBUSY));
+    if (r != -1)
+      return r;
+    if (errno != ENOSYS)
+      return -errno;
+    /* Fall through. */
+    no_dup3 = 1;
+  }
+#endif
+  {
+    int err;
+    do
+      r = dup2(oldfd, newfd);
+#if defined(__linux__)
+    while (r == -1 && (errno == EINTR || errno == EBUSY));
+#else
+    while (r == -1 && errno == EINTR);
+#endif
+
+    if (r == -1)
+      return -errno;
+
+    err = uv__cloexec(newfd, 1);
+    if (err) {
+      uv__close(newfd);
+      return err;
+    }
+
+    return r;
+  }
 }
--- a/src/unix/linux-syscalls.c
+++ b/src/unix/linux-syscalls.c
@@ -219,6 +219,16 @@
 # endif
 #endif /* __NR_pwritev */
 
+#ifndef __NR_dup3
+# if defined(__x86_64__)
+#  define __NR_dup3 292
+# elif defined(__i386__)
+#  define __NR_dup3 330
+# elif defined(__arm__)
+#  define __NR_dup3 (UV_SYSCALL_BASE + 358)
+# endif
+#endif /* __NR_pwritev */
+
 
 int uv__accept4(int fd, struct sockaddr* addr, socklen_t* addrlen, int flags) {
 #if defined(__i386__)
@@ -407,6 +417,7 @@ int uv__utimesat(int dirfd,
 #endif
 }
 
+
 ssize_t uv__preadv(int fd, const struct iovec *iov, int iovcnt, off_t offset) {
 #if defined(__NR_preadv)
   return syscall(__NR_preadv, fd, iov, iovcnt, offset);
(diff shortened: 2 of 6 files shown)


saghul:
So, I un-reverted and adapted the un-revert, tried the failing case `node | cat` and ran the test suite and it seems to work!

/cc @indutny, @bnoordhuis: can you guys review?

PS: I still need to look into making the newdf close-on-exec.

indutny:
Needs close-on-exec, otherwise harmful :)

saghul:
Force pushed!

src/unix/tty.c
```diff
@@ -35,26 +35,70 @@
...
+   * with us.
+   *
+   * Example: `node | cat` - if we put our fd 0 in non-blocking mode, it also
+   * affects fd 1 of `cat` because both file descriptors refer to the same
+   * struct file in the kernel. When we reopen our fd 0, it points to a
+   * different struct file, hence changing its properties doesn't affect
+   * other processes.
+   */
+  if (isatty(fd)) {
+    newfd = open("/dev/tty", O_RDWR);
```
bnoordhuis: There is a `uv__open_cloexec()` in src/unix/stream.c that you can use.

src/unix/tty.c
```diff
@@ -35,26 +35,70 @@
...
+
+    if (newfd == -1)
+      return -errno;
+
+    do
+      r = dup2(newfd, fd);
```
bnoordhuis: This should use dup3() or fcntl(F_DUP2FD_CLOEXEC) where possible.

bnoordhuis: And use uv__cloexec() when those are unavailable, of course.

saghul:
@bnoordhuis I addressed your feedback, can you check again? I added a `uv__dup_cloexec` helper funcion, should we ever need it. It uses:
- dup3 on Linux (if possible)
- dup3 on FreeBSD 10
- fcntl with F_DUP2FD_CLOEXEC (if possible)
- dup2 + uv__cloexec as a last resort

Hope I got it right :-)

saghul:
Oh, I'll rename uv__dup_cloexec to uv__dup2_cloexec, kinda makes more sense.

src/unix/stream.c
```diff
@@ -93,6 +93,52 @@ static int uv__open_cloexec(const char* path, int flags) {
...
+  static int no_dup3;
+  if (!no_dup3) {
+    do
+      r = uv__dup3(oldfd, newfd, UV__O_CLOEXEC);
+    while (r == -1 && (errno == EINTR || errno == EBUSY));
+    if (r == -1 && errno == ENOSYS)
+      no_dup3 = 1;
+    else if (r == -1)
+      return -errno;
+    return 0;
```
bnoordhuis: Should return r.

src/unix/stream.c
```diff
@@ -93,6 +93,52 @@ static int uv__open_cloexec(const char* path, int flags) {
...
+  }
+#elif defined(__FreeBSD__)
+# if __FreeBSD__ >= 10
+  do
+    r = dup3(oldfd, newfd, O_CLOEXEC);
+  while (r == -1 && errno == EINTR);
+  if (r == -1)
+    return -errno;
+  return 0;
```
bnoordhuis: Ditto.

src/unix/stream.c
saghul: I thought of that, but what should the fallback case (dup2 + cloexec)
return?

src/unix/tty.c
```diff
@@ -35,26 +35,61 @@
...
+   * other processes.
+   */
+  if (isatty(fd)) {
+    newfd = uv__open_cloexec("/dev/tty", O_RDWR);
+
+    if (newfd == -1)
+      return -errno;
+
+    r = uv__dup_cloexec(newfd, fd);
+    if (r && r != -EINVAL) {
```
bnoordhuis: When you update uv__dup_cloexec(), you need to update this check as well.

src/unix/stream.c
bnoordhuis: In the inconceivable case that uv__cloexec() fails, the return value of uv__cloexec() (and it should close the file descriptor).  Else it should return the new file descriptor.

bnoordhuis:
Maybe move uv__dup_cloexec() and uv__open_cloexec() to src/unix/core.c, that's where most of the system call polyfills live.

src/unix/stream.c
saghul: Sounds reasonable, thanks a lot for the help Ben.

saghul:
@bnoordhuis addressed your last comments.

src/unix/core.c
```diff
@@ -821,3 +821,87 @@ int uv_getrusage(uv_rusage_t* rusage) {
...
+
+#if defined(__linux__)
+  fd = open(path, flags | UV__O_CLOEXEC);
+  if (fd != -1)
+    return fd;
+
+  if (errno != EINVAL)
+    return -errno;
+
+  /* O_CLOEXEC not supported. */
```
bnoordhuis: It would be good to cache the result in a no_cloexec static variable so you don't keep repeating the open(O_CLOEXEC) call when it's unsupported.

src/unix/core.c
```diff
@@ -821,3 +821,87 @@ int uv_getrusage(uv_rusage_t* rusage) {
...
+  static int no_dup3;
+  if (!no_dup3) {
+    do
+      r = uv__dup3(oldfd, newfd, UV__O_CLOEXEC);
+    while (r == -1 && (errno == EINTR || errno == EBUSY));
+    if (r == -1 && errno == ENOSYS)
+      no_dup3 = 1;
+    else if (r == -1)
+      return -errno;
+    return r;
```
bnoordhuis: Returns -1 when the system call failed with ENOSYS.  Probably not what you intended, I suspect you want it to fall through.

src/unix/core.c
```diff
@@ -821,3 +821,87 @@ int uv_getrusage(uv_rusage_t* rusage) {
...
+# elif !defined(F_DUP2FD_CLOEXEC) && defined(_F_DUP2FD_CLOEXEC)
+#  define F_DUP2FD_CLOEXEC  _F_DUP2FD_CLOEXEC
+# endif
+# if defined(F_DUP2FD_CLOEXEC)
+  do
+    r = fcntl(oldfd, F_DUP2FD_CLOEXEC, newfd);
+  while (r == -1 && errno == EINTR);
+  if (r == -1)
+    return -errno;
+  return r;
```
bnoordhuis: You may want to consider falling through on EINVAL in case libuv is compiled against kernel headers that are newer than the actual kernel.

src/unix/core.c
```diff
@@ -821,3 +821,87 @@ int uv_getrusage(uv_rusage_t* rusage) {
...
+# endif
+#endif
+  {
+    int err;
+    do
+      r = dup2(oldfd, newfd);
+    while (r == -1 && (errno == EINTR || errno == EBUSY));
```
bnoordhuis, line 900: Minor caveat: EBUSY is a Linux-ism.  The BSDs never return it and I'm pretty sure Solaris doesn't either.

Probably doesn't hurt now but might be good to exclude it, just in case dup2() in the future does start returning EBUSY but with semantics that are different from Linux.

src/unix/linux-syscalls.c
```diff
@@ -422,3 +434,12 @@ ssize_t uv__pwritev(int fd, const struct iovec *iov, int iovcnt, off_t offset) {
   return errno = ENOSYS, -1;
 #endif
 }
+
+
+int uv__dup3(int oldfd, int newfd, int flags) {
+#if defined(__NR_dup3)
+  return syscall(__NR_dup3, oldfd, newfd);
```
bnoordhuis: You forgot the flags argument.

src/unix/tty.c
```diff
@@ -35,26 +35,61 @@
...
   }
+#endif
 
+  if (readable)
+    flags = UV_STREAM_READABLE;
+  else
+    flags = UV_STREAM_WRITABLE;
+
+  uv__nonblock(fd, 1);
+  uv__stream_open((uv_stream_t*)tty, fd, flags);
```
bnoordhuis: Style is inconsistent with a few lines up: `(uv_stream_t*)tty` vs. `(uv_stream_t*) tty`

src/unix/linux-syscalls.c
saghul: What the... :-/

saghul:
Thanks for the detailed feeback @bnoordhuis! I think I now coevered all those blind spots :-)

src/unix/core.c
```diff
@@ -821,3 +821,101 @@ int uv_getrusage(uv_rusage_t* rusage) {
...
+    return r;
+  }
+#elif defined(__FreeBSD__)
+# if __FreeBSD__ >= 10
+  do
+    r = dup3(oldfd, newfd, O_CLOEXEC);
+  while (r == -1 && errno == EINTR);
+  if (r == -1)
+    return -errno;
+  return r;
```
bnoordhuis, line 871: This will make the compiler complain about unreachable code and the 'fallback' label being unused.  I think you can rewrite the error checking logic in all three cases to something like this:

``` c
if (r != -1)
  return r;
if (errno != EINVAL)
  return -errno;
/* Fall through. */
```

saghul, line 871: Changed. How do you suggest I initialize `err`? In case of FreeBSD 10 the fallthrough case will never be reached.

bnoordhuis, line 871: Maybe like this?

``` c
int uv__dup2_cloexec(int oldfd, int newfd) {
#if defined(__FreeBSD__) && __FreeBSD__ >= 10
  int r;
  do
    r = dup3(oldfd, newfd, O_CLOEXEC);
  while (r == -1 && errno == EINTR);
  if (r < 0)
    return -errno;
  return r;
#else
# if defined(__linux__)
  // ...
# elif /* etc. */
  // ...
# endif
#endif
/* fall through goes here */
#endif
}
```

saghul, line 871: Updated, how about now?

src/unix/core.c
```diff
@@ -821,3 +823,93 @@ int uv_getrusage(uv_rusage_t* rusage) {
...
+
+int uv__open_cloexec(const char* path, int flags) {
+  int err;
+  int fd;
+
+#if defined(__linux__)
+  static int no_cloexec;
+
+  if (!no_cloexec) {
+    fd = open(path, flags | UV__O_CLOEXEC);
```
bnoordhuis, line 837: I believe FreeBSD 10 also supports open(O_CLOEXEC) but apart from that LGTM.

saghul, line 837: On 03/10/2014 06:04 PM, Ben Noordhuis wrote:

> In src/unix/core.c:
> 
> > @@ -821,3 +823,93 @@ int uv_getrusage(uv_rusage_t\* rusage) {
> > 
> >    return 0;
> >  }
> > +
> > +
> > +int uv__open_cloexec(const char\* path, int flags) {
> > -  int err;
> > -  int fd;
> >   +
> >   +#if defined(**linux**)
> > -  static int no_cloexec;
> >   +
> > -  if (!no_cloexec) {
> > -    fd = open(path, flags | UV__O_CLOEXEC);
> 
> I believe FreeBSD 10 also supports open(O_CLOEXEC) but apart from that LGTM.

Updated. Online man pages say it's supported since version 9, but I 
could only tested in 10, where it does work.

bnoordhuis:
I'll take your word for it. ;-)  LGTM.

saghul:
Thanks for your invaluable help Ben, I really appreciate it. Landed!

PR #1186, opened 2014-03-10
linux: always deregister closing fds from epoll

indutny:
This is a backport of 780d8ad8e58c675dc60f5b73ea22a203760da2f2.

cc @saghul

(I'm going to verify it manually on linux).


saghul:
Patch seems to be the exact same, right? LGTM.

indutny:
Thanks, landed in 84f3059

PR #1187, opened 2014-03-10
kqueue: invalidate fd in uv_fs_event_t

indutny:
Invalidate file descriptor when closing `uv_fs_event_t` handle. Note
that `uv__io_stop` is just removing `fd` from `loop->watchers` and not
actually invalidating all consequent events in a `kevent()` results.

fix joyent/node#1101


indutny:
cc @saghul

saghul:
LGTM in principle. I guess there is no way to test this, right?

indutny:
It is rather complicated and definitely won't be stable.

indutny:
Thanks, landed in 9b38c01b540b60408f8eb1d9c288656405e25f7e

PR #1188, opened 2014-03-10
build: freebsd link with libelf if dtrace enabled

tjfontaine:
as it says on the tin -- add `-lelf` if we're compiling on freebsd in autoconf land


indutny:
LGTM

saghul:
LGTM. Interestingly the gyp build doesn't seem to need this.

saghul:
Landed in https://github.com/joyent/libuv/commit/bbcc0c963c4f2255c24ee9bf84c182f19b32b3a1 :sparkles:

Issue #40, opened 2011-05-17
remove some of these unnecessary status codes and int return values


ry:
e.g on the uv_loop_cb

what else?

piscisaureus:
uv_async_cb has the same useless status code.

On windows uv_prepare|check|idle|timer_init can never fail, I don't know about unix.

On windows ev_timer_start can fail if the btree is in inconsistent state. However I kill the loop if that happens, because it would probably mean that there is a bug in libuv or the user specified a nonexisting handle.

We need to think about what checks uv functions do. E.g. do they check if the handle type is correct, do they check if the handle isn't closing, do they check if the socket is connected when doing a read, does it check whether a req is improperly reused?

saghul:
#1190 landed, closing... :-)

PR #1190, opened 2014-03-11
unix, windows: removed unused status parameter

saghul:
async, timer, prepare, idle and check handles don't need the status parameter.

So, this closes issue #40.

I understand this will require many changes on dependant projects (mine included), but I'd go for it because it's just more correct. Also changes are pretty simple :-)

Now, since we seem to be pretty close to v0.12, we could merge this after, I'd also be ok with that.

/cc @indutny, @piscisaureus, @bnoordhuis

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -405,12 +405,11 @@ typedef void (*uv_shutdown_cb)(uv_shutdown_t* req, int status);
 typedef void (*uv_connection_cb)(uv_stream_t* server, int status);
 typedef void (*uv_close_cb)(uv_handle_t* handle);
 typedef void (*uv_poll_cb)(uv_poll_t* handle, int status, int events);
-typedef void (*uv_timer_cb)(uv_timer_t* handle, int status);
-/* TODO: do these really need a status argument? */
-typedef void (*uv_async_cb)(uv_async_t* handle, int status);
-typedef void (*uv_prepare_cb)(uv_prepare_t* handle, int status);
-typedef void (*uv_check_cb)(uv_check_t* handle, int status);
-typedef void (*uv_idle_cb)(uv_idle_t* handle, int status);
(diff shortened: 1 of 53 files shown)


indutny:
Let's do it after v0.12, I'd like to have some space for ABI trickery :P

saghul:
On 03/11/2014 09:22 AM, Fedor Indutny wrote:

> Let's do it after v0.12, I'd like to have some space for ABI trickery :P
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/joyent/libuv/pull/1190#issuecomment-37272309

Sure, I just wanted to get it out there and get some early feedback.
Look at the branch name ;-)

## 

Saúl Ibarra Corretgé
bettercallsaghul.com

indutny:
Haha, ok. Looks good.

bnoordhuis:
FWIW, I have no qualms about landing it before v0.12 but I don't have a strong opinion either.  The delta between v0.10 and v0.12 is already huge so a few more changes is just another drop in the ocean.  PR LGTM BTW.

saghul:
On 03/11/2014 05:15 PM, Ben Noordhuis wrote:

> FWIW, I have no qualms about landing it before v0.12 but I don't have a
> strong opinion either. The delta between v0.10 and v0.12 is already huge
> so a few more changes is just another drop in the ocean. PR LGTM BTW.

I just wanted to be cautious and not put work in other's shoulders. Now,
I wouldn't mind landing it before 0.12 (in fact I'd love to) and making
the required changes to Node, if @tjfontaine is ok with that.

## 

Saúl Ibarra Corretgé
bettercallsaghul.com

saghul:
Here is the companion node PR: https://github.com/joyent/node/pull/7298

saghul:
Thanks for the comments, it landed! Let there be dragons!

PR #1192, opened 2014-03-12
mingw: fix build with autotools

saghul:
/cc @piscisaureus


saghul:
The problem is with `CommandLineToArgvW`. It's defined in shellapi.h, but in MinGW some base types are not defined unless basetyps.h is included, which is part of the w32api MinGW package.

saghul:
I put the Works on Saghul's Computer (TM) sticker and landed it :-) /me ducks.

PR #1194, opened 2014-03-12
stream: do not leak `alloc_cb` buffers on error

indutny:
Always pass previously allocated (if any) buffer, when calling
`read_cb()` on error.

cc @saghul


indutny:
@saghul: if you have some time, could you please verify that it doesn't break anything on windows?

saghul:
I will, no problem.

saghul:
Changes LGTM by following code paths inside the Windows side. I'll run the tests just in case, but I'm pretty sure you go it right.

indutny:
@saghul cool, this was a bug, right? :)

src/win/pipe.c
```diff
@@ -1505,7 +1505,7 @@ void uv_process_pipe_read_req(uv_loop_t* loop, uv_pipe_t* handle,
           break;
         }
       } else {
-        uv_pipe_read_error_or_eof(loop, handle, GetLastError(), uv_null_buf_);
+        uv_pipe_read_error_or_eof(loop, handle, GetLastError(), &buf);
```
saghul, line 1508: This should be `buf`, it takes the struct by value.

saghul:
Yes, it was a bug, unless memory leaks are a feature in some parallel universe! :-P

Just fix that passing-by-value thing and it's good to go, tested and Works On My Computer (TM).

indutny:
Landed in b597d157791d83ad49130bfec3718231da2e0aef

indutny:
And backported in 4ac8c42

k0dek0mmand0:
Is it possible that the same kind of problem exists in the UDP implementation (at least on Windows)? 

If a datagram is sent from 192.168.0.100:10000 to 192.168.0.100:12345 and no-one is listening on port 12345, an alloc callback is triggered on the sending handle, but the allocated memory is never used anywhere (no read callback comes with this pointer). I'm assuming an ICMP error message (destination unreachable?) is being received, but not handled.

saghul:
@k0dek0mmand0 I checked the code on the Windows side and I only see one possible code path leading there: a WSAECONNRESET or WSAENETRESET error here: https://github.com/joyent/libuv/blob/master/src/win/udp.c#L508. alloc_cb was called but recv_cb won't be. This was introduced here: https://github.com/joyent/libuv/commit/ea63f069877eb20a5847add289b4b959c70db25e

Can you please open a new issue? A test case proving the problem would definitely help.

k0dek0mmand0:
Done. Thanks for the suggestion. 
I'm hoping this isn't an issue that's already been solved since version shipped with node 0.10.30.

saghul:
I don't think so, I checked the code in master.

On Tuesday, August 19, 2014, k0dek0mmand0 notifications@github.com wrote:

> Done. Thanks for the suggestion.
> I'm hoping this isn't an issue that's already been solved since version
> shipped with node 0.10.30.
> 
> —
> https://github.com/joyent/libuv/pull/1194#issuecomment-52566316.

## 

/Saúl
bettercallsaghul.com

PR #1196, opened 2014-03-13
Adding source code of ifaddrs to be available to Android

javihernandez:
The full story is https://github.com/joyent/node/issues/7291.
The original source code of ifaddrs has been taken from https://github.com/kmackay/android-ifaddrs.


saghul:
Hi! Thanks for submitting this! Can you please squash all commits into one and also update the autotools makefile? (Makefile.am)

src/unix/linux-core.c
```diff
@@ -36,21 +36,20 @@
 #include <fcntl.h>
 #include <time.h>
 
-#ifndef __ANDROID__
-#define HAVE_IFADDRS_H 1
```
saghul, line 40: Leave the define, we might add support for a platform which doesn't have it, I guess.

src/unix/linux-core.c
```diff
@@ -716,9 +715,6 @@ void uv_free_cpu_info(uv_cpu_info_t* cpu_infos, int count) {
 
 int uv_interface_addresses(uv_interface_address_t** addresses,
   int* count) {
-#ifndef HAVE_IFADDRS_H
-  return -ENOSYS;
-#else
```
saghul, line 721: Leave this, as you can see above if certain versions of UCLIBC are used this function is not supported.

linusmartensson:
Awesome Javiér! Nice job fixing that up!

javihernandez:
Thanks @linusmartensson, almost there!! :)

@saghul, thanks for your review. I have re-added the HAVE_IFADDRS_H condition as requested, but still have a couple of questions.

First one is about Makefile.am. Maybe I'm wrong, but we would like to have an ANDROID condition in configure.ac in order to include that files according to that condition from Makefile.am. The point is that I'm not sure how I should determine if we're building for Android by using $host_os, how should I proceed? 

The second one is about the failing tests, was it me? Isn't the new code only happening when building for Android?  It's curious that after merging master into my branch I got another extra fail. --- As a brief resume, how can I make these tests bite the dust?

Thanks again!

saghul:
Thanks for your work!

> First one is about Makefile.am. Maybe I'm wrong, but we would like to have an ANDROID condition in configure.ac in order to include that files according to that condition from Makefile.am. The point is that I'm not sure how I should determine if we're building for Android by using $host_os, how should I proceed?

Have a look at configure.ac, I believe just by adding a AM_CONDITIONAL like this:

```
AM_CONDITIONAL([ANDROID],     [AS_CASE([$host_os], [android*],     [true], [false])])
```

Should be enough. The in Makefile.am you can check and add the files as we do with other operating systems.

> The second one is about the failing tests, was it me? Isn't the new code only happening when building for Android? It's curious that after merging master into my branch I got another extra fail. --- As a brief resume, how can I make these tests bite the dust?

We have some failing tests in some platforms, by looking at the output it doesn't look like you introduces any failures :-)

src/unix/linux-core.c
```diff
@@ -715,93 +718,93 @@ void uv_free_cpu_info(uv_cpu_info_t* cpu_infos, int count) {
...
-  struct sockaddr_ll *sll;
-
-  if (getifaddrs(&addrs))
-    return -errno;
-
-  *count = 0;
+  #ifndef HAVE_IFADDRS_H
+    return -ENOSYS;
+  #else
+    int* count) {
```
saghul: Unintended change? This function doesn't actually need to be modified and it's actually broken as is :-)

javihernandez:
Ok

re configure.ac; yes, it was me putting the wrong value in _--host_ when running _configure_ - I was getting nothing in $host_os :/
After having fixed that, I'll wait for your _ok_ to squash all the commits into a single one.

Thanks for all your help :)

saghul:
Great work! LGTM. Once you squash the commits this is ready to land.

saghul:
Thanks Javier! Landed in https://github.com/joyent/libuv/commit/a2506c936296006abe7b8a7517c3f37691191858.

javihernandez:
Thanks Saúl,

closing report in node as well.

Cheers!

PR #1201, opened 2014-03-17
build: check for glibtoolize or libtoolize command

### LINKED ISSUE #1200: glibtoolize command not found when installing using autotools on OSX Mavericks
Using autotools on OSX Mavericks 10.9.2 to install libuv got this error message:

```
$ sh autogen.sh
+ glibtoolize
autogen.sh: line 43: glibtoolize: command not found
```

In [autogen.sh](https://github.com/joyent/libuv/blob/master/autogen.sh), changing [line 20](https://github.com/joyent/libuv/blob/master/autogen.sh#L20) from `LIBTOOLIZE=glibtoolize` to `LIBTOOLIZE=libtoolize` fixed the problem.

(libtool-2.4.2 was installed using `./configure`, `make`, `make install` on OSX Mavericks 10.9.2)

BioticPixels:
glibtool or libtool can be installed on OSX systems.

Fixes #1200.


autogen.sh
```diff
@@ -17,7 +17,14 @@
 cd `dirname "$0"`
 
 if [ "$LIBTOOLIZE" = "" ] && [ "`uname`" = "Darwin" ]; then
-  LIBTOOLIZE=glibtoolize
+    if command -v "glibtoolize" >/dev/null; then
+        LIBTOOLIZE=glibtoolize
+    elif command -v "libtoolize" >/dev/null; then
```
saghul, line 22: I think the code beyond this point is not needed because it would be handled by this: https://github.com/joyent/libuv/blob/master/autogen.sh#L26

BioticPixels:
In reference to https://github.com/joyent/libuv/issues/1200#issue-29533376, the script only functions if `LIBTOOLIZE=libtoolize`. If this is not set, then it causes the error described in #1200.

saghul:
> In reference to #1200 (comment), the script only functions if LIBTOOLIZE=libtoolize. If this is not set, then it causes the error described in #1200.

Yes, but that's because `LIBTOOLIZE` is unconditionally set on OSX. If `glibtoolize` is not found we souldn't do anything, and `libtoolize` will be used, AFAIS.

saghul:
Nobody had any problem with the current approach, so I'm going to close this then. If you manage to get the CLA signed and the patch updated, ping me and we can reopen this PR.

PR #1203, opened 2014-03-18
error: add EMLINK mapping

indutny:
fix joyent/node#7307


saghul:
LGTM.

indutny:
Thanks, landed in 295882e

PR #1206, opened 2014-03-19
unix, windows: add IPv6 support for uv_udp_multicast_interface

saghul:
/cc @indutny, @bnoordhuis


saghul:
Ok, Jenkins, your turn.

saghul:
Sigh, not sure how reliable Jenkins is for this :-/ I manually ran the tests on Linux and Windows and it passes. I'm getting a failure on OSX 10.6, I'll investigate that one...

src/unix/udp.c
```diff
@@ -610,28 +610,33 @@ int uv_udp_set_multicast_loop(uv_udp_t* handle, int on) {
 }
 
 int uv_udp_set_multicast_interface(uv_udp_t* handle, const char* interface_addr) {
-  struct in_addr addr;
-  int err;
+  struct sockaddr_in addr4;
```
indutny: `struct sockaddr_storage addr_st;` and two pointers to it, instead of two structs?

src/win/udp.c
```diff
@@ -660,14 +660,12 @@ int uv_udp_set_membership(uv_udp_t* handle,
 
 
 int uv_udp_set_multicast_interface(uv_udp_t* handle, const char* interface_addr) {
-  struct in_addr addr;
   int err;
+  struct sockaddr_in addr4;
```
indutny: Ditto.

indutny:
Few nits, otherwise LGTM

indutny:
But recheck jenkins output after fixes ;)

saghul:
Will fix, thanks for the review Fedor!

saghul:
Fixed nits, I'll test in on OSX and see what happens...

saghul:
Ok, I know what the issue is on OSX. When we try to set the interface we are using '::', which will result in having a scope id of 0. Something like 'fe80::1%lo0' would end up having s scope id of 1 (on OSX, because whatever comes after the % symbol is the interface name). On Linux setting the multicast interface to the 0 index (sin6_scope_id) works, but on OSX doesn't, it always gives back EADDRNOTAVAIL, which is surprising, because if you do `getaddrinfo` and try the results in turns, '::1' has a scope id of 0. Fun times.

I'll add an if-def specific to OSX.

Note to self: I'm learning a lot more about IPv6 than I originally intended to.

saghul:
@indutny I changed the approach a bit, and fixed the case when you pass NULL, since that should take the default. Can you please review again? Thanks!

saghul:
So, even if the man pages say that in order to reset the interface a 0 should be passed: https://developer.apple.com/library/mac/documentation/Darwin/Reference/Manpages/man4/ip6.4.html

OSX decides to return EADDRNOTAVAIL. Seriously, WAT.

bnoordhuis:
EADDRNOTAVAIL means that the interface doesn't exist or that it isn't multicast-capable/enabled.  I suspect in your case it's the latter.

saghul:
I tried the following: enable multicast on interface with index 1, then try
to reset it by passing a 0 (as per the docs) and I get the error. I should
also mention that if I check the initial state with getsockopt I get a 0
back. So IMHO the behavior OSX is showing here is not kosher.

Or did I miss anything?

Thanks Ben!

bnoordhuis:
No, that sounds about right.  Wouldn't be the first xnu kernel bug we have to paper over in libuv.

saghul:
Rebased. Can you guys please take a look?

saghul:
> No, that sounds about right. Wouldn't be the first xnu kernel bug we have to paper over in libuv.

Do you know where/if that can be reported?

indutny:
LGTM

bnoordhuis:
> Do you know where/if that can be reported?

https://bugreport.apple.com/‎ - but you need an Apple developer account and it's unlikely that you get actual feedback.

bnoordhuis:
Alternatively, you could try the darwin-kernel mailing list - I know at least a handful of Apple kernel engineers read that list.

saghul:
Thanks Ben, I'll give those a shot.

saghul:
I ended up reporting it :-) http://lists.apple.com/archives/darwin-kernel/2014/Mar/msg00012.html

saghul:
Fixed the test and landed it.

PR #1208, opened 2014-03-20
test: fix multicast join test on OSX when using IPv6

saghul:


indutny:
Needs comment that explains things, otherwise LGTM.

saghul:
OK, I'll amend the commit message with an explanation.

Issue #1207, opened 2014-03-20
Assertion `fd > -1' failed


Unitech:
Since node 0.11.12 this errors happens to any pm2 command : 

```
../deps/uv/src/unix/core.c:431: uv__close: Assertion `fd > -1' failed.
```

https://github.com/Unitech/pm2/issues/354

Do you have any clue about this failed assertion ?

Thanks

indutny:
Do you know any way to reproduce it?

indutny:
Also, what OS?

Unitech:
I'm under Ubuntu 12.04, using node 0.11.12.

To reproduce it, just install pm2 `npm install pm2@latest -g` and do a simple `pm2 list` and the error will be thrown

saghul:
Can you please try applying https://github.com/joyent/libuv/pull/1210 ? I tried myself but I'm not sure I tested it correctly, since I couldn't manage to reproduce the issue.

saghul:
Ping

saghul:
Closing, I'm going to assume the issue was fixed with #1210 :-)

PR #1210, opened 2014-03-21
unix: fallback to blocking writes if reopening a tty fails

saghul:
In case reopening the file descriptor fails, fallback to the old
behavior where we do blocking writes, to avoid disrupting other
processes potentially using the file descriptor.

### DIFF
--- a/src/unix/tty.c
+++ b/src/unix/tty.c
@@ -39,6 +39,7 @@ int uv_tty_init(uv_loop_t* loop, uv_tty_t* tty, int fd, int readable) {
   int newfd;
   int r;
 
+  flags = 0;
   newfd = -1;
 
   uv__stream_init(loop, (uv_stream_t*) tty, UV_TTY);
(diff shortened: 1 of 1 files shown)


src/unix/tty.c
```diff
@@ -54,10 +55,16 @@ int uv_tty_init(uv_loop_t* loop, uv_tty_t* tty, int fd, int readable) {
...
    */
   if (isatty(fd)) {
-    newfd = uv__open_cloexec("/dev/tty", O_RDWR);
+    r = uv__open_cloexec("/dev/tty", O_RDWR);
 
-    if (newfd < 0)
-      return newfd;    /* returned value is the error */
+    if (r < 0) {
+      /* fallback to using blocking writes */
+      if (!readable)
```
bnoordhuis, line 62: Why only when `!readable`?  Turning on O_NONBLOCK only for stdin still affects other processes that share that file descriptor (which is probably mapped to stdout and stderr as well.)

indutny:
LGTM, but please run tests!

src/unix/tty.c
indutny, line 62: That's how it was before that `dup` commit.

saghul, line 62: > Why only when !readable? Turning on O_NONBLOCK only for stdin still affects other processes that share that file descriptor (which is probably mapped to stdout and stderr as well.)

It used to be that way before, so I fallback to that. If we can safely do it for readable streams as well I can change that. Maybe it was done like that for consistency with Windows?

saghul:
> LGTM, but please run tests!

Works On My Computer (TM) :-P

On a serious note, I'm not sure how I can simulate the fallfack. Do you?

indutny:
You could try rebuilding node.js with your changes and running that test case, provided by user in https://github.com/joyent/libuv/issues/1207

src/unix/tty.c
saghul, line 62: @bnoordhuis ping :-)

indutny:
Seems to be working fine now! Feel free to land it.

indutny:
Oh, also it has stopped failing :)

indutny:
I mean even without this patch, anyway the change seems to be looking good.

PR #1213, opened 2014-03-25
enable to configure with older versions of autoconf

nori0428:
https://github.com/joyent/libuv/issues/1158
(already closed)

configure.ac at master: AC_PREREQ(2.57)
This means that libuv is enable to configure with autoconf version 2.57 or newer.

But AS_CASE is enable to use autoconf 2.60 or newer version.And AS_CASE has a BUG until version 2.69.
So AC_PREREQ(2.69) might be more better solution.

If you can support older versions of autoconf, please check this patch.
Regards and great thanks for your very useful software.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -16,6 +16,7 @@ AC_PREREQ(2.57)
 AC_INIT([libuv], [0.11.23], [https://github.com/joyent/libuv/issues])
 AC_CONFIG_MACRO_DIR([m4])
 m4_include([m4/libuv-extra-automake-flags.m4])
+m4_include([m4/as_case.m4])
 AM_INIT_AUTOMAKE([-Wall -Werror foreign subdir-objects] UV_EXTRA_AUTOMAKE_FLAGS)
 AC_CANONICAL_HOST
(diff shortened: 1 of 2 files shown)


saghul:
Thanks @nori0428! I'll give it a shot on OSX 10.6, where autoconf is borked.

nori0428:
I'm sorry. I forgot to write LICENCE.
An attached as_case.m4 file is a part of /usr/local/share/autoconf/m4sugar/m4sh.m4.
'Copyright (C) 2000-2012 Free Software Foundation, Inc.'

saghul:
Hi @nori0428, I tested your patch with autoconf 2.63 (version that shipped with OSX 10.6) and it's working great!

Can you please sign the CLA (http://nodejs.org/cla.html) so I can merge your changes? Thanks!

indutny:
@nori0428 ping ;)

saghul:
Checked the CLA, it's signed. Landed in https://github.com/joyent/libuv/commit/1759781c909cad86ea5387d3c4fa2c87b99a5d55 Thanks!

PR #1214, opened 2014-03-25
Support android cross compile on mac.

eddyb:
_Upstreaming @yichoi's patch from @rust-lang [libuv](https://github.com/rust-lang/libuv/tree/800b56fe6af21ffd8e56aee8cf12dd758f5bbdf1)._

### DIFF
--- a/gyp_uv.py
+++ b/gyp_uv.py
@@ -94,6 +94,9 @@ if __name__ == '__main__':
   if not any(a.startswith('-Dcomponent=') for a in args):
     args.append('-Dcomponent=static_library')
 
+  if sys.platform == 'darwin' and '-DOS=android' in sys.argv[1:]:
+    sys.platform='linux'    
+    
   gyp_args = list(args)
   print gyp_args
   run_gyp(gyp_args)


saghul:
HAve you tried using these instructions? https://github.com/joyent/libuv#android

alexcrichton:
I tried using those instructions, but they didn't quite work out for me. I ran into numerous problems running the script, but when I got it configured I get make errors such as https://gist.github.com/alexcrichton/9858498.

I'm not entirely sure why gcc isn't being invoked correctly, but it looks like it's attempting to build with `-arch x86_64` even though the `target_arch` to gyp is `arm`. I don't really fully understand gyp though, so I may be doing something wrong!

saghul:
Do you have these binaries in your toolchain? https://github.com/joyent/libuv/blob/master/android-configure#L11-L14

I'll give it a try myself, but it'll take me a bit...

alexcrichton:
I do indeed, I think that gyp may just be passing through the wrong flags, or expecting ARM on OSX to be for iOS rather than android?

saghul:
Note to self: http://stackoverflow.com/questions/22719843/generate-makefileor-android-mk-for-android-ndk-using-gyp-on-os-x

xudejian:
yeah, it's work for me

abhinavsingh:
nvmind, I see attached patch helps me proceed, thanks.

saghul:
@abhinavsingh the attached patch is a bit of a hack. We need a better way to this, alas I couldn't find the time to dig into it. If you have any idea on how to do it properly, that would be great :-)

bnoordhuis:
I think this can be closed as no longer relevant now that rust has dropped libuv?  Or is it applicable outside of cross-compiling rust?

abhinavsingh:
Ben, I encountered this when I was compiling C code (depending upon libuv)
using NDK build system for Android.

eddyb:
Closing old PRs of mine.

PR #1217, opened 2014-03-26
process: don't close stdio fds during spawn

### LINKED ISSUE #1211: process: don't close stdio fds during spawn
Original Node issue: joyent/node#7331
Simpler testcase for the same issue: https://gist.github.com/tonistiigi/9721405

What is going on here(afaik) is that when stdio fd's are already closed during spawn then they are reused for creating pipes between the processes. Spawn doesn't handle this case, duplicates the pipe endpoint back to stdio and closes the old fd(effectively closing its own stdio).

The regression is at #1083.

One way to fix this would be to only close non-stdio fd's. Others will be closed automatically by `dup`. 42408065

This does not happen in `v0.11` because 359d6678 already forbids ever closing stdio. So the other way would be to get this under `v0.10` as well.

cc @indutny

tonistiigi:
This is needed when closed stdio fd is reused for uv_spawn pipe.
Fixes #1211


test/test-spawn.c
```diff
@@ -676,6 +676,46 @@ static void timer_counter_cb(uv_timer_t* handle, int status) {
...
 
 
+TEST_IMPL(spawn_closed_process_io) {
+  int r;
+  uv_pipe_t in;
+  uv_write_t write_req;
+  uv_buf_t buf;
+  uv_stdio_container_t stdio[2];
+  uv_tty_t stdin;
+  char buffer[] = "hello-from-spawn_stdin";
```
indutny: `static char buffer[]`

test/test-spawn.c
```diff
@@ -676,6 +676,46 @@ static void timer_counter_cb(uv_timer_t* handle, int status) {
...
+
+  r = uv_tty_init(uv_default_loop(), &stdin, 0, 1);
+  ASSERT(r == 0);
+  uv_close((uv_handle_t*)&stdin, close_cb);
```
indutny: `(uv_handle_t*) &stdin`

test/test-spawn.c
```diff
@@ -676,6 +676,46 @@ static void timer_counter_cb(uv_timer_t* handle, int status) {
...
+TEST_IMPL(spawn_closed_process_io) {
+  int r;
+  uv_pipe_t in;
+  uv_write_t write_req;
+  uv_buf_t buf;
+  uv_stdio_container_t stdio[2];
+  uv_tty_t stdin;
+  char buffer[] = "hello-from-spawn_stdin";
+
+  r = uv_tty_init(uv_default_loop(), &stdin, 0, 1);
```
indutny: `ASSERT(0 == uv_tty_init())`

test/test-spawn.c
```diff
@@ -676,6 +676,46 @@ static void timer_counter_cb(uv_timer_t* handle, int status) {
...
+
+  init_process_options("spawn_helper1", exit_cb);
+
+  uv_pipe_init(uv_default_loop(), &in, 0);
+  options.stdio = stdio;
+  options.stdio[0].flags = UV_CREATE_PIPE | UV_READABLE_PIPE;
+  options.stdio[0].data.stream = (uv_stream_t*)&in;
+  options.stdio_count = 1;
+
+  r = uv_spawn(uv_default_loop(), &process, options);
```
indutny: `ASSERT(0 == uv_spawn())` and in the rest of diff.

test/test-spawn.c
```diff
@@ -676,6 +676,46 @@ static void timer_counter_cb(uv_timer_t* handle, int status) {
...
+  r = uv_tty_init(uv_default_loop(), &stdin, 0, 1);
+  ASSERT(r == 0);
+  uv_close((uv_handle_t*)&stdin, close_cb);
+
+  init_process_options("spawn_helper1", exit_cb);
+
+  uv_pipe_init(uv_default_loop(), &in, 0);
+  options.stdio = stdio;
+  options.stdio[0].flags = UV_CREATE_PIPE | UV_READABLE_PIPE;
+  options.stdio[0].data.stream = (uv_stream_t*)&in;
```
indutny: `(uv_stream_t*) &in` - and in the rest of diff.

test/test-spawn.c
```diff
@@ -676,6 +676,46 @@ static void timer_counter_cb(uv_timer_t* handle, int status) {
...
+  options.stdio_count = 1;
+
+  r = uv_spawn(uv_default_loop(), &process, options);
+  ASSERT(r == 0);
+
+  buf.base = buffer;
+  buf.len = sizeof(buffer);
+  r = uv_write(&write_req, (uv_stream_t*)&in, &buf, 1, write_cb);
```
indutny: ditto.

test/test-spawn.c
indutny: Also, I don't think that this test will work with libuv 0.11. `uv_close()` won't `close()` file descriptors that are less than 3. Perhaps we could use higher fd here?

indutny:
Looks good, but needs some fixes. Thank you! :)

test/test-spawn.c
saghul: Maybe just do a `close(0)` instead?

indutny: Or that.

tonistiigi:
Style issues are fixed and `uv_close` is replaced with `close`. Now the test works on both `v0.10` and `v0.11`. 

Should I be worried about Jenkins? Local tests pass fine on OSX and Linux x64.

Also, this doesn't apply cleanly to master. Not sure how you deal with that. Should I push another version in separate branch?

test/test-spawn.c
```diff
@@ -676,6 +676,41 @@ static void timer_counter_cb(uv_timer_t* handle, int status) {
...
 
+TEST_IMPL(spawn_closed_process_io) {
+  uv_pipe_t in;
+  uv_write_t write_req;
+  uv_buf_t buf;
+  uv_stdio_container_t stdio[2];
+  uv_tty_t stdin;
+  static char buffer[] = "hello-from-spawn_stdin";
+
+  ASSERT(0 == uv_tty_init(uv_default_loop(), &stdin, 0, 1));
```
indutny: Do you still need this line and `stdin` declaration?

indutny:
@trentm thanks man! Looks very good. I'll handle merge issues myself, please fix the last nit! ;)

test/test-spawn.c
tonistiigi: I couldn't get it to fail without it(at least on OSX). I could debug to see more precisely whats going on there.

indutny: Yes, that would be great. Thank you.

tonistiigi:
Stdin initialization is removed now. Had to move the `close()` call right before spawn so the fd wouldn't get reused in `uv_pipe_init()` and `uv_default_loop()`.

indutny:
Thank you, landed in fd77a5d!

PR #1219, opened 2014-03-28
unix: add UV_HANDLE_IPV6 flag to tcp and udp handles

saghul:
Aligns behavior with Windows. Currently the flag is used to pick the default multicast interface when resetting it.

/cc @indutny, @bnoordhuis


src/unix/tcp.c
```diff
@@ -89,8 +89,11 @@ int uv__tcp_bind(uv_tcp_t* tcp,
   errno = 0;
   if (bind(tcp->io_watcher.fd, addr, addrlen) && errno != EADDRINUSE)
     return -errno;
-
```
indutny, line 92: No need to change whitespace.

src/unix/udp.c
```diff
@@ -619,8 +622,13 @@ int uv_udp_set_multicast_interface(uv_udp_t* handle, const char* interface_addr)
 
   if (!interface_addr) {
     memset(&addr_st, 0, sizeof addr_st);
-    addr_st.ss_family = AF_INET;
-    addr4->sin_addr.s_addr = htonl(INADDR_ANY);
+    if (!(handle->flags & UV_HANDLE_IPV6)) {
```
indutny: May be inverse the condition? I doubt that `!` is really needed here.

indutny: Also seems like a behavior change, does it need any test?

indutny:
Looks good, see comments.

src/unix/udp.c
saghul: I'd rather write it like this because we always do IPv4 first and IPv6 after. There is no behavior "change" as setting the default interface for IPv6 never worked before this patch :-)

src/unix/tcp.c
saghul, line 92: I thought I'd group the errono handlign statements there and add my condition afterwards, but I can revert that part.

indutny, line 92: oh, ok then.

src/unix/udp.c
indutny: Could you add a test for default interface then? ;)

saghul: Won't work on OSX, unfortunately, but I can write one.

indutny: You could ifdef it if needed, that would be fine.

src/unix/tcp.c
```diff
@@ -89,8 +89,11 @@ int uv__tcp_bind(uv_tcp_t* tcp,
...
   tcp->delayed_error = -errno;
+
+  if (addr->sa_family == AF_INET6)
+    tcp->flags |= UV_HANDLE_IPV6;
```
bnoordhuis, line 95: Here's an edge case to consider: what if I bind the the socket to a IPv4 address first, then connect to a IPv6 address (or vice versa)?  Should UV_HANDLE_IPV6 be set or not?

src/unix/udp.c
bnoordhuis: > I'd rather write it like this because we always do IPv4 first and IPv6 after.

The counter-argument is that negation requires more mental effort to parse.  Just a tiny bit more, of course, but tiny bits add up.

src/unix/tcp.c
saghul, line 95: I'd say no. The meaning the flag currently has is basically "did we create the socket with AF_INET6 or not?". Right now the socket is always created right before binding, should that change in the future, we might need to revisit.

saghul:
I rewrote the flag test (looked better once I wrote it that way :-)) and updated the multicast test to use NULL, that is, the default interface. (except on OSX, because it doesn't work)

saghul:
No (new) failing tests :-)

saghul:
Any further comments?

bnoordhuis:
None from me.

saghul:
Kewl, thanks for confirming, Ben!

saghul:
Landed in https://github.com/joyent/libuv/compare/3558d65d2f17...7a4c42a0548b

PR #1220, opened 2014-03-28
fs: avoid using readv/writev where possible

indutny:
The VirtualBox shared file system doesn't seem to be capable of
`readv()` syscall. Since we don't need them in the most of the cases
anyway, it won't hurt to just use `read()` and `write()` when we are
reading just one buffer.

see https://github.com/joyent/node/issues/7343 for details.


indutny:
cc @saghul

saghul:
LGTM, assuming it actually fixes the issue reported by the user.

indutny:
Verified by me, landed in b48ca33

PR #1222, opened 2014-03-31
unix, windows: do not set SO_REUSEADDR by default on udp

saghul:
Add UV_UDP_REUSEADDR flag instead, which can be passed to uv_udp_bind.
If the udp handle is unbound when uv_udp_set_memberhsip or
uv_udp_set_multicast_interface is called, the handle will be bound with
UV_UDP_REUSEADDR set.

Moar love for UDP. /cc @bnoordhuis, @indutny, @piscisaureus

Hello https://github.com/joyent/node/issues/5425

PS: I'll ran the test suite on Windows tonight, so far tests pass on my Linux machine, lets see what you got Jenkins!


src/unix/udp.c
```diff
@@ -547,15 +553,23 @@ int uv_udp_set_membership(uv_udp_t* handle,
...
 
-  if (uv_ip4_addr(multicast_addr, 0, &addr4) == 0)
+  if (uv_ip4_addr(multicast_addr, 0, &addr4) == 0) {
+    err = uv__udp_maybe_deferred_bind(handle, AF_INET, UV_UDP_REUSEADDR);
+    if (err)
+      return err;
     return uv__udp_set_membership4(handle, &addr4, interface_addr, membership);
-  else if (uv_ip6_addr(multicast_addr, 0, &addr6) == 0)
+  } else if (uv_ip6_addr(multicast_addr, 0, &addr6) == 0) {
+    err = uv__udp_maybe_deferred_bind(handle, AF_INET6, UV_UDP_REUSEADDR);
```
indutny, line 566: Why reusing it here?

src/win/udp.c
```diff
@@ -168,13 +159,17 @@ void uv_udp_endgame(uv_loop_t* loop, uv_udp_t* handle) {
...
+static int uv_udp_maybe_bind(uv_udp_t* handle,
+                             const struct sockaddr* addr,
+                             unsigned int addrlen,
+                             unsigned int flags) {
   int r;
   int err;
   DWORD no = 0;
+  DWORD yes = 1;
+
+  if (handle->flags & UV_HANDLE_BOUND)
```
indutny, line 170: Is it the same on unix side? Does it always return zero on both unix and windows when already bound?

src/win/udp.c
```diff
@@ -193,6 +188,19 @@ static int uv_udp_try_bind(uv_udp_t* handle,
       return err;
     }
 
+    if (flags & UV_UDP_REUSEADDR) {
```
indutny, line 190: Move `yes` here.

src/unix/udp.c
saghul, line 566: Because you need it for multicast if you are bound to "any".

test/test-watcher-cross-stop.c
```diff
@@ -73,7 +73,9 @@ static void close_cb(uv_handle_t* handle) {
 
   for (i = 0; i < ARRAY_SIZE(sockets); i++) {
     ASSERT(0 == uv_udp_init(loop, &sockets[i]));
-    ASSERT(0 == uv_udp_bind(&sockets[i], (const struct sockaddr*) &addr, 0));
+    ASSERT(0 == uv_udp_bind(&sockets[i],
+    			    (const struct sockaddr*)
```
indutny: Indent.

src/win/udp.c
```diff
@@ -704,6 +702,12 @@ int uv_udp_set_multicast_interface(uv_udp_t* handle, const char* interface_addr)
       return uv_translate_sys_error(WSAGetLastError());
     }
   } else if (addr_st.ss_family == AF_INET6) {
+    err = uv_udp_maybe_bind(handle,
+			    (const struct sockaddr*) &uv_addr_ip6_any_,
```
indutny: Indent.

src/win/udp.c
```diff
@@ -696,6 +688,12 @@ int uv_udp_set_multicast_interface(uv_udp_t* handle, const char* interface_addr)
   }
 
   if (addr_st.ss_family == AF_INET) {
+    err = uv_udp_maybe_bind(handle,
+			    (const struct sockaddr*) &uv_addr_ip4_any_,
```
indutny: Indent.

src/win/udp.c
saghul, line 170: No, unix doesn't have UV_HANDLE_BOUND, we check for fd being -1.

src/unix/udp.c
indutny, line 566: Ah, actually, makes sense. :)

indutny:
Some style issues, otherwise LGTM

src/win/udp.c
saghul, line 190: You mean the declaration?

indutny, line 190: Yes.

saghul, line 190: Didn't we do declarations at the top?

indutny, line 190: At the top of the block ;)

saghul:
Fixed style nits.

saghul:
Thanks for having a look @indutny!

saghul:
Tests seem to pass, also on Windows. I'll double check FreeBSD though, the Jenkins environment doesn't seem to work so well with regards to IPv6.

saghul:
Things are fine on FreeBSD as well. There are some unrelated issues which I'll address as sepparate PRs.

@bnoordhuis, any chance you could take a look? Thanks!

include/uv.h
```diff
@@ -921,18 +929,11 @@ struct uv_udp_send_s {
...
  * Returns:
  *  0 on success, or an error code < 0 on failure.
- *
- * This sets the SO_REUSEPORT socket flag on the BSDs and OS X. On other
- * UNIX platforms, it sets the SO_REUSEADDR flag.  What that means is that
- * multiple threads or processes can bind to the same address without error
- * (provided they all set the flag) but only the last one to bind will receive
- * any traffic, in effect "stealing" the port from the previous listener.
- * This behavior is something of an anomaly and may be replaced by an explicit
- * opt-in mechanism in future versions of libuv.
```
bnoordhuis, line 935: I was quite the forward thinker there.

test/test-watcher-cross-stop.c
```diff
@@ -73,7 +73,9 @@ static void close_cb(uv_handle_t* handle) {
 
   for (i = 0; i < ARRAY_SIZE(sockets); i++) {
     ASSERT(0 == uv_udp_init(loop, &sockets[i]));
-    ASSERT(0 == uv_udp_bind(&sockets[i], (const struct sockaddr*) &addr, 0));
+    ASSERT(0 == uv_udp_bind(&sockets[i],
+                            (const struct sockaddr*)
+                            &addr, UV_UDP_REUSEADDR));
```
bnoordhuis: Style: the `&addr` should go on the previous line.

bnoordhuis:
LGTM, both the Windows and the UNIX side.

include/uv.h
saghul, line 935: Indeed you were!

test/test-watcher-cross-stop.c
saghul: Ah, thanks for catching that. Will fix when GitHub decides to work:

```
fatal: protocol error: expected sha/ref, got 'GitHub SSH access is temporarily unavailable'
```

saghul:
Thanks for the help guys, landed in https://github.com/joyent/libuv/commit/3558d65d2f173ff7a8148d340a31741aaa5b35f9

PR #1223, opened 2014-03-31
windows: fix opening of read-only stdin pipes

orangemocha:
Fix https://github.com/joyent/node/issues/7345

Google Chrome is launching native messaging hosts by invoking cmd.exe
with input/output redirected from/to named pipes. The host ends up
with a read-only handle to the stdin pipe. This is causing
SetNamedPipeHandleState to fail.

### DIFF
--- a/src/win/pipe.c
+++ b/src/win/pipe.c
@@ -229,14 +229,35 @@ static int uv_set_pipe_handle(uv_loop_t* loop, uv_pipe_t* handle,
   IO_STATUS_BLOCK io_status;
   FILE_MODE_INFORMATION mode_info;
   DWORD mode = PIPE_TYPE_BYTE | PIPE_READMODE_BYTE | PIPE_WAIT;
-
+  DWORD current_mode = 0;
+  DWORD err = 0;
+  
   if (!SetNamedPipeHandleState(pipeHandle, &mode, NULL, NULL)) {
-    /* If this returns ERROR_INVALID_PARAMETER we probably opened something */
-    /* that is not a pipe. */
-    if (GetLastError() == ERROR_INVALID_PARAMETER) {
-      SetLastError(WSAENOTSOCK);
+    err = GetLastError();
+    if (err == ERROR_ACCESS_DENIED) {
+      /* 
+       * SetNamedPipeHandleState can fail if the handle doesn't have either
+       * GENERIC_WRITE  or FILE_WRITE_ATTRIBUTES.
+       * But if the handle already has the desired wait and blocking modes
+       * we can continue.
+       */
+      if (!GetNamedPipeHandleState(pipeHandle, &current_mode, NULL, NULL,
+                                   NULL, NULL, 0)) {
+        return -1;
+      } else if (current_mode != mode) {
+        SetLastError(ERROR_ACCESS_DENIED);
+        return -1;
+      } else {
+        duplex_flags &= ~UV_HANDLE_WRITABLE;
+      }
+    } else {
+      /* If this returns ERROR_INVALID_PARAMETER we probably opened */
+      /* something that is not a pipe. */
+      if (err == ERROR_INVALID_PARAMETER) {
+        SetLastError(WSAENOTSOCK);
+      }
+      return -1;
     }
-    return -1;
   }
 
   /* Check if the pipe was created with FILE_FLAG_OVERLAPPED. */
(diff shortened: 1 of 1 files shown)


src/win/pipe.c
```diff
@@ -231,12 +231,28 @@ static int uv_set_pipe_handle(uv_loop_t* loop, uv_pipe_t* handle,
   DWORD mode = PIPE_TYPE_BYTE | PIPE_READMODE_BYTE | PIPE_WAIT;
 
   if (!SetNamedPipeHandleState(pipeHandle, &mode, NULL, NULL)) {
-    /* If this returns ERROR_INVALID_PARAMETER we probably opened something */
-    /* that is not a pipe. */
-    if (GetLastError() == ERROR_INVALID_PARAMETER) {
-      SetLastError(WSAENOTSOCK);
+    DWORD err = GetLastError();
```
saghul: Please define it at the top of the function.

src/win/pipe.c
```diff
@@ -231,12 +231,28 @@ static int uv_set_pipe_handle(uv_loop_t* loop, uv_pipe_t* handle,
...
+    if (err == ERROR_ACCESS_DENIED) {
+      /* SetNamedPipeHandleState can fail if the handle doesn't have either */
+      /* GENERIC_WRITE  or FILE_WRITE_ATTRIBUTES. */
+      /* But if the handle already has the desired wait and blocking modes */
+      /* we can continue. */
+      DWORD current_mode = 0;
```
saghul: Ditto.

orangemocha:
Updated, thanks.

saghul:
Would `uv_is_writable` return false, or do we need to reset the writable flag? Also, any chance this could have a test?

Code looks ok to me, my Windows kung-fu is pretty limited though :-/

orangemocha:
Updated to reset writable flag. Added a test to node.js here: https://github.com/joyent/node/pull/7433

src/win/pipe.c
```diff
@@ -1749,15 +1768,14 @@ static void eof_timer_close_cb(uv_handle_t* handle) {
...
 
   if (os_handle == INVALID_HANDLE_VALUE ||
-      uv_set_pipe_handle(pipe->loop, pipe, os_handle, 0) == -1) {
+      uv_set_pipe_handle(pipe->loop, pipe, os_handle, duplex_flags) == -1) {
     return UV_EINVAL;
   }
 
   uv_pipe_connection_init(pipe);
-  pipe->handle = os_handle;
-  pipe->flags |= UV_HANDLE_READABLE | UV_HANDLE_WRITABLE;
```
saghul, line 1760: Where are the flags and handle set in the pipe now? I can't see it in the diff.

orangemocha, line 1760: See line 1771:
-  DWORD duplex_flags = UV_HANDLE_READABLE | UV_HANDLE_WRITABLE;

uv_set_pipe_handle is setting the flags, clearing UV_HANDLE_WRITABLE should the pipe be read-only.

saghul, line 1760: Yes, but where do we save those flags into pipe->flags now?

orangemocha, line 1760: https://github.com/joyent/libuv/blob/master/src/win/pipe.c#L267

saghul, line 1760: Ah, I see now, sorry.

saghul:
LGTM, though I'd like @piscisaureus's blessing :-)

src/win/pipe.c
```diff
@@ -229,14 +229,33 @@ static int uv_set_pipe_handle(uv_loop_t* loop, uv_pipe_t* handle,
...
+  DWORD err = 0;
+  
   if (!SetNamedPipeHandleState(pipeHandle, &mode, NULL, NULL)) {
-    /* If this returns ERROR_INVALID_PARAMETER we probably opened something */
-    /* that is not a pipe. */
-    if (GetLastError() == ERROR_INVALID_PARAMETER) {
-      SetLastError(WSAENOTSOCK);
+    err = GetLastError();
+    if (err == ERROR_ACCESS_DENIED) {
+      /* SetNamedPipeHandleState can fail if the handle doesn't have either */
```
indutny: I know it is lame, but could you please make it a single block comment?

src/win/pipe.c
```diff
@@ -229,14 +229,33 @@ static int uv_set_pipe_handle(uv_loop_t* loop, uv_pipe_t* handle,
...
+      /* GENERIC_WRITE  or FILE_WRITE_ATTRIBUTES. */
+      /* But if the handle already has the desired wait and blocking modes */
+      /* we can continue. */
+      if (!GetNamedPipeHandleState(pipeHandle, &current_mode, NULL, NULL,
+          NULL, NULL, 0)) {
```
indutny: Please align NULL with paren after `GetNamedPipeHandleState`.

indutny:
LGTM

orangemocha:
addressed style issues

saghul:
Fixed a couple of style nits since I was there and landed it as https://github.com/joyent/libuv/commit/ba47e68824dbd00c51d2c659ab05a83989283aab

Thanks @orangemocha!

PR #1224, opened 2014-03-31
windows: Fix an infinite loop in uv_spawn

### LINKED ISSUE #909: windows: PATH=";.;" causes uv_spawn to infinitely loop
Using this code:

``` c
#include "uv.h"
#include <stdio.h>

uv_loop_t *loop;
uv_process_t child_req;
uv_process_options_t options;

void on_exit(uv_process_t *p, int exit_status, int term_status) {
    printf("%p %d %d\n", p, exit_status, term_status);
}

int main() {
    loop = uv_default_loop();

    char* args[3];
    args[0] = "mkdir";
    args[1] = "test-dir";
    args[2] = NULL;

    options.exit_cb = on_exit;
    options.file = "mkdir";
    options.args = args;

    int ret;
    if ((ret = uv_spawn(loop, &child_req, options)) != 0) {
        fprintf(stderr, "%s\n", uv_strerror(ret));
        return 1;
    }

    return uv_run(loop, UV_RUN_DEFAULT);
}
```

The code will infinitely loop on windows when run with `PATH=";.;"`. I ran this in mingw, but it may be a problem elsewhere. We found this when working on rust, and @vadimcn fixed the issue with this patch:

``` patch
diff --git a/src/win/process.c b/src/win/process.c
index 477e5e4..e2eb3a3 100644
--- a/src/win/process.c
+++ b/src/win/process.c
@@ -374,15 +374,6 @@ static WCHAR* search_path(const WCHAR *file,
                                   name_has_ext);

     while (result == NULL) {
-      if (*dir_end == L'\0') {
-        break;
-      }
-
-      /* Skip the separator that dir_end now points to */
-      if (dir_end != path) {
-        dir_end++;
-      }
-
       /* Next slice starts just after where the previous one ended */
       dir_start = dir_end;

@@ -393,27 +384,33 @@ static WCHAR* search_path(const WCHAR *file,
       }

       /* If the slice is zero-length, don't bother */
-      if (dir_end - dir_start == 0) {
-        continue;
+      if (dir_end - dir_start > 0) {
+        dir_path = dir_start;
+        dir_len = dir_end - dir_start;
+
+        /* Adjust if the path is quoted. */
+        if (dir_path[0] == '"' || dir_path[0] == '\'') {
+          ++dir_path;
+          --dir_len;
+        }
+
+        if (dir_path[dir_len - 1] == '"' || dir_path[dir_len - 1] == '\'') {
+          --dir_len;
+        }
+
+        result = path_search_walk_ext(dir_path, dir_len,
+                                      file, file_len,
+                                      cwd, cwd_len,
+                                      name_has_ext);
       }

-      dir_path = dir_start;
-      dir_len = dir_end - dir_start;
-
-      /* Adjust if the path is quoted. */
-      if (dir_path[0] == '"' || dir_path[0] == '\'') {
-        ++dir_path;
-        --dir_len;
-      }
-
-      if (dir_path[dir_len - 1] == '"' || dir_path[dir_len - 1] == '\'') {
-        --dir_len;
+      /* Stop if reached end of the line */
+      if (*dir_end == L'\0') {
+        break;
       }
-
-      result = path_search_walk_ext(dir_path, dir_len,
-                                    file, file_len,
-                                    cwd, cwd_len,
-                                    name_has_ext);
+      
+      /* Skip the separator that dir_end now points to */
+      dir_end++;
     }
   }

```

I haven't looked too much into it, but I figured that I should report it here as well to make sure you guys are aware of it!

alexcrichton:
The PATH-parsing code for windows erroneously contained an infinite
loop when the PATH started with a leading semicolon. Each iteration of
the loop usually bumped over the separator, but if the first character
was a semicolon then it would never skip it, causing the infinite
loop.

Closes #909

### DIFF
--- a/test/test-spawn.c
+++ b/test/test-spawn.c
@@ -935,6 +935,28 @@ TEST_IMPL(environment_creation) {
 
   return 0;
 }
+
+// Regression test for issue #909
+TEST_IMPL(spawn_with_an_odd_path) {
+  int r;
+
+  char newpath[2048];
+  char *path = getenv("PATH");
+  ASSERT(path != NULL);
+  snprintf(newpath, 2048, ";.;%s", path);
+  SetEnvironmentVariable("PATH", path);
+
+  init_process_options("", exit_cb);
+  options.file = options.args[0] = "program-that-had-better-not-exist";
+  r = uv_spawn(uv_default_loop(), &process, &options);
+  ASSERT(r == UV_ENOENT || r == UV_EACCES);
+  ASSERT(0 == uv_is_active((uv_handle_t*) &process));
+  uv_close((uv_handle_t*) &process, NULL);
+  ASSERT(0 == uv_run(uv_default_loop(), UV_RUN_DEFAULT));
+
+  MAKE_VALGRIND_HAPPY();
+  return 0;
+}
 #endif
 
 #ifndef _WIN32
--- a/test/test-spawn.c
+++ b/test/test-spawn.c
@@ -171,6 +171,7 @@ TEST_IMPL(spawn_fails) {
   r = uv_spawn(uv_default_loop(), &process, &options);
   ASSERT(r == UV_ENOENT || r == UV_EACCES);
   ASSERT(0 == uv_is_active((uv_handle_t*) &process));
+  uv_close((uv_handle_t*) &process, NULL);
   ASSERT(0 == uv_run(uv_default_loop(), UV_RUN_DEFAULT));
 
   MAKE_VALGRIND_HAPPY();
--- a/src/win/process.c
+++ b/src/win/process.c
@@ -378,7 +378,7 @@ static WCHAR* search_path(const WCHAR *file,
       }
 
       /* Skip the separator that dir_end now points to */
-      if (dir_end != path) {
+      if (dir_end != path || *path == L';') {
         dir_end++;
       }
 
--- a/test/test-list.h
+++ b/test/test-list.h
@@ -244,6 +244,7 @@ TEST_DECLARE   (environment_creation)
 TEST_DECLARE   (listen_with_simultaneous_accepts)
 TEST_DECLARE   (listen_no_simultaneous_accepts)
 TEST_DECLARE   (fs_stat_root)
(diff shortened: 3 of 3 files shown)


alexcrichton:
I remember signing the CLA awhile ago, did it change recently, or do I need to renew it? I certainly don't mind signing it again!

saghul:
@alexcrichton the jenkins bot is acting weird, you don't need to renew the CLA.

alexcrichton:
ping?

saghul:
Can we have (windows only) test for this? The fix looks simple, but as Bert mentioned earlier some other changes might be needed in the path parsing code (for other reasons) so it would help catch regressions.

alexcrichton:
Good point! I've added a test.

test/test-spawn.c
```diff
@@ -935,6 +935,27 @@ TEST_IMPL(environment_creation) {
...
+  ASSERT(path != NULL);
+  snprintf(newpath, 2048, ";.;%s", path);
+  SetEnvironmentVariable("PATH", path);
+
+  init_process_options("", exit_cb);
+  options.file = options.args[0] = "program-that-had-better-not-exist";
+  r = uv_spawn(uv_default_loop(), &process, &options);
+  ASSERT(r == UV_ENOENT || r == UV_EACCES);
+  ASSERT(0 == uv_is_active((uv_handle_t*) &process));
+  ASSERT(0 == uv_run(uv_default_loop(), UV_RUN_DEFAULT));
```
saghul, line 956: Close the handle before calling uv_run here, otherwise we'd get an assert in debug mode IIRC, because the loop cannot be destroyed.

alexcrichton, line 956: Should I also update the `spawn_fails` test above? (I just copy/pasted from there)

saghul, line 956: Yeah. Please do it as a separated commit.

alexcrichton, line 956: Updated!

saghul:
LGTM.

saghul:
Landed, thanks Alex!

PR #1227, opened 2014-04-01
bsd: support IPv6 qualified link-local addresses

saghul:
AFAICT (online man pages) all BSDs we support have `if_nametoindex` in net/if.h, so use it. Tested on FreeBSD 10.

/cc @indutny @bnoordhuis

### DIFF
--- a/include/uv-bsd.h
+++ b/include/uv-bsd.h
@@ -31,4 +31,6 @@
 
 #define UV_HAVE_KQUEUE 1
 
+#define UV_PLATFORM_HAS_IP6_LINK_LOCAL_ADDRESS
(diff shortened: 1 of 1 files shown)


bnoordhuis:
LGTM

saghul:
Landed! Thanks Ben!

tjfontaine:
sunos also `if_nametoindex` http://illumos.org/man/3xnet/if_nametoindex

saghul:
Thanks for the heads up, I'll push a fix!0

saghul:
https://github.com/joyent/libuv/pull/1229

PR #1229, opened 2014-04-07
sunos: support IPv6 qualified link-local addresses

saghul:
/cc @tjfontaine

AFAIS all SunOS variants seem to support it.


saghul:
Hum, we have now added this flag to all systems now. Maybe we can just get rid of it? Thoughts @indutny? I don't know of any system libuv supports which doesn't have `if_nametoindex`, is there any?

PR #1231, opened 2014-04-09
linux: reduce file descriptor count of async pipe

bnoordhuis:
Reopen one of the pipe descriptors as read/write through the procfs.
Allows us to close the original pipe file descriptors, saving a file
descriptor on kernels that don't support eventfd(2).


indutny:
I thought you had no interest in libuv ;) Will look into it soon.

saghul:
Hi Ben, thanks for the PR! Can you point me to some docs on why/how this (the 'special' fds) works? My google-fu is not helping :-/

indutny:
@saghul because it is the same inode, essentially. Looking good.

indutny:
@saghul if you are ok with it - please land it or at least let me know. ;)

indutny:
Thank you, landed in 86831fe99c625f3a8180c02e628758eb25a0bd3b

PR #1232, opened 2014-04-10
windows: console signal handler, add refs not correct

chopdown:
If the console control handler has already been hooked, add a reference.
pool english, sorry.


saghul:
LGTM.

saghul:
Thanks! Landed in https://github.com/joyent/libuv/commit/0c726e76a09efb4b27b0b2ca18b489053c1b50ad

PR #1233, opened 2014-04-10
test: fix EPIPE race in spawn_closed_process_io

indutny:
The just created stdin should not be closed by a child process before
writing to it, otherwise EPIPE happens.


saghul:
Landed!

PR #1234, opened 2014-04-13
inet: allow scope in `uv_inet_ntop` ip6 check

indutny:
see https://github.com/joyent/node/issues/7395

@saghul, basically we already support link local interface name/scope in `uv_ip6_addr`, but `uv_inet_ntop` does not recognize it. With this patch - it will not fail on encountering it.

There are two questions about it:
- Is it a new feature or a bug fix? (I think the latter one)
- Should it fail if `UV_PLATFORM_HAS_IP6_LINK_LOCAL_ADDRESS` is not present

@tjfontaine: are you fine with this in node v0.10 ?

### DIFF
--- a/src/inet.c
+++ b/src/inet.c
@@ -229,7 +229,7 @@ static uv_err_t inet_pton6(const char *src, unsigned char *dst) {
   curtok = src;
   seen_xdigits = 0;
   val = 0;
-  while ((ch = *src++) != '\0') {
+  while ((ch = *src++) != '\0' && ch != '%') {
     const char *pch;
 
     if ((pch = strchr((xdigits = xdigits_l), ch)) == NULL)
@@ -260,7 +260,18 @@ static uv_err_t inet_pton6(const char *src, unsigned char *dst) {
       continue;
     }
     if (ch == '.' && ((tp + sizeof(struct in_addr)) <= endp)) {
-      uv_err_t err = inet_pton4(curtok, tp);
+      uv_err_t err;
+
+      /* Scope id present, parse ipv4 addr without it */
+      pch = strchr(curtok, '%');
+      if (pch != NULL) {
+        char tmp[sizeof "255.255.255.255"];
+
+        memcpy(tmp, curtok, pch - curtok);
+        curtok = tmp;
+        src = pch;
+      }
+      err = inet_pton4(curtok, tp);
       if (err.code == 0) {
         tp += sizeof(struct in_addr);
         seen_xdigits = 0;
--- a/build.mk
+++ b/build.mk
@@ -87,6 +87,7 @@ TESTS= \
 	test/test-getsockname.o \
 	test/test-hrtime.o \
 	test/test-idle.o \
+	test/test-ip6-addr.o \
 	test/test-ipc.o \
 	test/test-ipc-send-recv.o \
 	test/test-loop-handles.o \
--- a/checksparse.sh
+++ b/checksparse.sh
@@ -107,6 +107,7 @@ test/test-getaddrinfo.c
 test/test-getsockname.c
 test/test-hrtime.c
 test/test-idle.c
+test/test-ip6-addr.c
 test/test-ipc-send-recv.c
 test/test-ipc.c
 test/test-loop-handles.c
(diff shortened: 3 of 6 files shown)


saghul:
The title really confused me :-/The patch changes inet_pton6, but you mentined ntop. Also, I'd say this is more like 'skipping' the scope id, since inet_pton6 doesn't really process it.

saghul:
The test doesn't actually test what this patch fixes :-) Please add some tests such as:

```
"fe80:0:0:0:2acf:daff:1.2.3.4%lo0"
"fe80:0:0:0:2acf:daff%lo0"
"fe80:0:0:0:2acf:daff%%%%"
"fe80:0:0:0:2acf:daff%lo0:1.2.3.4"
```

indutny:
Ooops, the title is really confusing, thanks! I'll look into your tests soon. But I think `%lo0:1.2.3.4` should be totally fine, right? and `%%%%` too.

indutny:
@saghul force pushed

src/inet.c
```diff
@@ -260,7 +260,18 @@ static uv_err_t inet_pton6(const char *src, unsigned char *dst) {
       continue;
     }
     if (ch == '.' && ((tp + sizeof(struct in_addr)) <= endp)) {
-      uv_err_t err = inet_pton4(curtok, tp);
+      uv_err_t err;
+
+      /* Scope id present, parse ipv4 addr without it */
+      pch = strchr(curtok, '%');
+      if (pch != NULL) {
+        char tmp[sizeof "255.255.255.255"];
```
saghul, line 268: Use INET_ADDRSTRLEN?

indutny, line 268: Copied it from above, actually. Is it present on windows?

saghul, line 268: Ah, ok, then leave it :-)

saghul:
LGTM.

indutny:
Now, the only thing that blocks it is a @tjfontaine word on it. @tjfontaine in my opinion this is a bug fix and should be landed in both v0.10 node and libuv.

saghul:
I agree, this is a bugfix. Also, no need to check UV_PLATFORM_HAS_IP6_LINK_LOCAL_ADDRESS (I'm even considering removing it).

indutny:
Ok, considering that we have consensus here and @tjfontaine is usually unavailable, I will land it in v0.10 libuv. We could always revert it before next release, anyway.

indutny:
Also look at the issue number! :)

indutny:
Thank you, landed in d30e3ab

src/inet.c
saghul, line 268: I'm having second thoughts about this part. I don't think it's valid to have a scope id in a IPv4-mapped address, since the scope id only needs to be specified for link local addresses. IMHO we should revert this part and adjust tests accordingly.

indutny, line 268: Link local address starts with `fe80::`, and every link local address should have it's scope. The way you specify the rest of the address should not affect it, IMHO.

saghul, line 268: What I mean is that fe80:some_valid_thing:1.2.3.4%lo0 is not a valid IPv6 address and this thing would say it is, AFAIS.

indutny, line 268: Ok, makes sense. We should also backport fd45f876, btw.

I'll submit a PR in a minute.

saghul, line 268: On 4/14/14 1:06 PM, Fedor Indutny wrote:

> In src/inet.c:
> 
> > @@ -260,7 +260,18 @@ static uv_err_t inet_pton6(const char *src, unsigned char *dst) {
> >        continue;
> >      }
> >      if (ch == '.' && ((tp + sizeof(struct in_addr)) <= endp)) {
> > -      uv_err_t err = inet_pton4(curtok, tp);
> > -      uv_err_t err;
> >   +
> > -      /\* Scope id present, parse ipv4 addr without it */
> > -      pch = strchr(curtok, '%');
> > -      if (pch != NULL) {
> > -        char tmp[sizeof "255.255.255.255"];
> 
> Ok, makes sense. We should also backport fd45f87
> https://github.com/joyent/libuv/commit/fd45f876, btw.

How far do you want to go with IPv6 "bugs vs features"? I fixed quite a 
few things in there, and not everything should be backported IMHO.

indutny, line 268: I want to get scopeid working in v0.10 branch of both node and libuv. And considering need of that backport - it is a new feature. Reverting it!

indutny, line 268: Reverted, will open a PR for master in a bit.

indutny:
See https://github.com/joyent/libuv/pull/1237

PR #1236, opened 2014-04-14
win: always leave crit section in get_proc_title

### LINKED ISSUE #1235: Windows: uv_get_process_title doesn't call LeaveCriticalSection in all code paths
LeaveCriticalSection should be called inside uv_get_process_title (win\util.c), the before all return statements. Currently one is missing before the line:

```
return uv__new_sys_error(GetLastError());
```

indutny:
fix #1235


indutny:
Thank you @tjfontaine , landed in ffb4922

PR #1237, opened 2014-04-14
inet: allow scopeid in uv_inet_pton

indutny:
We already support it in `uv_ip6_addr` anyway.

See https://github.com/joyent/node/issues/7395

### DIFF
--- a/src/inet.c
+++ b/src/inet.c
@@ -225,7 +225,7 @@ static int inet_pton6(const char *src, unsigned char *dst) {
   curtok = src;
   seen_xdigits = 0;
   val = 0;
-  while ((ch = *src++) != '\0') {
+  while ((ch = *src++) != '\0' && ch != '%') {
     const char *pch;
 
     if ((pch = strchr((xdigits = xdigits_l), ch)) == NULL)
@@ -256,7 +256,19 @@ static int inet_pton6(const char *src, unsigned char *dst) {
       continue;
     }
     if (ch == '.' && ((tp + sizeof(struct in_addr)) <= endp)) {
-      int err = inet_pton4(curtok, tp);
+      int err;
+
+      /* Scope id present, parse ipv4 addr without it */
+      pch = strchr(curtok, '%');
+      if (pch != NULL) {
+        char tmp[sizeof "255.255.255.255"];
+
+        memcpy(tmp, curtok, pch - curtok);
+        curtok = tmp;
+        src = pch;
+      }
+
+      err = inet_pton4(curtok, tp);
       if (err == 0) {
         tp += sizeof(struct in_addr);
         seen_xdigits = 0;
--- a/test/test-ip6-addr.c
+++ b/test/test-ip6-addr.c
@@ -97,3 +97,45 @@ TEST_IMPL(ip6_addr_link_local) {
   RETURN_SKIP("Qualified link-local addresses are not supported.");
 #endif
 }
+
+
+#define GOOD_ADDR_LIST(X)                                                     \
+    X("::")                                                                   \
+    X("::1")                                                                  \
+    X("fe80::1")                                                              \
+    X("fe80::")                                                               \
+    X("fe80::2acf:daff:fedd:342a")                                            \
+    X("fe80:0:0:0:2acf:daff:fedd:342a")                                       \
+    X("fe80:0:0:0:2acf:daff:1.2.3.4")                                         \
+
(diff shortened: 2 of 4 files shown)


src/inet.c
```diff
@@ -256,7 +256,14 @@ static int inet_pton6(const char *src, unsigned char *dst) {
...
     if (ch == '.' && ((tp + sizeof(struct in_addr)) <= endp)) {
-      int err = inet_pton4(curtok, tp);
+      int err;
+
+      /* Scope id is not allowed after ipv4 address */
+      pch = strchr(curtok, '%');
+      if (pch != NULL)
+        return UV_EINVAL;
+
+      err = inet_pton4(curtok, tp);
```
saghul, line 271: inet_pton4 will fail, I'd let it fail there.

saghul:
uv_ip6_addr actually works with scoped addresses already. If you look at the code, it calls uv_inet_pton, so I wonder if we actually have anything to fix here. Or I'm in low-monday-mode...

src/inet.c
indutny, line 271: Oh, good point, actually.

indutny, line 271: Fixed!

saghul:
Ah, no, we pass a copied area which removes the scope id. I think it's ok the way we have it, I'd leave it that way.

indutny:
So, let me describe this problem on a node level: we use `uv_inet_pton` in `net.isIP` function, the test isn't passing right now. But it'll work if manually passed in the `bind()` or `connect()` binding method.

saghul:
I see. I propose the following: check if the system's inet_pton returns ok or not for an IP address with a scope id (I believe it doesn't). If it does, we adjust our implementation. Else, you can use uv_ip4_addr / uv_ip6_addr, or cut the original string at the '%' character. What do you think?

indutny:
It does work:

``` C
#include <arpa/inet.h>
#include <stdio.h>

int main(int argc, char** argv) {
  struct in6_addr addr;
  int r;
  if (argc < 2) {
    fprintf(stdout, "Usage: %s <addr>\n", argv[0]);
    return 1;
  }

  r = inet_pton(AF_INET6, argv[1], &addr);
  fprintf(stdout, "Result: %d\n", r);
  return r;
}
```

``` bash
/tmp $ ./1 ::1
Result: 1
/tmp $ ./1 ::1%
Result: 1
/tmp $ ./1 "::1%ef0"
Result: 1
/tmp $ ./1 "::127.0.0.1%ef0"
Result: 1
/tmp $ ./1 "ff::ff::127.0.0.1%ef0"
Result: 0
```

saghul:
Looks like I was wrong. At lest on OSX inet_pton seems to ignore the scope id. So half of your fix is actually correct :-)

indutny:
Actually, initial version was better than current one, it seems :)

saghul:
Ok, don't allow me near a computer today :-/ Sorry about that Fedor.

indutny:
No problem, I haven't dug that deep myself. So basically it was my fault.

LGTY now?

test/test-ip6-addr.c
```diff
@@ -97,3 +97,45 @@ TEST_IMPL(ip6_addr_link_local) {
...
+    X("abcde::1")                                                             \
+    X("fe80:0:0:0:2acf:daff:fedd:342a:5678")                                  \
+    X("fe80:0:0:0:2acf:daff:abcd:1.2.3.4")                                    \
+    X("fe80:0:0:2acf:daff:1.2.3.4.5")                                         \
+
+#define TEST_GOOD(ADDR)                                                       \
+    ASSERT(0 == uv_inet_pton(AF_INET6, ADDR, &addr));                         \
+    ASSERT(0 == uv_inet_pton(AF_INET6, ADDR "%en1", &addr));                  \
+    ASSERT(0 == uv_inet_pton(AF_INET6, ADDR "%%%%", &addr));                  \
+    ASSERT(0 == uv_inet_pton(AF_INET6, ADDR "%en1:1.2.3.4", &addr));          \
```
saghul, line 122: I don't think this is a valid address.

indutny, line 122: Why not? Is there any specification of the format of the interface name?

saghul, line 122: Apparently doesn't give an error on OSX either. LGTM.

indutny:
Thank you, landed in 2b21419

PR #1239, opened 2014-04-14
include: reorder UV_HANDLE_PRIVATE_FIELDS

trevnorris:
The fields for UV_HANDLE_PRIVATE_FIELDS in uv-win.h and uv-unix.h are
the same, but in different order. Simply swap them so uv_handle_t are
the same on both platforms.

---

Completely unnecessary I know, but has just been sticking in the back of my head. So decided to create the PR.


indutny:
LGTM

indutny:
Thank you, landed in f212cef

PR #1241, opened 2014-04-16
unix: free watchers on uv_loop_close()

trevnorris:
Simple addition to make valgrind happier on `uv_loop_close()`. Since
`uv_loop_close()` does not free() the loop if it's the `uv_default_loop()`
then the issue still appears. Here's the simplest script to show the
issue:

``` c
#include "uv.h"

int main() {
  int er = uv_run(uv_default_loop(), UV_RUN_DEFAULT);
  uv_loop_delete(uv_default_loop());
  return er;
}
```

And here's the sample valgrind output:

```
HEAP SUMMARY:
    in use at exit: 128 bytes in 1 blocks
  total heap usage: 1 allocs, 0 frees, 128 bytes allocated

128 bytes in 1 blocks are still reachable in loss record 1 of 1
   at 0x4C2A2DB: malloc (...)
   by 0x4C2C74F: realloc (...)
   by 0x403EB6: uv__io_start (core.c:679)
   by 0x406639: uv_signal_init (signal.c:225)
   by 0x4045BC: uv__loop_init (loop.c:138)
   by 0x40444E: uv_default_loop (loop.c:41)
```

---

I'm not sure if it's possible, but it would be nice if there was a way to allow proper cleanup of the `uv_default_loop()`.


trevnorris:
I just noticed that `uv_loop_delete()` is deprecated. nm this PR.

PR #1242, opened 2014-04-17
uv_tcp_get{peer/sock}name and uv_ip{4/6}_name should accept first argument by const pointer

ghost:
Aforementioned functions do not modify first argument in any way, so passing first argument by non-const pointer is not required. I'm developing my own C++11 wrapper around libuv and have to keep mutable private member variables to expose wrapper functions as const member functions due to this issue. I really hope that I have done all required changes for OSX. Do not have any machine to test them unfortunately.

I know that second commit's message exceeds 50 chars limit, but I do not see any way to fit it in without even more cryptic abbreviations :)


ghost:
It is strange to receive CLA warning, I have signed it before forking the repo and creating pull request.

saghul:
Disregard the CLA warning thing, there seems to be a problem with the Google Docs API.

src/unix/stream.c
```diff
@@ -1482,8 +1482,8 @@ int uv_is_writable(const uv_stream_t* stream) {
 
 
 #if defined(__APPLE__)
-int uv___stream_fd(uv_stream_t* handle) {
-  uv__stream_select_t* s;
+int uv___stream_fd(const uv_stream_t* handle) {
```
saghul, line 1485: Not that I'm against it, but does this give you any trouble? It's an internal function after all.

saghul:
LGTM, what do you think @indutny?

src/unix/stream.c
ghost, line 1485: This internal function is used by uv__stream_fd macro on OSX, and uv__stream_fd is called inside uv_tcp_getsockname and uv_tcp_getpeername. If I leave it as is, we'll cast constness away, and I don't see why should we do it here.

saghul, line 1485: I see, ok.

indutny:
-1for `uv_tcp_getsockname` and `uv_tcp_getpeername` . Why they clearly not changing handles now:
- Adding `const` qualifier to handle doesn't match the rest of API
- We won't be able to change it in stable version, and it'll generally give us more trouble than help

I'm ok with making internal method accept `const handle`, though.

ghost:
What API are you talking about? uv_pipe_getsockname accepts handle by const pointer, uv_tcp_get_name functions do not. uv_ip__addr functions accept ip by const pointer, uv_ip*_name do not do it for sockadrr that is not modified. There is simply no logic here. How does it break API for clients anyway? If someone passes non-const pointer to function that accepts const pointer, there is no error at all. All current code has to pass non-const pointers. It means that all current code will work the same way after this change.

indutny:
Oh, right! I think I have missed a couple of recent changes :) Ok, LGTM

saghul:
Thanks @Astellar! Rebased and landed in https://github.com/joyent/libuv/commit/4018f726150f060e73d578a48a7d24914e06d899, https://github.com/joyent/libuv/commit/886e2bcd55f974198832cb0887f9521fdd7a8b9e and https://github.com/joyent/libuv/commit/20d092edc8dc8f4a8a6198f9fa31d21fdec92d7b.

PR #1243, opened 2014-04-21
windows: add invert ANSI (7 / 27) emulation

Qix-:
Since `COMMON_LVB_REVERSE_VIDEO` is not supported on any modern windows system, it is used as a flag and emulated through the bitwise flipping of fg/bg to create the inverse (reverse video) graphics in the console.

Uses the standard ANSI codes `[7;` and `[27;` for negative/positive.


Qix-:
Done.

src/win/tty.c
```diff
@@ -1238,6 +1258,23 @@ static int uv_tty_set_style(uv_tty_t* handle, DWORD* error) {
...
 
+  if (inverse != 0) {
+    if (inverse > 0) {
+      info.wAttributes |= COMMON_LVB_REVERSE_VIDEO;
+    } else {
+      info.wAttributes &= ~COMMON_LVB_REVERSE_VIDEO;
+    }
+  }
+
+  if ((info.wAttributes & COMMON_LVB_REVERSE_VIDEO) > 0) {
```
indutny, line 1272: Is there any way to deduplicate it by macro or something like that?

src/win/tty.c
```diff
@@ -1158,6 +1161,10 @@ static int uv_tty_set_style(uv_tty_t* handle, DWORD* error) {
       /* Background bright on */
       bg_bright = 1;
 
+    } else if (arg == 7) {
+      /* Inverse: on (negative) */
+      inverse = 1;
```
saghul, line 1174: This reads a bit odd. "on (negative)" yet the value is positive. Maybe you can reword the comment a bit?

src/win/tty.c
```diff
@@ -1166,6 +1173,10 @@ static int uv_tty_set_style(uv_tty_t* handle, DWORD* error) {
       /* Background bright off */
       bg_bright = 0;
 
+    } else if (arg == 27) {
+      /* Inverse: off (positive) */
+      inverse = -1;
```
saghul: Ditto.

src/win/tty.c
```diff
@@ -1126,6 +1126,7 @@ static int uv_tty_set_style(uv_tty_t* handle, DWORD* error) {
 
   char fg_color = -1, bg_color = -1;
   char fg_bright = -1, bg_bright = -1;
+  char inverse = 0;
```
piscisaureus: The pattern used in this function is that if `fg_color`, `bg_bright` etc remain unchanged their value is -1.
You should do the same for `inverse`: initialize to -1 and then change the value to either 0 or 1 if it has to change.

src/win/tty.c
piscisaureus: So this should become `inverse = 0`.

src/win/tty.c
```diff
@@ -1198,7 +1209,7 @@ static int uv_tty_set_style(uv_tty_t* handle, DWORD* error) {
   }
 
   if (fg_color == -1 && bg_color == -1 && fg_bright == -1 &&
-      bg_bright == -1) {
+      bg_bright == -1 && inverse == 0) {
```
piscisaureus: inverse == -1

src/win/tty.c
Qix-, line 1272: Sure, what is the preference on location of macro defines relative to the function?

src/win/tty.c
Qix-, line 1174: Yea I thought I'd get such a response. I will update it to `-1`=unchanged, `0`=normal, `1`=inverted.

src/win/tty.c
Qix-: Yep; see [my comment above](https://github.com/joyent/libuv/pull/1243#r11830283).

Qix-:
Updated the values to more closely follow the convention of the other flag values.

Just waiting on a response from @indutny about a possible macro and its location.

src/win/tty.c
indutny, line 1272: Right before the function should be fine ;)

Qix-:
Not entirely sure why GitHub isn't showing the diff'd lines, but...
- [1120-1125](https://github.com/Qix-/libuv/blob/8aefacc68b255526566b156ff658f3622b8de799/src/win/tty.c#L1120-L1125): macro definition
- [1229](https://github.com/Qix-/libuv/blob/8aefacc68b255526566b156ff658f3622b8de799/src/win/tty.c#L1229): initial flip
- [1271](https://github.com/Qix-/libuv/blob/8aefacc68b255526566b156ff658f3622b8de799/src/win/tty.c#L1271): secondary flip

@indutny, I went ahead and put the macro right before the function. I assume that is where you meant.

EDIT: Is the diff issue (plus the fact Github thinks I've changed thousands of lines) a line ending issue?

saghul:
All the line endings have been changed so the diff looks like it replaces the whole file. Please stick to Unix style line endings.

Qix-:
No problem, text editor settings must have been changed. I'll fix it.

Qix-:
@saghul Fixed line endings. :)

src/win/tty.c
```diff
@@ -1117,6 +1117,12 @@ static int uv_tty_clear(uv_tty_t* handle, int dir, char entire_screen,
   return 0;
 }
 
+#define FLIP_FGBG \
```
saghul: maybe use the do..while(0) construct here

Qix-: Done :)

src/win/tty.c
```diff
@@ -1117,6 +1117,14 @@ static int uv_tty_clear(uv_tty_t* handle, int dir, char entire_screen,
...
+    do { \
```
indutny: Pleas align `\` at 79 column

indutny:
LGTM after fixing few nits

Qix-:
@indutny Aligned :) Thank you for the review!

saghul:
@Qix- LGTM too. Can you squash and rebase?

Qix-:
@saghul Done.

saghul:
Landed in https://github.com/joyent/libuv/commit/e56717ae988416ddaebadbb1c23c07f466bab046. Thanks @Qix-!

Issue #1244, opened 2014-04-22
test echo server aborts with SIGPIPE


trevnorris:
Build: v0.11.22 Debug

---

Reproduce:

From one terminal do:

```
$ ./run-tests tcp4_echo_server
```

From another do:

```
$ nc localhost 9123 < /dev/zero > /dev/null
```

Then terminate with `Ctrl+C`.

---

Backtrace:

```
Program received signal SIGPIPE, Broken pipe.
0x00007ffff6fa1c20 in __write_nocancel () at ../sysdeps/unix/syscall-template.S:81

(gdb) bt
#0  0x00007ffff6fa1c20 in __write_nocancel () at ../sysdeps/unix/syscall-template.S:81
#1  0x0000000000462832 in uv__write (stream=0x7d7180) at ../src/unix/stream.c:794
#2  0x0000000000460e2e in uv__stream_io (loop=0x7d6488 <default_loop_struct>, w=0x7d71e8, events=4) at ../src/unix/stream.c:1220
#3  0x0000000000455180 in uv__run_pending (loop=0x7d6488 <default_loop_struct>) at ../src/unix/core.c:643
#4  0x0000000000454fd6 in uv_run (loop=0x7d6488 <default_loop_struct>, mode=UV_RUN_DEFAULT) at ../src/unix/core.c:288
#5  0x0000000000404658 in run_helper_tcp4_echo_server () at ../test/echo-server.c:349
#6  0x000000000040651d in run_test (test=0x7d7052 "tcp4_echo_server", benchmark_output=0, test_count=1) at ../test/runner.c:215
#7  0x0000000000405eca in maybe_run_test (argc=2, argv=0x7d7010) at ../test/run-tests.c:166
#8  0x00000000004057bd in main (argc=2, argv=0x7d7010) at ../test/run-tests.c:54

(gdb) f 1
#1  0x0000000000462832 in uv__write (stream=0x7d7180) at ../src/unix/stream.c:794
794             n = write(uv__stream_fd(stream), iov[0].iov_base, iov[0].iov_len);
(gdb) l
789         }
790         while (n == -1 && errno == EINTR);
791       } else {
792         do {
793           if (iovcnt == 1) {
794             n = write(uv__stream_fd(stream), iov[0].iov_base, iov[0].iov_len);
795           } else {
796             n = writev(uv__stream_fd(stream), iov, iovcnt);
797           }
798         }

(gdb) info locals
iov = 0x857880
q = 0x857858
req = 0x857820
[...]

PR #1249, opened 2014-04-22
test: fix for echo-server.c assert failure; addition to print uv error information

trevnorris:
First commit simply fixes https://github.com/joyent/libuv/issues/1244

Second one is an additional macro that will print a lot of information about libuv specific errors. This patch replaces a few areas that simply did a `return 1;` with always running `abort()`. If not acceptable then I'll work around it.

### DIFF
--- a/test/echo-server.c
+++ b/test/echo-server.c
@@ -51,24 +51,20 @@ static void after_write(uv_write_t* req, int status) {
   /* Free the read/write buffer and the request */
   wr = (write_req_t*) req;
   free(wr->buf.base);
+
+  if (status == 0) {
+    free(wr);
+    return;
+  }
+
+  fprintf(stderr,
+          "uv_write error: %s - %s\n",
+          uv_err_name(status),
+          uv_strerror(status));
+
+  if (!uv_is_closing((uv_handle_t*) req->handle))
+    uv_close((uv_handle_t*) req->handle, on_close);
   free(wr);
-
-  if (status == 0)
-    return;
-
-  fprintf(stderr, "uv_write error: %s\n", uv_strerror(status));
-
-  if (status == UV_ECANCELED)
-    return;
-
-  ASSERT(status == UV_EPIPE);
-  uv_close((uv_handle_t*)req->handle, on_close);
-}
-
-
-static void after_shutdown(uv_shutdown_t* req, int status) {
-  uv_close((uv_handle_t*)req->handle, on_close);
-  free(req);
 }
 
 
@@ -77,7 +73,6 @@ static void after_read(uv_stream_t* handle,
                        const uv_buf_t* buf) {
   int i;
   write_req_t *wr;
-  uv_shutdown_t* req;
 
   if (nread < 0) {
     /* Error or EOF */
@@ -87,9 +82,7 @@ static void after_read(uv_stream_t* handle,
       free(buf->base);
     }
 
-    req = (uv_shutdown_t*) malloc(sizeof *req);
-    uv_shutdown(req, handle, after_shutdown);
-
+    uv_close((uv_handle_t*) handle, on_close);
     return;
   }
 


saghul:
I'm -1 on the second commit. All tests are written with ASSERT already and checking for return codes is done all over the place. I'd rather have a consistent way of doing things.

trevnorris:
OK. No worries about the second commit.

As for the first, i'll do some testing and see if it ever gets past the `if()` to the `EPIPE` assert.

saghul:
> As for the first, i'll do some testing and see if it ever gets past the if() to the EPIPE assert.

Ok, cool!

trevnorris:
As for the second commit, reason I put that in was because of the comments in the test code. If I did the work to replace all instances of that type of check, would it be worth it, or just too much code turn?

saghul:
Personally, I think it's too much code turn.

trevnorris:
That's cool.

As for the `EPIPE` assert. I just put in an `fprintf()` after the if, and it indeed does get to that assert. Any ideas why?

saghul:
So, that means you do get EPIPE then. Are you on OSX? On Linux you'd get SIGPIPE and the process would exit.

It might be a good idea to simplify that code even further and just do the printf. That is, rely on the read callback to close the handle. This way, the `uv_close` is done in a single place and we avoid checking if the handle is closed or closing it twice by mistake.

trevnorris:
nope. i'm running linux.

trevnorris:
also, in `after_read()` it does always `uv_close()` if `nread < 0`. Not sure how else to simplify that.

saghul:
I ment to simplify after_write, sorry if I misled you :-)

trevnorris:
ah, yeah. just realized that.

trevnorris:
removed the `uv_close()` in the `nread < 0` case. though there are still two instances of `uv_close()` below i'm not sure how to deal with.

regardless, I simplified the end of `after_write()` to look like so:

``` c
  free(wr);

  if (status == 0)
    return;

  fprintf(stderr, "uv_write error: %s\n", uv_strerror(status));
  if (!uv_is_closing((uv_handle_t*)req->handle))
    uv_close((uv_handle_t*)req->handle, on_close);
```

but it segfaults. from gdb:

```
(gdb) l
56        if (status == 0)
57          return;
58
59        fprintf(stderr, "uv_write error: %s\n", uv_strerror(status));
60        if (!uv_is_closing((uv_handle_t*)req->handle))
61          uv_close((uv_handle_t*)req->handle, on_close);
62      }

(gdb) p req
$1 = (uv_write_t *) 0xbadb00
```

so seems the checks and early return must be put into place?

saghul:
I'm not sure I follow what you did. It's "mandatory" that you close the handle if nread is < 0, the code expects you to do so: https://github.com/joyent/libuv/blob/master/src/unix/stream.c#L1114-L1117

Now, what I meant is that in the write callback you'd just fprintf in case there was an error, and bail out otherwise. So no uv_close there.

Is there any other place where the handle is closed appart from the read callback? (assuming you make the change I mentioned, that is).

trevnorris:
ok. force pushed what I think you mean. here's some example output from the test:

```
uv_write error: ECONNRESET - connection reset by peer
uv_write error: EPIPE - broken pipe
uv_write error: EPIPE - broken pipe
uv_write error: EPIPE - broken pipe
uv_write error: ECANCELED - operation canceled
uv_write error: ECANCELED - operation canceled
uv_write error: EPIPE - broken pipe
```

trevnorris:
also, i'm a little confused by the old code. how it would resolve the `uv_stream_t` as a `write_req_t`, but then below would `uv_close()` on `req->handle`. but i'll figure out why that works later.

edit: oh wait. yeah. I see it now. duh. :P

saghul:
Hum, I don't get why all those EPIPE errors show up if we close the socket. Can you also fprintf (just for debugging) when you call uv_close?

> also, i'm a little confused by the old code. how it would resolve the uv_stream_t as a write_req_t, but then below would uv_close() on req->handle. but i'll figure out why that works later.

The write callback gets the write request, which has the associated handle in req->handle. The confusion can come becuase there is that write_req_t auxiliary struct just to hold on to the buffer. There also seemd to be a use-after-free bug in there, since req->handle was accessed after wr was freed, which freed the storage for req in turn.

trevnorris:
In `after_read()` there's `ASSERT(nread == UV_EOF);`. so that solidifies that that's the only error reaching the `uv_read_cb`.

The other two places where `uv_close()` is called is below. Explained by the following comment:

``` c
  /*
   * Scan for the letter Q which signals that we should quit the server.
   * If we get QS it means close the stream.
   */
```

but since I'm just `nc localhost 9123 < /dev/zero` neither of those cases should ever be true. I'll double check that to be certain.

edit: just verified. neither of those cases are ever hit.

trevnorris:
ok, force pushed what I think is the safest thing. still getting the following output:

```
uv_write error: ECONNRESET - connection reset by peer
uv_write error: EPIPE - broken pipe
uv_write error: ECANCELED - operation canceled
```

saghul:
Looks ok. I'll have another deeper look in the morning, with a fresh mind :-) Thanks for working on this Trevor!

trevnorris:
np :)

saghul:
LGTM.

trevnorris:
Awesome. Thanks. :-)

I don't have the commit bit so I'll leave it to you.

trevnorris:
ping. I can't merge this. :)

@indutny

saghul:
Sorry Trevor, I've been between planes... Will merge it asap.

trevnorris:
no worries. not a not a critical bug fix or anything. :)

saghul:
Thanks, Trevor, landed!

PR #1250, opened 2014-04-23
windows: fix build error

stakach:
Previously using Windows DevKit 8.1
`src\win\tcp.c(250): error C2275: 'SOCKET' : illegal use of this type as an expression`

### DIFF
--- a/src/win/tcp.c
+++ b/src/win/tcp.c
@@ -243,11 +243,13 @@ static int uv_tcp_try_bind(uv_tcp_t* handle,
   int r;
 
   if (handle->socket == INVALID_SOCKET) {
+    SOCKET sock;
+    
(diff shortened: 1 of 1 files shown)


stakach:
CLA signed

src/win/tcp.c
```diff
@@ -241,13 +241,14 @@ static int uv_tcp_try_bind(uv_tcp_t* handle,
                            unsigned int flags) {
   DWORD err;
   int r;
+  SOCKET sock;
```
saghul: Can you declare it inside the if (line 245) instead? It's only used there.

saghul:
Sorry, I introduced this, my bad :-(

stakach:
ok, should be ready to merge
Also, no worries, you do great work!

saghul:
Landed, thanks!

PR #1253, opened 2014-04-23
unix: handle case when addr is not initialized after recvmsg

### LINKED ISSUE #1252: addr in uv_udp_recv_cb() is garbage when using uv_udp_open() in Linux
I pass a non-blocking AF_UNIX SOCK_DGRAM socket to `uv_udp_open(fd)`, so I can use the UDP API of libuv. 

It works perfectly. But... when I inspect the `struct sockaddr* addr` given in the `uv_udp_recv_cb()`: 
- In OSX (Mavericks): I inspect the family of `addr` and, indeed, it is `AF_UNIX` (OK). 
- Linux (Debian 7): I inspect the family of `addr` and it is unknown. This is, the given `addr` is not a `struct sockaddr` at all, but garbage.

I expect this is a bug (in Linux) given that libuv should not provide a garbage `struct sockaddr`.

saghul:
If the handle was opened using `uv_udp_open` ift's possible that the
kernel doesn't fill in the msg_name field, so return NULL as the address
in that case.

fixes #1252

### DIFF
--- a/src/unix/udp.c
+++ b/src/unix/udp.c
@@ -221,16 +221,17 @@ static void uv__udp_recvmsg(uv_loop_t* loop,
         handle->recv_cb(handle, -errno, &buf, NULL, 0);
     }
     else {
-      flags = 0;
+      const struct sockaddr *addr;
+      if (h.msg_namelen == 0)
+        addr = NULL;
+      else
+        addr = (const struct sockaddr*) &peer;
 
+      flags = 0;
       if (h.msg_flags & MSG_TRUNC)
         flags |= UV_UDP_PARTIAL;
 
-      handle->recv_cb(handle,
-                      nread,
-                      &buf,
-                      (const struct sockaddr*) &peer,
-                      flags);
+      handle->recv_cb(handle, nread, &buf, addr, flags);
     }
   }
   /* recv_cb callback may decide to pause or close the handle */


src/unix/udp.c
```diff
@@ -229,7 +229,7 @@ static void uv__udp_recvmsg(uv_loop_t* loop,
       handle->recv_cb(handle,
                       nread,
                       &buf,
-                      (const struct sockaddr*) &peer,
+                      (h.msg_namelen == 0 ? NULL : (const struct sockaddr*) &peer),
```
indutny: 80 column limit applies?

indutny: Perhaps it could be assigned to a variable before passing there?

indutny:
Could the test be written for it?

src/unix/udp.c
saghul: It's a tad longer yes, but aren't rules ment to be broken? ;-) On a more serious note, writing it otherwise looks ugly.

saghul:
Yes, but I'm not sure what systems are affected by this, the fix is a defensive mechanism, you can't get into this problem is you stick to libuv handles... So I'm not very motivated to write one :-P

indutny:
Ok, I think this is fine then :) Could you please move that long expression to a variable?

indutny:
Otherwise LGTM

saghul:
Force pushed.

src/unix/udp.c
```diff
@@ -221,15 +221,20 @@ static void uv__udp_recvmsg(uv_loop_t* loop,
...
+      if (h.msg_namelen == 0)
+          addr = NULL;
+      else
+          addr = (const struct sockaddr*) &peer;
 
+      flags = 0;
       if (h.msg_flags & MSG_TRUNC)
         flags |= UV_UDP_PARTIAL;
 
       handle->recv_cb(handle,
```
indutny: Will it fit in one line now?

src/unix/udp.c
```diff
@@ -221,15 +221,20 @@ static void uv__udp_recvmsg(uv_loop_t* loop,
         handle->recv_cb(handle, -errno, &buf, NULL, 0);
     }
     else {
-      flags = 0;
+      const struct sockaddr *addr;
+      if (h.msg_namelen == 0)
+          addr = NULL;
```
indutny: two spaces indent.

indutny:
Two nits, otherwise LGTM

saghul:
Fixed the nits.

saghul:
Landed, thanks @indutny!

PR #1262, opened 2014-04-30
windows: add VT100 codes ?25l and ?25h (hide/show cursor)

Qix-:
One of the first machines to support ANSI was the VT100; two of its extension codes, `?25l` and `?25h` (hide/show cursor, respectively) are still in use today (and are supported by most/all ANSI-compliant terminals).

This patch adds support for this in Windows using `Get`/`SetConsoleCursorInfo` and introduces a skipped character `?` to the ANSI parser.

> An initial `?` is innocuous to _any_ escape sequence; this means this will not break if `CSI [?31m` is outputted; as expected, dark red text is shown.


Qix-:
> The following commiters were not found in the CLA

I'm in the CLA. [I promise](https://github.com/joyent/libuv/pull/1243#issuecomment-40914261). :)

> First line of commit message must be no longer than 50 characters

Is this fixed by a `commit --amend` / `push --force`?

src/win/tty.c
```diff
@@ -1527,6 +1545,13 @@ static int uv_tty_write_bufs(uv_tty_t* handle,
               continue;
             }
 
+          } else if (utf8_codepoint == '?' && !(ansi_parser_state & ANSI_IN_ARG) &&
+                     handle->ansi_csi_argc == 0) {
+            /* Ignores '?' if it is the first character after CSI[ */
```
saghul, line 1550: Comment the whole block, not each line.

saghul:
> Is this fixed by a commit --amend / push --force?

Yes.

Code LGTM, but I have no experience in terminal handling, so I'll wait for @indutny or @piscisaureus to comment as well.

piscisaureus:
Commits to be squashed, but the code is otherwise flawless. Thanks.

src/win/tty.c
Qix-, line 1550: Check 1517/1535 - That seems to be the commenting scheme.

Qix-:
@piscisaureus Fixed the commit message. :) Thanks for the review.

saghul:
@Qix- rebased and landed in https://github.com/joyent/libuv/commit/21b1b87d12a36e7912a8237241267402c91e5162 Thanks a lot for this and the other pending PRs, keep 'em coming! ;-)

PR #1264, opened 2014-04-30
thread: add uv_thread_cb typedef

trevnorris:
To keep with convention, add a typedef for the function called that
initializes the new thread.


include/uv.h
```diff
@@ -2204,8 +2204,14 @@ UV_EXTERN void uv_key_delete(uv_key_t* key);
...
-UV_EXTERN int uv_thread_create(uv_thread_t *tid,
-    void (*entry)(void *arg), void *arg);
+/*
+ * Callback that is invoked to initialize thread execution.
+ *
+ * `arg` is the same value that was passed to uv_thread_create().
+ */
+typedef void (*uv_thread_cb)(void* arg);
+
+UV_EXTERN int uv_thread_create(uv_thread_t *tid, uv_thread_cb, void *arg);
```
indutny: Mind changing the signature at the implementation site too?

indutny: Also, please make it `uv_thread* tid` and `void* arg` while we are here

indutny:
Few comments, otherwise LGTM

include/uv.h
trevnorris: derp. feel stupid for missing the first one.

trevnorris:
@indutny both things fixed. thanks much.

indutny:
Thanks, landed in 32597100b81a03cc827c164ca8250fba91599593 !

Issue #1205, opened 2014-03-18
Some love for UDP and IPv6


saghul:
While working on adding IPv6 support to `uv_udp_set_membership` I noticed some inconsistencies / missing stuff in the UDP support which I plan to fix. Here is the list of things I'll be working on:
- [x] Extend `uv_udp_bind` to accept a `UV_UDP_REUSEADDR` flag and don't set `SO_REUSEADDR` by default.
- [x] Add IPv6 support to `uv_udp_set_multicast_interface`
- [x] ~~Only do implicit binding on `uv_udp_set_membership` (binding to 'any' and setting `UV_UDP_REUSEADDR`, other `uv_udp_set_*` functions should probably return `UV_EBADF`. Currently Windows does implicit binding but Unix doesn't. I wonder if `uv_udp_send` and `uv_recv_start` shouldn't bind the socket either.~~ Review where implicit binding happens and align Unxi and Windows.
- [x] Mark handles which were created with AF_INET6 with some flag, like Windows does.

Comments are welcome! /cc @indutny @bnoordhuis.

bnoordhuis:
> I wonder if uv_udp_send and uv_recv_start shouldn't bind the socket either.

There is no real need for that, I think, you just need a valid socket file descriptor.  UDP sockets on POSIX platforms are always implicitly bound to _some_ address/port combo.

(I suppose that also applies to uv_udp_set_membership().)

saghul:
> There is no real need for that, I think, you just need a valid socket file descriptor. UDP sockets on POSIX platforms are always implicitly bound to some address/port combo.
> 
> (I suppose that also applies to uv_udp_set_membership().)

But until we bind to it the fd will be -1, won't it? We already to the implicit bind on `uv_udp_send`: https://github.com/joyent/libuv/blob/master/src/unix/udp.c#L405-L407 (also on `uv_udp_recv_start).

I was thinking about removing it, so the user is forced to bind it explicitly. That way, there would only be 2 ways to bind a socket: `uv_udp_bind` and `uv_udp_set_membership`, which would bind it to the 'any' address. What do you think?

bnoordhuis:
> But until we bind to it the fd will be -1, won't it?

In the current implementation, yes.  But you don't have to bind a socket in order for it to be usable: just `socket(AF_INET, SOCK_DGRAM, 0)` is enough to be able to call sendmsg() or whatever.  You can bind it later if so desired.

[...]

PR #1270, opened 2014-05-07
unix, windows: be consistent when binding implictly in UDP

saghul:
Only these functions will trigger an implicit binding of a UDP handle:
- uv_udp_send
- uv_udp_recv_start
- uv_udp_set_membership

All other functions will return UV_EBADF in case the socket was not
bound.

Note: currently the socket is created and bound at the same time. This
may change in the future.

This is the last piece for https://github.com/joyent/libuv/issues/1205

/cc @indutny


saghul:
I'll add some tests after first review.

indutny:
Looks good

saghul:
Updated with tests.

test/test-udp-options.c
```diff
@@ -86,3 +86,38 @@ TEST_IMPL(udp_options) {
...
+
+TEST_IMPL(udp_no_autobind) {
+  uv_loop_t* loop;
+  uv_udp_t h;
+  int r;
+
+  loop = uv_default_loop();
+
+  r = uv_udp_init(loop, &h);
+  ASSERT(r == 0);
```
indutny: Please merge with the line above.

saghul:
Updated. Is that what you meant?

indutny:
Not really, I meant:

``` C
ASSERT(UV_EBADF == uv_udp_set_multicast_interface(&h, "0.0.0.0"));
```

Instead of two lines.

saghul:
Ah, I see. Updated.

test/test-udp-options.c
```diff
@@ -86,3 +86,27 @@ TEST_IMPL(udp_options) {
...
+  ASSERT(0 == uv_udp_init(loop, &h));
+  ASSERT(UV_EBADF == uv_udp_set_multicast_ttl(&h, 32));
+  ASSERT(UV_EBADF == uv_udp_set_broadcast(&h, 1));
+  ASSERT(UV_EBADF == uv_udp_set_ttl(&h, 1));
+  ASSERT(UV_EBADF == uv_udp_set_multicast_loop(&h, 1));
+  ASSERT(UV_EBADF == uv_udp_set_multicast_interface(&h, "0.0.0.0"));
+
+  uv_close((uv_handle_t*) &h, NULL);
+
+  r = uv_run(loop, UV_RUN_DEFAULT);
```
indutny: Forgot one? :)

indutny: Also you don't need `int r;` declaration anymore.

saghul:
Done.

indutny:
LGTM, thank you!

saghul:
Landed in https://github.com/joyent/libuv/commit/386d2141e44a0d007e6d5a8535f82f9191dc9d47, thanks Fedor!

PR #1271, opened 2014-05-07
unix: fix handling error on UDP socket creation

saghul:
Thanks @ibc for the report and helping with testing :-)

/cc @indutny

### DIFF
--- a/src/unix/udp.c
+++ b/src/unix/udp.c
@@ -320,9 +320,10 @@ int uv__udp_bind(uv_udp_t* handle,
 
   fd = handle->io_watcher.fd;
   if (fd == -1) {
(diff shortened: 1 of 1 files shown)


saghul:
Writing a test for this is not easy as we'd need to make uv__socket return an error. It works as it says on the tin :-)

indutny:
You could make it return error by lowering fd limit and triggering ENOENT... But it is rather tricky, indeed.

LGTM, land it! :)

saghul:
Thanks Fedor! :-)

ibc:
I've tested and re-tested with all the possible values for "ulimit -n" (I've open ~65000 UDP sockets right now XD). It just works :)

:cake:

PR #1272, opened 2014-05-08
unix, windows: remove unneeded define

saghul:
We define it for all supported platforms. I doubt there is a system which doesn't have `if_nametoindex`.

/cc @indutny


indutny:
Right, I remember we discussed it previously. LGTM

saghul:
Thanks, landed!

PR #1276, opened 2014-05-10
windows: required environment variables

vtjnash:
adds more of the undocumented-but-seemingly-required environment variables for uv_spawn

written to address https://github.com/JuliaLang/julia/issues/5574

this list was created with the help of https://github.com/Alexpux/Cygwin/blob/b266b04fbbd3a595f02ea149e4306d3ab9b1fe3d/winsup/cygwin/environ.cc#L955 to get a more complete list

I notice that cygwin file also ensures certain variables are uppercase for compatibility "for posix and backwards compatibility". Would you be interested in a pull request (written from scratch, but using the same variable list for reference) so that libuv also also does this? Or does this seem like a non-existent issue for libuv?

### DIFF
--- a/src/win/process.c
+++ b/src/win/process.c
@@ -616,6 +616,11 @@ static void check_required_vars_contains_var(env_var_t* required, int count,
  * TEMP. SYSTEMDRIVE is probably also important. We therefore ensure that
  * these get defined if the input environment block does not contain any
  * values for them.
+ *
+ * Also add variables known to Cygwin to be required for correct
+ * subprocess operation in many cases:
+ * https://github.com/Alexpux/Cygwin/blob/b266b04fbbd3a595f02ea149e4306d3ab9b1fe3d/winsup/cygwin/environ.cc#L955
+ *
  */
 int make_program_env(char* env_block[], WCHAR** dst_ptr) {
   WCHAR* dst;
@@ -630,6 +635,14 @@ int make_program_env(char* env_block[], WCHAR** dst_ptr) {
     E_V("SYSTEMROOT"),
     E_V("SYSTEMDRIVE"),
     E_V("TEMP"),
+    E_V("HOMEDRIVE"),
+    E_V("HOMEPATH"),
+    E_V("USERDOMAIN"),
+    E_V("USERNAME"),
+    E_V("USERPROFILE"),
+    E_V("WINDIR"),
+    E_V("PATH"),
+    E_V("LOGONSERVER"),
   };
 
   for (env = env_block; *env; env++) {
@@ -653,13 +666,12 @@ int make_program_env(char* env_block[], WCHAR** dst_ptr) {
 
   for (i = 0; i < ARRAY_SIZE(required_vars); ++i) {
     if (!required_vars[i].supplied) {
-      env_len += required_vars[i].len;
       var_size = GetEnvironmentVariableW(required_vars[i].wide, NULL, 0);
-      if (var_size == 0) {
-        return GetLastError();
-      }
       required_vars[i].value_len = var_size;
-      env_len += var_size;
+      if (var_size != 0) {
+        env_len += required_vars[i].len;
+        env_len += var_size;
+      }
     }
   }
 
@@ -684,7 +696,7 @@ int make_program_env(char* env_block[], WCHAR** dst_ptr) {
   }
 
   for (i = 0; i < ARRAY_SIZE(required_vars); ++i) {
-    if (!required_vars[i].supplied) {
+    if (!required_vars[i].supplied && required_vars[i].value_len!=0) {
       wcscpy(ptr, required_vars[i].wide);
       ptr += required_vars[i].len - 1;
       *ptr++ = L'=';
--- a/src/win/process.c
+++ b/src/win/process.c
@@ -25,6 +25,7 @@
 #include <stdlib.h>
 #include <signal.h>
 #include <limits.h>
+#include <malloc.h>
 
 #include "uv.h"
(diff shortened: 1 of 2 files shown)


orangemocha:
While these environment variables seem generally safe, I am concerned because uv_spawn would now fail if they were not defined. Since you are passing in a custom environment, my recommendation would be to make sure you are setting those variables in your application.

vtjnash:
would it be better to make it non-fatal if these variables didn't exist, but make a best attempt? I expect most users are not aware of all of these, especially if trying to move an application from linux to windows

orangemocha:
Yes, I think that would be better.

vtjnash:
To confirm, I would propose changing this line (and surrounding code): https://github.com/joyent/libuv/pull/1276/files#diff-ea628f72ae5d0872acddec8be5b07ad0L660
to silently skip these "required" environment variables that don't exist in the callee's (libuv's) environment.

orangemocha:
That sounds reasonable to me. I was initially thinking of having a required set and a best-effort set, but it's probably not worth the extra complexity.

vtjnash:
I updated the pull request to make the required set best-effort.

However, in doing so, I realized libuv has not been sorting the env variables as is required by CreateProcess:(http://msdn.microsoft.com/en-us/library/windows/desktop/ms682425(v=vs.85).aspx). Would it be preferable for me to update this pull request to sort them according to the requirements (http://msdn.microsoft.com/en-us/library/windows/desktop/ms682009(v=vs.85).aspx) or to open a new pull request based on this one?

saghul:
You could add another commit to this PR, that should do.

src/win/process.c
```diff
@@ -684,7 +696,7 @@ int make_program_env(char* env_block[], WCHAR** dst_ptr) {
   }
 
   for (i = 0; i < ARRAY_SIZE(required_vars); ++i) {
-    if (!required_vars[i].supplied) {
+    if (!required_vars[i].supplied && required_vars[i].value_len!=0) {
```
saghul: Add some spaces between the !=

saghul:
Ping.

vtjnash:
alright, it think that should cover everything, be reasonably flexible, and have good test coverage

vtjnash:
Apparently CompareStringOrdinal is not available on XP, and has the incorrect calling convention for 32-bit in mingw, so this needs to be updated with a custom compare function before merging.

Investigating the behavior of windows, we see that the desired behavior is UTF-16 byte-order, with characters converted to uppercase before sorting (the numbers indicate what would have been the UTF-32 sort order):

```
  "Ba"                      => "1"
  "BB"                      => "2"
  "B_"                      => "3"
  "B\u20ac"                 => "4"
  "B\U10002"                => "6"
  "B\U23456"                => "7"
  "B\uff61"                 => "5"
```

(thse unicode value were selected from the table at https://ssl.icu-project.org/docs/papers/utf16_code_point_order.html)

vtjnash:
i believe this should now work on XP also, with the latest commit

saghul:
Thanks, Jameson. I'll try it out.

tkelman:
Hope you don't mind, I cherry-picked including `malloc.h` into this branch, which I found was necessary to get this to compile (or, rather, link) in MSVC.

src/win/process.c
```diff
@@ -616,91 +659,148 @@ static void check_required_vars_contains_var(env_var_t* required, int count,
...
+        if (var_size != len-1) { /* race condition? */
+          uv_fatal_error(GetLastError(), "GetEnvironmentVariableW");
+        }
       }
-      ptr += required_vars[i].value_len;
+      i++;
+    } else {
+      /* copy var from env_block */
+      len = wcslen(*ptr_copy)+1;
+      DWORD r = wmemcpy_s(ptr, (env_len - (ptr - dst)), *ptr_copy, len);
```
tkelman, line 792: this would have to be declared at the top of the function for MSVC < 2013

vtjnash, line 792: top of the scope block, yes. such an annoying restriction, but any `{` should also count as the start of a new scope block, IIRC, so this doesn't need to move far (just after the `else`)

saghul, line 792: Yeah, that will do.

saghul:
@vtjnash I fixed some style nits, rebased and squashed, the result is here: https://github.com/saghul/libuv/tree/win-env-vars. I'm ready to land it, but when compiling with Visual Studio or MinGW-w64 I get a compilation warning about LOCALE_NAME_INVARIANT being redefined (the default seems to be L""). It seems to be defined in winnls.h. I'm not familiar with those APIs, but according to http://msdn.microsoft.com/en-us/library/windows/desktop/dd373795%28v=vs.85%29.aspx LOCALE_INVARIANT is 0x007f.

I removed the define and tests continue to pass. Am I really naive or can we drop it?

vtjnash:
We can drop it. I added that when I was attempting to use a different sort API, but it was giving the wrong sort order (the windows API's appear to typically convert to lowercase, rather than uppercase, before sorting, so `a` and `_` would end up in the wrong order)

The `LOCALE_INVARIANT` const also happens to have the same value (`0x007f`), and was the correct one to use for the `LCMapStringW` function.

saghul:
It (finally) landed on https://github.com/joyent/libuv/commit/2ce14cfab45eb62017e39017fd156e3cf1cd1060 and https://github.com/joyent/libuv/commit/8db42383ad247659d73f9e6812ec4339c0bf4e64.

Sorry about the delay @vtjnash, everyone else is busy and this patch was way outside of my confort zone :-/

PR #1279, opened 2014-05-12
unix: fix uv__open_cloexec usage

saghul:
It returns the fd or the negated errno.


ibc:
This fixes a crash when more than `ulimit -n` TPC connections are requested from a peer. With this commit the `on_connection` callback properly gives a "EMFILE" error instead of crashing.

So :cake: for @saghul.

indutny:
LGTM

saghul:
Thanks @indutny for having a lookg and @ibc for reporting and testing. Landed.

PR #1280, opened 2014-05-12
doc: refer to README.md to find out more information about tests

lperkov:
Reason for this change is because 'make test' command does not work.

Signed-off-by: Luka Perkov luka.perkov@sartura.hr


saghul:
Thanks Luka! Landed in https://github.com/joyent/libuv/commit/85e17ca66589444d5949337456d5d4ce9dee1441

PR #1282, opened 2014-05-12
windows: fix handling closed socket while poll handle is closing

### LINKED ISSUE #1278: Assertion failed: ((handle)->flags & UV__HANDLE_CLOSING) == 0
I'm having this issue with the [libcurl bindings for node](https://github.com/JCMais/node-libcurl) that I'm working on. The same issue is reproducible using the [code present on the uvbook](https://github.com/nikhilm/uvbook/blob/master/code/uvwget/main.c).

It happens only on Windows, the issue is that, when a connection timeout, libcurl try to create a new connection to the next ip, however it fails with the given assertion error.

Example of output:

```
>node examples\simple-request.js www.google.com:81
* Rebuilt URL to: www.google.com:81/
* Hostname was NOT found in DNS cache
*�� Trying 200.149.119.118...
Action: 2
Socket: 00000000
* After 2390ms connect time, move on!
* connect to 200.149.119.118 port 81 failed: Timed out
*�� Trying 200.149.119.89...
Action: 4
Socket: 006A0898
Action: 2
Socket: 00000000
Assertion failed: ((handle)->flags & UV__HANDLE_CLOSING) == 0, file src\win\poll.c, line 190 
```

From what I can see from the curl source code, the connection socket is closed (https://github.com/bagder/curl/blob/master/lib/connect.c#L584), and the socket callback is called (https://github.com/bagder/curl/blob/master/lib/connect.c#L1245 and https://github.com/bagder/curl/blob/master/lib/multi.c#L2092), however when the next connection is made, libuv tries to use a handler that was closed before. (because of the given socket callback being called).

I don't know how the handler is, somewhat, being reused, because the socket callback is called correctly (you can check that on the ouput above, Action 2 means register the uv_handle and 4 means destroy it), and it does close the handler, check: https://github.com/JCMais/node-libcurl/blob/develop/src/Curl.cc#L447.

Anyway, I don't know if this is a bug on libuv, libcurl, on with my code, probably the last one.

saghul:
fixes #1278

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -166,6 +166,7 @@ test_run_tests_SOURCES = test/blackhole-server.c \
                          test/test-pipe-server-close.c \
                          test/test-platform-output.c \
                          test/test-poll-close.c \
+                         test/test-poll-closesocket.c \
                          test/test-poll.c \
                          test/test-process-title.c \
                          test/test-ref.c \
--- a/src/win/poll.c
+++ b/src/win/poll.c
@@ -187,7 +187,8 @@ static void uv__fast_poll_process_poll_req(uv_loop_t* loop, uv_poll_t* handle,
     if (afd_poll_info->Handles[0].Events & AFD_POLL_LOCAL_CLOSE) {
       /* Stop polling. */
       handle->events = 0;
-      uv__handle_stop(handle);
+      if (uv__is_active(handle))
+        uv__handle_stop(handle);
     }
 
     if (events != 0) {
--- a/test/test-list.h
+++ b/test/test-list.h
@@ -240,6 +240,9 @@ TEST_DECLARE   (dlerror)
 TEST_DECLARE   (poll_duplex)
 TEST_DECLARE   (poll_unidirectional)
 TEST_DECLARE   (poll_close)
+#ifdef _WIN32
+TEST_DECLARE   (poll_closesocket)
+#endif
 TEST_DECLARE   (ip4_addr)
 TEST_DECLARE   (ip6_addr_link_local)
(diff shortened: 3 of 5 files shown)


indutny:
Test?

saghul:
I think I can write one, if my assumption about the problem actually stands. Will write one sson.

JCMais:
@saghul Any updates? (:

saghul:
I still need to write a test for this, but I'm a bit busy these days. Can you help with that?

JCMais:
I was just pinging, since it's common to forget when you have many issues to take care.

I can help, but if it's to write it, you are probably going to have more work if I do so, since I don't know where to start (4 weeks of experience with C/C++). =P

saghul:
No worries. I hope I find time for writing it next week.

saghul:
Added test. Can you PTAL, @indutny?

indutny:
LGTM, thank you!

sequoiar:
can back port to v0.8 ?

saghul:
To 0.10 yes, but I believe 0.8 is EOLd. Nevertheless, the patch should apply cleanly.

saghul:
Thanks Fedor, landed!

mscdex:
@saghul Any love for v0.10?

saghul:
Thanks for the reminder! Will do in a few days!

saghul:
Here goes the love: https://github.com/joyent/libuv/commit/12bb46c095032b5106ffbd6feb1931917dcdd9d9 :sparkling_heart:

mscdex:
:thumbsup:

PR #1283, opened 2014-05-12
unix, windows: avoid error in uv_fs_event_stop

saghul:
For consistency with other handle types, if stop is called while the
handle wasn't started yet, it's a no-op.


trevnorris:
This makes several checks in `test/test-fs-event.c` pointless. Might as well be removed?

saghul:
You mean the checks for the return value? Maybe not right now, but it's possible (in theory) that the function could fail further down and return something other than 0.

trevnorris:
ah, true. I only looked at `unix/kqueue.c`.

indutny:
LGTM

saghul:
Landed!

PR #1284, opened 2014-05-13
heap: fix `heap_remove()`

### LINKED ISSUE #1267: Question about heap_remove() in "src/heap-inl.h"
Two part of this function:
    1. Place the "max" node to be where the node to be removed is.
    2. And then try to move down the "max" node through heap as far as possible.

If I am right about what it does, my question is that there is no guarantee that the "max" node is greater than its parent. Thus moving up should also be tried. Right?

One of the case is like this:

```
      0
  5       1
6   7   2   N
```

`6` is to be removed and `2` is the "max" node. After moving `2` to `6`: 

```
      0
  5       1
2   7   N   N
```

`2` should move up.

the source code of this function is here:

```
HEAP_EXPORT(void heap_remove(struct heap* heap,
                             struct heap_node* node,
                             heap_compare_fn less_than)) {
  struct heap_node* smallest;
  struct heap_node** max;
  struct heap_node* child;
  unsigned int path;
  unsigned int k;
  unsigned int n;

  if (heap->nelts == 0)
    return;

  /* Calculate the path from the min (the root) to the max, the left-most node
   * of the bottom row.
   */
  path = 0;
  for (k = 0, n = heap->nelts; n >= 2; k += 1, n /= 2)
    path = (path << 1) | (n & 1);

  /* Now traverse the heap using the path we calculated in the previous step. */
  max = &heap->min;
  while (k > 0) {
    if (path & 1)
      max = &(*max)->right;
    else
      max = &(*max)->left;
    path >>= 1;
    k -= 1;
  }

  heap->nelts -= 1;

  /* Unlink the max node. */
  child = *max;
  *max = NULL;

  if (child == node) {
    /* We're removing either the max or the last node in the tree. */
    if (child == heap->min) {
      heap->min = NULL;
    }
    return;
  }

  /* Replace the to be deleted node with the max node. */
  child->left = node->left;
  child->right = node->right;
  child->parent = node->parent;

  if (child->left != NULL) {
    child->left->parent = child;
  }

  if (child->right != NULL) {
    child->right->parent = child;
  }

  if (node->parent == NULL) {
    heap->min = child;
  } else if (node->parent->left == node) {
    node->parent->left = child;
  } else {
    node->parent->right = child;
  }

  /* Walk down the subtree and check at each node if the heap property holds.
   * It's a min heap so parent < child must be true.  If the parent is bigger,
   * swap it with the smallest child.
   */
  for (;;) {
    smallest = child;
    if (child->left != NULL && less_than(child->left, smallest))
      smallest = child->left;
    if (child->right != NULL && less_than(child->right, smallest))
      smallest = child->right;
    if (smallest == child)
      break;
    heap_node_swap(heap, child, smallest);
  }
}
```

indutny:
Remove should shuffle items in both directions, not just down. It is
required, because `max` node could be not the actual maximum value in
the tree.

fix #1267


indutny:
cc @saghul

saghul:
Ok, I took a look at this and I think I have an idea of what's going on. Yes, it's true that when a node is removed we don't re-heapify upwards. However, if I'm not mistaken, there is one thing that holds true: the misplaced node is bigger than the heap min.

So, considering how we use it, aka, always remove the min when timers are processed, looks to me like we'd fix the heap anyway, because when removing the min we'd re-heapify downwards, from the root. Maybe it was a performance trick by @bnoordhuis?

bnoordhuis:
I don't remember, it's too long ago.  The patch itself - apart from the lack of capitalization and punctuation in the comment - looks trivially correct to me.

saghul:
The million timer benchmark doesn't show any noticeable slowdown, @indutny go and merge! :-)

indutny:
Thank you, landed in e002340

PR #1289, opened 2014-05-18
initial implementation of uv_promise_t

avalanche123:
uv_promise_t implementation as proposed in https://groups.google.com/forum/#!topic/libuv/nXZpATk7zI0 and https://gist.github.com/avalanche123/98cdbc79feb0feaea33f.


indutny:
Have you received any feedback from core developers?

saghul:
I haven't been able to go through the email yet, sorry about that. Will do as soon as I possibly can.

indutny:
The thing is that I don't see any connection or dependency to the loop in this `promise.c` file, I wonder if there is a point of having it within libuv?

saghul:
Folks, lets have the discussion about the concept on the mailing list and lets come back here in case we need to look at the code :-)

avalanche123:
@indutny you are right, there is not inherent connection to the event loop, it is more of a pattern implementation to make integration with application threads simpler.

saghul:
As I mentioned on the mailing list thread, there is no reason for this to be in libuv. It is a useful construct, however, so it would be nice to maybe have some libuv-extras repo where this could live. I'll ping you on that.

Closing since we won't integrate it in the libuv core library. Thanks for your effort though, I think this could be useful to others.

saghul:
@avalanche123 Can you send a PR to the libuv-extras repo? (https://github.com/libuv/libuv-extras) We'll use your contribution as the fist example for others :-)

(Just create a 'promise' directory with the C and H files, not sure what to do about tests yet)

avalanche123:
will do later today

saghul:
Thanks a lot!

PR #1293, opened 2014-05-20
socks5-proxy won't compile under windows

agrandville:
compile the sample under windows is not naturaly.
I propose several modifications to get this sample functional.


samples/socks5-proxy/build.gyp
```diff
@@ -19,9 +19,8 @@
 # IN THE SOFTWARE.
 
 {
-  'targets': [
+	'targets': [
     {
-      'dependencies': ['../../uv.gyp:libuv'],
```
indutny, line 24: You should probably try placing this above the `targets`:

``` python
'variables': {
  'uv_library': 'static'
}
```

samples/socks5-proxy/defs.h
```diff
@@ -59,11 +61,11 @@ typedef struct {
   } handle;
   uv_timer_t timer_handle;  /* For detecting timeouts. */
   uv_write_t write_req;
+  uv_getaddrinfo_t addrinfo_req;
```
indutny, line 64: Why?

samples/socks5-proxy/main.c
```diff
@@ -63,9 +63,9 @@ const char *_getprogname(void) {
 static void parse_opts(server_config *cf, int argc, char **argv) {
   int opt;
 
-  while (-1 != (opt = getopt(argc, argv, "H:hp:"))) {
+  while (-1 != (opt = getopt(argc, argv, "b:hp:"))) {
```
indutny, line 66: Why?

agrandville, line 66: D:...\socks5-proxy\build\Debug>s5-proxy.exe -h
Usage:

  s5-proxy.exe [-b <address>] [-h] [-p <port>]

Options:

  -b <hostname|address>  Bind to this address or hostname.
Default: "127.0.0.1"
  -h                     Show this help message.
  -p <port>              Bind to this port number.  Default: 1080

samples/socks5-proxy/defs.h
agrandville, line 64: the function conn_connect (client.c@634) calls
return uv_tcp_connect(&c->t.connect_req,
                        &c->handle.tcp,
                        &c->t.addr,
                        conn_connect_done);

uv_tcp_connect will reset the connect_req struct (tcp.c@725) but this memory address is share with  addr,addr4,addr6, doing so will leads in rewritting the tcp address with junk data, that's why I moved the three req structs outside the union to prevent this behavior.

samples/socks5-proxy/build.gyp
agrandville, line 24: all right

samples/socks5-proxy/build.gyp
```diff
@@ -36,7 +35,12 @@
       'conditions': [
         ['OS=="win"', {
           'defines': ['HAVE_UNISTD_H=0'],
-          'sources': ['getopt.c']
+          'sources': ['getopt.c'],
+		  'link_settings': {
+			'libraries': ['-llibuv.lib','-ladvapi32.lib','-lWs2_32.lib','-lPsapi.lib','-lIphlpapi.lib'],
+			'library_dirs': ['../../Debug/lib']
+			},
+		  'include_dirs': ['../../include']
```
saghul: Are these necessary if uv.gyp is included again?

agrandville: well seen. I just do the test, if dependency is working, there is no more need to link libuv.lib

PR #1295, opened 2014-05-22
unix, windows: fix parsing scoped IPv6 addresses

saghul:


saghul:
Test is failing and I could reproduce it. Basically, inet_pton4 needs the buffer to be NULL terminated in order not to fail, so do that. As for why it works on Linux, I have no idea.

/cc @indutny

src/inet.c
```diff
@@ -175,7 +175,7 @@ static int inet_pton4(const char *src, unsigned char *dst) {
 
   saw_digit = 0;
   octets = 0;
-  *(tp = tmp) = 0;
+  memset((tp = tmp), '\0', sizeof tmp);
```
indutny: better split it into two lines

indutny: `sizeof(tmp)`

indutny: I guess it should be fine to use `0` instead of `'\0'`, we do it everywhere anyway.

indutny: Actually, wouldn't it be simpler to add one zero after the loop?

indutny: Ah, gosh. nvm.

indutny: After some reading, it seems to be totally irrelevant change.

saghul: We do it in inet_pton6, so I applied it here as well, and it made the tests pass.

saghul:
Style is not our usual, I just made it look like the suroundings.

saghul:
Here is Jenkins passing tests: http://jenkins.nodejs.org/job/libuv-pullrequest/620/ see previous builds for failures.

src/inet.c
```diff
@@ -264,8 +264,8 @@ static int inet_pton6(const char *src, unsigned char *dst) {
         char tmp[sizeof "255.255.255.255"];
 
         memcpy(tmp, curtok, pch - curtok);
+        tmp[pch-curtok] = '\0';
```
indutny: `pch - curtok`

src/inet.c
indutny: I guess the line below should be responsible for fixing the failure, if it is not - could you please explain why it is required here?

hasufell:
Where can I see the build log in jenkins?

saghul:
@hasufell here: http://jenkins.nodejs.org/job/libuv-pullrequest/620/

hasufell:
I mean the "make output". Either I am blind or it is not there.

saghul:
You can click on each build and check the console: http://jenkins.nodejs.org/job/libuv-pullrequest/620/DESTCPU=ia32,label=linux/console

hasufell:
Afair these kind of memsets are likely to get optimized out. OpenSSL had similar trouble too, see https://github.com/jmhodges/libssl/commit/a44713614e5b5716e4cc01194408b59eb63f4360

saghul:
Hum, SmartOS 64 bit seems to be failing regardless. Sigh.

hasufell:
I think I fixed it, sec.

saghul:
@hasufell kewl! What did you do? I tried removing the memset, and it fails on Linux again. I also tried to rewrite the test and memset the dst buffer, to no avail.

hasufell:
https://gist.github.com/hasufell/7fead0237c77114acdd8

hasufell:
line 35 of the patch was crucial

saghul:
> line 35 of the patch was crucial

But that can be fixed by doing this: https://github.com/saghul/libuv/commit/b8e879d55503e249fe0d80699a2b15812f9105ba#diff-b67ba0a4ee7f362844a46a9aefb38f62R267

Let me try to combine both and run jenkins.

hasufell:
> But that can be fixed by doing this:

I don't think so. I tried that one and it seems to get optimized out.

hasufell:
Of course the patch is not mergeable the way I posted it. explicit_bzero shouldn't be exposed like that and it may also be already defined in BSD systems.

saghul:
Can you guys test again now? I defined uv__bzero, internal to inet.c and the test passes on all jenkins bots.

hasufell:
That worked for me.

saghul:
@indutny thoughts?

saghul:
@hasufell thanks for confirming and all the help testing / figuring it out!

indutny:
@saghul oh gosh, I guess it is fine, if it works. @bnoordhuis: do you have any suggestions?

bnoordhuis:
Suggestions on what?

saghul:
> Suggestions on what?

The approach. What seems to be happening is that some compilers seem to optimize out our memset and as a result we might copy some garbage around. The current patch which seems to solve the problem for the reporter (I couldn't repiably reproduce it myself) is based on some `uv__bzero` internal function, which is basically like `explicit_bzero` in BSDs.

My experience with compiler woes is quite limited, I'd love your 2 cents if you've got the time :-)

bnoordhuis:
Right, I see what you mean.

I'm somewhat distrustful of the current approach.  Yes, a volatile pointer will invoke memset() unconditionally but that's probably only addressing the symptoms, not curing the actual disease.

Ask yourself why the compiler optimizes away the call to memset.  There are two possibilities: either there is a bug in the compiler (possible but unlikely) or there is a bug in the code.  At a quick glance, there are at least two places that have undefined behavior so I'm betting on the second one.  As a point in case:

``` c
       if (tp + sizeof(uint16_t) > endp)
         return UV_EINVAL;
```

A pointer is only allowed to point one element after the end of the array so this is UB.  A compiler can (and modern compilers will) take advantage of that.  I suggest auditing the code _very_ carefully.

saghul:
Thanks for your input @bnoordhuis, much appreciated. I'll take a deeper look!

saghul:
Ok, I took a different approach at the problem. Here is what was happening (to the best of my knowledge):  we wanted to skip anything beyond '%' because that's the scope ID delimiter. In doing so, we become a bit more forgiving, examples:

fe80::21e:58ff:fe30:cebe (valid)
fe80::21e:58ff:fe30:cebe%eth0 (now valid)
fe80::21e:58ff:fe30:cebe:1.2.3.4 (valid)
fe80::21e:58ff:fe30:cebe:1.2.3.4%eth0 (now valid)
fe80::21e:58ff:fe30:cebe:1.2.3.4%eth0:hello-world (now valid, WAT)
fe80::21e:58ff:fe30:cebe:1.2.3.4%%%% (now valid, WAT)

The problem was when handling the 4th case in the list. In that case, when a '.' is found, the rest is parsed with inet_pton4, so it needs to be a valid IPv4 address. "1.2.3.4%eth0" is not a valid IP address, so the code looked for the '%' char, and copied the memory before it to a temporary buffer, which was used to do the IPv4 address parsing.

For some reasons (unkown to me, really) properly NULL-terminating the buffer didn't cut it. For the life of me, I cannot see why. GCC 4.8 on my Debian (64 bit) seemed to do the right thing, whereas the GCC 4.4 on the Jenkins buildbots swallowed it. The SmartOS 64 bit builbot also misbehaved.

So I had a look at the only implementation that seems to handle this case properly: Apple's libc. The current approach is based on what Apple's libc does: if the source buffer contains the '%' char, copy the buffer, NULL-terminate at the right place, operate on the copy and be done with it. My approach here is a bit more forgiving because it strips anything after the first '%', Apple's looks at the last one. I'm undecided but we should probably make it a bit more strict so that the last case on the above list gives an error.

Thoughts?

PS: I ran a Jenkins build and test passes on all machines, which only did when I tried the explicit_bzero approach. So, I think we are good.

saghul:
Fore the curious, Apple's inet_pton.c: http://www.opensource.apple.com/source/Libc/Libc-997.90.3/net/inet_pton.c

src/inet.c
```diff
@@ -156,11 +157,28 @@ static int inet_ntop6(const unsigned char *src, char *dst, size_t size) {
...
     return (inet_pton4(src, dst));
-  case AF_INET6:
-    return (inet_pton6(src, dst));
+  case AF_INET6: {
+    int status;
+    char *s, *p;
+    s = (char*) src;
+    p = strchr(src, '%');
+    if (p != NULL) {
+      s = strdup(src);
```
bnoordhuis: This could be an on-stack buffer, IPv6 addresses are only INET6_ADDRSTRLEN bytes long.

bnoordhuis:
If invalid interface names are an issue, you could validate them with if_nametoindex().  It's in POSIX and I'm reasonably sure all supported platforms support it, even Windows.

src/inet.c
saghul: > This could be an on-stack buffer, IPv6 addresses are only INET6_ADDRSTRLEN bytes long.

Yeah, I considered it, but somehow that was the problem that brought me here so well, I I not so big allocation shouldn't be that bad :-)

indutny: Hm... -1 for this

indutny:
I'm really afraid of this change, it looks generally correct, but it bothers me that we don't know actual cause of this problem

saghul:
> If invalid interface names are an issue, you could validate them with if_nametoindex(). It's in POSIX and I'm reasonably sure all supported platforms support it, even Windows.

I don't think they are an issue really. We just ignore whatever is after the first '%', which is a bit more forgiving. If we would take the last '%' we wouldn't allow some invalid values.

As for using if_nametoindex, the recommended way of doing it is using getaddrinfo with AI_PASSIVE and AI_NUMERICHOST, AFAIK, so skipping if here seems reasonable, seems it's not stored in the in6_addr structure.

saghul:
> I'm really afraid of this change, it looks generally correct, but it bothers me that we don't know actual cause of this problem

The (current) change is straightforward. The reason behind it is in my (long) note. It has to do with what compilers do, so it's somewhat out of my control, I'm just trying to get things right.

The current state is bad, we either fix it or revert https://github.com/joyent/libuv/commit/2b2141924282a046b9f199374da2aa0da058656a because right now it just fails arbitrarily, based in the phase of the moon :-/

src/inet.c
saghul: Why?

indutny: Because it doesn't need to allocate, really.

hasufell:
If this would only happen on "-O3" I would say "don't bother", but the fact that this fails on all optimization levels (in case you are affected) is really a bit weird. If you want, I can try to figure out the exact "-ffoo" flag(s) that is causing this. That might help with understanding the issue.

saghul:
Updated with an on stack temporary buffer. FWIW, I liked the strdup version more, it was simpler, but whatever.

saghul:
@hasufell mind trying the new patch(es)? It would indeed be interesting to know in which cases it failed, but TBH I like the current approach (passing a sanitized buffer to inet_pton6) better and I spent a lot more time than I originally had on this :-(

hasufell:
I can't figure it out, now it even fails with -O0, lol (I mean the master branch here). Can't even run the test binary under valgrind. It errors out with "unrecognised instruction...".

saghul:
> I can't figure it out, now it even fails with -O0, lol (I mean the master branch here). Can't even run the test binary under valgrind. It errors out with "unrecognised instruction...".

With or without this patch?

hasufell:
without, I was testing the master branch from this repository

indutny:
@saghul could you paste a disasm of this method on a failing build?

indutny:
I meant build with a test failure

saghul:
@indutny no, I can't. I could only reproduce the obvious initial bug on OSX, but after null terminating the buffer it goes away. On the jenkins buildots, however, that's not enough so that's why I went down the rabbit hole, but I could never reproduce it myself.

indutny:
Ok, whatever... LGTM :) Thank you for all your efforts!

saghul:
Thanks @indutny, this was driving me nuts. @hasufell @mtorromeo can you guys please test the latest patch? Thanks!

hasufell:
works for me with both gcc and clang

saghul:
Thanks for confirming!

hasufell:
I guess I can safely backport this to 0.11.25?

saghul:
Sure.

saghul:
Thank you all for the help. fixes landed in master.

PR #1296, opened 2014-05-22
Fix build on mingw

mattn:
- add getnameinfo.c to Makefile.mingw
- define WSAPOLLFD on mingw
- call getnameinfo if GetNameInfo is not defined, and it out localized characters, so convert to wide characters.
- define COMMON_LVB_REVERSE_VIDEO if not defined

### DIFF
--- a/src/win/getnameinfo.c
+++ b/src/win/getnameinfo.c
@@ -27,18 +27,6 @@
 #include "internal.h"
 #include "req-inl.h"
 
-#ifndef GetNameInfo
-int WSAAPI GetNameInfoW(
-  const SOCKADDR *pSockaddr,
-  socklen_t SockaddrLength,
-  PWCHAR pNodeBuffer,
-  DWORD NodeBufferSize,
-  PWCHAR pServiceBuffer,
-  DWORD ServiceBufferSize,
-  INT Flags
-);
-#endif
-
 /* getnameinfo worker thread implementation */
 static DWORD WINAPI getnameinfo_thread_proc(void* parameter) {
   uv_getnameinfo_t* req = (uv_getnameinfo_t*)parameter;
@@ -49,6 +37,7 @@ static DWORD WINAPI getnameinfo_thread_proc(void* parameter) {
 
   assert(req != NULL);
 
+#ifdef GetNameInfo
   ret = GetNameInfoW((struct sockaddr*)&req->storage,
                      sizeof(req->storage),
                      host,
@@ -56,6 +45,32 @@ static DWORD WINAPI getnameinfo_thread_proc(void* parameter) {
                      service,
                      sizeof(service),
                      req->flags);
+#else
+  {
+    char _host[NI_MAXHOST];
+    char _service[NI_MAXSERV];
+    ret = getnameinfo((struct sockaddr*)&req->storage,
+                       sizeof(req->storage),
+                       _host,
+                       sizeof(_host),
+                       _service,
+                       sizeof(_service),
+                       req->flags);
+    /* convert multi-bytes to wide chars */
+    MultiByteToWideChar(CP_THREAD_ACP,
+                       0,
+                       _host,
+                       sizeof(_host),
+                       host,
+                       sizeof(host));
+    MultiByteToWideChar(CP_THREAD_ACP,
+                       0,
+                       _service,
+                       sizeof(_service),
+                       service,
+                       sizeof(service));
+  }
+#endif
   req->retcode = uv__getaddrinfo_translate_error(ret);
 
   /* convert results to UTF-8 */
--- a/src/win/getnameinfo.c
+++ b/src/win/getnameinfo.c
@@ -27,6 +27,17 @@
 #include "internal.h"
 #include "req-inl.h"
 
+#ifndef GetNameInfo
+int WSAAPI GetNameInfoW(
+  const SOCKADDR *pSockaddr,
+  socklen_t SockaddrLength,
+  PWCHAR pNodeBuffer,
+  DWORD NodeBufferSize,
+  PWCHAR pServiceBuffer,
+  DWORD ServiceBufferSize,
+  INT Flags
+);
+#endif
 
 /* getnameinfo worker thread implementation */
 static DWORD WINAPI getnameinfo_thread_proc(void* parameter) {
--- a/include/uv-win.h
+++ b/include/uv-win.h
@@ -31,7 +31,7 @@ typedef intptr_t ssize_t;
 
 #include <winsock2.h>
 
-#if defined(__MINGW32__) && !defined(__MINGW64__)
+#if defined(__MINGW32__) && !defined(__MINGW64_VERSION_MAJOR)
 typedef struct pollfd {
   SOCKET fd;
   short  events;
--- a/include/uv-win.h
+++ b/include/uv-win.h
@@ -30,6 +30,15 @@ typedef intptr_t ssize_t;
 #endif
 
 #include <winsock2.h>
+
+#if defined(__MINGW32__) && !defined(__MINGW64__)
+typedef struct pollfd {
+  SOCKET fd;
+  short  events;
+  short  revents;
+} WSAPOLLFD, *PWSAPOLLFD, *LPWSAPOLLFD;
+#endif
+
 #include <mswsock.h>
 #include <ws2tcpip.h>
 #include <windows.h>
--- a/src/win/tty.c
+++ b/src/win/tty.c
@@ -30,6 +30,10 @@
 # include <stdint.h>
 #endif
 
+#ifndef COMMON_LVB_REVERSE_VIDEO
+# define COMMON_LVB_REVERSE_VIDEO 0x4000
+#endif
+
 #include "uv.h"
 #include "internal.h"
 #include "handle-inl.h"


saghul:
Hey @mattn, thank you for this! Can you please explain why some of the changes are needed? We never needed to define WSAPOLLFD before, nor COMMON_LVB_REVERSE_VIDEO, and AFAIS GetNameInfoW is supported on Windows >= XP SP2, which is enough. What MinGW version are you using? and what Windows version?

mattn:
mingw32 doesn't have definition of WSAPOLLFD. Also COMMON_LVB_REVERSE_VIDEO.
I know GetNameInfoW can be used after XP SP2, But mingw32 doesn't have definition of that.

```
gcc (gcc-4.6.3 release with patches [build 20121012 by perlmingw.sf.net]) 4.6.3
Copyright (C) 2011 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

```

```
C:\mingw\include>grep WSAPOLLFD *
mswsock.h:int WSAAPI WSAPoll(WSAPOLLFD, ULONG, INT);

C:\mingw\include>
```

It don't have definition it self of WSAPOLLFD.

```
C:\mingw\include>grep COMMON_LVB_REVERSE_VIDEO *
```

this too.

saghul:
I have a quite fresh copy of mingw-w64 and I do have all those definitions. Given the sorry state of MinGW (MinGW.org vs MinGW-w64 vs whatever other pack) I'm not sure what is the correct way of pleasing everyone.

I'll be away for a couple of weeks, but when I'm back I'll make sure everything works correctly on MinGW-w64 (both 32 and 64 bits), update the README and call it a day.

Long story short: you need a better MinGW.

I'm keeping this open as a reminder to myself, but I won't be merging the changes for the aforementioned reasons, sorry. It's not your fault, it's MinGW's.

mattn:
NP. :)

saghul:
Ok, I'm closing this now. I just pushed a fix and there are now no compilation warnings on MinGW-w64, both on 32 and 64 bits. The error in this error seems to happen with the older MinGW toolchains, I can reproduce it with:

```
$ i586-mingw32msvc-cc -v
Using built-in specs.
Target: i586-mingw32msvc
Configured with: /home/ron/devel/debian/mingw32/mingw32-4.2.1.dfsg/build_dir/src/gcc-4.2.1-2-dfsg/configure -v --prefix=/usr --target=i586-mingw32msvc --enable-languages=c,c++ --enable-threads --enable-sjlj-exceptions --disable-multilib --enable-version-specific-runtime-libs
Thread model: win32
gcc version 4.2.1-sjlj (mingw32-2)
```

So, for future reference, my advice is to use MinGW-w64.

mattn:
This issue is not fixed yet.

```
cc -Wall -Wextra -Wno-unused-parameter -Iinclude -Isrc -Isrc/win -DWIN32_LEAN_AND_MEAN -D_WIN32_WINNT=0x0600 -c -o src/fs-poll.o src/fs-poll.c
In file included from include/uv-win.h:33:0,
                 from include/uv.h:59,
                 from src/fs-poll.c:22:
c:\mingw\include\mswsock.h:125:31: error: expected ')' before 'ULONG'
 int WSAAPI WSAPoll(WSAPOLLFD, ULONG, INT);
                               ^
Makefile.mingw:82: recipe for target 'src/fs-poll.o' failed
mingw32-make: *** [src/fs-poll.o] Error 1
```

mattn:
This caused by a reason that WSAPOLLFD is not defined on mingw32

saghul:
As I said here: https://github.com/joyent/libuv/pull/1296#issuecomment-44375281 you need a better MinGW.

mattn:
I'm thinking that there are many people using mingw32 (not mingw-w64) still in the world.

saghul:
I'll send an email to the mailing list requesting some feedback about this. At least Debian has packages for both cross compilers, so testing is not that hard, but everything adds up...

mattn:
OK, currently, to build libuv on mingw32, it need a patch following.

``` diff
diff --git a/include/uv-win.h b/include/uv-win.h
index 2f24dfe..86e8413 100644
--- a/include/uv-win.h
+++ b/include/uv-win.h
@@ -30,6 +30,15 @@ typedef intptr_t ssize_t;
 #endif

 #include <winsock2.h>
+
+#if defined(__MINGW32__) && !defined(__MINGW64__)
+typedef struct pollfd {
+  SOCKET fd;
+  short  events;
+  short  revents;
+} WSAPOLLFD, *PWSAPOLLFD, *LPWSAPOLLFD;
+#endif
+
 #include <mswsock.h>
 #include <ws2tcpip.h>
 #include <windows.h>
diff --git a/src/win/tty.c b/src/win/tty.c
index 87e3eb5..6b8297c 100644
--- a/src/win/tty.c
+++ b/src/win/tty.c
@@ -30,6 +30,10 @@
 # include <stdint.h>
 #endif

+#ifndef COMMON_LVB_REVERSE_VIDEO
+# define COMMON_LVB_REVERSE_VIDEO 0x4000
+#endif
+
 #include "uv.h"
 #include "internal.h"
 #include "handle-inl.h"
```

saghul:
I reopened the PR, mind force-pushing your changes?

mattn:
No problem.

mattn:
If you need a patch against master latest master branch, I'll make it

saghul:
That would be ideal :-)

saghul:
FTR: https://groups.google.com/forum/#!topic/libuv/iKn0sBXuPgM

mattn:
Pushed forcely.

src/win/getnameinfo.c
```diff
@@ -45,6 +46,32 @@ static DWORD WINAPI getnameinfo_thread_proc(void* parameter) {
...
+    char _service[NI_MAXSERV];
+    ret = getnameinfo((struct sockaddr*)&req->storage,
+                       sizeof(req->storage),
+                       _host,
+                       sizeof(_host),
+                       _service,
+                       sizeof(_service),
+                       req->flags);
+    /* convert multi-bytes to wide chars */
+    MultiByteToWideChar(CP_THREAD_ACP,
```
saghul, line 60: Is this conversion required, given that we seem to do the opposite a bit down below?

mattn:
What you do mean `bit down`? About conversion failure?

saghul:
On line 78 we conver the wide chars to multibyte. Here we convert multibyte to wide char. Is this double conversion necessary when using `getnameinfo` instead of `GetNameInfoW`?

mattn:
In japanese, we are using cp932 encoding. And CJK people uses others. This part convert `locale encoding` characters to `wide characters`, and convert `wide characters` into `utf-8 encoding` characters.
Windows doesn't provide APIs which can convert `locale characters` to `utf-8 characters` directly.

mattn:
Ah, sorry

http://msdn.microsoft.com/en-us/library/windows/desktop/ms738531(v=vs.85).aspx

GetNameInfoW is provided after Windows XP. So it shouldn't put the covert. it should be only adding prototype of missing definition.

mattn:
I'll fix in soon.

mattn:
Sorry about noisy replies. Done.

saghul:
Thanks!

k-takata:
```
#if defined(__MINGW32__) && !defined(__MINGW64__)
```

This line might cause a problem on MinGW-w64 32-bit version because `__MINGW64__` isn't defined. (It is only defined on 64-bit version.)
Maybe the following is better?:

```
#ifndef POLLERR
```

`POLLERR` will be defined only when `WSAPOLLFD` is typedef'ed.

saghul:
This seems to be the correct check for MinGW32 / w64:

```
#if defined(__MINGW32__) && !defined(__MINGW64_VERSION_MAJOR)
```

k-takata:
Using `__MINGW64_VERSION_MAJOR` seems better than mine.

saghul:
Hi there,

I tested this, but unfortunately it's not enough :-( I had to disable CRT debugging on MinGW32, though it works on MinGW-w64: https://github.com/saghul/libuv/commit/922d6e4c5774f32d1cdba0a04ba25285a3f05f44

Also, aligned allocation function are not present, so: https://github.com/saghul/libuv/commit/772b9ae2e4f03341899644e65cb147241e0c3586

Now, after adding those to your changes, I get:

```
src/win/.libs/libuv_la-getaddrinfo.o:getaddrinfo.c:(.text+0x4d): undefined reference to `__imp__GetAddrInfoW@16'
src/win/.libs/libuv_la-getaddrinfo.o:getaddrinfo.c:(.text+0x3d7): undefined reference to `__imp__FreeAddrInfoW@4'
src/win/.libs/libuv_la-getnameinfo.o:getnameinfo.c:(.text+0x6a): undefined reference to `_GetNameInfoW@28'
```

I guess defining GetNameInfoW is not enough. Can you add the fallback to plain ascii `getnameinfo` like you did before?

Thanks!

saghul:
Ping :-)

mattn:
It's wired thing...
However using ANSI API fixes this issue.

saghul:
Thanks! Almost there! Looks like GetAddrInfoW and FreeAddrInfoW are also missing:

```
src/win/.libs/libuv_la-getaddrinfo.o:getaddrinfo.c:(.text+0x2d): undefined reference to `__imp__GetAddrInfoW@16'
src/win/.libs/libuv_la-getaddrinfo.o:getaddrinfo.c:(.text+0xf3): undefined reference to `__imp__FreeAddrInfoW@4'
src/win/.libs/libuv_la-getaddrinfo.o:getaddrinfo.c:(.text+0x3e4): undefined reference to `__imp__FreeAddrInfoW@4'
```

Mind provising a patch for that too?

mattn:
Really? It's too strange.

Those APIs are containing after Windows XP.

http://msdn.microsoft.com/en-us/library/windows/desktop/ms738531(v=vs.85).aspx

saghul:
Yeah, AFAIS they should be there :-/ Not sure if it could be due to the mingw32 toolchain on Debian, unfortunately I have no way of testing it on Windows ATM.

saghul:
Hum, looks related to the runtime version or something, which seems to be ancient on Debian. I downloaded a more recent one from https://wiki.videolan.org/Win32Compile/#Mingw32 but got the same.

This probably means that the change for GetNameInfoW was unnecessary :-( I'll try to setup a MinGW32 environment on Windows and test this out, not to waste your time.

Thanks for your effort!

saghul:
Finally managed to get around this. Landed https://github.com/joyent/libuv/commit/636b108d3780006a5dbed14b48bb9acfa3d61957 and https://github.com/joyent/libuv/commit/b17223cd607718abcd79610a6bd2057f523df971, which fix the build in MinGW32.

Unfortunately, the cross-compiler toolchain in Debian and Ubuntu is too old so I cannot create a Jenkins configuration to test compilation. At least I did get the environment up and running on my Windows 7 machine. Sigh.

Anyway, my advice to you, future reader, is to stick to MinGW-w64.

mattn:
Awesome! thank you

PR #1300, opened 2014-05-29
unix: block SIGPROF in worker threads

bnoordhuis:
Ensure that SIGPROF-based profilers like V8's never wake up on one of
our worker threads.  V8 handles it graciously by simply dropping the
sample but that means the aggregated result becomes increasingly
unreliable.

/cc @indutny - should be back-ported to v0.10.


bnoordhuis:
On a related note, it's worth considering blocking all signals in the worker threads.

indutny:
This is good for node, but what about other users? What if people do want to profile worker threads with SIGPROF-based profiler?

+1 for it, but it should be an option, probably enabled by default.

bnoordhuis:
I suspect that I'm hitting an OS X kernel bug.  V8 sends the signals with pthread_kill() but a small fraction still end up being delivered to other threads.  It works okay on Linux so I'm going to close this for now.

PR #1304, opened 2014-06-02
unix, windows: make uv_barrier_wait() return bool

bnoordhuis:
Make uv_barrier_wait() return a boolean value indicating whether this
thread was the "serializer".

From `man pthread_barrier_wait`:

```
Upon successful completion, the pthread_barrier_wait() function
shall return PTHREAD_BARRIER_SERIAL_THREAD for a single (arbitrary)
thread synchronized at the barrier and zero for each of the other
threads.
```

Exposing that information from libuv is useful because it can make
cleanup significantly easier:

```
if (uv_barrier_wait(&barrier) > 0)
  uv_barrier_destroy(&barrier);
```


src/unix/thread.c
```diff
@@ -411,7 +413,8 @@ void uv_barrier_wait(uv_barrier_t* barrier) {
   uv_sem_post(&barrier->turnstile1);
 
   uv_mutex_lock(&barrier->mutex);
-  if (--barrier->count == 0) {
+  serial_thread = (--barrier->count == 0);
```
indutny, line 416: Why parens?

test/test-barrier.c
```diff
@@ -38,7 +40,11 @@ static void worker(void* arg) {
   if (c->delay)
     uv_sleep(c->delay);
 
-  uv_barrier_wait(&c->barrier);
+  c->worker_barrier_wait_rval = uv_barrier_wait(&c->barrier);
+  if (c->worker_barrier_wait_rval) {
```
indutny: hrm... either do `== 1` here or drop it below.

src/unix/thread.c
bnoordhuis, line 416: Readability.

test/test-barrier.c
bnoordhuis: Oh, good point.  I'll add that.

src/unix/thread.c
indutny, line 416: Ok...

indutny:
Few nits, otherwise LGTM

bnoordhuis:
Updated, PTAL.

indutny:
Landed in 91985d48f6b5477bf198fd6f92a4f3434ba16310, b64c359557edd7bac3f133a67a4f9ffeda1a9628, and 6c6c9221679c6cc66ac30fbad5ce086658113ea3. Thank you!

PR #1306, opened 2014-06-03
uv_fs_req_result method

gigi81:
I added this method because it makes easier and safer on a c# libuv binding library to get the result of an fs request on a callback. Without this method I would have to replicate and then marshal the entire uv_fs_t structure. This is prone to errors and difficult to maintain: if you ever change the internals of uv_fs_t my library will break. As long as it's a really small change in libuv, can you please consider to merge this patch? I'm pretty much sure this will come useful to other binding libraries. Thanks


txdv:
ebaaf52 and 6de082b are doing what you describe and they should be one commit, the rest is not the targeted issue in this pull request.

txdv:
By the way, you can calculate the offset easily with uv_req_sizeof and use the struct definition with unsafe code like here: https://github.com/txdv/LibuvSharp/blob/master/LibuvSharp/Internal/FileSystemRequest.cs#L17

If you go the route of making a function for every field which should be accessible, you get a lot of 5 liner code added like in your two commits which were of relevance to your problem. Makes it easier for FFI bindings though.

gigi81:
I messed up a little with my fork. I will have another pull request just with the relevant commit.
The purpose of the method is exactly to avoid the structure declaration that is useless. This is the only situation where it's needed. For tcp and other structures this is not needed.

PR #1307, opened 2014-06-04
Added uv_fs_req_result method

gigi81:
This is single commit of the pull request #1306.

I added this method because it makes easier and safer on a c# libuv binding library to get the result of an fs request on a callback. Without this method I would have to replicate and then marshal the entire uv_fs_t structure. This is prone to errors and difficult to maintain: if you ever change the internals of uv_fs_t my library will break. As long as it's a really small change in libuv, can you please consider to merge this patch? I'm pretty much sure this will come useful to other binding libraries. Thanks


saghul:
Hum. This would make life of ffi users easy, since we have https://github.com/joyent/libuv/blob/master/include/uv.h#L554 to get the size. However, there are other fields on that structure which are also needed on different fs operations. @indutny how do you feel about this? Should we add a one-liner for each meaningful structure member here?

txdv:
Would be quite a lot of functions, but, it would make FFI functions independent of any struct change.

I am personally for this, because currently my binding code looks ugly as hell, with hardcoded offsets and such.

PR #1311, opened 2014-06-07
windows: fix PYTHON environment variable usage

jay:
- (README.md) The PYTHON environment variable should point to python.exe
  not the python directory.
- (vcvars.bat) The PYTHON environment variable should be quoted on
  expansion in case there's spaces in the path.

---

The PYTHON environment variable should be quoted in case there's spaces in the path. For example assume `set PYTHON=C:\some path\python.exe` you'd want the command to be `"C:\some path\python.exe"` so PYTHON would need to be expanded in quotes `"%PYTHON%"`. This is typical in Windows because environment variables usually don't contain quotes and instead you're expected to quote on expansion.

Your contributing document says bugfixes go into the stable branch not master but I don't see a branch 'stable'.


saghul:
LGTM. Have you signed the CLA?

jay:
Yeah I just submitted it as Ray Satiro my legal name. If you need it to match the name on my commit (my nickname Jay) I can submit again.

saghul:
No worries, that should be fine. I currently don't have access to a proper
computer, I'll merge it as soon as I do. Thanks!

saghul:
Thanks @jay, landed in https://github.com/joyent/libuv/commit/90e4afd4fa3fe4910a686f8bf50cb6d8cc7ba5e9 :+1:

PR #1312, opened 2014-06-11
doc: remove requirement for signing the CLA

saghul:
I guess we can get rid of this now :-)

/cc @tjfontaine @indutny


indutny:
yay!

indutny:
LGTM

saghul:
It's done! :sparkles:

PR #1316, opened 2014-06-16
unix, windows: return system error on EAI_SYSTEM

saghul:
This should fix this problem: https://github.com/joyent/node/issues/7742#issuecomment-46074338

Windows doesn't have EAI_SYSTEM apparently.

I'm not providing a test since this error is only triggered for system failures, which could be anything really...

### DIFF
--- a/include/uv-errno.h
+++ b/include/uv-errno.h
@@ -39,7 +39,6 @@
 #define UV__EAI_OVERFLOW    (-3009)
 #define UV__EAI_SERVICE     (-3010)
 #define UV__EAI_SOCKTYPE    (-3011)
-#define UV__EAI_SYSTEM      (-3012) /* TODO(bnoordhuis) Return system error. */
 #define UV__EAI_BADHINTS    (-3013)
 #define UV__EAI_PROTOCOL    (-3014)
 
--- a/include/uv.h
+++ b/include/uv.h
@@ -83,7 +83,6 @@ extern "C" {
   XX(EAI_PROTOCOL, "resolved protocol is unknown")                            \
   XX(EAI_SERVICE, "service not available for socket type")                    \
   XX(EAI_SOCKTYPE, "socket type not supported")                               \
-  XX(EAI_SYSTEM, "system error")                                              \
   XX(EALREADY, "connection already in progress")                              \
   XX(EBADF, "bad file descriptor")                                            \
   XX(EBUSY, "resource busy or locked")                                        \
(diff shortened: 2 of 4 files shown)


saghul:
Not sure if we should "fix" this for v0.10, but here is my suggestion: if the return code is EAI_SYSTEM here: https://github.com/joyent/libuv/blob/v0.10/src/unix/getaddrinfo.c#L33, set the result to 'errno'. Then the UV error will be UV_EADDRINFO, and the sys_errno_ the actual errno.

indutny:
LGTM, not sure about v0.10 either.

saghul:
Here is the fix for 0.10:

```
diff --git a/src/unix/getaddrinfo.c b/src/unix/getaddrinfo.c
index 283d295..f6d8a78 100644
--- a/src/unix/getaddrinfo.c
+++ b/src/unix/getaddrinfo.c
@@ -34,6 +34,8 @@ static void uv__getaddrinfo_work(struct uv__work* w) {
                              req->service,
                              req->hints,
                              &req->res);
+  if (req->retcode == EAI_SYSTEM)
+    req->retcode = errno;
 }
```

Would this be enough for 0.10 @tjfontaine? err.sys_errno_ will contain the actual errno.

saghul:
I came up with another way for fixing it on 0.10:

```
diff --git a/src/unix/getaddrinfo.c b/src/unix/getaddrinfo.c
index 283d295..a4b7b3d 100644
--- a/src/unix/getaddrinfo.c
+++ b/src/unix/getaddrinfo.c
@@ -34,6 +34,8 @@ static void uv__getaddrinfo_work(struct uv__work* w) {
                              req->service,
                              req->hints,
                              &req->res);
+  if (req->retcode == EAI_SYSTEM)
+    req->retcode = -errno;
 }


@@ -67,7 +69,10 @@ static void uv__getaddrinfo_done(struct uv__work* w, int status) {
   req->service = NULL;
   req->hostname = NULL;

-  if (req->retcode == 0) {
+  if (req->retcode < 0) {
+    /* EAI_SYSTEM error */
+    uv__set_sys_error(req->loop, -req->result);
+  } else if (req->retcode == 0) {
     /* OK */
 #if defined(EAI_NODATA) /* FreeBSD deprecated EAI_NODATA */
   } else if (req->retcode == EAI_NONAME || req->retcode == EAI_NODATA) {
```

This time around the error will be the actual UV errno, not UV_EADDRINFO, so there is no digging to be done in err.sys_errno_. Thoughts?

saghul:
I merged the patch for master, will open another PR for the v0.10 branch.

PR #1318, opened 2014-06-17
unix: return system error on EAI_SYSTEM (for v0.10)

saghul:


indutny:
LGTM

saghul:
Thanks Fedor, landed!

PR #1319, opened 2014-06-18
unix: allow uv_getnameinfo to be cancelled

saghul:
Looks like this one slipped on the initial implementation.

/cc @indutny


zerhacken:
@saghul oups, nice spotted.

test/test-threadpool-cancel.c
```diff
@@ -202,6 +210,44 @@ TEST_IMPL(threadpool_cancel_getaddrinfo) {
...
+  loop = uv_default_loop();
+  saturate_threadpool();
+
+  r = uv_getnameinfo(loop, reqs + 0, getnameinfo_cb, (const struct sockaddr*)&addr4, 0);
+  ASSERT(r == 0);
+
+  r = uv_getnameinfo(loop, reqs + 1, getnameinfo_cb, (const struct sockaddr*)&addr4, 0);
+  ASSERT(r == 0);
+
+  r = uv_getnameinfo(loop, reqs + 2, getnameinfo_cb, (const struct sockaddr*)&addr4, 0);
```
indutny, line 233: `ASSERT(0 == ...)`?

indutny:
LGTM

test/test-threadpool-cancel.c
saghul, line 233: Makes the line too long. Also, this test is copied verbatim from the getaddrinfo one. The new ones I do write them like that :-)

saghul:
Thanks, will merge!

saghul:
Landed!

PR #1326, opened 2014-06-19
unix, windows: use the same threadpool implementation

saghul:
UNIFY ALL THE THINGS!

With this patch we'll use a single threadpool implementation, the one that we were using on the Unix side. When changes are done to it both platforms will benefit, thus reducing the maintenance burden.

The first visible advantage is that `uv_cancel` now also works on Windows!


saghul:
@indutny please review! :-)

saghul:
Ping @indutny

txdv:
You could unify more: 

https://github.com/saghul/libuv/blob/acf5d7beeb37c669e92436cab2ea90eed2f2f682/src/win/core.c#L163-L170 

https://github.com/saghul/libuv/blob/acf5d7beeb37c669e92436cab2ea90eed2f2f682/src/unix/loop.c#L146-L153

saghul:
I have plans to unify more things, but I'll focus on each individually. Thanks for the reminder though!

include/uv-win.h
```diff
@@ -307,7 +308,11 @@ RB_HEAD(uv_timer_tree_s, uv_timer_s);
   /* Counter to keep track of active udp streams */                           \
   unsigned int active_udp_streams;                                            \
   /* Counter to started timer */                                              \
-  uint64_t timer_counter;
+  uint64_t timer_counter;                                                     \
+  /* Threadpool */                                                            \
+  void* wq[2];                                                                \
```
indutny, line 313: Could be a define in `uv-threadpool.h`?

src/threadpool.c
```diff
@@ -19,7 +19,24 @@
...
+# include "win/req-inl.h"
+/* TODO(saghul): unify internal req functions */
+static void uv__req_init(uv_loop_t* loop,
+                         uv_req_t* req,
+                         uv_req_type type) {
+  uv_req_init(loop, req);
+  req->type = type;
+  uv__req_register(loop, req);
+}
+#define uv__req_init(loop, req, type) \
```
indutny: Guess it should be `# define`

include/uv-win.h
saghul, line 313: You mean all threadpool related fields? Could be, but I thought I'd do that later since I want to unify the structures more (use QUEUEs as well on Windows) and I'd do it in a single shot.

indutny, line 313: No, I meant this three fields. It'd be easier to put a define here.

saghul, line 313: you mean:

```
#define UV_LOOP_THREADPOOL_FIELDS \
void* wq[2];  \
uv_mutex_t wq_mutex;       \
uv_async_t wq_async;
```

src/win/fs.c
```diff
@@ -1547,9 +1538,41 @@ static DWORD WINAPI uv_fs_thread_proc(void* parameter) {
     default:
       assert(!"bad uv_fs_type");
   }
+}
 
-  POST_COMPLETION_FOR_REQ(loop, req);
-  return 0;
+
+static void uv__fs_done(struct uv__work* w, int status) {
```
indutny, line 1544: I wonder if you could move these functions to their original positions to minimize amount of noise in a diff :)

indutny, line 1544: Ah, don't worry much about it.

src/win/getaddrinfo.c
```diff
@@ -178,11 +178,8 @@ void uv_process_getaddrinfo_req(uv_loop_t* loop, uv_getaddrinfo_t* req) {
         }
       }
     } else {
-      err = UV_EAI_MEMORY;
+      req->retcode = UV_EAI_MEMORY;
```
indutny, line 181: Wouldn't it be less fuzz to still store `err` here and update `req->retcode` at `complete`?

include/uv-win.h
indutny, line 313: Ah, just realized that it was windows-only thing.

indutny:
LGTM with one very minor style nit about `# define`.

include/uv-win.h
saghul, line 313: Actually unix also has it, but I'll do it as a bigger refactor, don't worry :-)

saghul:
Thanks a lot Fedor! Will fix and land later then.

saghul:
Thanks for the review Fedor! Landed in 8d11aacb79472009189c35e8953d0fa316bfd202 :sparkles:

PR #1327, opened 2014-06-20
windows: API for manually associating file handles with IOCP.

cjdelisle:
I'm opening this PR to begin a conversation about manual IOCP. For cjdns this is critical because I need to open a TAP device and the OpenWRT TAP demands that I read >= 1500 bytes at a time, trying to trick it into being a pipe is no good but I still need to poll on it with IOCP. This adds a function so that anyone can associate any OVERLAPPED file handle with the IOCP handle used in libuv so that "weird" (device) file handles can be polled on when it makes more sense to do the ReadFile/WriteFile calls manually.

This code is already working in cjdns, here is the consumer of this api:
https://github.com/cjdelisle/cjdns/blob/crashey/interface/tuntap/windows/TAPInterface.c

Side note: I'm not available on weekends, I am not intentionally ignoring your comments :)


saghul:
Quick check before diving deeper: why is uv_poll_t not suitable for your use case?

cjdelisle:
This type of file descriptor is a TAP device, not a socket nor an actual file or pipe
so it can't be dealt with using WSA socket polling which prevents uv_poll_t from working
and it can't be dealt with as a stream file descriptor so opening it as a named pipe
will not work.

saghul:
I'll need to think about this a bit. libuv is about platform abstraction, and I wouldn't like to open the door for platform specific handles and UV_ENOSYS everywhere else. Perhaps there is a way to extend `uv_poll_t` to also support this, with a different constructor maybe.

cjdelisle:
hmm perhaps it could be worked into uv_poll_t, it might be a cost-more-than-benefit since it ends up doing entirely different things.. I'll give it a bit of thought on my end as well, the very nice thing about libuv is that there is not the religious attachment to backward compatibility so if we determine later that we took the wrong road, we can fix it.

cjdelisle:
After some thought, I don't think it's sensible to merge this with uv_poll_t, poll is fundamentally about being informed when a file is "ready to read/write" and this is about being informed when an action is complete. If we merged them, we would spend twice the time explaining to people that poll means two entirely different things.

saghul:
Yeah, that was a long shot. Unfortunately the approach proposed in this PR is too platform specific :-/ I'm not a Windows expert, and alas I don't have an alternative I can suggest at the moment.

txdv:
Could you use uv_fs to do the reads? You can even specify how big the read should be.

When you use the stream API you have to specify the amount you want to read in the allocator.

indutny:
I do really object this change in a way it is right now. It'd be really hard to write cross-platform apps that would depend on it.

Let's take a look from the other side of this thing. What stuff do you need to be exposed by libuv in order to be able to write a separate library that does support it?

cjdelisle:
`What stuff do you need to be exposed by libuv in order to be able to write a separate library that does support it?`

This is putting the very least of functionality in libuv perhaps even too little.
It's just a "hole" into the inner IOCP which allows me to inject the event and get
called occurs. I can't imagine any way to move more of the logic out of libuv.

I need some type of request structure because the IOCP poll in the core of the loop
received an OVERLAPPED and expects that to be contained inside of some kind of request
structure and then my codepath is about the shortest way I can think of to get control
out of libuv and back into my code.

perhaps the GetOverlappedResult() call should logically happen inside of libuv instead of my code but that would be moving _more_ logic into libuv, not less. At any rate I don't know enough about windows to know for sure that one always wants to call GetOverlappedResult() after an IOCP poll returns so I think it's best to let the caller decide.

saghul:
@cjdelisle is the use of OVERLAPPED mandatory?Is it a handle? Can we get a fd out of it with `os_getfhandle` (or however that was called)?

cjdelisle:
@saghul the OVERLAPPED is needed by the ReadFile/WriteFile call (which we will be polling on) so it needs to exist, internally libuv calls GetQueuedCompletionStatus ( https://github.com/joyent/libuv/blob/master/src/win/core.c#L255 ) which returns a pointer to the OVERLAPPED and then libuv uses offsetof() to find the request object containing the OVERLAPPED and calls the callback associated with that request so there is not much way around this design.

saghul:
Let me rephrase: you need to read and write from that TAP thing, but do you _need_ to use a OVERLAPPED, that is, can you open that in non-overlapped mode? If so, using a uv_poll_t handle could work.

cjdelisle:
You can't use a uv_poll_t in windows for anything other than a SOCKET and this is a different type of handle, oh I wish it was so easy :)

saghul:
Ok, so no poll handles :-( But that is is a HANDLE, right? Have you tried opening it in non-overlapped mode and using uv_pipe_open? Or the fs functions, except open.

cjdelisle:
uv_poll does 0 byte reads to check for data which blows up the OpenVPN tap device (it tries to send the packet but truncates it to 0 bytes).
I didn't try fs stuff but that's getting very complicated :S

saghul:
Alas, that's all I got on my bag of tricks :-/

indutny:
```
[19:18:48] <@indutny>    cjd: the hole is too wide :)
[19:19:03] <@indutny>    the obvious alternative to it
[19:19:07]   c4milo (~c4milo@207-38-151-213.c3-0.avec-ubr2.nyr-avec.ny.cable.rcn.com) joined the channel.
[19:19:23] <@indutny>    would be spinning another thread with IOCP
[19:19:31] <@indutny>    and sending back events with async_send
[19:19:58] <@indutny>    s/back events/events back/
[19:20:03] <@indutny>    saghul: what do you think?
[19:20:14] <@indutny>    we could also just expose the IOCP
[19:20:27] <@indutny>    but that wouldn't suffice, right?
[19:20:32] <cjd>     you can't really expose the IOCP without exposing the way to get the callback
[19:20:32] <@indutny>    you'll need some other stuff like OVERLAPPED
[19:20:36] <@indutny>    yeah
[19:20:36] <cjd>     which is basically what my patch does
[19:23:33]   c4milo (~c4milo@207-38-151-213.c3-0.avec-ubr2.nyr-avec.ny.cable.rcn.com) left IRC. (Ping timeout: 240 seconds)
[19:26:45] <@indutny>    ok
[19:26:46] <@indutny>    another idea
[19:26:52] <@indutny>    could we just add a flag on handle?
[19:26:58] <@indutny>    or flag for uv_poll_t
[19:27:16] <@indutny>    like UV_WIN_WIERD_HANDLE
[19:27:16] <@indutny>    :D
[19:27:20] <cjd>     :D
[19:28:00] <@indutny>    and encapsulate all of your code inside uv_poll_t implementation
[19:28:06] <cjd>     the OVERLAPPED needs to be contained within the req structure in order for libuv to find the right function to call
[19:28:21] <@indutny>    we could alloc it
[19:28:41] <@indutny>    and add ptr field to uv_poll_t
[19:28:54] <cjd>     as long as uv_poll_ex() call to offsetof() works ok
[19:29:01] <@indutny>    aaah
[19:29:02] <@indutny>    offsetof
[19:29:03]   a_le (~a_le@c-98-210-111-144.hsd1.ca.comcast.net) joined the channel.
[19:29:06] <@indutny>    well
[19:29:11] <@indutny>    heh
[19:29:19] <cjd>     it takes the OVERLAPPED and gets the req
[19:29:23] <@indutny>    I know
[19:29:31] <@indutny>    you could work-around this too
[19:30:05] <cjd>     so I take it the main objection is the change to uv.h, horrors inside of src/win are ok as long as the user doesn't see them (?)
[19:30:40] <@indutny>    yes
[19:30:48] <@indutny>    but not as much horrors as you describe it :)
[19:31:03] <@indutny>    the way I describe it
[19:31:10] <@indutny>    it is still possible to write cross-platform uv code
[19:31:15] <@tjfontaine>     heh
[19:31:21] <@indutny>    tjfontaine: ohai
[19:31:22] <@tjfontaine>     first thing I see is horrors
[19:31:24] <cjd>     mmm not really
[19:31:28] <@indutny>    cjd: why not?/
[19:31:34] <@indutny>    this is just a flag on uv_poll_t
[19:31:37] <@indutny>    you could ignore it on unixes
[19:31:39] <@indutny>    and voila
[19:31:56] <cjd>     oh, I thought you meant "same code works for both windows TAP and linux TAP"
[19:32:04] <@indutny>    no, that's not what I meant
[19:32:12] <@indutny>    ok, anyway
[19:32:13] <cjd>     which it will never work unless we write uv_tun() which I know you won't go for :D
[19:32:13]   kazupon (~kazupon@ntkngw356229.kngw.nt.ftth.ppp.infoweb.ne.jp) left IRC. (Remote host closed the connection)
[19:32:24]   a_le (~a_le@c-98-210-111-144.hsd1.ca.comcast.net) left IRC. (Remote host closed the connection)
[19:32:25] <@indutny>    that depends :)
[19:32:34] <@indutny>    couldn't it just work via uv_poll_t ? :D
[19:33:01] <cjd>     once you call ReadFile() to find out whether there's a packet ready to read, you have the packet
[19:33:10] <@indutny>    wait
[19:33:13] <@indutny>    I mean unix side
[19:33:23] <@indutny>    could it just work via plain uv_poll-t
[19:33:25] <@indutny>    uv_poll_t*
[19:33:27] <@indutny>    without any flags
[19:33:30] <cjd>     yup
[19:33:32] <@indutny>    ok
[19:33:35] <@indutny>    so here were are!
[19:33:49] <cjd>     uv_poll() and then read()
[19:34:08] <cjd>     in windows:   ReadFile() and then uv_poll(UV_WEIRD_MAGIC)
[19:34:20] <cjd>     and the overlapped stuff :|
[19:34:47] <@indutny>    ok, that's fine
[19:34:48] <@indutny>    :)
[19:34:55] <@indutny>    hm...
[19:34:57] <@indutny>    which makes me think
[19:35:07] <@indutny>    the TAP on linux
[19:35:19] <@indutny>    does it behave more like a file, or more like a socket?
[19:35:23] <@indutny>    I mean
[19:35:28] <@indutny>    could you use it in uv_stream_t
[19:35:37] <@indutny>    (sorry, I have no idea, how it works internally)
[19:35:38] <cjd>     mmm iirc I do
[19:36:26] <cjd>     https://github.com/cjdelisle/cjdns/blob/master/interface/tuntap/TUNInterface_linux.c
[19:36:29] <cjd>     aha pipe()
[19:36:30] <@indutny>    ok
[19:36:33]   chiper (~chiper@216.98.4.170) left the channel.
[19:36:34] <cjd>     uv_pipe()
[19:36:40] cjd   using lazy/sneeky method
[19:36:47] <@indutny>    and
[19:36:49]   calvinfo (~Adium@c-50-174-77-78.hsd1.ca.comcast.net) left IRC. (Quit: Leaving.)
[19:37:04] <@indutny>    could we just make uv_pipe work with windows TAP?
[19:37:06] <@indutny>    by adding flag?
[19:37:36] <cjd>     hmm
[19:37:41] <@indutny>    yeah
[19:37:57] <cjd>     uv_pipe wants an int fileno, how do we get that from a HANDLE ?
[19:38:10] <@indutny>    we could make it accept HANDLE
[19:38:12] <@tjfontaine>     osf_ functions
[19:38:22] <@tjfontaine>     we already have them in there
[19:38:29] <cjd>     does osf_ allow converting *any* handle into an int?
[19:38:46]   Ralith (~ralith@c-24-56-252-127.customer.broadstripe.net) left IRC. (Quit: Lost terminal)
[19:38:46] <@indutny>    cjd: it is easier to check this
[19:38:55] <@indutny>    than figure out from docs :)
[19:38:58] <cjd>     hehe
[19:39:10] <cjd>     my only windows vm is at work and I usually only develop at home
[19:39:13] cjd   hardcore
[19:39:28] <@indutny>    let me google it
[19:39:40] <@tjfontaine>     (for you)
[19:40:25] <cjd>     xD
[19:40:35] <@indutny>    I think it'll definitely work this way
[19:40:37] <@indutny>    fd => HANDLE
[19:40:41] <cjd>     I have a horrific memory but I seem to recall there were some kind of problems with this
[19:40:47] <@indutny>    if you managed to get fd
[19:41:06] <@tjfontaine>     it's windows how are there not horrific problems? :)
[19:41:16] <cjd>     heh
[19:41:26] <@indutny>    https://github.com/Arty0/tapcfg/blob/master/src/lib/tapcfg_windows.c#L232
[19:41:54] <@indutny>    I have no idea why this line is commented
[19:42:11] <cjd>     interesting replacement
[19:42:20] <cjd>     just pass the handle as an int
[19:42:29] <@indutny>    yeah
[19:42:34] <cjd>     handles in windows will never be >2**24 so it's safe to cast them to int
[19:42:48] <@indutny>    but
[19:42:56] <@indutny>    you'll need to distinguish them
[19:42:59] <@indutny>    from fds
[19:43:26] <cjd>     return 0-((uintptr_t)handle);  // and go take a shower
[19:43:33]   Ralith (~ralith@c-24-56-252-127.customer.broadstripe.net) joined the channel.
[19:43:54] <@indutny>    ok, this probably is a way to go
[19:44:03] <@indutny>    the highest bit couldn't ever be 1 for both handles and fds
[19:44:07] <@indutny>    you could just set it
[19:44:14] <@indutny>    saghul: I need you
[19:44:25] <cjd>     that would actually be super cool because we could document that you can always pass a HANDLE just by negating it
[19:44:25] <@indutny>    saghul: any thoughts on this? :)
[19:44:40] <@indutny>    cjd: no, we will need to add api method
[19:44:46] <@indutny>    cjd: to convert handle to int
[19:44:59] <@indutny>    I don't want to lock ourselves into this negating thing
```

indutny:
So the latest idea is to introduce a windows-specific flag for `uv_poll_t` and make it work with `HANDLE`s.

cjdelisle:
tl;dr
We might be able to make uv_pipe() do what we want with a special flag.
It will be less generic but it will avoid addition of new APIs.
We need to be able to pass arbitrary HANDLEs to uv_pipe() which takes ints so we need an API to convert HANDLE to int (it can internally do something silly like set a high bit because HANDLE is guaranteed to never use more than 24 bits).

Unfortunately uv_poll() doesn't seem like it will work because we need both libuv and the user to have access to the OVERLAPPED and we end up back where this PR started.

saghul:
Hey,

I went through the notes and did a bit of digging, here are is what I gathered, though some things are still not 100% clear to me.

We cannot use `uv_poll_t` because this thing is not a `SOCKET`. So, moving on!

Now, you say apparently it could be made to work with `uv_pipe_t` and some flag, but how about this note from earlier? https://github.com/joyent/libuv/pull/1327#issuecomment-47236456 (I understand uv_poll should mean uv_pipe there) Where do we do those 0 byte reads?

Moreover, if this thing is a `HANDLE`, you should be able to get a CRT file descriptor with `_open_osfhandle` and then pass it to `uv_pipe_open` and just use it normally. Would that work for you? Why not?

Did I miss something?

piscisaureus:
I agree with the TS that this is desirable. I made a proposal once on the mailing list, see https://groups.google.com/d/msg/libuv/IO06T7I25Rc/tX2Llz9yuSQJ.

I know that Tomasz Janczuk's http_sys module for node (ab)used an uv_async_t to accomplish this but we should add a proper way.

zhaozg:
If we made a uv_device_t as a stream type, which support /dev/xxx on linux and \.\devices on windows, that will better support SERIAL, PRINT, TAP/TUN, PPP?
@saghul @indutny @cjdelisle are you think so, I need it, but I not family with libuv.

@zhaozg

zhaozg:
Please review this PR https://github.com/joyent/libuv/pull/1580

saghul:
Closing, we'll go with `uv_device_t`. as it's a more integrated and cross platform approach for the particular problem of TUN/TAP and character devices.

PR #1328, opened 2014-06-20
unix: define NI_MAXHOST and NI_MAXSERV if needed

saghul:
OSX only exposes them if _POSIX_C_SOURCE is not defined, for instance.

### DIFF
--- a/include/uv-unix.h
+++ b/include/uv-unix.h
@@ -55,6 +55,14 @@
 # include "uv-bsd.h"
 #endif
 
+#ifndef NI_MAXHOST
+# define NI_MAXHOST 1025
+#endif
+
+#ifndef NI_MAXSERV
+# define NI_MAXSERV 32
+#endif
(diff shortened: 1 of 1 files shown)


indutny:
Hm... Are you sure that this couldn't be solved by adding proper defines in some places?

saghul:
Not that I know of. Quite the opposite, if some application using libuv
defines _POSIX_C_SOURCE  then OSX doesn't define it.

indutny:
Hm... but how are they propagated to the compiled `.c` files?

saghul:
Not sure what you mean there. I defined it in our Unix internal include, so
all Unix files get it.

You can test it by trying to compile Node with latest libuv on OSX.

indutny:
Aaaah, it is used in headers.... Ok, LGTM

saghul:
Landed!

PR #1331, opened 2014-06-23
windows: fix mingw build error

retep998:
COMMON_LVB_REVERSE_VIDEO might not be defined on mingw


mattn:
ref #1296

saghul:
Duplicated. See https://github.com/joyent/libuv/pull/1296

PR #1333, opened 2014-06-23
Fix build for mingw32

mattn:
This patch close #1296


saghul:
What about the GetNameInfoW problem?

mattn:
Ooops, Sorry. I mistaken make a patch. I'll re-make a patch.

saghul:
You can force-push #1296, that's fine :-)

Issue #1334, opened 2014-06-23
sizeof loop in the unstable version


txdv:
I see that the API for creating a loop instance has been changed.

It used to be uv_loop_t\* uv_loop_new() which would allocate the memory on its own, but now it is uv_loop_init(uv_loop_t\* loop) so the user has to manage the memory of the loop struct.

The problem here is that the size of uv_loop_t is not exposed through ffi calls (compared to all the request and handle sizes). This makes it more difficult to write bindings with languages which only support ffi.

uv_loop_t is neither a handle nor a request so I propose to add a function uv_loop_size which returns sizeof(uv_loop_t).

saghul:
Sounds good to me! Mind writing the patch?

txdv:
Here you go

indutny:
Landed?

PR #1335, opened 2014-06-23
Expose functionality to lookup the uv_loop_t struct size.

txdv:
We add a function called uv_loop_size which returns the size of the loop.
It is an additional function which complements uv_handle_size and uv_req_size
in order to provide full FFI support.


txdv:
Original issue: https://github.com/joyent/libuv/issues/1334

include/uv.h
```diff
@@ -288,6 +288,11 @@ UV_EXTERN uv_loop_t* uv_loop_new(void);
 UV_EXTERN void uv_loop_delete(uv_loop_t*);
 
 /*
+ * Returns size of the loop struct, useful for dynamic lookup with FFI
+ */
+UV_EXTERN size_t uv_loop_size();
```
saghul: (void)

src/uv-common.c
```diff
@@ -67,6 +67,11 @@ size_t uv_req_size(uv_req_type type) {
 #undef XX
 
 
+size_t uv_loop_size() {
```
saghul: ditto.

txdv:
Should have checked the surrounding code... But I still don't understand why we need to annotate the stuff that doesn't exist. Whatever, I will just role with the old C style.

saghul:
IIRC it's something about C89, not sure though.

indutny:
Hm... how do you allocate `uv_handle_t` in your FFI bindings?

indutny:
Oops, nvm

indutny:
Thank you, landed in e392a1a

indutny:
And LGTM :)

Issue #122, opened 2011-08-01
uv_spawn env+path semantics


piscisaureus:
uv_spawn uses the `PATH` environment variable to locate the executable to be spawned. 

On unix, if the user explicitly supplies an environment block, the `PATH` from this environment is used. If the supplied environment does not contain a `PATH` variable, `:/bin:/usr/bin` is used as a fallback. 

On windows the search path from the _parent_ process is always used.

This semantic difference may need to be resolved.

piscisaureus:
Dropping the path search code entirely and using the search mechanics built into CreateProcess should also be considered.

The downside is that it uses a different search rules than what (I think) people are familiar with from cmd.exe / spawnvpe(). 
On the upside, we could delete quite some code. Also the CreateProcess search rules are well-documented.

DrPizza:
There are many options. They all suck to a greater or lesser extent. Windows has a bunch of rules in different places. I'm not really sure users will have any consistent expectations.
- `ShellExecute()` is probably the most complex API. It checks working directories, PATH, various registry keys, and probably more. Awful.
- cmd.exe's behaviour is perhaps even more complex than `ShellExecute()`, since it seems to do most of the things `ShellExecute()` does, with the added bonus of also considering cmd.exe built-ins and the PATHEXT environment variable.
- `CreateProcess()`'s rules with lpApplicationName are very simple and easy to understand. They're also utterly unforgiving: no filename extension is added (no automatic .exe or .com or anything else), no PATH searching. If you give it an absolute path to a filename it'll try that absolute path and no other; if you give it a bare filename (or maybe a relative path? It doesn't specify) it'll try $CWD\filename. And that's it.
- `CreateProcess()` without lpApplicationName does a search; application folder, $CWD, system32, system, \windows, $PATH, in that order. It will try each location with and without a .exe extension.
- VC++'s spawn-with-path `_spawnvpe()` will first try `CreateProcess()`'s lpApplicationName strategy. If that fails, and if a bare filename was supplied (with neither an absolute nor a relative path), it will then try each directory in the parent process's $PATH. Even if a child environment $PATH is specified, it will not be examined.

Not directly used, but another Win32 lookup behaviour is found in `SearchPath()`, which will check $CWD and then $PATH, or $PATH followed by $CWD, depending on options.

With the exception of `SearchPath()`, none of these functions has a search API available. `_spawnvpe()` is, I believe, directly equivalent to:

```
if(filename contains '\' or '/') {
[...]

PR #1337, opened 2014-06-24
windows: Read the PATH env var of the child

alexcrichton:
The unix and windows process implementations diverge in their behavior
when dealing with subprocesses that are spawned with a relative path.
With unix the _child's_ PATH environment variable is read, whereas
with windows the _parent's_ environment variable is read.

This commit brings the two implementation in line with respect to
their behavior of reading PATH by having both read the _child's_ PATH
environment variable. This involves looking into the user-provided
environment on windows and extracting the PATH variable specifically
so it can be inspected later on.

I'll cc #122 here as well, although I ended up discovering that after
writing this. It appears that there's more to this than I originally
thought, but I figured I'd at least put this out there!


saghul:
Thanks Alex! I'll take a look. It will take a bit since I'm a bit flooded right now, but I'll get to it (if nobody does beforehand).

saghul:
@piscisaureus this is a bit over my head, can you help please? :-)

piscisaureus:
@saghul 

No time today but I will try to find a spare hour or so.
It's not super difficult: when the user specifies the environment for a child process (via the uv_process_options struct), we should use the PATH in that environment, and not the PATH of the parent process. 

The affected code is here: https://github.com/joyent/libuv/blob/c11a598d7cac6f450277261d39ce887727a433b2/src/win/process.c#L878-L899.

In pseudocode it should look like:

```
if (environment specified in process_options) {
  // get path from options;
  // if not found set path to the empty string;
} else {
  // use path from parent process  
  path = GetEnvironmentVariableW('path');
}
```

saghul:
@piscisaureus Thanks! I'll also go through it.

saghul:
Hi @alexcrichton, sorry for the delay on this. Unfortunately your patch no longer applies because `make_program_env` was heavily modified to require extra env variables and sort them in the right order. I think it would be best so search for "PATH=" in the provessed env rather than doing it as part of `make_program_env`, since it's already quite complex.

Can you rebase and work on a patch that applies to current master?

alexcrichton:
Sure! I've rebased and tested on my local mingw installation, so hopefully it's working!

src/win/process.c
```diff
@@ -813,6 +813,30 @@ int make_program_env(char* env_block[], WCHAR** dst_ptr) {
...
 }
 
+/*
+ * Attempt to find the value of the PATH environment variable in the child's
+ * preprocessed environment.
+ *
+ * If found, `*path` and `*path_len` will be the value of PATH and the lenth,
+ * respectively, and 1 will be returned. If not found, then 0 will be returned.
+ */
+static int find_path(WCHAR *env, WCHAR **path, DWORD *path_len) {
```
saghul: Can you make this set path to NULL and path_len to 0 in case it doesn't find it instead? You can also make it return *path to make the if check nicer.

src/win/process.c
```diff
@@ -813,6 +813,30 @@ int make_program_env(char* env_block[], WCHAR** dst_ptr) {
   return 0;
 }
 
+/*
+ * Attempt to find the value of the PATH environment variable in the child's
+ * preprocessed environment.
+ *
+ * If found, `*path` and `*path_len` will be the value of PATH and the lenth,
```
saghul: typo: length.

src/win/process.c
```diff
@@ -1131,7 +1157,8 @@ int uv_spawn(uv_loop_t* loop,
   free(arguments);
   free(cwd);
   free(env);
-  free(path);
+  if (path_owned)
+    free(path);
```
saghul: Maybe use another variable "env_path", which we initialize to NULL and always free anyway?

test/test-spawn.c
```diff
@@ -1291,3 +1291,39 @@ TEST_IMPL(closed_fd_events) {
...
+
+  /* Set up the process, but make sure that the file to run is relative and */
+  /* requires a lookup into PATH */
+  init_process_options("spawn_helper1", exit_cb);
+  options.file = "run-tests";
+  args[0] = "run-tests";
+
+  /* Set up the PATH env variable */
+  len = strlen(exepath);
+  for (len = strlen(exepath);
```
saghul, line 1308: isn't len initialized twice here?

test/test-spawn.c
```diff
@@ -1291,3 +1291,39 @@ TEST_IMPL(closed_fd_events) {
...
+       exepath[len - 1] != '/' && exepath[len - 1] != '\\';
+       len--);
+  exepath[len] = 0;
+  strcpy(path, "PATH=");
+  strcpy(path + 5, exepath);
+
+  char *env[2] = {path, NULL};
```
saghul: declare this at the top

saghul:
> Sure! I've rebased and tested on my local mingw installation, so hopefully it's working!

Nice, thanks! Left a few more coments, nothing big though, the approach seems fine :-)

alexcrichton:
Thanks for taking a look! I've updated the patch.

src/win/process.c
```diff
@@ -1131,7 +1147,7 @@ int uv_spawn(uv_loop_t* loop,
   free(arguments);
   free(cwd);
   free(env);
-  free(path);
+  free(alloc_path);
```
saghul, line 1150: Is this correct? We set path to alloc_path, so we should be freeing that one, right?

src/win/process.c
```diff
@@ -993,11 +1008,12 @@ int uv_spawn(uv_loop_t* loop,
       goto done;
     }
 
-    path = (WCHAR*) malloc(path_len * sizeof(WCHAR));
-    if (path == NULL) {
+    alloc_path = (WCHAR*) malloc(path_len * sizeof(WCHAR));
```
saghul, line 1011: I know you didn't introduce this, but can you remove the unnecessary cast? :-)

saghul:
Left a couple more comments, we are almost there :-)

src/win/process.c
piscisaureus, line 1011: I vote for leaving the cast in :)

src/win/process.c
alexcrichton, line 1150: `alloc_path` is the variable storing the allocation, and `path` is just a pointer to something which is the actual path value (may or may not be an allocation). If `PATH` is found in `env` then `path` doesn't need to be free'd (it's in the middle of an allocation).

saghul, line 1150: Ah, right. LGTM then.

alexcrichton:
@saghul, looks ok?

saghul:
@alexcrichton yeah, will merge now! :+1:

saghul:
I just ran the tests on my Windows 7 machine and `uv_spawn` fails with ENOENT for me. Does the test pass for you?

alexcrichton:
Running the tests locally causes the `fs_file_loop` test to fail, but other than that all other tests pass. What does the failure look like for you?

saghul:
How are you running your tests? I'm running `vcbuild.bat test` in a PowerShell prompt. Test fails as follows:

```
[%  72|+ 165|-   2|T   0|S   0]: spawn_reads_child_path
`spawn_reads_child_path` failed: exit code 3
Output from process `spawn_reads_child_path`:
Assertion failed in test\test-spawn.c on line 1318: r == 0
```

I printed the error and it's ENOENT, so I assume it comes from here: https://github.com/joyent/libuv/blob/master/src/win/process.c#L1018

alexcrichton:
Ah I may be running the tests incorrectly then. I'm using mingw and I don't know how to run the tests otherwise so it looks something like this:

```
$ gcc -Iinclude -Itest $(ls tests/*.c | grep -v unix | grep -v benchmarks) -L. -luv -lws2_32 -lpsapi -liphlpapi
$ ./a.exe
```

Could you add a print to `search_path` to see what path is being searched? I could also try to install power shell and friends to reproduce if you're busy!

saghul:
That's an unusual way to run tests :-) I just ran them on a MSYS2/MinGW environment (built it with autotools), and the failure is the same. cmd.exe gives the same error :-S

I'll add some prints.

saghul:
Hum, here are the paths being searched:

```
-XX C:\Users\saghul\src\libuv\run-tests.com
-XX C:\Users\saghul\src\libuv\run-tests.exe
-XX C:\Users\saghul\src\libuv\%SystemRoot%\system32\WindowsPowerShell\v1.0\run-tests.com
-XX C:\Users\saghul\src\libuv\%SystemRoot%\system32\WindowsPowerShell\v1.0\run-tests.exe
-XX C:\Windows\system32\run-tests.com
-XX C:\Windows\system32\run-tests.exe
-XX C:\Windows\run-tests.com
-XX C:\Windows\run-tests.exe
-XX C:\Windows\System32\Wbem\run-tests.com
-XX C:\Windows\System32\Wbem\run-tests.exe
-XX C:\Windows\System32\WindowsPowerShell\v1.0\run-tests.com
-XX C:\Windows\System32\WindowsPowerShell\v1.0\run-tests.exe
-XX C:\Python27\run-tests.com
-XX C:\Python27\run-tests.exe
-XX C:\Python27\Scripts\run-tests.com
-XX C:\Python27\Scripts\run-tests.exe
-XX C:\Program Files\Microsoft Windows Performance Toolkit\run-tests.com
-XX C:\Program Files\Microsoft Windows Performance Toolkit\run-tests.exe
-XX C:\ProgramData\chocolatey\bin\run-tests.com
-XX C:\ProgramData\chocolatey\bin\run-tests.exe
```

Of course it fails, the path where the exe is located is `C:\Users\saghul\src\libuv\Debug`. Maybe the bug is in `search_path`...

saghul:
Got it! The PATH wasn't actually properly found:

```
for (; env != NULL && *env != 0; env += wcslen(env)) {
```

vs

```
for (; env != NULL && *env != 0; env += wcslen(env) + 1) {
```

We also need to skip the null byte at the end of each env entry.

I'll amend your patch and merge.

alexcrichton:
Oh whoops, thanks @saghul for tracking that down!

saghul:
Landed in https://github.com/joyent/libuv/commit/c7e4b31444c94623060979126e29648c0b940aec, thanks Alex!

PS: Nice PR number ;-)

PR #1338, opened 2014-06-25
doc: removed duplicate word in uv.h documentation

ghost:
Removed duplicate "other" in:

> Using uv_poll_t for any other other purpose is not recommended


indutny:
Thank you!

Please take a look at https://github.com/joyent/libuv/blob/master/CONTRIBUTING.md#commit and fix the commit to make it match the requirements.

txdv:
the commit message should be:

doc: remove duplicate word

txdv:
Please use present tense.

It is like a journal. You already have a date so you don't need to specify that it was a in the past.

saghul:
Thanks Tim! Landed in 546fa7d508ba741c106d0161a31e9f8d257a5d56

PR #1342, opened 2014-06-26
include: const-ify arguments to uv_getnameinfo_cb

saghul:
Does what it says on the tin.

/cc @indutny


indutny:
LGTM

saghul:
That was quick! Thanks Fedor, landed!

PR #1344, opened 2014-06-29
add uv udp try send

txdv:
Please review the commits and give me comments on them.


src/unix/udp.c
```diff
@@ -441,6 +448,43 @@ int uv__udp_send(uv_udp_send_t* req,
...
+  QUEUE* q;
+  struct msghdr h;
+  ssize_t size;
+  int status;
+
+  /* already sending a message */
+  if (handle->send_queue_count != 0) {
+    return 0;
+  }
+  assert(nbufs > 0);
```
saghul, line 460: Move the assert to before the if.

src/unix/udp.c
```diff
@@ -441,6 +448,43 @@ int uv__udp_send(uv_udp_send_t* req,
...
+  h.msg_namelen = addrlen;
+  h.msg_iov = (struct iovec*) &bufs;
+  h.msg_iovlen = nbufs;
+
+  do {
+    size = sendmsg(handle->io_watcher.fd, &h, 0);
+  }
+  while (size == -1 && errno == EINTR);
+
+  return (size == -1 ? -errno : size);
```
saghul: We shouldn't propagate EAGAIN to the user. Thou should get 0 instead. Not sure about EMSGSIZE.

include/uv.h
```diff
@@ -1068,6 +1070,17 @@ UV_EXTERN int uv_udp_send(uv_udp_send_t* req,
                           uv_udp_send_cb send_cb);
 
 /*
+ * Same as `uv_udp_send()`, but won't queue a send request if it can't be completed
+ * immediately.
+ * Will return either:
+ * - >= 0: number of bytes written (can be less than the supplied buffer size)
```
saghul: Since datagrams are truncated if too big, how can we possibly get a value lower than the buffer size?

src/win/udp.c
```diff
@@ -860,3 +869,12 @@ int uv__udp_send(uv_udp_send_t* req,
...
   return 0;
 }
+
+
+int uv__udp_try_send(uv_udp_t* handle,
+                     const uv_buf_t bufs[],
+                     unsigned int nbufs,
+                     const struct sockaddr* addr,
+                     unsigned int addrlen) {
+  return -1;
```
saghul: Return UV_ENOSYS.

saghul:
I made a quick review, will do it more in deep tomorrow.

saghul:
Oh, and thanks a lot for working on this Andrius!

include/uv.h
ibc: Right, replace with:

```
* - >= 0: number of bytes sent (same as supplied buffer size)
```

txdv: quote by bnoordhuis from last thread

> uv_try_write() returns the number of bytes written. That would be kind of superfluous for uv_udp_try_send() because datagram writes are atomic: everything is written or nothing is but partial writes don't happen. A boolean is probably sufficient but then again, returning the number of bytes written doesn't actually hurt and is "bool-y" to

ibc:
One question: why do we need to keep the number of bytes to be sent (I mean the `send_queue_size`)? In UDP the number of bytes does not matter but just the number of pending datagrams (Regardless how long they are).

Would not make more sense a `send_queue_datagrams` or `send_queue_packets`?

include/uv.h
saghul: Sure, I'm ok with the return value, but I think the comment needs to be updated, not to mislead anyone.

saghul:
> One question: why do we need to keep the number of bytes to be sent (I mean the send_queue_size)? In UDP the number of bytes does not matter but just the number of pending datagrams (Regardless how long they are).
> 
> Would not make more sense a send_queue_datagrams or send_queue_packets?

Your application may want to stop trying to write until the queue is down to certain number of bytes, perhaps in a memory constrained environment. A long shot? Maybe yes.

ibc:
> Your application may want to stop trying to write until the queue is down to certain number of bytes, perhaps in a memory constrained environment.

Are we assuming then that `send_queue_size`is public API? if so it should be documented.

include/uv.h
ibc: Please avoid "write" word in UDP as it is confusing. Use "send" :)

src/unix/udp.c
txdv: If we return the written size in the function, the user can check for himself for udp truncation?

What should we do with EAGAIN?

saghul: I think we should do it like in `uv_try_write`: modify `uv__udp_send` to always try to write immediately and return 0 on EAGAIN. Then `uv_udp_try_send` would call `uv_udp_send` with a on-stack request and dequeue it if the send didn't happen.

include/uv.h
txdv: @saghul I want this function to return the size that has actually been written.

I mean, if it is > 0, the user knows that the packet has been sent, but the user can now check if the packet was truncated using the returned size.

saghul:
Yes, it is part of the public API. We don't have a good track record of documenting struct members though.

include/uv.h
```diff
@@ -900,6 +900,8 @@ typedef void (*uv_udp_recv_cb)(uv_udp_t* handle,
 /* uv_udp_t is a subclass of uv_handle_t */
 struct uv_udp_s {
   UV_HANDLE_FIELDS
+  size_t send_queue_size;
```
saghul, line 907: Maybe put a `/* public  */` here?

ibc:
2014-06-30 10:23 GMT+02:00 Saúl Ibarra Corretgé notifications@github.com:

> We don't have a good track record of documenting struct members though.

That's not true, there is no documentation about struct members at all XD.

Yes, patches are welcome.

## 

Iñaki Baz Castillo
ibc@aliax.net

src/unix/udp.c
txdv: try_write modifies all queues and then tries to patch the queues again so nothing is present there.

Why touch the queues at all?

saghul: Hum, after looking at what we have right now, it might be better not to do that. Just handle EAGAIN and EWOULDBLOCK then. (return 0 in that case).

txdv:
All current comments have been addressed.

I am going to squash the additions into the original commits.

txdv:
The only one I didn't address was the name of the field send_queue_count. What would be the best name for it?

I don't like send_queue_datagrams or send_queue_packets because they have an unclear meaning.

A testcase for try_write is missing as well. I'll add one in the evening, i'll just copy the try_write test case and adjust it to test udp.

src/unix/udp.c
```diff
@@ -441,6 +448,51 @@ int uv__udp_send(uv_udp_send_t* req,
...
+  int err;
+  QUEUE* q;
+  struct msghdr h;
+  ssize_t size;
+  int status;
+
+  assert(nbufs > 0);
+
+  /* already sending a message */
+  if (handle->send_queue_count != 0) {
```
saghul: Drop braces for single statement ifs.

saghul:
I havent run the tests, but LGTM. Can you add a test for `uv_udp_try_send`?

ibc:
> I havent run the tests, but LGTM. Can you add a test for uv_udp_try_send?

How? It is difficult to test that. Same for `uv_try_write` for which there is a test unit (test-tcp-try-write.c) but it does not test that it is able to write at the moment, but just that it somehow seems to work.

saghul:
By brute-forcing it. Look at the test for `uv_try_write`, we do it in an infinite loop. If it doesn't succeed in the test-timeout time, the test will be killed anyway.

ibc:
Opps, missed that. OK.

txdv:
@saghul on stream sockets you can just not read on the other side until there is too much congestion for the os buffer to handle and the call returns something.

With udp if I use 127.0.0.1 it will just send the thing regardless. I fear that it might be possible that I won't be able to generate enough packets on the loopback interface to actually block it. But we will see when I get home.

txdv:
I need some help on debugging @saghul . Tried to write a test but ain't nothing working.

ibc:
Please let me know how to test it. Must I compile your branch and just run the tests?

saghul:
I'll see if I can debug it. At a first glance I don't see any obvious mistake :-O

src/unix/udp.c
```diff
@@ -441,6 +448,50 @@ int uv__udp_send(uv_udp_send_t* req,
 }
 
 
+int uv__udp_try_send(uv_udp_t* handle,
+                     const uv_buf_t bufs[],
+                     unsigned int nbufs,
+                     const struct sockaddr* addr,
+                     unsigned int addrlen) {
+  int err;
+  QUEUE* q;
```
saghul: Unused.

src/unix/udp.c
```diff
@@ -441,6 +448,50 @@ int uv__udp_send(uv_udp_send_t* req,
...
+  struct msghdr h;
+  ssize_t size;
+  int status;
```
saghul: Unused.

txdv:
bash command on unix:

``` bash
make -j8 test/run-tests && test/run-tests udp_try_send udp_try_send
```

will compile and run the test

test/test-udp-try-send.c
```diff
@@ -0,0 +1,221 @@
...
+  ASSERT(addr != NULL);
+  ASSERT(nread == 4);
+  ASSERT(!memcmp("PONG", buf->base, nread));
+
+  cl_recv_cb_called++;
+
+  //uv_close((uv_handle_t*) handle, close_cb);
+}
+
+#define RESULT(x) printf("%s = %u\n", #x, x)
```
saghul: Error codes are negative, but you are formatting is as unsigned, that's why you see the weird number.

txdv: O yeah, forgot that. Shouldn't ignore the error codes of the compiler...

ibc:
Which branch should I test? I see various branches about "udp".

txdv:
Look at the top of the page:  Open   txdv  wants to merge 9 commits into joyent:master  from txdv:uv_udp_try_send

saghul:
Edit the .git/config file in your libuv directory and make sure origin looks like this:

```
[remote "origin"]
        url = git@github.com:joyent/libuv.git
        fetch = +refs/heads/*:refs/remotes/origin/*
        fetch = +refs/pull/*/head:refs/remotes/origin/pr/*
```

Then:

```
git fetch
git checkout pr/1344
```

saghul:
You can use the HTTPS url

ibc:
I've tryed `uv_udp_try_send()` in my code. I get a "address not available" error (which does not happen with `uv_udp_send()` or `uv_try_write()`).

I guess that the sockaddr\* pointer being passed to `uv_udp_try_send()` is not being correctly handled.

txdv:
Fixed it. https://github.com/txdv/libuv/commit/278a1b10bbcf1ceef845027bfdf9719ee044083c

Going to work now on the test now.

ibc:
It works! Great :)

txdv:
@saghul https://github.com/txdv/libuv/commit/ece947e090e404f1493f678e0934e980da7d647d

> With udp if I use 127.0.0.1 it will just send the thing regardless. I fear that it might be possible that I won't be able to generate enough packets on the loopback interface to actually block it. But we will see when I get home.

This is the case on my machine. max udp packet size is 65507 

I can't get it to spit out an EAGAIN or EWOULDBLOCK when using the loopback. SO I can't test for that specifically.

saghul:
I cannot even compile it. Can you please use /\* */ for comments?

txdv:
Its not a clean patch, just for showcasing. removing 3 comments is not the hardest thing to do :/

But I pushed a patch anyway.

saghul:
Doesn't work here either. I'd go for trying to write and bailing out the first time we succeed, instead of saturating. A test for EMSGSIZE would also be nice :-)

ibc:
@saghul it compiles for me (clang). Which compiler are you using?

saghul:
GCC.

ibc:
@txdv one question about the return value of `uv_udp_try_sent()`. In case of success it returns "size" which is the return value of `sendmsg()` (which means the number of bytes sent).

So, in case I try to send too big data, what is the current behavior of both `uv_udp_send()` and `uv_udp_try_send()`:
1. The former indicates UV_EMSGSIZE in the send_cb and the latter returns UV_EMSGSIZE error.
2. The former indicates OK in the send_cb and the latter returns the number of bytes sent... but, shouldn't `sendmsg()` just return EMSGSIZE in case of a datagram instead of the "truncated" size??

According to the spec:

> ```
>   EMSGSIZE
>          The socket type requires that message be sent atomically, and
>          the size of the message to be sent made this impossible.
> ```

txdv:
#### current uv_udp_try_sent

x is the buffer size
- if there is something in the queue => 0
- error => -error code
- x > 65507 => EMSGSIZE
- 0 < x  < 65507 =>  truncated sent buffer value

If you try to send buffer with the size of 0, it will return 0 regardless if the queue is empty and the message has been sent succesfully or the queue is full.

#### uv_udp_send

I don't know how it works. I have to look into it ... and we should adjust the behavior of uv_udp_try_send to uv_udp_send I guess.

ibc:
In fact, I'm testing:

```
char long_data[60000];
len = 60000;

buffer = uv_buf_init((char*)data, len);
sent = uv_udp_try_send(udp, &buffer, 1, addr);
```

and `uv_udp_try_send()` clearly returns "message too long" (so EMSGSIZE error). This is: I'm PRETTY sure that it won't NEVER return a positive value less than the given buffer(s) length, so I suggest changing the documentation in uv.h:

```
/*
 * Same as `uv_udp_send()`, but won't queue a send request if it can't be completed
 * immediately.
 * Will return either:
 * - >= 0: matches the supplied buffer size.
 * - < 0: negative error code
 */
```

ibc:
@txdv 

> 0 < x < 65507 => truncated sent buffer value

Not in my host. I call `uv_udp_try_send()` with a 60000 bytes length buffer and get EMSGSIZE. I'm not able to reproduce any case in which the return value is positive but less than the given buffer.

txdv:
What OS are you using?

The test case should then fail for you.

ibc:
OSX 10.9.3.

ibc:
> Does the test fail for you?

No, but... not sure. If I run "make check" I get an error about an unrelated test unit. If I run `test/run-tests udp_try_send udp_try_send` I get no output and return code is 0, so seems to work, yes.

txdv:
I have updated the test case, pull again and run the test.

You can see 3 of the 4 different behaviors in the test case here:

 https://github.com/txdv/libuv/blob/32b84f240ff20785d4ba9d8fb1fe79c757ec50b7/test/test-udp-try-send.c#L113-L123

In order to test the particular test you have to execute this command:

``` bash
make -j8 test/run-tests && test/run-tests udp_try_send udp_try_send
```

ibc:
Sorry @txdv, not sure how you are doing commits, but every time I do `git pull` I get a Git conflict (and I've not modified the code at all):

```
ibc @ibc-macbook in ~/src/libuv.uv_udp_try_send/test git:uv_udp_try_send $ git pull
remote: Counting objects: 4, done.
remote: Compressing objects: 100% (4/4), done.
remote: Total 4 (delta 0), reused 0 (delta 0)
Unpacking objects: 100% (4/4), done.
From https://github.com/txdv/libuv
 + 08d8a4b...32b84f2 uv_udp_try_send -> origin/uv_udp_try_send  (forced update)
Auto-merging test/test-udp-try-send.c
CONFLICT (add/add): Merge conflict in test/test-udp-try-send.c
Automatic merge failed; fix conflicts and then commit the result.
```

Are you using "git push --force"??

txdv:
if txdv is my repo name then:

``` bash
git fetch txdv
git checkout txdv/uv_udp_try_send
autoconf
configure
make -j8 test/run-tests && test/run-tests udp_try_send udp_try_send
```

Yes I am using push --force, I am constantly squashing changes into one nice patch.

I'll try to not squash immediately from now on, only after the discussion about the current patch is over so you can use pull.

ibc:
```
Assertion failed in test/test-udp-try-send.c on line 119: r > 0
Abort trap: 6
```

This is because `sendmsg()` is returning EMSGSIZE rather than a size less than MAX_UDP_PACKET_SIZE.

txdv:
It doesn't fail on my Ubuntu 14.04 64bit.

So we need to tweak the flags of the udp creation I guess?

Maybe there is an ioctl call or something for MAX_UDP_PACKET_SIZE on all unixes?

Don't know, we need @saghul or someone else who has experience with macosx.

ibc:
According to the man:

> If the message is too long to pass atomically through the underlying protocol, the error EMSGSIZE is returned, and the message is not transmitted.

May be that depends on the OS? is there any flag to set such an "atomicy" request?

BTW: I will test the same for `uv_udp_send()`. If I get same results it means that your `uv_udp_try_send()` is consistent with `uv_udp_send()` (regardless the behavior is different on different OSes).

ibc:
OK, here the send status of both `uv_udp_try_send` and `uv_udp_send` in OSX 10.9.3:
- len = 500 => OK
- len = 65536 => error EMSGSIZE
- len = 65000 => error EMSGSIZE

So, the behavior is consistent between both API calls. The fact that the result is different on OSX and Linux does not matter (as the doc currently says that return value may be less than the given buffer length).

ibc:
...but please, remove the ASSERT in the test :)

txdv:
I will leave it, but I will add a check for UV_EMSGSIZE as well like this https://github.com/txdv/libuv/commit/d82542fba2ce95387831822cf0f79af997a5334d

ibc:
@saghul it compiles in my Debian with gcc version 4.7.2 (Debian 4.7.2-5).

txdv:
Now check Solaris, BSD and Windows!

ibc:
@txdv You have al ready added a test for EMSGSIZE, but anyhow, it should be UV_EMSGSIZE:

```
// static char buffer[64 * 1024];

buf = uv_buf_init(buffer, sizeof(buffer));
r = uv_udp_try_send(&client, &buf, 1, (const struct sockaddr*) &addr);
ASSERT(r == UV_EMSGSIZE);
```

txdv:
I already pushed it in: https://github.com/txdv/libuv/commit/ec1f69f4f036995f6711f4d1428e72c7db644925 and https://github.com/txdv/libuv/commit/d82542fba2ce95387831822cf0f79af997a5334d

Something wrong with the timezone setting I guess, because it shows the commits only behind our posts which already happened.

ibc:
Interesting: on Debian I can send a 65000 bytes datagram (with both API calls). And get EMSGSIZE if 65536. I'm not able at all to generate a truncated datagram :)

txdv:
The loopback interface won't truncate, you have to use your ethernet card.

Try changing 127.0.0.1 to the ip address that your ethernet card has.

ibc:
Using non-local interface now. Same on OSX. Never get a truncated datagram, but EMSGSIZE.

Anyhow, really, this **also** happens with `uv_udp_send()`, nothing to fix here :)

saghul:
I'll have Jenkins run tests a bit later.

saghul:
IIRC truncation can only happen on the receiving side, if the provided buffer is not big enough to for the whole datagram. On the sending side you'd get an error.

txdv:
If that is the case, then the code is correct, because you will receive a UV_EMSGSIZE

ibc:
Wow, so it is done, right? anything missing? good work!

saghul:
Jenkins is on it: http://jenkins.nodejs.org/job/libuv-pullrequest/626/

ibc:
I do not understand the result.

## 

Iñaki Baz Castillo
ibc@aliax.net

saghul:
> I do not understand the result.

Click around, there is not much to figure out :-)

Test fails on SmartOS: http://jenkins.nodejs.org/job/libuv-pullrequest/626/DESTCPU=ia32,label=smartos/tapTestReport/test.tap-76/

src/unix/udp.c
```diff
@@ -441,6 +448,48 @@ int uv__udp_send(uv_udp_send_t* req,
...
+
+  memset(&h, 0, sizeof h);
+  h.msg_name = (struct sockaddr*) addr;
+  h.msg_namelen = addrlen;
+  h.msg_iov = (struct iovec*) bufs;
+  h.msg_iovlen = nbufs;
+
+  do {
+    size = sendmsg(handle->io_watcher.fd, &h, 0);
+  }
```
saghul: Put the while on the same line.

test/test-udp-try-send.c
```diff
@@ -0,0 +1,136 @@
...
+  ASSERT(r == UV_EMSGSIZE);
+
+  buf = uv_buf_init(buffer, MAX_UDP_PACKET_SIZE);
+  r = uv_udp_try_send(&client, &buf, 1, (const struct sockaddr*) &addr);
+  ASSERT(r > 0 || r == UV_EMSGSIZE);
+
+  buf = uv_buf_init("EXIT", 4);
+  r = uv_udp_try_send(&client, &buf, 1, (const struct sockaddr*) &addr);
+  ASSERT(r == 4);
+
```
saghul, line 120: Extra newline.

uv.gyp
```diff
@@ -402,6 +402,7 @@
         'test/test-ip6-addr.c',
         'test/test-udp-multicast-interface.c',
         'test/test-udp-multicast-interface6.c',
+        'test/test-udp-try-send.c',
```
saghul, line 405: Also add the test to Makefile.am

test/test-udp-try-send.c
```diff
@@ -0,0 +1,136 @@
+/* Copyright Joyent, Inc. and other Node contributors. All rights reserved.
```
saghul, line 1: The test needs to be skipped on Windows.

test/test-udp-try-send.c
```diff
@@ -0,0 +1,136 @@
...
+  sv_recv_cb_called++;
+}
+
+
+TEST_IMPL(udp_try_send) {
+  struct sockaddr_in addr;
+  static char buffer[64 * 1024];
+  uv_buf_t buf;
+  int r;
+  int count;
```
saghul: Unused.

uv.gyp
txdv, line 405: It is already there, at the top https://github.com/joyent/libuv/pull/1344/files#r14430235

saghul, line 405: Oh, I managed to miss it, sorry about that!

src/unix/udp.c
txdv: you mean

``` c
do {

} while (true);
```

?

ibc: Yes.

test/test-udp-try-send.c
txdv, line 120: Remove the extra new line or add an extra new line? Please be more specific.

ibc, line 120: Remove empty line number 124 (or 125).

txdv:
Whats with the test on SmartOS? Do you want me to install SmartOS and investigate?

ibc:
Info here: http://jenkins.nodejs.org/job/libuv-pullrequest/626/DESTCPU=ia32,label=smartos/console

saghul:
> Whats with the test on SmartOS? Do you want me to install SmartOS and investigate?

It hangs and the test runner kills it. You can add debug commits and I can trigger Jenkins builds.

txdv:
@ibc 

```
not ok 76 - udp_try_send
#Output from process `udp_try_send`: (no output)
```

This is the exact opposite of info, it is lack of info.

@saghul 

The round trip time for modify, check, debug, compile is too long using Jenkins. I can't wait that long. My hdd is already like sliced bread consisting of partitions, so I'll just hit one with SmartOS or eventually run it in a VM when I get home.

Sorry for closing and reopening, I clicked the wrong button.

saghul:
> @saghul https://github.com/saghul
> 
> The round trip time for check, debug, compile is too long using Jenkins.
> I can't wait that long. My hdd is already like sliced bread consisting
> of partitions, so I'll just hit one with SmartOS or eventually run it in
> a VM when I get home.

If that works better for you, go ahead! :-)

> Sorry for closing and reopening, I clicked the wrong button.

No worries!

## 

Saúl Ibarra Corretgé
bettercallsaghul.com

txdv:
Squashed everything, re-review, I'll check smartos today out when I come home.

src/unix/udp.c
```diff
@@ -441,6 +448,47 @@ int uv__udp_send(uv_udp_send_t* req,
...
+
+  do {
+    size = sendmsg(handle->io_watcher.fd, &h, 0);
+  } while (size == -1 && errno == EINTR);
+
+  if (size == -1) {
+    if (errno == EAGAIN || errno == EWOULDBLOCK)
+      return 0;
+    else
+      return -errno;
```
saghul, line 484: I guess here we have the same problem, in case sending an empty packet returns EAGAIN. Not sure how likely that is though.

src/unix/udp.c
```diff
@@ -441,6 +448,47 @@ int uv__udp_send(uv_udp_send_t* req,
...
+                     unsigned int addrlen) {
+  int err;
+  struct msghdr h;
+  ssize_t size;
+
+  assert(nbufs > 0);
+
+  /* already sending a message */
+  if (handle->send_queue_count != 0)
+    return UV_EAGAIN;
```
saghul: Use -EAGAIN, since this code is on unix/

src/unix/udp.c
txdv, line 484: This has to return -EAGAIN.

ibc, line 484: 2014-07-02 14:25 GMT+02:00 Andrius Bentkus notifications@github.com:

> > -  do {
> > -    size = sendmsg(handle->io_watcher.fd, &h, 0);
> > -  } while (size == -1 && errno == EINTR);
> >   +
> > -  if (size == -1) {
> > -    if (errno == EAGAIN || errno == EWOULDBLOCK)
> > -      return 0;
> > -    else
> > -      return -errno;
> 
> This has to return -EAGAIN.

Why should it return -EAGAIN? If sendmsg() fails due to EMSGSIZE it must
return -EMSGSIZE (which becomes UV_MSGSIZE). Do I miss something?

ibc:
Can the `uv_udp_try_send()` implementation guarantee than it would just fail (this is, return a UV_ error) in the **same** cases in which `uv_udp_send()` would fail?

Of course `uv_udp_try_send()` may return 0 (so the user may use `uv_udp_send()` after that) but it should just fail in the same cases in which `uv_udp_send()` fails.

Probably this is already true. I'm just wondering.

txdv:
I don't understand the question.

src/unix/udp.c
ibc, line 484: 2014-07-02 14:33 GMT+02:00 Iñaki Baz Castillo ibc@aliax.net:

> > > -  do {
> > > -    size = sendmsg(handle->io_watcher.fd, &h, 0);
> > > -  } while (size == -1 && errno == EINTR);
> > >   +
> > > -  if (size == -1) {
> > > -    if (errno == EAGAIN || errno == EWOULDBLOCK)
> > > -      return 0;
> > > -    else
> > > -      return -errno;
> > 
> > This has to return -EAGAIN.
> 
> Why should it return -EAGAIN? If sendmsg() fails due to EMSGSIZE it must
> return -EMSGSIZE (which becomes UV_MSGSIZE). Do I miss something?

Sorry, I wrongly understood that you meant the return value after the else.

saghul:
@ibc No. `uv_usdp_send` only fails if the suplied parameters are incorrect, otherwise failures are reported in the callback, whereas `uv_udp_try_send` reports erros synchronously. Also, `uv_udp_send` would never report EAGAIN.

ibc:
> I don't understand the question.

I mean that if I use the same buffer and destination address for both `uv_udp_try_send()` and `uv_udp_send()` they both must behave identically, which means that there should be no one case in which any of them fails and the other not.

NOTE: Let me make clear that I don't consider an error when `uv_udp_try_send()` returns 0.

ibc:
@saghul:

> No. uv_usdp_send only fails if the suplied parameters are incorrect, otherwise failures are reported in the callback, whereas uv_udp_try_send reports erros synchronously. Also, uv_udp_send would never report EAGAIN.

OK, let me improve the question:

I mean that if I use the same buffer and destination address for both uv_udp_try_send() and uv_udp_send() they both must behave identically, which means that there should be no one case in which any of them fails and the other not. And of course, "fail" in `uv_udp_send()` may happen in two places (the return code when supplied parameters are incorrect, and the `uv_udp_send_cb`).

txdv:
uv_udp_try_send will fail when there is something in the queue and will return UV_EAGAIN while uv_udp_send will queue it and won't return an error.

How do you propose we should change this?

ibc:
> uv_udp_try_send will fail when there is something in the queue and will return UV_EAGAIN while uv_udp_send will queue it and won't return an error.

That is wrong. If sendmsg() fails with EAGAIN or EWOULDBLOCK that means that it cannot immediately send the datagram and thus the function uv_udp_try_send() should reply 0 rather than an error. This is what the doc says and this is how uv_try_write() behaves (it does never return UV_EAGAIN to the user but 0, which means "0 bytes written so try uv_write() if you want").

saghul:
> That is wrong. If sendmsg() fails with EAGAIN or EWOULDBLOCK that means that it cannot immediately send the datagram and thus the function uv_udp_try_send() should reply 0 rather than an error. This is what the doc says and this is how uv_try_write() behaves (it does never return UV_EAGAIN to the user but 0, which means "0 bytes written so try uv_write() if you want").

@ibc, no, you are wrong. In UDP you can send an empty datagram, so sendmsg would return 0. That's why we return the error in this case. uv_try_write is different, because on TCP you can't send an empty packet.

ibc:
So lines 464 and 483 must be `return 0` rather than `return -EAGAIN`: https://github.com/txdv/libuv/blob/8e2c76fc033a8284726e509eef90b151c0650711/src/unix/udp.c#L464

txdv:
Think about the corner case of empty udp packets. It sends only the header, no content. The returned length is 0. If try_send returns 0 on not immediate writes as well - it will be ambiguous whether the empty packet has been sent successfully or the queue is full and we need to wait.

ibc:
@saghul I don't think your comment is related to what I said. Please check lines 464 and 483 in [udp.c](https://github.com/txdv/libuv/blob/8e2c76fc033a8284726e509eef90b151c0650711/src/unix/udp.c#L464) and let me know whether they should be `return 0` or not.

saghul:
As a side note, I'll probably also make uv_try_write return EAGAIN.

txdv:
@saghul I found this issue, I will do the patch!

saghul:
@ibc it is related. If we return 0 on EAGAIN and 0 on empty datagram, how do we know if the empty datagram was sent or not?

ibc:
OK, may we please refactor the API documentation for `uv_udp_try_send()` so then it is easier to understand how the code must behave? :)

saghul:
It looks perfectly clear to me as it is right now.

ibc:
Oh, it has been already updated:

```
/*
 * Same as `uv_udp_send()`, but won't queue a send request if it can't be completed
 * immediately.
 * Will return either:
 * - >= 0: number of bytes written (can be less than the supplied buffer size if the
 *         packet is truncated)
 * - < 0: negative error code (UV_EAGAIN is returned when the message can't be sent
 *        immediately)
 */
```

OK, I should have known this before checking the code :)

@saghul if you plan to do the same for `uv_try_write()` please clarify it as some of us currently rely on 0 return code ;)

saghul:
@ibc sure, I'll update the docs too.

ibc:
That would be an API change. I propose to wait for `uv_udp_try_send()` to be merged so you can release a new libuv version.

txdv:
Do you think this pull request will get into the stable branch?

saghul:
It's 0.11 aka unstable, so we are allowed to make changes. I'm not going to propose the change right away, of course.

saghul:
> Do you think this pull request will get into the stable branch?

Next stable, 0.10 is in bugfix mode only.

txdv:
According to Semantic Versioning you can add APIs in stable branches, but not remove or change.

saghul:
We don't do that in libuv.

saghul:
@txdv doh, looks like I missed your last message :-S I modified a bit your test indeed sidestepping that part and tests pass on SmartOS. Here is your branch with my changes (plus some cosmetic ones): https://github.com/joyent/libuv/tree/test-try-send and the jenkins build: http://jenkins.nodejs.org/job/libuv-review/150/

Great work!!

I'll squash the commits before merging, I think this is ready for landing. @indutny, mind taking a quick look at this? If you are too busy you can trust my judgement :-)

src/unix/udp.c
```diff
@@ -441,6 +448,47 @@ int uv__udp_send(uv_udp_send_t* req,
...
+  /* already sending a message */
+  if (handle->send_queue_count != 0)
+    return -EAGAIN;
+
+  err = uv__udp_maybe_deferred_bind(handle, addr->sa_family, 0);
+  if (err)
+    return err;
+
+
+  memset(&h, 0, sizeof h);
```
indutny, line 470: `sizeof(h)`

src/unix/udp.c
```diff
@@ -441,6 +448,47 @@ int uv__udp_send(uv_udp_send_t* req,
...
+  assert(nbufs > 0);
+
+  /* already sending a message */
+  if (handle->send_queue_count != 0)
+    return -EAGAIN;
+
+  err = uv__udp_maybe_deferred_bind(handle, addr->sa_family, 0);
+  if (err)
+    return err;
+
```
indutny, line 469: Unnecessary line break

src/uv-common.c
```diff
@@ -446,6 +466,19 @@ int uv__getaddrinfo_translate_error(int sys_err) {
   return 0;  /* Pacify compiler. */
 }
 
+
+size_t uv__count_bufs(const uv_buf_t bufs[], unsigned int nbufs) {
```
indutny, line 470: Hm... it was inlined on windows before this patch-set, should we consider transforming this into a define with a code wrapped in `do { } while (0)`?

indutny:
The patch looks solid, except few style nits.

src/unix/udp.c
txdv, line 470: Hey, I copied that code from the same file.

src/uv-common.c
saghul, line 470: It wasn't on unix though. I don't think this would be a problem.

src/unix/udp.c
saghul, line 470: We are a bit inconsistent with sizeof, I agree. Both are actually correct in this case so I personally don't mind which one we use.

saghul:
@txdv are you going to squash / address the comments or should I on merge?

txdv:
@saghul 

Please respond to this comment: https://github.com/joyent/libuv/pull/1344#commitcomment-6884948

You just removed the MAX_UDP_PACKET size send completely. I guess that is one way of solving the problem, but the function still behaves in 3 different ways on 3 platforms.
- linux: it sends the packet
- macosx: emsgsize
- solaris: silently fucks everything up

txdv:
@saghul: I'm going to address them. How do you want me to squash them? One big patch?

saghul:
I removed it because it's an unreliable test, as you discovered :-) Thus testing that doesn't tell us anything useful, does it? We do have a test for EMSGSIZE and for try_send itself, I'd say we have our bases well covered.

saghul:
> @saghul: I'm going to address them. How do you want me to squash them? One big patch?

1 patch for the uv__count_bufs, 1 patch for send_queue_size/count, 1 patch for try_send

and please rebase, since you are there :-)

src/unix/udp.c
txdv, line 469: I patched this one

txdv:
so what is it with the sizeof? Do you want me to force to put some ( ) and make the code inconsistent in the same file or do you want me to leave it like it is?

saghul:
I'd go for no parens, keeping it consistent with the rest of the file.

txdv:
@saghul: Can I squash your test changes into my commit?

saghul:
@txdv sure thing!

src/uv-common.c
txdv, line 470: What do I have to do with the `do { } while (0)`?

saghul, line 470: Nothing, let's leave it as is. It was inlined on Windows, but not anymore.
I think its OK.

saghul:
Thank you so much @txdv! Landed in :sparkles: https://github.com/joyent/libuv/commit/f63e073be24c05180c634df7d8d35195b6e942df, https://github.com/joyent/libuv/commit/b769484ca3678208343c938820cdb00cc2732398 and https://github.com/joyent/libuv/commit/0d43992c3b268b756d3018b7e55663f2b056daf3 :sparkles:

ibc:
Excellent work @txdv!! Congrats.

txdv:
:heart: :+1:  and shit

PR #1345, opened 2014-06-29
style: use "(type) value" for all conversations

txdv:
indutny said that `(type) value` is the actual style guide line, so i cooked up this patch


saghul:
We generally don't accept linting patches, because they just add noise to git history and make git archeology harder.

Im -1 on this but I'll let @indutny decide.

indutny:
-1 on it too. Noise in a git blame. Sorry!

PR #1347, opened 2014-06-30
unix: use struct sockaddr_storage for target UDP addr

saghul:
Quick review, @indutny ?


indutny:
LGTM, if it builds.

saghul:
It does! Thanks for the quick review Fedor!

PR #1350, opened 2014-07-02
doc: add documentation to to uv_udp_start_recv

txdv:
This clarifies the behaviour of uv_udp_recv_start.

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -1079,6 +1079,11 @@ UV_EXTERN int uv_udp_send(uv_udp_send_t* req,
  *
  * Returns:
  *  0 on success, or an error code < 0 on failure.
+ *
+ * Note: The receive callback will be called with nread == 0
+ *       and addr == NULL when the there was nothing to read and
+ *       with nread == 0 and addr != NULL when an empty udp
+ *       packet is received.
  */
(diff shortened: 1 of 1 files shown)


saghul:
Thanks! This is not entirely correct, here is when nread==0 and addr==NULL can happen: https://github.com/joyent/libuv/blob/master/src/unix/udp.c#L219 As you can see, it's not related to closing the handle. In UDP, since we read datagrams, when the fd is readable we read until we get EAGAIN, but because alloc_cb was called to allocate the buffer for a datagram, we need to return it back, so the recv callback is called with nread==0 and addr==NULL.

txdv:
Why does it call alloc_cb if there is nothing to read?

saghul:
Because we don't know what there is nothing to read. Once the socket is readable we read until we get EAGAIN, but each time re call recvmsg we need to allocate a buffer first. So, when we get EAGAIN we already called alloc_cb and thus we need to give it back to the user in order for him to free the memory.

txdv:
Makes sense.

saghul:
LGTM now. Can you update the patch title? (to to).

saghul:
Thanks a lot Andrius! Landed! :+1:

PR #1354, opened 2014-07-04
unix: return UV_EAGAIN if uv_try_write cannot write any data

saghul:
For consistency with `uv_udp_try_send`.

Yes, UDP and TCP are different but I think it reasonable to think users will expect the same behavior on both cases. The reason why we return EAGAIN on UDP is bbecause sending an empty datagram is allowed, so we need to know if what happened is that the datagram could not be sent vs an empty datagram was sent. We cannot send an empty TCP packet, but still IMHO consistent behavior is a Good Thing (TM) in this case.

/cc @indutny

@txdv come and agree with me here :-)

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -707,8 +707,9 @@ UV_EXTERN int uv_write2(uv_write_t* req,
  * Same as `uv_write()`, but won't queue write request if it can't be completed
  * immediately.
  * Will return either:
- * - >= 0: number of bytes written (can be less than the supplied buffer size)
- * - < 0: negative error code
+ * - > 0: number of bytes written (can be less than the supplied buffer size)
+ * - < 0: negative error code (UV_EAGAIN is returned if no data can be sent
+ *   immediately)
  */
 UV_EXTERN int uv_try_write(uv_stream_t* handle,
                            const uv_buf_t bufs[],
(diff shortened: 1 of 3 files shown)


indutny:
LGTM, but please remind me about it when we'll upgrade libuv in node.

saghul:
Will do!

txdv:
We **need** to do it for UDP, we don't _need_ to do it for TCP.

But I am for this because of consistency.  :+1:

saghul:
That's what I tried to say in the description. Will merge when brain is fully functional.

txdv:
Your description says exactly the same , I just made a shorter conclusion of it.

saghul:
Landed in https://github.com/joyent/libuv/commit/40ad12e5bedb7f328328f5fb634203e3b84b9af9. I also fixed the test so it would actually test the functionality, in https://github.com/joyent/libuv/commit/b018dc5b72452e3db6fbab00ac1f2b71c4747c43. Before the test would also bail out if the return value was 0, which was always the case on OSX. Now we use a small buffer which makes the test pass.

saghul:
@indutny here is the fix for node: https://github.com/joyent/node/pull/7899

ibc:
:cake:

PR #1356, opened 2014-07-05
doc: grammar, udp documentation fix up

txdv:
FIX ALL THE GRAMMAR :trollface:


include/uv.h
```diff
@@ -901,12 +901,14 @@ typedef void (*uv_udp_recv_cb)(uv_udp_t* handle,
 struct uv_udp_s {
   UV_HANDLE_FIELDS
   /* read-only */
-  /* number of bytes queued for sending, ay send
-   * actually less since udp packets are truncated to the MTU size
+  /* Number of bytes queued for sending. Note that packets can be
+   * truncated and the actual size of the sent information can be less.
```
saghul: How / when does the truncation happen, don't we get EMSGSIZE if we try to send too much? I believe truncation can only happen when receiving a packet, if the supplied buffer is not big enough to accomodate the whole datagram.

txdv:
You prevented me from misleading multiple libuv users :(

saghul:
Cheers , Andrius, landed in https://github.com/joyent/libuv/commit/b9b386ac5a8bf772af95e7238c3f0b6673626923 :+1:

PR #1359, opened 2014-07-06
unix: guarantee write queue cb execution order in streams

txdv:
The problem is that currently the canceled write callbacks
are being called before the already written callbacks, even
though they were queued after them.


saghul:
Thanks Andrius! In this case, since we are changing behavior we'll need a test which checks the ordering.

saghul:
Also, since you are here, can you please add an assert which checks that the write bytes is 0 after running all the completed callbacks?

txdv:
I have done that.

In udp we set recv_cb and alloc_cb to NULL after closing. Should we do it consistently in the entire library?

saghul:
Those will be going away, so please don't make the change now.

txdv:
removed that commit, fixed a typo, please review and merge

saghul:
`uv__write_callbacks` also includes this: https://github.com/joyent/libuv/blob/master/src/unix/stream.c#L895-L897 which doesn't apply in case you are destrying the stream because it already happened. Please move that code after here: https://github.com/joyent/libuv/blob/master/src/unix/stream.c#L1213

saghul:
Also, since we are changing behavior and making guarantees now, we need a test to make sure we never break our promise in the future.

txdv:
Good catch dude  :fish:  :+1:

I should have seen this myself after adding that assert.

txdv:
Added the test, @saghul review review review.

I think a test for udp is hard to make ... we can't fill the queue on a the loopback interface, it will just send regardless.

With tcp we can just not read and wait until it gets filled.

test/test-tcp-queue-order.c
```diff
@@ -0,0 +1,134 @@
...
+  close_cb_called++;
+}
+
+void write_cb(uv_write_t* req, int status) {
+  ASSERT(write_cancelled_callbacks == 0);
+  write_callbacks++;
+}
+
+void write_cb_cancelled(uv_write_t* req, int status) {
+  write_cancelled_callbacks++;
```
saghul: ASSERT that status is UV_ECANCELLED.

test/test-tcp-queue-order.c
```diff
@@ -0,0 +1,134 @@
...
+
+  ASSERT(0 == uv_ip4_addr("0.0.0.0", TEST_PORT, &addr));
+
+  ASSERT(0 == uv_tcp_init(uv_default_loop(), &server));
+  ASSERT(0 == uv_tcp_bind(&server, (struct sockaddr*) &addr, 0));
+  ASSERT(0 == uv_listen((uv_stream_t*) &server, 128, connection_cb));
+}
+
+
+TEST_IMPL(tcp_queue_order) {
```
saghul: Please rename it to tcp_write_queue_order or tcp_write_callback_order.

Makefile.am
```diff
@@ -196,6 +196,7 @@ test_run_tests_SOURCES = test/blackhole-server.c \
                          test/test-tcp-write-to-half-open-connection.c \
                          test/test-tcp-writealot.c \
                          test/test-tcp-try-write.c \
+                         test/test-tcp-queue-order.c \
```
saghul: Also add it to uv.gyp.

txdv:
i addressed your comments for test and squashed them

saghul:
Test fails for me:

```
`tcp_write_queue_order` failed: exit code 134
Output from process `tcp_write_queue_order`:
Assertion failed in ../test/test-tcp-write-queue-order.c on line 55: status == UV_ECANCELED
```

Looks like teh request actually completed:

```
(gdb) bt
#0  0x00007f761440b407 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0x00007f761440c7e8 in __GI_abort () at abort.c:89
#2  0x0000000000434dd8 in write_cb_cancelled (req=0x7e6380 <write_requests+80000>, status=0) at ../test/test-tcp-write-queue-order.c:55
#3  0x0000000000451464 in uv__write_callbacks (stream=0x7d2900 <client>) at ../src/unix/stream.c:872
#4  0x0000000000450341 in uv__stream_destroy (stream=0x7d2900 <client>) at ../src/unix/stream.c:387
#5  0x00000000004457d4 in uv__finish_close (handle=0x7d2900 <client>) at ../src/unix/core.c:198
#6  0x00000000004458db in uv__run_closing_handles (loop=0x86bb00 <default_loop_struct>) at ../src/unix/core.c:228
#7  0x0000000000445ac7 in uv_run (loop=0x86bb00 <default_loop_struct>, mode=UV_RUN_DEFAULT) at ../src/unix/core.c:296
#8  0x0000000000435291 in run_test_tcp_write_queue_order () at ../test/test-tcp-write-queue-order.c:125
#9  0x0000000000406802 in run_test_part (test=0x21bf046 "tcp_write_queue_order", part=0x21bf05c "tcp_write_queue_order") at ../test/runner.c:395
#10 0x0000000000405436 in main (argc=3, argv=0x21bf010) at ../test/run-tests.c:55
(gdb) 
```

txdv:
Try to make the base buffer bigger.

``` c
static char base[1024];
```

what os are you on?

saghul:
I'm on Linux. It passes if I increase FILL_COUNT to 5000.

txdv:
You must have like 32gb of ram or something. I don't know how to reliably scale this.

We could do it with try_write (filling up the buffer), but that doesn't fly with windows.

saghul:
I have 8 GBs of RAM, not sure if it's RAM related. Put it to 5000 and let's see what Jenkins has to say.

txdv:
http://www.cyberciti.biz/faq/linux-tcp-tuning/

```
The default maximum Linux TCP buffer sizes are way too small.
TCP memory is calculated automatically based on system memory.
```

Yes it is! The more you ram have, the more the default size will be.

I think we need a cross-platform function which returns the  maximum amount for the receive socket memory.

Otherwise the test will keep failing if someone puts into 128gb and such.

txdv:
After googling a bit, the solution for windows: http://smallvoid.com/article/winnt-winsock-buffer.html http://stackoverflow.com/questions/4257410/what-are-so-sndbuf-and-so-recvbuf

txdv:
@saghul I found a way to test this without uv_recv_buffer_size.

We just run the loop with NOWAIT until the the normal callback doesn't get called anymore.

txdv:
I think this is it for the this patch. Though before squashing the last two commits a test on windows would be nice, because I still want to have the FILL_COUNT patch around if this doesn't work on windows.

test/test-tcp-write-queue-order.c
```diff
@@ -0,0 +1,139 @@
...
+
+  buf = uv_buf_init(base, sizeof(base));
+
+
+
+  while (1) {
+    r = uv_write(&write_request, req->handle, &buf, 1, write_cb);
+    ASSERT(r == 0);
+
+    uv_run(uv_default_loop(), UV_RUN_NOWAIT);
```
saghul: This is not right. Calling uv_run again while it's running is not supported, even if it works in some corner cases. It's a can of worms.

txdv: I can get it out of there. Wait a second.

saghul:
I'd revert to the previous test and find a high enough FILL_COUNT value to make Jenkins happy. It can be updated to use the buffer size functions afterwards. Or just hold on to it until we merge the other PR.

txdv:
You do understand that FILL_COUNT will one day fail if you put in 128GB RAM.

This solution is so much more beautiful.

review now

test/test-tcp-write-queue-order.c
```diff
@@ -0,0 +1,149 @@
...
+  ASSERT(0 == uv_ip4_addr("127.0.0.1", TEST_PORT, &addr));
+
+  ASSERT(0 == uv_tcp_init(uv_default_loop(), &client));
+  ASSERT(0 == uv_tcp_connect(&connect_req,
+                             &client,
+                             (struct sockaddr*) &addr,
+                             connect_cb));
+
+  ASSERT(1 == uv_run(uv_default_loop(), UV_RUN_ONCE));
+
```
saghul: Assert that connection_cb was called

test/test-tcp-write-queue-order.c
```diff
@@ -0,0 +1,149 @@
...
+  buf = uv_buf_init(base, sizeof(base));
+
+  /* fill up the recv buffer */
+  while (1) {
+    r = uv_write(&write_request, connect_req.handle, &buf, 1, write_cb);
```
saghul: Use client, not connect_req.handle.

txdv: but client is uv_tcp_t and connect_req.handle is stream!

saghul: On 07/07/2014 04:35 PM, Andrius Bentkus wrote:

> In test/test-tcp-write-queue-order.c:
> 
> > +
> > -  ASSERT(0 == uv_ip4_addr("127.0.0.1", TEST_PORT, &addr));
> >   +
> > -  ASSERT(0 == uv_tcp_init(uv_default_loop(), &client));
> > -  ASSERT(0 == uv_tcp_connect(&connect_req,
> > -                             &client,
> > -                             (struct sockaddr*) &addr,
> > -                             connect_cb));
> >   +
> > -  ASSERT(1 == uv_run(uv_default_loop(), UV_RUN_ONCE));
> >   +
> > -  buf = uv_buf_init(base, sizeof(base));
> >   +
> > -  /\* fill up the recv buffer */
> > -  while (1) {
> > -    r = uv_write(&write_request, connect_req.handle, &buf, 1, write_cb);
> 
> but client is uv_tcp_t and connect_req.handle is stream!

But it looks odd. Please just cast it.

txdv: already did it

test/test-tcp-write-queue-order.c
```diff
@@ -0,0 +1,149 @@
...
+static void close_cb(uv_handle_t* handle) {
+  close_cb_called++;
+}
+
+void write_cb(uv_write_t* req, int status) {
+  ASSERT(write_cancelled_callbacks == 0);
+
+  /* we set the last write to have content in req
+   * data, so we check if it was canceled as well */
+  if (req->data)
```
saghul: assert it equals to (void*)1.

test/test-tcp-write-queue-order.c
```diff
@@ -0,0 +1,149 @@
...
+  ASSERT(0 == uv_ip4_addr("127.0.0.1", TEST_PORT, &addr));
+
+  ASSERT(0 == uv_tcp_init(uv_default_loop(), &client));
+  ASSERT(0 == uv_tcp_connect(&connect_req,
+                             &client,
+                             (struct sockaddr*) &addr,
+                             connect_cb));
+
+  ASSERT(1 == uv_run(uv_default_loop(), UV_RUN_ONCE));
+  ASSERT(connection_cb_called == 1);
```
saghul: This might need more than one iteration to happen, run it in a loop until connection_cb is called. Test failure: http://jenkins.nodejs.org/job/libuv-pullrequest/lastCompletedBuild/DESTCPU=x64,label=smartos/tapTestReport/test.tap-38/

txdv: thought the same while going home. Technically you have two events, one connect and one connection cb.

test/test-tcp-write-queue-order.c
txdv: I can't do that.

Because its only 1 time equal 1.

However I can assert that it gets executed one time with req->data == 1

saghul: That's what I meant.

test/test-tcp-write-queue-order.c
txdv: It returns 1 when more events are expected.

So probably on my machine it ran the connection cb of the client first.

txdv:
run the test with jenkins again please. does jenkins test windows as well?

saghul:
Afk, will do in a bit.

saghul:
Tests pass on Unixes: http://jenkins.nodejs.org/job/libuv-pullrequest/628/ will test on Windows later tonight.

test/test-tcp-write-queue-order.c
```diff
@@ -0,0 +1,151 @@
...
+
+  /* fill up the recv buffer */
+  while (1) {
+    r = uv_write(&write_request, (uv_stream_t*) &client, &buf, 1, write_cb);
+    ASSERT(r == 0);
+
+    uv_run(uv_default_loop(), UV_RUN_NOWAIT);
+    /* the buffer must be full if write_callbacks
+     * doesnt increase anymore */
+    if (write_callbacks == last_write_callbacks) {
```
saghul: last_write_callbacks is used unitialized.

saghul: I'm a bit uncertain about how this should work. If no write happens immediately you'll bail out in the first loop iteration, and if the request succeeds on the first try and the callback is called, we might call uv_write again with the same write_request object.

txdv: I don't understand you.

txdv: If we succeed immediately, then the write callback has been called in the uv_run(NOWAIT) and the write_requeust is free to use.

txdv: And it can't bail on the first try, unless your tcp send buffer is < 1024 and the receive buffer of the other side is as well.

txdv: have you tested it on windows?

saghul: I have tested it on Windows. It fails. What is the initial value of last_write_callbacks supposed to be? You cannot guarantee that the write callback will complete on the first loop iteration. As a result, you'll bail on the first while loop iteration (assuming last_write_callbacks is initialized to 0, but even then the test fails on Windows).

So, looks to me that the test is basically sending one write request, waiting until it's completed and then sendign a another one which will be canceled. This doesn't quite cover the scenario we need.

txdv: If we can't guarantee that, there is no way to test it.

txdv: Do we have something which returns the number of outstanding operations?

We might need something like that.

saghul: > Do we have something which returns the number of outstanding operations?
> 
> We might need something like that.

No, we don't have such a thing, and I don't want to expose it. Just brute force it, it doesn't need to be the prettiest test, it needs to test that the behavior is correct. When our Jenkins machines get 128GBs of RAM we can adjust it :-)

txdv:
This test is not better for the same reason why it failed on windows: Not a single write will happen on windows will happen, everything will be cancelled.

txdv:
The timeout should ensure that there is at least something written by windows, even though it doesn't make the test fail if nothing is written.

test/test-tcp-write-queue-order.c
```diff
@@ -0,0 +1,141 @@
...
+  uv_close((uv_handle_t*) &incoming, close_cb);
+}
+
+void write_cb(uv_write_t* req, int status) {
+  ASSERT(write_cancelled_callbacks == 0);
+  write_callbacks++;
+}
+
+void write_cb_cancelled(uv_write_t* req, int status) {
+  ASSERT(status == UV_ECANCELED);
```
saghul: Test blows up here on Windows. I don't think we can guarantee that all REST_COUNT requests are going to be cancelled. Check if status is 0 or ECANCELLED and increase either counter. Line 57 will make sure the order is preserved. At the end, assert that both write_callbacks and write_cancelled_callbacks are > 0, and their sum is equal to FILL_COUNT + REST_COUNT.

txdv: You want me to tell that on Windows 7 the socket buffers 10mb of information?

txdv: It can't blow up there. We are writing in the other requests 10mb/s of data.

How does it not block?

saghul:
I just tried it again, and now I just get no output :-/

```
[%  15|+  36|-   0|T   0|S   0]: tcp_write_queue_order
`tcp_write_queue_order` failed: exit code 3
Output from process `tcp_write_queue_order`: (no output)
=============================================================
```

saghul:
I moved the call to `uv_timer_start` to the connection callback, and now I do get the failure just where I said:

```
PS C:\Users\saghul\src\libuv> .\Debug\run-tests.exe tcp_write_queue_order

`tcp_write_queue_order` failed: exit code 3
Output from process `tcp_write_queue_order`:
Assertion failed in test\test-tcp-write-queue-order.c on line 62: status == UV_ECANCELED
=============================================================
PS C:\Users\saghul\src\libuv>
```

saghul:
Here is an attempt to make it work: https://gist.github.com/saghul/989ec675c839ae9a6bef

It sometimes passes, but when it doesn't it leaves no trace, no assert output, nothing :-(

txdv:
It runs fine with the msvc compiler, just tested windows 8/msvc2012

txdv:
I get this error when building:

```
make  all-am
make[1]: Entering directory '/home/Andrius/Projects/c/libuv'
  CC       src/libuv_la-fs-poll.lo
In file included from ./include/uv.h:61:0,
                 from src/fs-poll.c:22:
./include/uv-unix.h:148:1: error: unknown type name 'pthread_barrier_t'
 typedef pthread_barrier_t uv_barrier_t;
 ^
Makefile:1510: recipe for target 'src/libuv_la-fs-poll.lo' failed
make[1]: *** [src/libuv_la-fs-poll.lo] Error 1
make[1]: Leaving directory '/home/Andrius/Projects/c/libuv'
Makefile:837: recipe for target 'all' failed
make: *** [all] Error 2
```

Can you help me with it? Im using cygwin/make on windows8

saghul:
> Im using cygwin/make on windows8

cygwin is not supported. I suggest you get MSYS2, it's really nice, it almost feels like home! http://msys2.github.io/

txdv:
I get absolutely the same error with msys2

saghul:
> I get absolutely the same error with msys2

How did you start the console, with mingw32_shell.bat? I just tried here and it compiled without issues.

txdv:
jup, absolutely the same

saghul:
What MinGW did you install, the one that comes with MSYS2? I installed mingw-w64-i686-gcc 4.9.0-4 with pacman.

Can you paste gcc -v ?

txdv:
```
$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/i686-pc-msys/4.8.2/lto-wrapper.exe
Target: i686-pc-msys
Configured with: /msys_repo/gcc/src/gcc-4.8.2/configure --build=i686-pc-msys --prefix=/usr --libexecdir=/usr/lib --enable-bootstrap --enable-shared --enable-shared-libgcc --enable-static --enable-version-specific-runtime-libs --with-arch=i686 --disable-multilib --disable-sjlj-exceptions --with-tune=generic --disable-__cxa_atexit --enable-languages=c,c++,fortran,lto --enable-graphite --enable-threads=posix --enable-libatomic --enable-libgomp --disable-libitm --enable-libquadmath --enable-libquadmath-support --enable-libssp --disable-win32-registry --disable-symvers --with-gnu-ld --with-gnu-as --disable-isl-version-check --enable-checking=release --without-libiconv-prefix --without-libintl-prefix --with-system-zlib
Thread model: posix
gcc version 4.8.2 (GCC)
```

I didnt install anything.

But the test, as I already said, passes on msvs2012 msvc

saghul:
That's not the right MinGW compiler. How did you install gcc?

Anyway, the test doesn't pass for me on Windows 7 (32 bits) on msvs2012 msvc.

txdv:
i installed gcc with pacman

how do i install the right compiler?

saghul:
Use the msys2_shell.bat and run:

```
pacman -S mingw-w64-i686-gcc 
```

That will get you the 32 bit gcc. Then close the shell and open mingw32_shell.bat. You should get something like this for gcc -v:

```
$ gcc -v
Using built-in specs.
COLLECT_GCC=C:\msys32\mingw32\bin\gcc.exe
COLLECT_LTO_WRAPPER=C:/msys32/mingw32/bin/../lib/gcc/i686-w64-mingw32/4.9.0/lto-wrapper.exe
Target: i686-w64-mingw32
Configured with: ../gcc-4.9.0/configure --prefix=/mingw32 --with-local-prefix=/mingw32/local --build=i686-w64-mingw32 --host=i686-w64-mingw32 --target=i686-w64-mingw32 --with-native-system-header-dir=/mingw32/i686-w64-mingw32/include --libexecdir=/mingw32/lib --with-gxx-include-dir=/mingw32/include/c++/4.9.0 --enable-bootstrap --with-arch=i686 --with-tune=generic --enable-languages=c,lto,c++,objc,obj-c++,fortran,ada --enable-shared --enable-static --enable-libatomic --enable-threads=posix --enable-graphite --enable-fully-dynamic-string --enable-libstdcxx-time=yes --disable-libstdcxx-pch --disable-libstdcxx-debug --enable-cloog-backend=isl --enable-version-specific-runtime-libs --disable-cloog-version-check --disable-isl-version-check --enable-lto --enable-libgomp --disable-multilib --enable-checking=release --disable-rpath --disable-win32-registry --disable-nls --disable-werror --disable-symvers --with-libiconv --with-system-zlib --with-gmp=/mingw32 --with-mpfr=/mingw32 --with-mpc=/mingw32 --with-isl=/mingw32 --with-cloog=/mingw32 --with-pkgversion='Rev4, Built by MSYS2 project' --with-bugurl=http://sourceforge.net/projects/msys2 --with-gnu-as --with-gnu-ld --disable-sjlj-exceptions --with-dwarf2
Thread model: posix
gcc version 4.9.0 (Rev4, Built by MSYS2 project)
```

txdv:
It just tries to compile the src/unix and throws a lot of errors.

```
$ make
make  all-am
make[1]: Entering directory '/home/Andrius/Projects/c/libuv'
  CC       src/unix/libuv_la-async.lo
In file included from src/unix/async.c:26:0:
src/unix/internal.h:161:36: warning: 'struct msghdr' declared inside parameter l                                                                                                                                                             ist
 ssize_t uv__recvmsg(int fd, struct msghdr *msg, int flags);
                                    ^
src/unix/internal.h:161:36: warning: its scope is only this definition or declar                                                                                                                                                             ation, which is probably not what you want
src/unix/internal.h:164:18: error: unknown type name 'uv__io_t'
 void uv__io_init(uv__io_t* w, uv__io_cb cb, int fd);
                  ^
src/unix/internal.h:164:31: error: unknown type name 'uv__io_cb'
 void uv__io_init(uv__io_t* w, uv__io_cb cb, int fd);
                               ^
src/unix/internal.h:165:36: error: unknown type name 'uv__io_t'
 void uv__io_start(uv_loop_t* loop, uv__io_t* w, unsigned int events);
                                    ^
src/unix/internal.h:166:35: error: unknown type name 'uv__io_t'
 void uv__io_stop(uv_loop_t* loop, uv__io_t* w, unsigned int events);
                                   ^
src/unix/internal.h:167:36: error: unknown type name 'uv__io_t'
 void uv__io_close(uv_loop_t* loop, uv__io_t* w);
                                    ^
src/unix/internal.h:168:35: error: unknown type name 'uv__io_t'
 void uv__io_feed(uv_loop_t* loop, uv__io_t* w);
                                   ^
src/unix/internal.h:169:25: error: unknown type name 'uv__io_t'
 int uv__io_active(const uv__io_t* w, unsigned int events);
                         ^
src/unix/internal.h:173:28: warning: 'struct uv__async' declared inside paramete                                                                                                                                                             r list
 void uv__async_send(struct uv__async* wa);
                            ^
src/unix/internal.h:174:28: warning: 'struct uv__async' declared inside paramete                                                                                                                                                             r list
 void uv__async_init(struct uv__async* wa);
                            ^
src/unix/internal.h:175:60: error: unknown type name 'uv__async_cb'
 int uv__async_start(uv_loop_t* loop, struct uv__async* wa, uv__async_cb cb);
                                                            ^
src/unix/internal.h:176:45: warning: 'struct uv__async' declared inside paramete                                                                                                                                                             r list
 void uv__async_stop(uv_loop_t* loop, struct uv__async* wa);
                                             ^
src/unix/internal.h:191:37: error: unknown type name 'uv__io_t'
 void uv__server_io(uv_loop_t* loop, uv__io_t* w, unsigned int events);
                                     ^
src/unix/async.c:37:29: warning: 'struct uv__async' declared inside parameter li                                                                                                                                                             st
                             unsigned int nevents);
                             ^
src/unix/async.c: In function 'uv_async_init':
src/unix/async.c:45:36: error: 'uv_loop_t' has no member named 'async_watcher'
   err = uv__async_start(loop, &loop->async_watcher, uv__async_event);
                                    ^
src/unix/async.c:51:9: error: 'uv_async_t' has no member named 'pending'
   handle->pending = 0;
         ^
In file included from ./src/uv-common.h:41:0,
                 from src/unix/internal.h:25,
                 from src/unix/async.c:26:
src/unix/async.c:53:50: error: 'uv_async_t' has no member named 'queue'
   QUEUE_INSERT_TAIL(&loop->async_handles, &handle->queue);
                                                  ^
./src/queue.h:22:47: note: in definition of macro 'QUEUE_NEXT'
 #define QUEUE_NEXT(q)       (*(QUEUE **) &((*(q))[0]))
                                               ^
src/unix/async.c:53:3: note: in expansion of macro 'QUEUE_INSERT_TAIL'
   QUEUE_INSERT_TAIL(&loop->async_handles, &handle->queue);
   ^
src/unix/async.c:53:26: error: 'uv_loop_t' has no member named 'async_handles'
   QUEUE_INSERT_TAIL(&loop->async_handles, &handle->queue);
                          ^
./src/queue.h:78:22: note: in definition of macro 'QUEUE_INSERT_TAIL'
     QUEUE_NEXT(q) = (h);                                                      \
                      ^
src/unix/async.c:53:50: error: 'uv_async_t' has no member named 'queue'
   QUEUE_INSERT_TAIL(&loop->async_handles, &handle->queue);
                                                  ^
./src/queue.h:23:47: note: in definition of macro 'QUEUE_PREV'
 #define QUEUE_PREV(q)       (*(QUEUE **) &((*(q))[1]))
                                               ^
src/unix/async.c:53:3: note: in expansion of macro 'QUEUE_INSERT_TAIL'
   QUEUE_INSERT_TAIL(&loop->async_handles, &handle->queue);
   ^
src/unix/async.c:53:26: error: 'uv_loop_t' has no member named 'async_handles'
   QUEUE_INSERT_TAIL(&loop->async_handles, &handle->queue);
                          ^
./src/queue.h:23:47: note: in definition of macro 'QUEUE_PREV'
 #define QUEUE_PREV(q)       (*(QUEUE **) &((*(q))[1]))
                                               ^
src/unix/async.c:53:3: note: in expansion of macro 'QUEUE_INSERT_TAIL'
   QUEUE_INSERT_TAIL(&loop->async_handles, &handle->queue);
   ^
src/unix/async.c:53:50: error: 'uv_async_t' has no member named 'queue'
   QUEUE_INSERT_TAIL(&loop->async_handles, &handle->queue);
                                                  ^
./src/queue.h:22:47: note: in definition of macro 'QUEUE_NEXT'
 #define QUEUE_NEXT(q)       (*(QUEUE **) &((*(q))[0]))
                                               ^
./src/queue.h:24:41: note: in expansion of macro 'QUEUE_PREV'
 #define QUEUE_PREV_NEXT(q)  (QUEUE_NEXT(QUEUE_PREV(q)))
                                         ^
./src/queue.h:80:5: note: in expansion of macro 'QUEUE_PREV_NEXT'
     QUEUE_PREV_NEXT(q) = (q);                                                 \
     ^
src/unix/async.c:53:3: note: in expansion of macro 'QUEUE_INSERT_TAIL'
   QUEUE_INSERT_TAIL(&loop->async_handles, &handle->queue);
   ^
src/unix/async.c:53:50: error: 'uv_async_t' has no member named 'queue'
   QUEUE_INSERT_TAIL(&loop->async_handles, &handle->queue);
                                                  ^
./src/queue.h:80:27: note: in definition of macro 'QUEUE_INSERT_TAIL'
     QUEUE_PREV_NEXT(q) = (q);                                                 \
                           ^
src/unix/async.c:53:26: error: 'uv_loop_t' has no member named 'async_handles'
   QUEUE_INSERT_TAIL(&loop->async_handles, &handle->queue);
                          ^
./src/queue.h:23:47: note: in definition of macro 'QUEUE_PREV'
 #define QUEUE_PREV(q)       (*(QUEUE **) &((*(q))[1]))
                                               ^
src/unix/async.c:53:3: note: in expansion of macro 'QUEUE_INSERT_TAIL'
   QUEUE_INSERT_TAIL(&loop->async_handles, &handle->queue);
   ^
src/unix/async.c:53:50: error: 'uv_async_t' has no member named 'queue'
   QUEUE_INSERT_TAIL(&loop->async_handles, &handle->queue);
                                                  ^
./src/queue.h:81:22: note: in definition of macro 'QUEUE_INSERT_TAIL'
     QUEUE_PREV(h) = (q);                                                      \
                      ^
src/unix/async.c: In function 'uv_async_send':
src/unix/async.c:61:37: error: 'uv_async_t' has no member named 'pending'
   if (uv__async_make_pending(&handle->pending) == 0)
                                     ^
src/unix/async.c:62:33: error: 'uv_loop_t' has no member named 'async_watcher'
     uv__async_send(&handle->loop->async_watcher);
                                 ^
In file included from ./src/uv-common.h:41:0,
                 from src/unix/internal.h:25,
                 from src/unix/async.c:26:
src/unix/async.c: In function 'uv__async_close':
src/unix/async.c:69:23: error: 'uv_async_t' has no member named 'queue'
   QUEUE_REMOVE(&handle->queue);
                       ^
./src/queue.h:22:47: note: in definition of macro 'QUEUE_NEXT'
 #define QUEUE_NEXT(q)       (*(QUEUE **) &((*(q))[0]))
                                               ^
./src/queue.h:24:41: note: in expansion of macro 'QUEUE_PREV'
 #define QUEUE_PREV_NEXT(q)  (QUEUE_NEXT(QUEUE_PREV(q)))
                                         ^
./src/queue.h:87:5: note: in expansion of macro 'QUEUE_PREV_NEXT'
     QUEUE_PREV_NEXT(q) = QUEUE_NEXT(q);                                       \
     ^
src/unix/async.c:69:3: note: in expansion of macro 'QUEUE_REMOVE'
   QUEUE_REMOVE(&handle->queue);
   ^
src/unix/async.c:69:23: error: 'uv_async_t' has no member named 'queue'
   QUEUE_REMOVE(&handle->queue);
                       ^
./src/queue.h:22:47: note: in definition of macro 'QUEUE_NEXT'
 #define QUEUE_NEXT(q)       (*(QUEUE **) &((*(q))[0]))
                                               ^
src/unix/async.c:69:3: note: in expansion of macro 'QUEUE_REMOVE'
   QUEUE_REMOVE(&handle->queue);
   ^
src/unix/async.c:69:23: error: 'uv_async_t' has no member named 'queue'
   QUEUE_REMOVE(&handle->queue);
                       ^
./src/queue.h:23:47: note: in definition of macro 'QUEUE_PREV'
 #define QUEUE_PREV(q)       (*(QUEUE **) &((*(q))[1]))
                                               ^
./src/queue.h:25:41: note: in expansion of macro 'QUEUE_NEXT'
 #define QUEUE_NEXT_PREV(q)  (QUEUE_PREV(QUEUE_NEXT(q)))
                                         ^
./src/queue.h:88:5: note: in expansion of macro 'QUEUE_NEXT_PREV'
     QUEUE_NEXT_PREV(q) = QUEUE_PREV(q);                                       \
     ^
src/unix/async.c:69:3: note: in expansion of macro 'QUEUE_REMOVE'
   QUEUE_REMOVE(&handle->queue);
   ^
src/unix/async.c:69:23: error: 'uv_async_t' has no member named 'queue'
   QUEUE_REMOVE(&handle->queue);
                       ^
./src/queue.h:23:47: note: in definition of macro 'QUEUE_PREV'
 #define QUEUE_PREV(q)       (*(QUEUE **) &((*(q))[1]))
                                               ^
src/unix/async.c:69:3: note: in expansion of macro 'QUEUE_REMOVE'
   QUEUE_REMOVE(&handle->queue);
   ^
src/unix/async.c: At top level:
src/unix/async.c:76:29: warning: 'struct uv__async' declared inside parameter li                                                                                                                                                             st
                             unsigned int nevents) {
                             ^
src/unix/async.c:74:13: error: conflicting types for 'uv__async_event'
 static void uv__async_event(uv_loop_t* loop,
             ^
src/unix/async.c:35:13: note: previous declaration of 'uv__async_event' was here
 static void uv__async_event(uv_loop_t* loop,
             ^
In file included from ./src/uv-common.h:41:0,
                 from src/unix/internal.h:25,
                 from src/unix/async.c:26:
src/unix/async.c: In function 'uv__async_event':
src/unix/async.c:80:25: error: 'uv_loop_t' has no member named 'async_handles'
   QUEUE_FOREACH(q, &loop->async_handles) {
                         ^
./src/queue.h:22:47: note: in definition of macro 'QUEUE_NEXT'
 #define QUEUE_NEXT(q)       (*(QUEUE **) &((*(q))[0]))
                                               ^
src/unix/async.c:80:3: note: in expansion of macro 'QUEUE_FOREACH'
   QUEUE_FOREACH(q, &loop->async_handles) {
   ^
src/unix/async.c:80:25: error: 'uv_loop_t' has no member named 'async_handles'
   QUEUE_FOREACH(q, &loop->async_handles) {
                         ^
./src/queue.h:32:37: note: in definition of macro 'QUEUE_FOREACH'
   for ((q) = QUEUE_NEXT(h); (q) != (h); (q) = QUEUE_NEXT(q))
                                     ^
./src/queue.h:29:54: error: 'uv_async_t' has no member named 'queue'
   ((type *) ((char *) (ptr) - ((char *) &((type *) 0)->field)))
                                                      ^
src/unix/async.c:81:9: note: in expansion of macro 'QUEUE_DATA'
     h = QUEUE_DATA(q, uv_async_t, queue);
         ^
src/unix/async.c:83:10: error: 'uv_async_t' has no member named 'pending'
     if (h->pending == 0)
          ^
src/unix/async.c:85:6: error: 'uv_async_t' has no member named 'pending'
     h->pending = 0;
      ^
src/unix/async.c: At top level:
src/unix/async.c:125:43: error: unknown type name 'uv__io_t'
 static void uv__async_io(uv_loop_t* loop, uv__io_t* w, unsigned int events) {
                                           ^
src/unix/async.c:169:28: warning: 'struct uv__async' declared inside parameter l                                                                                                                                                             ist
 void uv__async_send(struct uv__async* wa) {
                            ^
src/unix/async.c:169:6: error: conflicting types for 'uv__async_send'
 void uv__async_send(struct uv__async* wa) {
      ^
In file included from src/unix/async.c:26:0:
src/unix/internal.h:173:6: note: previous declaration of 'uv__async_send' was he                                                                                                                                                             re
 void uv__async_send(struct uv__async* wa);
      ^
src/unix/async.c: In function 'uv__async_send':
src/unix/async.c:177:10: error: dereferencing pointer to incomplete type
   fd = wa->wfd;
          ^
src/unix/async.c: At top level:
src/unix/async.c:203:28: warning: 'struct uv__async' declared inside parameter l                                                                                                                                                             ist
 void uv__async_init(struct uv__async* wa) {
                            ^
src/unix/async.c:203:6: error: conflicting types for 'uv__async_init'
 void uv__async_init(struct uv__async* wa) {
      ^
In file included from src/unix/async.c:26:0:
src/unix/internal.h:174:6: note: previous declaration of 'uv__async_init' was he                                                                                                                                                             re
 void uv__async_init(struct uv__async* wa);
      ^
src/unix/async.c: In function 'uv__async_init':
src/unix/async.c:204:5: error: dereferencing pointer to incomplete type
   wa->io_watcher.fd = -1;
     ^
src/unix/async.c:205:5: error: dereferencing pointer to incomplete type
   wa->wfd = -1;
     ^
src/unix/async.c: At top level:
src/unix/async.c:209:60: error: unknown type name 'uv__async_cb'
 int uv__async_start(uv_loop_t* loop, struct uv__async* wa, uv__async_cb cb) {
                                                            ^
src/unix/async.c:256:45: warning: 'struct uv__async' declared inside parameter l                                                                                                                                                             ist
 void uv__async_stop(uv_loop_t* loop, struct uv__async* wa) {
                                             ^
src/unix/async.c:256:6: error: conflicting types for 'uv__async_stop'
 void uv__async_stop(uv_loop_t* loop, struct uv__async* wa) {
      ^
In file included from src/unix/async.c:26:0:
src/unix/internal.h:176:6: note: previous declaration of 'uv__async_stop' was he                                                                                                                                                             re
 void uv__async_stop(uv_loop_t* loop, struct uv__async* wa);
      ^
src/unix/async.c: In function 'uv__async_stop':
src/unix/async.c:257:9: error: dereferencing pointer to incomplete type
   if (wa->io_watcher.fd == -1)
         ^
src/unix/async.c:260:9: error: dereferencing pointer to incomplete type
   if (wa->wfd != -1) {
         ^
src/unix/async.c:261:11: error: dereferencing pointer to incomplete type
     if (wa->wfd != wa->io_watcher.fd)
           ^
src/unix/async.c:261:22: error: dereferencing pointer to incomplete type
     if (wa->wfd != wa->io_watcher.fd)
                      ^
src/unix/async.c:262:19: error: dereferencing pointer to incomplete type
       uv__close(wa->wfd);
                   ^
src/unix/async.c:263:7: error: dereferencing pointer to incomplete type
     wa->wfd = -1;
       ^
src/unix/async.c:266:24: error: dereferencing pointer to incomplete type
   uv__io_stop(loop, &wa->io_watcher, UV__POLLIN);
                        ^
src/unix/async.c:267:15: error: dereferencing pointer to incomplete type
   uv__close(wa->io_watcher.fd);
               ^
src/unix/async.c:268:5: error: dereferencing pointer to incomplete type
   wa->io_watcher.fd = -1;
     ^
src/unix/async.c: At top level:
src/unix/async.c:35:13: warning: 'uv__async_event' used but never defined
 static void uv__async_event(uv_loop_t* loop,
             ^
Makefile:1720: recipe for target 'src/unix/libuv_la-async.lo' failed
make[1]: *** [src/unix/libuv_la-async.lo] Error 1
make[1]: Leaving directory '/home/Andrius/Projects/c/libuv'
Makefile:837: recipe for target 'all' failed
make: *** [all] Error 2
```

Can you please give me the full list of steps I have to do?

You have clearly already gone through all of them.

txdv:
NEVERMIND, had to re autogen autoconfigure configure

txdv:
```
$ make test/run-tests
gcc -g -O2    test/run-tests.c   -o test/run-tests
test/run-tests.c:32:16: fatal error: uv.h: No such file or directory
 #include "uv.h"
                ^
compilation terminated.
<builtin>: recipe for target 'test/run-tests' failed
make: *** [test/run-tests] Error 1
```

I can compile the library but not the run-tests. What now?

txdv:
I'm fucking bewildered by this patch.

I just got a failure on windows with msvc.

```
Assertion failed in test\test-tcp-write-queue-order.c on line 62: status == UV_ECANCELED
```

It probably didn't fill the IOCP queue.

with msys2 it crashes with: 

```
This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.
```

AND THE MOST CONFUSING THING is that it affects the audio output of iTunes. Running the test tcp_write_queue_order will affect my audio output.

saghul:
I'll try to debug this tonight.

saghul:
Hey @txdv, sorry for the delay. I think I finally understand what's going on. Here is a simplified version of the test, which passes on Windows: https://github.com/saghul/libuv/commit/6efddc9c9b853f4e241c8936e79a1252ec89f469

The reason why this didn't work on Windows turns out to be pretty simple: we will never get UV_ECANCELED there, period. On Windows you can apparently throw as much shit as you want to the kernel and it will be queued: https://github.com/joyent/libuv/blob/master/src/win/tcp.c#L836-L868 so we don't keep the queue, the kernel does. Gotta love IOCP. Then, when the stream is closed on the timer, all pending callbacks in the kernel are called with ECONNABORTED.

Feel free to incorporate my changes. Patch looks solid, can you rebase since you are at it?

About the iTunes weird thing, watch out, it might have become self-aware! :-P

txdv:
It makes sense now.

Shoudln't we return ECONNABORTED on unix as well? We close the socket, the remaining writing operations are queued with ECANCELLED here: https://github.com/joyent/libuv/pull/1359/files#diff-c9c9805d800eb132a191a9e40da6d813R382 

We close the handle, all the remaining queued requests should be called with ECONNABORTED.

Either that or we map ECONNABORTED to ECANCELED on windows.

We need consistent API.

What do you think?

saghul:
ECANCELED makes more sense to me, since the request was actually canceled.
The fact that we don't do it on Windows can be considered a limitation
because of how IOCP works (or how we use it).

We can discuss that on another issue.

txdv:
CANCELED and ABORTED is basically the same, the operation has been cancelled, but with ABORTED it gives more information to us: the socket has been closed or some failure happened on the socket, it is not just a simple cancel.

This is the case in the https://github.com/joyent/libuv/pull/1359/files#diff-c9c9805d800eb132a191a9e40da6d813R382

We are about to close the socket and flushing the queue in order to invoke the callbacks. It totally makes sense to me that we use ABORTED instead of CANCELED.

IOCP even supports cancelling operations: http://msdn.microsoft.com/en-us/library/windows/desktop/aa363789(v=vs.85).aspx

> You can cancel asynchronous I/O from any thread in the process that issued the I/O operation. You must specify the handle which the I/O was performed on and, optionally, the overlapped structure that was used to perform the I/O. You can determine if the cancel occurred by examining the status returned in the overlapped structure or in the completion callback. A status of ERROR_OPERATION_ABORTED indicates that the operation was canceled.

ERROR_OPERATION_ABORTED = ECANCELLED

Even from a binding for C# perspective it makes sense to me: I would make Aborted a subclass of Cancelled.

saghul:
Please, let's focus on this PRs initial purpose, and open a new issue/PR
for the other change.

txdv:
squashed your commit, please review, ill add another thread

saghul:
Thanks, Andrius. Will review in detail when my Internet provider stops
sucking. :-/

txdv:
I think we should not be using the idfef.

I already told you that the test passed as it was on the msvc compiler, so this is a safer bet.

saghul:
I can reproduce the failure 100% of the time on Windows 7, using MSVC.

txdv:
I'm using microsofts c(c++?) compiler.

Anyway, the assert has still the same guarantee, doesn't it? And it is one macro less.

saghul:
Same here. How are you building / running the tests? I'm doing vcbuild.bat
test on a PowerShell prompt.

txdv:
When I ran the tests a week ago I didn't get the failure, I build it with the IDE. Anyway, I don't care how it looks, I just want to get through your review wall.

Are you ok with the removal of that ifdef?

saghul:
No, please leave it. Let's discuss converting ECONNABORTED into ECANCELED
afterwards, if we make that change the ifdef will go.

saghul:
Thanks Andrius! Landed in :sparkles: 960eefb :sparkles:

PR #1360, opened 2014-07-06
win: set SO_REUSEADDR when binding TCP sockets

retep998:
We already did this for unix, so this brings parity to Windows.
Hopefully it works the same on Windows as it does on unix.


saghul:
Hi,

Unfortunately, they behave totally different and you can consider SO_REUSEADDR harmful in Windows. Usually we use SO_REUSEADDR on Unix to be able to bind to the same address and port of a socket in the TIME_WAIT state. Windows, however does't check for the state if SO_REUSEADDR is set, so you can effectively hijack that port from another process. You can use SO_EXCLUSIVEADDRUSE to prevent that, but then sockets in TIME_WAIT state are also checked, which is not what we want. Here is a link explaining SO_EXCLUSIVEADDRUSE: http://msdn.microsoft.com/en-us/library/windows/desktop/ms740621%28v=vs.85%29.aspx Here is a more thorough explanation: http://stackoverflow.com/a/14388707

So, we don't really need to set SO_REUSEADDR on Windows because sockets in TIME_WAIT are not checked when binding.

Now, did you run into any problem related to this? Do you have a test case I can reproduce?

retep998:
The reason I reported the original issue was because I was trying to figure out what was causing a libuv upgrade to Rust to sometimes fail on certain tests on certain windows machines. One of the tests which was causing the most issues was a test for quickly rebinding to the same socket, so after investigating I found it weird that SO_REUSEADDR was enabled for unix but not windows. Since as you just stated, there is a valid reason for this, then it should probably be documented in the code to clarify to any future readers who wonder the same thing.

saghul:
@retep998 I just committed this: https://github.com/joyent/libuv/commit/9e921ed1e518ddcc27d6b1ddb0ad60b168bfebc3 Hope it helps next time someone wonder why we don't use SO_REUSEADDR. Thanks for reporting!

PR #1363, opened 2014-07-07
add uv_send/recv_buffer_size

txdv:
This will let the user inspect the buffer sizes of the socks and adjust their sizes.

This is useful if you have high volume traffic applications, since you can adjust the buffer to be much bigger and let the app therefore make less receives or writes(which are costly operations, since you cross from userspace to kernelspace). 

Goes hand in hand with try_write and try_send.

I'm going to add the windows code later today.


include/uv.h
```diff
@@ -601,6 +601,10 @@ UV_EXTERN void uv_walk(uv_loop_t* loop, uv_walk_cb walk_cb, void* arg);
  */
 UV_EXTERN void uv_close(uv_handle_t* handle, uv_close_cb close_cb);
 
+UV_EXTERN int uv_recv_buffer_size(uv_handle_t* handle, int* value);
```
saghul, line 614: Docs!

saghul:
Looks good! Needs a test and a Windows side, but you know that already ;-)

src/unix/core.c
```diff
@@ -159,6 +159,38 @@ void uv_close(uv_handle_t* handle, uv_close_cb close_cb) {
   uv__make_close_pending(handle);
 }
 
+static int uv__sockopt(uv_handle_t* handle, int* value, int optname) {
```
saghul: Please reorder the parameters:

```
static int uv__sockopt(uv_handle_t* handle, int optname, int* value) {
```

txdv:
It returns double the value on linux(or lets say better, it sets double the value): https://bugzilla.redhat.com/show_bug.cgi?id=170694

Kinda strange behaviour, however it does not on bsd and other unices.

```
 SO_RCVBUF
              Sets or gets the maximum socket receive buffer in bytes.  The
              kernel doubles this value (to allow space for bookkeeping
              overhead) when it is set using setsockopt(2), and this doubled
              value is returned by getsockopt(2).  The default value is set
              by the /proc/sys/net/core/rmem_default file, and the maximum
              allowed value is set by the /proc/sys/net/core/rmem_max file.
              The minimum (doubled) value for this option is 256.
```

What should we do? Leave it like that or divide by 2?

src/win/core.c
```diff
@@ -395,3 +395,28 @@ int uv_run(uv_loop_t *loop, uv_run_mode mode) {
...
+  int r;
+  int len;
+  SOCKET socket;
+
+  if (handle->type == UV_TCP)
+    socket = ((uv_tcp_t*) handle)->socket;
+  else if (handle->type == UV_UDP)
+    socket = ((uv_udp_t*) handle)->socket;
+  else
+    return -ENOTSUP;
```
saghul: UV_ENOTSUP, on Windows errors are not -errno.

src/win/core.c
```diff
@@ -395,3 +395,28 @@ int uv_run(uv_loop_t *loop, uv_run_mode mode) {
...
+
+  len = sizeof(*value);
+
+  if (*value == 0)
+    r = getsockopt(socket, SOL_SOCKET, SO_RCVBUF, (char*)value, &len);
+  else
+    r = setsockopt(socket, SOL_SOCKET, SO_RCVBUF, (const char*)value, len);
+
+  if (r < 0)
+    return -errno;
```
saghul: I think you need to use WSAGetLastError here. http://msdn.microsoft.com/en-us/library/aa916172.aspx

saghul:
> What should we do? Leave it like that or divide by 2?

Leave it as is.

include/uv.h
```diff
@@ -601,6 +601,26 @@ UV_EXTERN void uv_walk(uv_loop_t* loop, uv_walk_cb walk_cb, void* arg);
...
+ * to *value, otherwise it will use *value to set the new buffer
+ * size.
+ */
+UV_EXTERN int uv_recv_buffer_size(uv_handle_t* handle, int* value);
+
+/*
+ * Returns or sets the size of the send buffer that the operating
+ * system uses for the socket.
+ *
+ * If *value == 0, it will change the current send buffer size
```
saghul: I guess its "If *value != 0".

src/unix/core.c
```diff
@@ -159,6 +159,30 @@ void uv_close(uv_handle_t* handle, uv_close_cb close_cb) {
   uv__make_close_pending(handle);
 }
 
+int uv__sockopt(uv_handle_t* handle, int optname, int* value) {
```
saghul: Call this uv__socket_sockopt, since it only operated on SOL_SOCKET.

src/uv-common.c
```diff
@@ -478,6 +478,13 @@ size_t uv__count_bufs(const uv_buf_t bufs[], unsigned int nbufs) {
   return bytes;
 }
 
+int uv_recv_buffer_size(uv_handle_t* handle, int* value) {
```
saghul, line 481: Check if value is NULL.

txdv, line 481: I put the check in sockopt, it induces less code clutter.

src/win/core.c
```diff
@@ -395,3 +395,31 @@ int uv_run(uv_loop_t *loop, uv_run_mode mode) {
...
+
+  len = sizeof(*value);
+
+  if (*value == 0)
+    r = getsockopt(socket, SOL_SOCKET, SO_RCVBUF, (char*)value, &len);
+  else
+    r = setsockopt(socket, SOL_SOCKET, SO_RCVBUF, (const char*)value, len);
+
+  if (r == SOCKET_ERROR)
+    return WSAGetLastError();
```
saghul: You need to translate the error.

src/unix/core.c
```diff
@@ -159,6 +159,33 @@ void uv_close(uv_handle_t* handle, uv_close_cb close_cb) {
...
+    fd = uv__stream_fd((uv_stream_t*) handle);
+  else if (handle->type == UV_UDP)
+    fd = ((uv_udp_t *)handle)->io_watcher.fd;
+  else
+    return -ENOTSUP;
+
+  len = sizeof(*value);
+
+  if (*value == 0)
+    r = getsockopt(fd, SOL_SOCKET, SO_RCVBUF, (void*)value, &len);
```
saghul: should be optname, not SO_RCVBUF.

src/unix/core.c
```diff
@@ -159,6 +159,33 @@ void uv_close(uv_handle_t* handle, uv_close_cb close_cb) {
   uv__make_close_pending(handle);
 }
 
+int uv__socket_sockopt(uv_handle_t* handle, int optname, int* value) {
+  int r;
+  int fd;
+  socklen_t len;
+
+  if (handle == NULL)
+    return -EINVAL;
```
saghul, line 168: Also check if value is NULL.

src/unix/core.c
txdv: My brain is full of fuck

txdv:
Should I add a UDP and NAMED PIPE test?

I wonder if windows supports named pipe SO_RECVBUF.

saghul:
I'm not sure this actually works on Windows, since named pipes are not sockets. So we'd need to know if the buffers can be adjusted for named pips on windows some other way, or return ENOTSUP.

test/test-socket-buffer-size.c
```diff
@@ -0,0 +1,127 @@
...
+TEST_IMPL(socket_buffer_size) {
+  uv_connect_t connect_req;
+  struct sockaddr_in addr;
+
+  start_server();
+
+  ASSERT(0 == uv_ip4_addr("127.0.0.1", TEST_PORT, &addr));
+
+  ASSERT(0 == uv_tcp_init(uv_default_loop(), &client));
+  ASSERT(0 == uv_tcp_connect(&connect_req,
```
saghul: Why don't you just bind the socket and check the buffers, without connecting anywhere?

txdv: I feared late socket creation.

By the way, less lines to delete.

saghul: > I feared late socket creation.

If you bind the handle, the socket is created on the spot.

txdv:
The test is now very precise and beautiful. thanks for suggesting that.

We can test tty and pipes on unix as well I guess?

saghul:
> We can test tty and pipes on unix as well I guess?

Sure. Also, it would be nice for that to also work on windows, even if the API is different there.

src/unix/core.c
saghul, line 168: This still needs to be addressed.

```
if (handle == NULL || value == NULL)
```

test/test-socket-buffer-size.c
```diff
@@ -0,0 +1,79 @@
...
+  ASSERT(value == 10000 || value == 20000);
+}
+
+
+TEST_IMPL(socket_buffer_size) {
+  struct sockaddr_in addr;
+
+  ASSERT(0 == uv_ip4_addr("127.0.0.1", TEST_PORT, &addr));
+
+
```
saghul: 1 too many newlines

test/test-socket-buffer-size.c
```diff
@@ -0,0 +1,79 @@
...
+  ASSERT(0 == uv_tcp_init(uv_default_loop(), &tcp));
+  ASSERT(0 == uv_tcp_bind(&tcp, (struct sockaddr*) &addr, 0));
+  check_buffer_size((uv_handle_t*) &tcp);
+  uv_close((uv_handle_t*) &tcp, close_cb);
+
+
```
saghul: Ditto.

txdv:
> Sure. Also, it would be nice for that to also work on windows, even if the API is different there.

It looks like the buffer size of named piped on windows has to be specified on creation.
Getting the value is not a problem, but setting it has to be done creation.

What do you mean with "even if the API is different there"?.

saghul:
Then we should at lest allow you to get it, even if it's not settable. What
I mean "even if the API is different" is that out uv_ functions should also
work with named pipes on Windows, even if they use an API other than
getsockopt/setsockopt.

On Wed, Jul 9, 2014 at 5:14 PM, Andrius Bentkus notifications@github.com
wrote:

> Sure. Also, it would be nice for that to also work on windows, even if the
> API is different there.
> 
> It looks like the buffer size of named piped on windows has to be
> specified on creation.
> Getting the value is not a problem, but setting it has to be done creation.
> 
> What do you mean with "even if the API is different there"?.
> 
> —
> https://github.com/joyent/libuv/pull/1363#issuecomment-48486390.

## 

/Saúl
bettercallsaghul.com

txdv:
We can set the size before binding, on unix you could bind before as well, and make the setopt operation during initialization of the socket itself.

saghul:
Right now the socket is created and bound in a single go, you cannot set any socket option between socket creation and binding.

saghul:
Tests pass on Unixes: http://jenkins.nodejs.org/job/libuv-pullrequest/630/

saghul:
Hum, I wonder what we should do with uv_tty handles. They are streams after all...

txdv:
> Right now the socket is created and bound in a single go, you cannot set any socket option between socket creation and binding.

You can save it in a variable and use that value when you are creating the socket...

saghul:
Not sure what you mean. We currently create and bind the socket at the same
time, there is no chance for the user to change the buffers before the
socket is bound, because its not created yet.

txdv:
uv_pipe_init
set_buffer_size
uv_pipe_bind <- i think the creation happens here?

saghul:
Yes. But we don't want to store the value for later, we'd have to do it for
many other options.

It's a limitation we currently have. There are plans to lift it by creating
sockets early, but I still have some doubts about how the API would look
like when accepting connections.

I'd say it's OK to have this limitation in named pipes for now, it will go
away in the future.

include/uv.h
```diff
@@ -601,6 +601,24 @@ UV_EXTERN void uv_walk(uv_loop_t* loop, uv_walk_cb walk_cb, void* arg);
...
+UV_EXTERN int uv_recv_buffer_size(uv_handle_t* handle, int* value);
+
+/*
+ * Returns or sets the size of the send buffer that the operating
+ * system uses for the socket.
+ *
+ * If *value == 0, it will return the current send buffer size,
+ * otherwise it will use *value to set the new send buffer size.
+ */
+UV_EXTERN int uv_send_buffer_size(uv_handle_t* handle, int* value);
```
bnoordhuis, line 626: The comments should probably mention that, depending on the platform, the effective send/recv buffer size may be different from what was set (c.f. Linux).

bnoordhuis, line 626: Also, `value` being an in/out parameter is somewhat incongruent with the other socket option functions, like uv_tcp_keepalive().  Not necessarily worse, just different.  No concrete suggestions, just an observation.

txdv, line 626: We are already multiplexing buffer_size on recv and send, do we really want to multiplex on set get as well?

That would be 4 functions.

saghul, line 626: I didn't feel a strong opposition from Ben ;-) I'm on with the current
approach, 4 functions sounds a bit too much.

saghul:
This LGTM, can you take a quick look @indutny?

src/win/core.c
```diff
@@ -395,3 +395,31 @@ int uv_run(uv_loop_t *loop, uv_run_mode mode) {
...
+    socket = ((uv_tcp_t*) handle)->socket;
+  else if (handle->type == UV_UDP)
+    socket = ((uv_udp_t*) handle)->socket;
+  else
+    return UV_ENOTSUP;
+
+  len = sizeof(*value);
+
+  if (*value == 0)
+    r = getsockopt(socket, SOL_SOCKET, optname, (char*)value, &len);
```
indutny: `(char*) value` and everywhere else

src/win/core.c
```diff
@@ -395,3 +395,31 @@ int uv_run(uv_loop_t *loop, uv_run_mode mode) {
...
+  else
+    r = setsockopt(socket, SOL_SOCKET, optname, (const char*)value, len);
```
indutny: `(const char*) value`

src/unix/core.c
```diff
@@ -159,6 +159,33 @@ void uv_close(uv_handle_t* handle, uv_close_cb close_cb) {
...
+    fd = uv__stream_fd((uv_stream_t*) handle);
+  else if (handle->type == UV_UDP)
+    fd = ((uv_udp_t *)handle)->io_watcher.fd;
+  else
+    return -ENOTSUP;
+
+  len = sizeof(*value);
+
+  if (*value == 0)
+    r = getsockopt(fd, SOL_SOCKET, optname, (void*)value, &len);
```
indutny: ditto.

src/unix/core.c
```diff
@@ -159,6 +159,33 @@ void uv_close(uv_handle_t* handle, uv_close_cb close_cb) {
...
+  int fd;
+  socklen_t len;
+
+  if (handle == NULL || value == NULL)
+    return -EINVAL;
+
+  if (handle->type == UV_TCP || handle->type == UV_NAMED_PIPE)
+    fd = uv__stream_fd((uv_stream_t*) handle);
+  else if (handle->type == UV_UDP)
+    fd = ((uv_udp_t *)handle)->io_watcher.fd;
```
indutny: `(uv_udp_t*) handle`

src/unix/core.c
```diff
@@ -159,6 +159,33 @@ void uv_close(uv_handle_t* handle, uv_close_cb close_cb) {
   uv__make_close_pending(handle);
 }
 
+int uv__socket_sockopt(uv_handle_t* handle, int optname, int* value) {
```
indutny, line 162: Btw, could we use this function in other places? I think we have a couple of `setsockopt` calls.

src/unix/core.c
indutny: Also, I don't think that you really need to convert it to `void*` here.

indutny:
Only some style nits, please fix them and LGTM!

indutny:
Thank you, btw ;)

src/unix/core.c
txdv, line 162: Do you want me to change all occurrences setsockopt using this function?

indutny, line 162: After a second thought - no, thanks for asking!

saghul:
This looks ready now, I'll have a final look and land it tonight. Thanks Andrius!

txdv:
still lacks the windows part for pipes. I'll add it this evening.

saghul:
> still lacks the windows part for pipes. I'll add it this evening.

Oh, I had forgotten about that. Thanks!

include/uv.h
```diff
@@ -608,6 +629,9 @@ UV_EXTERN void uv_close(uv_handle_t* handle, uv_close_cb close_cb);
  * Due to platform differences the user cannot rely on the ordering of the
  * base and len members of the uv_buf_t struct. The user is responsible for
  * freeing base after the uv_buf_t is done. Return struct passed by value.
+ *
+ * NOTE: linux will set double the size and return double the size
```
saghul: This note seems misplaced.

txdv: This kinda shit happens when I am rushing those patches too fast, have to sit down and review it myself.

Didn't have any time yesterday.

saghul: > This kinda shit happens when I am rushing those patches too fast, have to sit down and review it myself.
> 
> Didn't have any time yesterday.

No worries, it just caught my eye :-) Take your time.

saghul:
Ping.

saghul:
Thanks for working on this Andrius! It landed in :sparkles: https://github.com/joyent/libuv/commit/0ecee213eac91beca141130cff2c7826242dab5a :sparkles:

mikinho:
Any reason this is a single function to get and set the buffer?  Using 0 to wrap getsockopt prevents use from setting SO_SNDBUF socket option to 0 (zero) which on Windows would allow the stack to send from your application buffer directly.

saghul:
@mikinho can you please open an issue at libuv/libuv?

txdv:
I hate myself for not thinking about this. But apparently nobody else thought about that at the time as well,

mikinho:
Issue opened https://github.com/libuv/libuv/issues/74

Will move any other discussion there.

PR #1364, opened 2014-07-10
landry@'s fix for uv_resident_set_memory

qbit:
On OpenBSD /dev/mem requires privileges. Use a sysctl to get the required info instead


saghul:
Thanks! I have an OpenBSD lying around here, I'll try it out and merge it if all goes well!

saghul:
Thanks, landed in https://github.com/joyent/libuv/commit/38323c9fbff49aee990528c145609c2a8241ad03 with some small changes.

PR #1366, opened 2014-07-12
doc: clarify the return value of uv_default_loop(void) function in uv.h.

ghmlee:
When I called the uv_default_loop(void) fucntion in my main function, I was confused about should I call the uv_loop_init(uv_loop_t\* loop) function or not.

So, I added a comment that is "Returns the `initialized` default loop".


indutny:
Btw, @saghul : this function may return `NULL` because of the allocation failure, should we assert internally? Or just document it?

indutny:
LGTM

saghul:
> Btw, @saghul : this function may return NULL because of the allocation failure, should we assert internally? Or just document it?

I'd go for documenting it. Can you also add a "NOTE:" about that as well @ghmlee?

ghmlee:
@saghul @indutny I wrote a new commit. Also, this is new pull request.
https://github.com/joyent/libuv/pull/1367

saghul:
Closing this one then. New time, @ghmlee you can just force push into the same branch and reuse the same PR :-)

PR #1367, opened 2014-07-13
doc: clarify the return value of uv_default_loop(void) function in uv.h.

ghmlee:
When I called the uv_default_loop(void) fucntion in my main function, I was confused about should I call the uv_loop_init(uv_loop_t\* loop) function or not.

So, I added a comment that is "Returns the initialized default loop".

From this pull request https://github.com/joyent/libuv/pull/1366, I added a note more.


saghul:
Thanks! Landed in https://github.com/joyent/libuv/commit/dc1ea27c736f0d21c7160c790bcd1b113d20abd9 :sparkles:

PR #1368, opened 2014-07-13
Implement uv_fs_mkdtemp

### LINKED ISSUE #1315: Implement cross-platform mkdtemp
We at https://github.com/neovim/neovim have faced with the need of porting glibc's mkdtemp to Windows. Would it make sense to make PR in your project for this? Perhaps introducing `uv_fs_mkdtemp` function.

Hinidu:
I implemented `uv_fs_mkdtemp` for unices and created a test for that. Could someone confirm that I am on the right way or provide some suggestions before I start implementing `uv_fs_mkdtemp` for Windows? Specifically I'm in doubt whether `uv_fs_t->ptr` is the right place for storing the name of temp directory which was created by underlying `mkdtemp`.

For Windows I want to use `glibc`'s [implementation](http://fossies.org/dox/glibc-2.19/mkdtemp_8c_source.html) which uses [`__gen_tempname`](http://fossies.org/dox/glibc-2.19/sysdeps_2posix_2tempname_8c_source.html#l00229).

Should close #1315


src/unix/fs.c
```diff
@@ -990,6 +991,15 @@ int uv_fs_mkdir(uv_loop_t* loop,
   POST;
 }
 
+int uv_fs_mkdtemp(uv_loop_t* loop,
+                  uv_fs_t* req,
+                  const char* template,
+                  uv_fs_cb cb) {
+  INIT(MKDTEMP);
+  req->ptr = strdup(template);
```
saghul: I'd use req->path.

src/unix/fs.c
```diff
@@ -779,6 +779,7 @@ static void uv__fs_work(struct uv__work* w) {
     X(LSTAT, uv__fs_lstat(req->path, &req->statbuf));
     X(LINK, link(req->path, req->new_path));
     X(MKDIR, mkdir(req->path, req->mode));
+    X(MKDTEMP, mkdtemp(req->ptr) ? 0 : -1);
```
saghul: The result is assigned to 'r', which is of type ssize_t, I think you'll need to wrap this into some `uv__fs_mkdtemp` which returns 0 on success and -1 on error.

Hinidu:
@saghul thank you for suggestions! Does it seem good now?

src/unix/fs.c
```diff
@@ -991,6 +997,16 @@ int uv_fs_mkdir(uv_loop_t* loop,
 }
 
 
+int uv_fs_mkdtemp(uv_loop_t* loop,
+                  uv_fs_t* req,
+                  const char* path,
```
saghul: Call this 'template'.

src/unix/fs.c
```diff
@@ -991,6 +997,16 @@ int uv_fs_mkdir(uv_loop_t* loop,
...
+                  uv_fs_cb cb) {
+  INIT(MKDTEMP);
+  PATH;
```
saghul: I'd do the strdup manually here, since you are storing the template, it's not a path yet :-) Also, remember to check for errors.

Hinidu: Ok, I thought about that but decided save some lines :-) I'll change it.

Hinidu:
@saghul fixed last two issues.

test/test-fs.c
```diff
@@ -787,6 +803,12 @@ TEST_IMPL(fs_async_dir) {
   uv_run(loop, UV_RUN_DEFAULT);
   ASSERT(mkdir_cb_count == 1);
 
+  r = uv_fs_mkdtemp(loop, &mkdtemp_req, "test_dir_XXXXXX", mkdtemp_cb);
```
saghul: Better add a new test, even if you need to copy some bolierplate. Also, if you create the file inside a temp directory you can just remove the entire directory, instead of saving the filename and removing that one then.

Hinidu: Ok, I'll add new test. I didn't create the file inside a temp directory yet, I'll add code for that in test too.
Btw how can I remove entire directory? Does `uv_fs_rmdir` remove directory recursively?

saghul: > Btw how can I remove entire directory? Does uv_fs_rmdir remove directory recursively?

I don't recall from the top of my head, but I think some test already does it...

Hinidu:
@saghul I've pushed the new version of test implementation. Do you (or anyone else) have other suggestions?

test/test-fs.c
```diff
@@ -376,6 +378,37 @@ static void mkdir_cb(uv_fs_t* req) {
...
+  uv_fs_req_cleanup(&open_req1);
+  unlink(filename);
+}
+
+
+static void mkdtemp_cb(uv_fs_t* req) {
+  ASSERT(req == &mkdtemp_req);
+  check_mkdtemp_result(req);
+  mkdtemp_cb_count++;
+  strcpy(buf, req->path);
```
saghul: Better remove the file here, instead of copying the path to buf.

test/test-fs.c
```diff
@@ -376,6 +378,37 @@ static void mkdir_cb(uv_fs_t* req) {
...
+
+  ASSERT(req->fs_type == UV_FS_MKDTEMP);
+  ASSERT(req->result == 0);
+  ASSERT(req->path);
+  ASSERT(strlen(req->path) == 15);
+  ASSERT(memcmp(req->path, "test_dir_", 9) == 0);
+  ASSERT(memcmp(req->path + 9, "XXXXXX", 6) != 0);
+  check_permission(req->path, 0700);
+
+  /* Create a file to check if req->path is actually a directory */
```
saghul: Better use uv_fs_stat here.

test/test-fs.c
```diff
@@ -376,6 +378,37 @@ static void mkdir_cb(uv_fs_t* req) {
...
+  int r;
+  char filename[32];
+
+  ASSERT(req->fs_type == UV_FS_MKDTEMP);
+  ASSERT(req->result == 0);
+  ASSERT(req->path);
+  ASSERT(strlen(req->path) == 15);
+  ASSERT(memcmp(req->path, "test_dir_", 9) == 0);
+  ASSERT(memcmp(req->path + 9, "XXXXXX", 6) != 0);
+  check_permission(req->path, 0700);
```
saghul, line 391: I it possible that default permissions change depending on the system?

test/test-fs.c
Hinidu: I did this to compare results of two `uv_fs_mkdtemp` calls. But I can also implement it by adding `mkdtemp_req2` and by calling `uv_fs_req_cleanup` in test body instead of `mkdtemp_cb`. Is this more appropriate way for you?

test/test-fs.c
Hinidu, line 391: These permissions are described in `man mkdtemp` on my system. I think this is the part of POSIX definition of `mkdtemp`.

test/test-fs.c
Hinidu: Could you please explain how to do it in a portable way? I think about `mode & S_IFDIR` but `S_IFDIR` seems to be not defined in Windows, though I'm not sure about it.

Hinidu: `S_IFDIR` is already used in `fs_symlink_dir` test, so perhaps I actually can use it too.

test/test-fs.c
saghul: Sounds good!

test/test-fs.c
saghul: Yep.

Hinidu:
@saghul thank you again for review! I've implemented your suggestions.

src/unix/fs.c
```diff
@@ -991,6 +997,18 @@ int uv_fs_mkdir(uv_loop_t* loop,
 }
 
 
+int uv_fs_mkdtemp(uv_loop_t* loop,
+                  uv_fs_t* req,
+                  const char* template,
+                  uv_fs_cb cb) {
+  INIT(MKDTEMP);
+  req->path = strdup(template);
```
txdv, line 1015: Are you freeing this anywhere?

saghul, line 1015: It's freed on request cleanup internally.

saghul, line 1015: https://github.com/joyent/libuv/blob/master/src/unix/fs.c#L1154

txdv, line 1015: https://github.com/Hinidu/libuv/blob/mkdtemp/src/unix/fs.c#L1173

everywhere we use free(var), but here we are using free((void *) var);

The user has to actually clean up after the operation is finished by calling uv_fs_req_cleanup.

saghul, line 1015: Of course the user has to call that function, that's how it works. Not sure
what your concern is.

Hinidu, line 1015: @txdv sorry, I don't understand what you want to say in your last comment, but I want to point you to https://github.com/Hinidu/libuv/blob/mkdtemp/src/unix/fs.c#L91
This `PATH` as you can see is used in many functions and it does the same thing as my code here. Actually I written it explicit by @saghul's suggestion because `PATH` works with `path` argument but in case of `uv_fs_mkdtemp` argument is `template`.

txdv, line 1015: @Hinidu nevermind, I was wrong

@saghul you said it would be done internally, but it is actually handled by the user. it is a small technical difference

saghul, line 1015: Well, what I meant is that you don't **explicitly** need to free it. It
happens as part of the cleanup process. It's an internal detail the user
doesn't really need to know about.

txdv, line 1015: Yes, I thought I saw a bug so I checked it out. Sometimes I am wrong.

saghul, line 1015: No worries! It's great to have an extra pair of eyes on the watch, thanks!
:-)

saghul:
@Hinidu ping, do you think you'll have time to complete the Windows part soon?

Hinidu:
@saghul actually I thought to ask tomorrow whether you have any other suggestions or I can proceed with Windows part :-) I hope tomorrow I will have time to implement `uv_fs_mkdtemp` for Windows. I think that it would be the most complex part of this PR.
Btw if you don't have objections I will squash last two commits for the unittest into one.

saghul:
Fantastic news! :-) Don't worry about squashing, it can be done at the end, once we have the code ready.

Hinidu:
@saghul I've pushed implementation for Windows. Make it work was simpler than I thought in the beginning :-) Now it passes my test locally.
Actually I ported `glibc`'s [implementation](http://fossies.org/dox/glibc-2.19/mkdtemp_8c_source.html) which uses [`__gen_tempname`](http://fossies.org/dox/glibc-2.19/sysdeps_2posix_2tempname_8c_source.html#l00229).
I have some questions which I'll post under particular lines shortly.

src/win/fs.c
```diff
@@ -721,6 +721,80 @@ void fs__mkdir(uv_fs_t* req) {
 }
 
 
+void fs__mkdtemp(uv_fs_t* req) {
+  static const WCHAR letters[] =
+    L"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
```
Hinidu, line 727: Does it make sense to use both upper and lower case? AFAIK most Windows file systems don't distinguish them in file names.

saghul, line 727: IIRC, NTFS can, so yes.

src/win/fs.c
```diff
@@ -721,6 +721,80 @@ void fs__mkdir(uv_fs_t* req) {
...
+  int len;
+  WCHAR* XXXXXX;
+  static uint64_t value;
+  unsigned int count;
+  int fd = -1;
+  int save_errno = errno;
```
Hinidu: How I should handle `errno`? Should I replace it with `_doserrno`? I didn't find `errno` usages in Windows code, but found one `_doserrno`.
**EDIT**: MSDN says that `_wmkdir` sets `errno` in failure cases.

src/win/fs.c
```diff
@@ -721,6 +721,80 @@ void fs__mkdir(uv_fs_t* req) {
 }
 
 
+void fs__mkdtemp(uv_fs_t* req) {
+  static const WCHAR letters[] =
+    L"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+  int len;
+  WCHAR* XXXXXX;
```
saghul: please use a better variable name here, like 'template'

src/win/fs.c
saghul: No need to save the errno here.

src/win/fs.c
```diff
@@ -721,6 +721,80 @@ void fs__mkdir(uv_fs_t* req) {
...
+      SET_REQ_RESULT(req, 0);
+      return;
+    } else if (errno != EEXIST) {
+      SET_REQ_RESULT(req, -1);
+      return;
+    }
+  }
+
+  /* We got out of the loop because we ran out of combinations to try.  */
+  errno = EEXIST;
```
saghul: no need to set errno

src/win/fs.c
Hinidu: I suggest `template_part` because it is not the entire template but just the `XXXXXX` part of it. Agree?

src/win/fs.c
```diff
@@ -721,6 +721,80 @@ void fs__mkdir(uv_fs_t* req) {
...
+  SET_REQ_UV_ERROR(req, UV_EEXIST, EEXIST);
```
saghul: SET_REQ_RESULT will do the job.

src/win/fs.c
```diff
@@ -721,6 +721,80 @@ void fs__mkdir(uv_fs_t* req) {
...
+    return;
+  }
+
+  /* This is where the Xs start.  */
+  XXXXXX = &req->pathw[len - 6];
+
+  /* Get some random data.  */
+  value += uv_hrtime() ^ _getpid();
+
+  for (count = 0; count < attempts; value += 7777, ++count) {
```
saghul, line 762: if you borrowed this implementation from glibc, please add a coment so that we know where to look for the magic numbers

src/win/fs.c
Hinidu: `SET_REQ_RESULT(req, -1)`? If I understand correctly it will use `_doserrno` which already should be equal to `EEXIST` because it is the only one way to get to that line. Am I right?

src/win/fs.c
saghul: Ok.

src/win/fs.c
saghul: SET_REQ_RESULT(req, fd), since fd will be -1 anyway. Also above.

Hinidu: Wouldn't it be more unclear? In the normal circumstances `fd` is not the result of the entire function.

saghul: Right, let's go with -1 then.

Hinidu:
@saghul I've pushed your suggestions.

saghul:
Nice! I'll take a deeper look tonight, thanks!

Hinidu:
Thank you, I'll wait final review!

src/win/fs.c
```diff
@@ -721,6 +721,78 @@ void fs__mkdir(uv_fs_t* req) {
...
+    return;
+  }
+
+  /* This is where the Xs start. */
+  template_part = &req->pathw[len - 6];
+
+  /* Get some random data. */
+  value += uv_hrtime() ^ _getpid();
+
+  /* Magic number 7777 was borrowed from the glibc's implementation */
```
saghul: Please move the comment to outside the function, just say it's borrowed from glibc.

src/win/fs.c
```diff
@@ -721,6 +721,78 @@ void fs__mkdir(uv_fs_t* req) {
...
+  /* The number of times to attempt to generate a temporary file. To
+     conform to POSIX, this must be no smaller than TMP_MAX. */
+#if ATTEMPTS_MIN < TMP_MAX
+  unsigned int attempts = TMP_MAX;
+#else
+  unsigned int attempts = ATTEMPTS_MIN;
+#endif
+
+  len = wcslen(req->pathw);
+  if (len < 6 || wcsncmp(&req->pathw[len - 6], L"XXXXXX", 6)) {
```
saghul, line 751: can you just use 6 there even if it's a wide-char string?

saghul, line 751: please check if wcsncmp != 0, it's more explicit

saghul, line 751: maybe we can have a static buffer with 6 "X"s, and its length, instead of having 6 everywhere?

saghul:
Other than clarifying to me the wide char thing, LGTM. Test pass: http://jenkins.nodejs.org/job/libuv-pullrequest/631/ Can you PTAL, @indutny?

src/win/fs.c
Hinidu, line 751: > can you just use 6 there even if it's a wide-char string?

Yes, it is count of characters to match, not bytes (docs says that and I checked it now).

> maybe we can have a static buffer with 6 "X"s, and its length, instead of having 6 everywhere?

Do you have idea how to name these variables? I have only two variants:
- `template_part_pattern` and `template_part_pattern_len` which is perhaps too long.
- `xs` and `xs_len` which also is not ideal but I like it more.

src/win/fs.c
Hinidu:
``` c
/* Some parts of the implementation was borrowed from glibc. */
```

Does this comment look good to you?

src/win/fs.c
saghul, line 751: Nah, leave them as is, I checked glibc and it also does it like that. Can
you squash all commits into one?

Hinidu:
@saghul fixed comment, squashed.

src/win/fs.c
```diff
@@ -721,6 +721,78 @@ void fs__mkdir(uv_fs_t* req) {
...
 
+/* Some parts of the implementation was borrowed from glibc. */
+void fs__mkdtemp(uv_fs_t* req) {
+  static const WCHAR letters[] =
+    L"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+  int len;
+  WCHAR* template_part;
+  static uint64_t value;
+  unsigned int count;
+  int fd = -1;
```
saghul: No need to initialize this, right?

Hinidu: :+1:

src/win/fs.c
```diff
@@ -721,6 +721,78 @@ void fs__mkdir(uv_fs_t* req) {
 }
 
 
+/* Some parts of the implementation was borrowed from glibc. */
```
Hinidu: Note to self: typo `was` -> `were`

src/win/fs.c
```diff
@@ -721,6 +721,78 @@ void fs__mkdir(uv_fs_t* req) {
...
+void fs__mkdtemp(uv_fs_t* req) {
+  static const WCHAR letters[] =
+    L"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+  int len;
```
Hinidu: I think better to change it to `size_t`. Do you agree @saghul?

saghul: yeah

Hinidu:
Should I do `git commit --amend && git push --force` for those small changes or should I add new commit for them to squash it after final review?

saghul:
Just force push, thanks!

Hinidu:
@saghul Done!

saghul:
@indutny Ping.

src/unix/fs.c
```diff
@@ -1001,6 +1007,18 @@ int uv_fs_mkdir(uv_loop_t* loop,
...
+  if (req->path == NULL)
+    return -ENOMEM;
```
indutny, line 1017: What about `ENOTDIR` ?

indutny, line 1017: nvm, it is returned for other functions of this family.

src/win/fs.c
```diff
@@ -721,6 +721,78 @@ void fs__mkdir(uv_fs_t* req) {
 }
 
 
+/* Some parts of the implementation were borrowed from glibc. */
+void fs__mkdtemp(uv_fs_t* req) {
+  static const WCHAR letters[] =
+    L"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
```
indutny, line 727: hahahaha!

src/unix/fs.c
saghul, line 1017: This is just checking the result after `strdup`.

src/win/fs.c
```diff
@@ -721,6 +721,78 @@ void fs__mkdir(uv_fs_t* req) {
...
+    return;
+  }
+
+  /* This is where the Xs start. */
+  template_part = &req->pathw[len - 6];
+
+  /* Get some random data. */
+  value += uv_hrtime() ^ _getpid();
+
+  for (count = 0; count < attempts; value += 7777, ++count) {
```
indutny, line 762: Not true randomness! :) You could do better!

indutny:
LGTM, if it works

src/win/fs.c
saghul, line 762: This is 95$ borrowed from glibc, FWIW. Given the number of attempts we make, I'd say it's Good Enough (TM) :-)

saghul:
Tests pass, will land tonight then. Thanks for the review Fedor!

src/win/fs.c
Hinidu, line 727: It was also borrowed from glibc :-)

Hinidu:
> LGTM, if it works

Thank you for review!

> Tests pass, will land tonight then.

Great! This code will go to 0.11.27? Is there an approximate release date of this version?

saghul:
> Great! This code will go to 0.11.27? Is there an approximate release date of this version?

Not sure if we'll release any libuv 0.11.x before 0.12. About the release date: it will be ready when it's ready :-)

Hinidu:
> Not sure if we'll release any libuv 0.11.x before 0.12. About the release date: it will be ready when it's ready :-)

Ok, then I'll try to keep an eye on this so I hope I don't miss next release :-)

saghul:
Landed in master, thanks for working on this @Hinidu!

Hinidu:
> Landed in master, thanks for working on this @Hinidu!

I'm very glad to hear it! Thank you and all other collaborators for your great library!

src/win/fs.c
```diff
@@ -721,6 +721,78 @@ void fs__mkdir(uv_fs_t* req) {
 }
 
 
+/* Some parts of the implementation were borrowed from glibc. */
```
sam-github, line 724: Is incorporating LGPL source into an MIT licensed project legal?

bnoordhuis, line 724: It's clearly recognizable as glibc's __gen_tempname() from sysdeps/posix/tempname.c, parts have been copied verbatim.  This seems like murky legal water.

saghul, line 724: Ouch. My bad since I did the reviewing :-( I'll look for an alternative with appropriate licensing. If anyone gets there beforehand, please open an issue / PR.

Hinidu, line 724: I apologize. It was my mistake. I hadn't much experience with copyright law.
I found slightly different OpenBSD version here http://openbsd.cs.toronto.edu/cgi-bin/cvsweb/src/lib/libc/stdio/mktemp.c It seems like their license permits copying and modification.
I can try change current implementation according to this version.
Sorry again, I hope I didn't create a lot of problems.

saghul:
I think we can go with the MIT licensed musl libc's implementation: https://github.com/runtimejs/musl-libc/blob/master/src/temp/mkdtemp.c and https://github.com/runtimejs/musl-libc/blob/master/src/temp/__randname.c.

Same license, so there should be no problem.

Hinidu:
@saghul I've pushed #1402 right now. But if musl's version is more desirable than I can use it.

saghul:
Oh, it's ok, will review now.

PR #1369, opened 2014-07-13
unix: fix inconsistent code style

dndx:
Fixed a line in `unix/tcp.c` that does not follow the style with rest of the project.


indutny:
Sorry, we do not accept the style fixes! But I'll try to remember to fix it during next rewrite of this portion of code. Thank you!

PR #1370, opened 2014-07-15
Update stream.c

huxingyi:
I think call uv_read_stop is also the right way, even the better way, to make the assert happy.
When stream->read_cb(status=-1), I might wont call uv_close, because there are some uv_write are not returned or some other operations are pending.

ps:No description was found about force call uv_close after stream->read_cb(status=-1) in uv.h


indutny:
Technically speaking, there could be no pending write calls anymore. If the kernel reports socket error it is definitely already down. You couldn't use the socket at this point.

saghul:
@indutny yes, but you don't necessarily need to uv_close it. Windows actually stops it for you and there is no assert, which I think it's a better thing to do.

huxingyi:
I will not touch the socket anymore, but I want keep the uv_tcp_t struct not been freed. 
Because I might have some pending operations(not libuv calls) associated with the uv_tcp_t handle.
If force call uv_close on read_cb(status=-1), then I need set a bit flag, for example, uv_handle_closed=1 to indicate that uv_close was called.
I usually call uv_close as the last step, before free the uv_tcp_t struct. 
Maybe force call uv_close on read_cb(status=-1) is more safely than just call uv_read_stop()?
I agree with @saghul, no assert and call uv_read_stop internally could be better.

saghul:
We have plans to change this behavior completely after 0.12, so for now it will stay as is. Closing.

PR #1372, opened 2014-07-19
stream write callbacks have different error codes on windows and unix when closing the socket

txdv:
If we close a socket and there are still some writes left then windows will call the callbacks of these writes with ECONNABORTED and unix with ECANCELED.

In https://github.com/joyent/libuv/pull/1359 we had to have a work around in the test case

CANCELED and ABORTED is basically the same, the operation has been cancelled, but with ABORTED it gives more information to us: the socket has been closed or some failure happened on the socket, it is not just a simple cancel.

Here we are closing the socket but returning just an ECANCELED in unix: https://github.com/txdv/libuv/blob/e3cb3d183ab3071bfc4fc45d20ea2148f42c4d85/src/unix/stream.c#L382

We are about to close the socket and flushing the queue in order to invoke the callbacks. It makes sense that we use ECONNABORTED instead of ECANCELED.

Even from a binding for C# perspective it makes sense to me(or any other language which has class orientated exception handling): I would make ABORTED a subclass of CANCELLED or create a class encompassing both classes. 

**Another case we have to check is what error code windows returns if the peer closes the socket**.

Another reason why we should return ABORTED on unix is because we have a CANCELED error code on windows:

IOCP supports cancelling operations: http://msdn.microsoft.com/en-us/library/windows/desktop/aa363789(v=vs.85).aspx

> You can cancel asynchronous I/O from any thread in the process that issued the I/O operation. You must specify the handle which the I/O was performed on and, optionally, the overlapped structure that was used to perform the I/O. You can determine if the cancel occurred by examining the status returned in the overlapped structure or in the completion callback. A status of ERROR_OPERATION_ABORTED indicates that the operation was canceled.

ERROR_OPERATION_ABORTED = ECANCELLED


saghul:
I need to think about this a bit more. AFAICT on Unix you can never get ECONNABORTED for a write, just on connect or accept. There an inportant difference between how we do things on Unix vs Windows: on Unix we don't actually write, we keep the queue ourselves, whereas on Windows we always write and let Windows itself handle it.

Now, when we add the ability to `uv_cancel()` a write request those will definitely get ECANCELED, but do we care about having a different error code as opposed to closing a handle which had pending writes? I'm uncertain.

Have you tested what errors we get when the remote closes the connection? On Unix probably ENOTCONN if you don't close the handle when you receive EOF. What does Windows give us?

txdv:
On UNIX we manage the write queue on our own, so if the remote closes the socket and we get that event, we have to just return the appropriate error code when flushing the write queue.

I haven't checked yet what windows returns. Whatever it returns we should just model the error message according to it.

saghul:
I looked at this, and Linux only seems to use ECONNABORTED on connect or accept. FreeBSD only uses it on accept, apparently. So, I'd say we do it the other way around: turn ERROR_OPERATION_ABORTED into UV_ECANCELED in the context of writes.

Also, FWIW, we'll eventually need to queue those writes in order to allow calling uv_write before uv_connect.

@indutny, thoughts?

indutny:
+1 for ECANCELED in write callbacks on `uv_close()`. It is good from an API view, and I was expecting it to be such too! :)

saghul:
Landed the alternate fix on https://github.com/joyent/libuv/commit/431b535a519e91260e9a643cbf4dad59a98961bb.

txdv:
I don't like that fix, it drops valuable error information.

saghul:
Well, Fedor and I agree on the other approach.

txdv:
I wrote an explanation why we should model it after the windows error code.

Does FreeBSD not report ECONNABORTED?

saghul:
Unices never report it for writes, AFAICT. That's the basis of my counter-argument. The write was cancelled, either by the user (though that's not currently possible) of as an implicit action of closing the handle, I don't think we loose any information.

txdv:
If we add uv_cancel for writes then we can't know whether the write was canceled or whether the remote disconnected and therfore the queue got flushed.

saghul:
That's an implementation detail. On Unix we do the queuing, on Windows the kernel does, but we need to build something consistent on top of the 2. In either case the write didn't happen, it's not relevant on what queue it happened to be in, so the effect is the same.

Can we move on?

txdv:
nope

saghul:
Please, stop. You proposed a change, I had a different idea, Fedor agreed with me, so we went with that. I might be wrong, (I'm confident, but you can never tell what the future might bring) I'm human after all, but the world won't collapse if we find that I was wrong in the future and change it back, it's a bugfix.

Right now there are bigger fish to fry, 0.12 is around the corner and this can wait.

PR #1374, opened 2014-07-19
Reentrant library

txdv:
A reentrant library is a library which does not abort on normal user input.

We are using a lot of asserts. Asserts should be used only in places where we are certain that they will never be called in order to be able to debug better.

We should return for example EINVAL instead of doing an assert(0) on bad user input.

Furthermore there are a lot of aborts. Some of them will never happen, but others are done when it should actually return a ENOMEM error.


saghul:
I've never heard of this concept as reentrancy, I thought you meant http://en.wikipedia.org/wiki/Reentrancy_%28computing%29 :-)

I tentatively marked this for 0.12, because I agree we assert a bit too much. Now, I'll go case by case having a look at your patches, since some asserts might be legit, specially if they are in internal functions which already get sanitized input and we are just being defensive.

Thanks!

txdv:
>  Now, I'll go case by case having a look at your patches, since some asserts might be legit, specially if they are in internal functions which already get sanitized input and we are just being defensive.

I said:

> We are using a lot of asserts. Asserts should be used only in places where we are certain that they will never be called in order to be able to debug better.

I understand that.

saghul:
I remember @bnoordhuis opened an issue related to this years ago, but I can't seem to find it (or maybe I saw it in my dreams). His idea was to have a macro which would assert on debug mode, and return an error in release mode. Something to think about.

include/uv.h
```diff
@@ -2218,22 +2218,22 @@ UV_EXTERN const char* uv_dlerror(const uv_lib_t* lib);
  * return type is void, of course).
  */
 UV_EXTERN int uv_mutex_init(uv_mutex_t* handle);
-UV_EXTERN void uv_mutex_destroy(uv_mutex_t* handle);
-UV_EXTERN void uv_mutex_lock(uv_mutex_t* handle);
+UV_EXTERN int uv_mutex_destroy(uv_mutex_t* handle);
+UV_EXTERN int uv_mutex_lock(uv_mutex_t* handle);
 UV_EXTERN int uv_mutex_trylock(uv_mutex_t* handle);
-UV_EXTERN void uv_mutex_unlock(uv_mutex_t* handle);
+UV_EXTERN int uv_mutex_unlock(uv_mutex_t* handle);
```
saghul: Please drop these changes to the synchronization primitives, I'd rather not change the API this late, we can revisit after.

include/uv.h
```diff
@@ -2218,22 +2218,22 @@ UV_EXTERN const char* uv_dlerror(const uv_lib_t* lib);
...
+UV_EXTERN int uv_rwlock_destroy(uv_rwlock_t* rwlock);
+UV_EXTERN int uv_rwlock_rdlock(uv_rwlock_t* rwlock);
 UV_EXTERN int uv_rwlock_tryrdlock(uv_rwlock_t* rwlock);
-UV_EXTERN void uv_rwlock_rdunlock(uv_rwlock_t* rwlock);
-UV_EXTERN void uv_rwlock_wrlock(uv_rwlock_t* rwlock);
+UV_EXTERN int uv_rwlock_rdunlock(uv_rwlock_t* rwlock);
+UV_EXTERN int uv_rwlock_wrlock(uv_rwlock_t* rwlock);
 UV_EXTERN int uv_rwlock_trywrlock(uv_rwlock_t* rwlock);
-UV_EXTERN void uv_rwlock_wrunlock(uv_rwlock_t* rwlock);
+UV_EXTERN int uv_rwlock_wrunlock(uv_rwlock_t* rwlock);
```
saghul: Ditto.

src/unix/stream.c
```diff
@@ -584,7 +584,7 @@ int uv_listen(uv_stream_t* stream, int backlog, uv_connection_cb cb) {
     break;
 
   default:
-    assert(0);
+    return err;
```
saghul: Just initialize error here and let the return below do its job.

src/unix/udp.c
```diff
@@ -365,8 +365,7 @@ static int uv__udp_maybe_deferred_bind(uv_udp_t* handle,
     break;
   }
   default:
-    assert(0 && "unsupported address family");
-    abort();
+    return -EAFNOSUPPORT;
```
saghul: There is an assert on line 343, remove that as well.

src/unix/udp.c
```diff
@@ -695,8 +694,7 @@ int uv_udp_set_multicast_interface(uv_udp_t* handle, const char* interface_addr)
       return -errno;
     }
   } else {
-    assert(0 && "unexpected address family");
-    abort();
+    return -EAFNOSUPPORT;
```
saghul: Undo this one. We set it above ourselves, it's just defensive programming :-P

saghul:
Most of the changes seem reasonable, sans the changes to mutex-es and such. I'll land the other changes, once you make the adjsutments.

saghul:
Ping.

src/fs-poll.c
```diff
@@ -78,20 +79,27 @@ int uv_fs_poll_start(uv_fs_poll_t* handle,
...
+    goto error;
 
   handle->poll_ctx = ctx;
   uv__handle_start(handle);
 
   return 0;
-}
+
+error:
+  if (ctx != NULL)
```
saghul: It doesn't seem possible to reach here with a NULL context

src/fs-poll.c
```diff
@@ -78,20 +79,27 @@ int uv_fs_poll_start(uv_fs_poll_t* handle,
...
+    free(ctx);
+  return err;
+};
```
saghul: Remove the trailing ';'

src/unix/stream.c
```diff
@@ -582,9 +582,6 @@ int uv_listen(uv_stream_t* stream, int backlog, uv_connection_cb cb) {
   case UV_NAMED_PIPE:
     err = uv_pipe_listen((uv_pipe_t*)stream, backlog, cb);
     break;
-
-  default:
```
saghul, line 586: Sigh, we get a compilation warning because not all enumeration values are handled in the swtich. Lets not initialize `err` at the top and do:

```
default:
  err = -EINVAL;
  break;
```

saghul:
Left a couple of comments. Please add a coment when you address them, GitHub doesn't send notifications if you just push new commits or force push :-S

txdv:
pong

saghul:
Thanks Andrius! LGTM, will merge later today.

saghul:
Thanks Andrius! Landed in 875814adc57cb9b607c8fd489837826d223707c2 :sparkles: 

PS: I undid the change in udp.c since it was wrong and all code paths leading to that funcion already checked the family type, so the assert/abort which comes after will catch internal problems anyway.

PR #1379, opened 2014-07-24
windows: keep uv__tcp_bind behavior identical with unix's version

HungMingWu:
The unix's version set SO_REUSEADDR, the windows' version follows it and erase process bind error.


saghul:
SO_REUSEADDR does not do what the Unix version does, it's essentially harmful on Windows. Here is an explanation: https://github.com/joyent/libuv/pull/1360

HungMingWu:
Sorry, I have one question

> So, we don't really need to set SO_REUSEADDR on Windows because sockets in TIME_WAIT are not checked when binding.

My code runs ok on windows 7 without setting SO_REUSEADDR, but Windows 8 fail..
I found the [document](http://msdn.microsoft.com/en-us/library/windows/desktop/ms740621%28v=vs.85%29.aspx)  you mention in [#1360](https://github.com/joyent/libuv/pull/1360)

I don't know whether the behavior changes at windows 8.

saghul:
If you can find info about it, we can ifdef it if necessary.

HungMingWu:
I cannot google enough information to support my assumption.
I can offer [test code](https://gist.github.com/HungMingWu/12787a934a407a52c061)
In windows 7 all connection pass.
But in windows 8, sometime it can pass, but sometime it would happen bind error.

HungMingWu:
the loop time is non-sense, even much less loop times can reproduce the issue.,

saghul:
There are a number of issues with that test:
- It reuses `connection`, try it on **different** connection structures
- It accesses `connection.bind_error`, which is private
- Some errors (EADDRINUSE) are not reported until `uv_listen` is called, so you should call that as well

HungMingWu:
- For pointer 1, my real project use different connections. Use one connection is just for demo
- For pointer 2, I know it is private member. But uv_tcp_bind would return 0 even it bind unsuccessfully.
  The demo only consider windows, so print it directly
- For pointer 3, my project is bind the specific address, and connect to somewhere. I found if the bind error, the connect process would invalid

saghul:
> For pointer 1, my real project use different connections. Use one connection is just for demo

Just create a statically allocated array for the test.

> For pointer 2, I know it is private member. But uv_tcp_bind would return 0 even it bind unsuccessfully. The demo only consider windows, so print it directly

The error will be returned on uv_listen or uv_connect, so call any of them after and you'll get it.

> For pointer 3, my project is bind the specific address, and connect to somewhere. I found if the bind error, the connect process would invalid

Don't guess, test :-)

HungMingWu:
I rewrite [my test code](https://gist.github.com/HungMingWu/12787a934a407a52c061)
And run it on windows 7 and windows 8.

Windows 7's result is similar as the following.
1: connect result = 0
1: connect status = 0
2: connect result = 0
2: connect status = -4091
and etc... The status may be 0 or -4091, it means the address have been used.
But the connect result is always be 0.

But Windows8 can run the unexpected result
45: connect result = -4091

Assertion failed: r == 0, file c:\test\bind.c line 24

Watch win/tcp.c uv_tcp_try_connect

if (handle->flag & UV_HANDLE_BIND_ERROR) {
       return handle->bind_error;
}

HungMingWu:
Sorry for my mistake.
I run the test in tvirtual machine before. And I assume the code works in windows 7.
And I borrwow a real one machine. The code failed as Windows 8.
Any suggestion?

saghul:
So, it apparently fails on Windows 7 and 8. That's expected since I wouldn't expect them to change such a thing between releases.

Now, as for why it fails, if does so because on Windows there is no proper way to set something similar to SO_REUSEADDR's behavior on Unix, that is, allow the address and port to be reused, but only if the other socket sharing the address is in TIME_WAIT state.

As I said earlier, setting SO_REUSEADDR on Windows is not an option because it's harmful, so I'm afraid you are confronted with a platform difference we cannot fix / abstract on libuv.

HungMingWu:
Whether it can  open a way to set the SO_REUSEADDR by the programmer decision.
Like
uv_tcp_bind(...., UV_TCP_USESAMEADDR);
The risk is responsibleby programmer.
Thanks for your help.

saghul:
I disagree. I see no reason whatsoever to allow literally stealing a port to an application. If you want do it yourself anyway, you are free to bind the socket manually and then use uv_tcp_open to use it.

HungMingWu:
Thanks for your replay, I know what I should do.

PR #1380, opened 2014-07-25
windows: return bind error immediately when implicitly binding

saghul:
If a TCP handle is bound automatically when calling uv_listen or
uv_connect, there is a remote possibility that it fails, return
the error if so.

### DIFF
--- a/src/win/tcp.c
+++ b/src/win/tcp.c
@@ -539,6 +539,8 @@ int uv_tcp_listen(uv_tcp_t* handle, int backlog, uv_connection_cb cb) {
                           0);
     if (err)
       return err;
+    if (handle->flags & UV_HANDLE_BIND_ERROR)
+      return handle->bind_error;
   }
 
   if (!handle->func_acceptex) {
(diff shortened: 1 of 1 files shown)


saghul:
Found it while checking the code for something else. /cc @indutny

src/win/tcp.c
```diff
@@ -539,6 +539,8 @@ int uv_tcp_listen(uv_tcp_t* handle, int backlog, uv_connection_cb cb) {
                           0);
     if (err)
       return err;
+    if (handle->flags & UV_HANDLE_BIND_ERROR)
```
indutny, line 542: Wait, isn't this a success branch?

saghul, line 542: Yes. `uv_tcp_try_bind` will return 0 in case of EADDRINUSE and set the flag and handle->bind_error to the actual error.

indutny, line 542: hm... and what if `bind_error` is `0`?

saghul, line 542: Can't be if `handle->flags & UV_HANDLE_BIND_ERROR` is set.

indutny:
Aaaah, I see. Sorry! LGTM

saghul:
No worries, thanks Fedor!

PR #1383, opened 2014-07-27
Proposal: Re-add travis configuration and badge

tarruda:
CI configuration was removed in @cf1dc613177 due to nondeterminism caused by travis. This PR introduces the `DISABLE_NONDETERMINISTIC_TESTS` ifndef guard to disable tests that may be affected by nondeterminism in the test environment.

Running the tests in a local ubuntu 12.04 virtualbox, I managed to get some random tests failures in the `tcp_writealot` test. The failures became frequent after a while and persisted even across reboots! It seems the operating system is saving some kind of state related to the network stack that is causing this problem.

I'm not sure what tests were randomly failing before, but for now I only wrapped `tcp_writealot` in the ifndef guard.

If this PR is accepted, I will send another that uses mingw/wine to cross-compile/test for windows. This pattern can be extended to other supported platforms through the use of prebuilt virtual machines and qemu

### DIFF
--- a/.ci/build-test.sh
+++ b/.ci/build-test.sh
@@ -1,0 +1,5 @@
+export CFLAGS="$CFLAGS -DDISABLE_NONDETERMINISTIC_TESTS"
+sh autogen.sh
+./configure
+make
+make check
--- a/.ci/build-test.sh
+++ b/.ci/build-test.sh
@@ -1,5 +1,5 @@
 export CFLAGS="$CFLAGS -DDISABLE_NONDETERMINISTIC_TESTS"
 sh autogen.sh
-./configure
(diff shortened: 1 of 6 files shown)


saghul:
Hi Thiago. We do already have a CI system inplace: http://jenkins.nodejs.org/ we don't have a "badge" on the README though, maybe we should. We also test cross-compiling on MinGW-w64: http://jenkins.nodejs.org/job/libuv-master-mingw-w64/

At the moment, I see no reason to re-add Travis-CI support, sorry. Yes, we know there are some failing test cases, we'll need to look into those.

@indutny, thoughts?

indutny:
Yeah, it is enough to have just one CI.

PR #1384, opened 2014-07-29
windows: relay TCP bind errors via ipc

orangemocha:
This is the libuv side of the fix for Node's cluster module on Windows.
https://github.com/joyent/node/issues/7691

Windows and Unix return certain socket errors (i.e. EADDRINUSE) at
different times: bind on Windows, and listen on Unix.
In an effort to hide this difference, libuv on Windows stores such
errors in the bind_error field of uv_tcp_t, to defer raising it at
listen time.
This worked fine except for the case in which a socket is shared in
a Node cluster and a bind error occurs.

A previous attempt to fix this (
https://github.com/joyent/libuv/commit/d1e6be1460f555a1f8a4063d7642696aa7238769
https://github.com/joyent/node/commit/3da36fe00e5d85414031ae812e473f16629d8645
) was flawed becaused in an attempt to relay the error at the JS level
it caused the master to start accepting connections.

With this new approach, libuv itself is relaying the bind errors,
providing for a uniform behavior of uv_tcp_listen.

### DIFF
--- a/src/win/internal.h
+++ b/src/win/internal.h
@@ -75,7 +75,6 @@ extern UV_THREAD_LOCAL int uv__crt_assert_enabled;
 /* Used by streams and UDP handles. */
 #define UV_HANDLE_READING                       0x00000100
 #define UV_HANDLE_BOUND                         0x00000200
-#define UV_HANDLE_BIND_ERROR                    0x00000400
 #define UV_HANDLE_LISTENING                     0x00000800
 #define UV_HANDLE_CONNECTION                    0x00001000
 #define UV_HANDLE_CONNECTED                     0x00002000
@@ -125,6 +124,12 @@ extern UV_THREAD_LOCAL int uv__crt_assert_enabled;
 /*
  * TCP
  */
+
+typedef struct {
+  WSAPROTOCOL_INFOW socket_info;
+  int delayed_error;
+} uv__ipc_socket_info_ex;
+
 int uv_tcp_listen(uv_tcp_t* handle, int backlog, uv_connection_cb cb);
 int uv_tcp_accept(uv_tcp_t* server, uv_tcp_t* client);
 int uv_tcp_read_start(uv_tcp_t* handle, uv_alloc_cb alloc_cb,
@@ -143,7 +148,7 @@ void uv_process_tcp_connect_req(uv_loop_t* loop, uv_tcp_t* handle,
 void uv_tcp_close(uv_loop_t* loop, uv_tcp_t* tcp);
 void uv_tcp_endgame(uv_loop_t* loop, uv_tcp_t* handle);
 
-int uv_tcp_import(uv_tcp_t* tcp, WSAPROTOCOL_INFOW* socket_protocol_info,
+int uv_tcp_import(uv_tcp_t* tcp, uv__ipc_socket_info_ex* socket_info_ex,
     int tcp_connection);
 
 int uv_tcp_duplicate_socket(uv_tcp_t* handle, int pid,
--- a/src/win/tcp.c
+++ b/src/win/tcp.c
@@ -156,6 +156,7 @@ int uv_tcp_init(uv_loop_t* loop, uv_tcp_t* handle) {
   handle->func_acceptex = NULL;
   handle->func_connectex = NULL;
   handle->processed_accepts = 0;
+  handle->delayed_error = 0;
 
   return 0;
 }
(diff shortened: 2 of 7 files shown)


saghul:
Can we have a libuv test case for this?

orangemocha:
cc @piscisaureus

src/win/internal.h
```diff
@@ -125,6 +125,12 @@ extern UV_THREAD_LOCAL int uv__crt_assert_enabled;
 /*
  * TCP
  */
+
+typedef struct {
+  WSAPROTOCOL_INFOW socket_info;
+  int bind_error;
+} uv_ipc_socket_info_ex;
```
saghul: Maybe call it `uv__ipc_socket_info_ex` as it's an internal thing and we lean towards that naming.

src/win/tcp.c
```diff
@@ -1156,6 +1155,11 @@ int uv_tcp_import(uv_tcp_t* tcp, WSAPROTOCOL_INFOW* socket_protocol_info,
   tcp->flags |= UV_HANDLE_BOUND;
   tcp->flags |= UV_HANDLE_SHARED_TCP_SOCKET;
 
+  if (socket_info_ex->bind_error) {
+    tcp->flags |= UV_HANDLE_BIND_ERROR;
+    tcp->bind_error = socket_info_ex->bind_error;
+  }
```
saghul: I wonder if we shouldn't rename this to `delayed_error`, like Unix does. And get rid of the flag since we are at it.

orangemocha:
Added test case (win32-only), and cleaned up per your suggestions. Thanks!

test/test-ipc.c
```diff
@@ -205,6 +204,70 @@ static void on_read(uv_stream_t* handle,
   free(buf->base);
 }
 
+#ifdef _WIN32
+static void on_read_listen_after_bound_twice(uv_stream_t* handle,
```
saghul, line 208: style: align arguments with the first one.

test/test-ipc.c
```diff
@@ -205,6 +204,70 @@ static void on_read(uv_stream_t* handle,
...
+      return;
+    }
+
+    printf("error recving on channel: %s\n", uv_strerror(nread));
+    abort();
+  }
+
+  fprintf(stderr, "got %d bytes\n", (int)nread);
+
+  pending = uv_pipe_pending_type(pipe);
```
saghul, line 236: Maybe assert that the pending count is 1 before?

test/test-ipc.c
```diff
@@ -657,3 +727,51 @@ int ipc_helper_tcp_connection(void) {
   MAKE_VALGRIND_HAPPY();
   return 0;
 }
+
+int ipc_helper_bind_twice(void) {
+  /*
+   * This is launched from test-ipc.c. stdin is a duplex channel that we
+   * over which two handles will be transmitted.
```
saghul, line 735: that we ?? over

test/test-ipc.c
```diff
@@ -657,3 +727,51 @@ int ipc_helper_tcp_connection(void) {
...
+  r = uv_tcp_init(uv_default_loop(), &tcp_server2);
+  ASSERT(r == 0);
+
+  r = uv_tcp_bind(&tcp_server, (const struct sockaddr*) &addr, 0);
+  ASSERT(r == 0);
+  r = uv_tcp_bind(&tcp_server2, (const struct sockaddr*) &addr, 0);
+  ASSERT(r == 0);
+
+  r = uv_write2(&write_req, (uv_stream_t*)&channel, &buf, 1,
+      (uv_stream_t*)&tcp_server, NULL);
```
saghul: style: align arguments with first one

test/test-ipc.c
```diff
@@ -657,3 +727,51 @@ int ipc_helper_tcp_connection(void) {
...
+  ASSERT(r == 0);
+  r = uv_write2(&write_req2, (uv_stream_t*)&channel, &buf, 1,
```
saghul, line 769: ditto

saghul:
Some style nits, otherwise LGTM. I'll run tests later.

test/test-ipc.c
orangemocha, line 236: Could it be higher than 1 given that two handles are being sent? That's why I didn't include it.
Should I assert > 0?

saghul, line 236: > Could it be higher than 1 given that two handles are being sent? That's why I didn't include it.
> Should I assert > 0?

It could be bigger, yes. Yeah, asserting if > 0 should be enough.

orangemocha:
Amended to address your latest feedback. Thanks!

saghul:
Thanks Alexis! Extra karma for refactoring the `bind_error` attribute :-) :+1:

Issue #1385, opened 2014-07-29
Making junctions via uv_fs_symlink is inconsistent on Windows


tkelman:
Built libuv from latest master, in MSYS2, with this MinGW-w64 version: http://sourceforge.net/projects/mingwbuilds/files/host-windows/releases/4.8.1/64-bit/threads-win32/seh/x64-4.8.1-release-win32-seh-rev5.7z

Test program, called `uv_symlink.c`:

``` c
#include <stdio.h>
#include "uv.h"

int main(int argc, char* argv[])
{
    uv_loop_t *jl_io_loop;
    uv_fs_t req;
    int ret, flags = UV_FS_SYMLINK_JUNCTION;
    if (argc != 3) {
        printf("Must provide 2 inputs\n");
        return 1;
    }
    jl_io_loop = uv_default_loop();
    ret = uv_fs_symlink(jl_io_loop, &req, argv[2], argv[1], flags, NULL);
    printf("ret = %d\n", ret);
    uv_fs_req_cleanup(&req);
    return ret;
}
```

Compiled as `gcc -Iinclude -L. -luv-11 uv_symlink.c -o uv_symlink`

Ran

```
$ ./uv_symlink "C:\Users\Tony\symlinktest" "D:\code\msys64\home\Tony\github\libuv"
ret = 0

$ ls -al /c/Users/Tony/symlinktest
lrwxrwxrwx 1 Tony None 24 Jul 29 16:03 /c/Users/Tony/symlinktest -> /home/Tony/github/libuv/

$ ./uv_symlink "D:\code\symlinktest" "D:\code\msys64\home\Tony\github\libuv"
ret = -4048
```

`D:\code` exists and is writable, there's no existing file `D:\code\symlinktest` in the way, so why does creating a junction work in `C:\Users\Tony` but not in `D:\code`? There are some other directories on my D: drive that I can create junctions in, but not all.

Note that using `mklink /j` works just fine:

```
$ rm /c/Users/Tony/symlinktest

$ cmd //c mklink //j "C:\Users\Tony\symlinktest" "D:\code\msys64\home\Tony\github\libuv"
Junction created for C:\Users\Tony\symlinktest <<===>> D:\code\msys64\home\Tony\github\libuv

$ ls -al /c/Users/Tony/symlinktest
lrwxrwxrwx 1 Tony None 23 Jul 29 16:08 /c/Users/Tony/symlinktest -> /home/Tony/github/libuv

[...]

PR #1388, opened 2014-07-31
change GENERIC_ALL to GENERIC_WRITE in fs__create_junction

tkelman:
fixes https://github.com/joyent/libuv/issues/1385, creating NTFS junctions in directories that are apparently missing certain permissions on Windows.


saghul:
A quick look shows that this is how ReactOS does it, right?

tkelman:
Yup. See https://github.com/mirror/reactos/blob/c69ae24f90028bf3bdba150885a924080e007117/reactos/base/shell/cmd/mklink.c#L84 - they also don't use `FILE_FLAG_OPEN_REPARSE_POINT`, but in my little test program that didn't make a difference as far as I could tell.

saghul:
I see. Calling the Windows cavalry: @piscisaureus, @orangemocha, can either of you give this a quick look? Looks reasonable to me.

saghul:
FTR: http://stackoverflow.com/a/10508724

piscisaureus:
Changing to GENERIC_WRITE lgtm.

saghul:
Thanks Tony! Landed in https://github.com/joyent/libuv/commit/3d73d5562791e8fe115732decd38db06a199e46a. Also thanks for confirming Bert!

tkelman:
Cheers. Keep up the good work guys. One of these days Julia (and Rust et al) won't need a fork anymore and we'll get everything we need all in the same place.

saghul:
> Cheers. Keep up the good work guys. One of these days Julia (and Rust et al) won't need a fork anymore and we'll get everything we need all in the same place.

Thanks! That's what we aim for.

PR #1389, opened 2014-07-31
unix, windows: fix uv_udp_getsockname signature

zerhacken:
Fixes uv_udp_getsockname handle argument to a similar signature
as uv_tcp_getsockname, handle isnt modifed and should be const.


saghul:
Thanks Rasmus! Landed in https://github.com/joyent/libuv/commit/e72d48ddab87a4168834b9f3449c998ce3600842 :sparkles:

PR #1390, opened 2014-08-01
win: match behavior of uv_backend_timeout with unix

vtjnash:
This rearranges some of the code in `uv_run` that computes the timeout value so that it more closely matches the structure (and function names) of the equivalent on unix. This enables the use of `uv_backend_timeout` for using libuv as an embedded event loops (for experimenting with libuv-Gtk integration. I fallback to a moderate rate of polling when uv_backend_fd returns -1, and use this function to give an upper-bound on that timeout.)

Note: this removes the call to `uv_update_time(loop)` from `uv_get_poll_timeout(uv_loop_t* loop)`, for consistency with unix (because `loop` is marked `const` there).


saghul:
@vtjnash thanks! This was on my list of things to clean up for a while :-) Now, on Unix the idea is to use this together with `uv_backend_fd` and poll that one manually on another thread. Since there is no such a thing on Windows, how if the Gtk integration done on Windows?

vtjnash:
Currently, I just accept the fact that new events may be handled after a slight delay (Currently set to 50ms). Eventually, it might be nice to use IOCP in a separate thread, and then use a condition variable (or NamedPipe) to feed these events back to the main thread (you can't wait on an IOCP – since IOCP notification is based upon keeping the number of running threads less than the number of cores, it doesn't make sense to have a non-blocking API for this – while you can wait on a condition variable handle as an event).

src/win/timer.c
```diff
@@ -193,15 +193,15 @@ uint64_t uv_timer_get_repeat(const uv_timer_t* handle) {
...
 
-DWORD uv_get_poll_timeout(uv_loop_t* loop) {
+DWORD uv__next_timeout(const uv_loop_t* loop) {
   uv_timer_t* timer;
   int64_t delta;
 
   /* Check if there are any running timers */
-  timer = RB_MIN(uv_timer_tree_s, &loop->timers);
+  /* Need to cast away const first, since RB_MIN doesn't know what we are */
+  /* going to do with this return value, it can't be marked const */
```
saghul: style: can you make this a multiline comment?

saghul:
Tiny style nit, but LGTM. A simple test for this would be nice. Just create a timer in the future and assert `uv_backend_timeout` returns > 0 before calling uv_run, that should do.

src/win/timer.c
vtjnash: I can. (I had changed it to match the style of the surrounding comments.)

saghul: > I can. (I had changed it to match the style of the surrounding comments.)

We prefer multiline comments, the Windows side is a bit inconsistent still, though :-S

vtjnash:
I cleaned up all of the comments in core.c and timer.c, for consistency (no changes to the text in those commits), and also added a simple test.

test/test-loop-time.c
```diff
@@ -32,3 +32,31 @@ TEST_IMPL(loop_update_time) {
...
+
+  r = uv_timer_start(&timer, cb, 1000, 0); /* 1 sec */
+  ASSERT(uv_backend_timeout(loop) > 100); /* 0.1 sec */
+  ASSERT(uv_backend_timeout(loop) <= 1000); /* 1.1 sec */
+
+  r = uv_run(uv_default_loop(), UV_RUN_DEFAULT);
+  ASSERT(!r);
+  ASSERT(uv_backend_timeout(loop) == 0);
+
+  uv_close((uv_handle_t*)&timer, NULL);
```
saghul: Close the timer on the timer callback, and use MAKE_VALGRIND_HAPPY at the end of the test.

vtjnash: i was showing that the `uv_run` loop would return even if the timer was still valid (but inactive).

should `loop_update_time` (above) have `MAKE_VALGRIND_HAPPY` also?

saghul: > i was showing that the uv_run loop would return even if the timer was still valid (but inactive).

I think we already have tests for that.

> should loop_update_time (above) have MAKE_VALGRIND_HAPPY also?

All tests should end like that, it actually does more than just "making valgrind happy", it also closes all pending handles and destroys the loop.

test/test-loop-time.c
```diff
@@ -30,5 +30,34 @@ TEST_IMPL(loop_update_time) {
...
+
+  r = uv_timer_init(loop, &timer);
+  ASSERT(!r);
+
+  ASSERT(!uv_loop_alive(loop));
+  ASSERT(uv_backend_timeout(loop) == 0);
+
+  r = uv_timer_start(&timer, cb, 1000, 0); /* 1 sec */
+  ASSERT(uv_backend_timeout(loop) > 100); /* 0.1 sec */
+  ASSERT(uv_backend_timeout(loop) <= 1000); /* 1.1 sec */
```
saghul: You meant 1100 here? (comment says 1.1 sec)

test/test-loop-time.c
```diff
@@ -30,5 +30,34 @@ TEST_IMPL(loop_update_time) {
...
+}
+
+TEST_IMPL(loop_backend_timeout) {
+  uint64_t start;
+  uv_loop_t *loop = uv_default_loop();
+  uv_timer_t timer;
+  int r;
+
+  r = uv_timer_init(loop, &timer);
+  ASSERT(!r);
```
saghul: Style: please use "== 0", it's the preferred way. (some tests written long ago may not yet comply to this, just leave them).

test/test-loop-time.c
vtjnash: no `== 1000` ended up being correct

test/test-loop-time.c
```diff
@@ -30,5 +30,34 @@ TEST_IMPL(loop_update_time) {
...
+
+  ASSERT(!uv_loop_alive(loop));
+  ASSERT(uv_backend_timeout(loop) == 0);
+
+  r = uv_timer_start(&timer, cb, 1000, 0); /* 1 sec */
+  ASSERT(uv_backend_timeout(loop) > 100); /* 0.1 sec */
+  ASSERT(uv_backend_timeout(loop) <= 1000); /* 1.1 sec */
+
+  r = uv_run(uv_default_loop(), UV_RUN_DEFAULT);
```
saghul: Use loop variable here.

test/test-loop-time.c
```diff
@@ -30,5 +30,34 @@ TEST_IMPL(loop_update_time) {
...
+  uv_timer_t timer;
+  int r;
+
+  r = uv_timer_init(loop, &timer);
+  ASSERT(!r);
+
+  ASSERT(!uv_loop_alive(loop));
+  ASSERT(uv_backend_timeout(loop) == 0);
+
+  r = uv_timer_start(&timer, cb, 1000, 0); /* 1 sec */
```
saghul, line 53: Assert r == 0.

test/test-loop-time.c
```diff
@@ -30,5 +30,35 @@ TEST_IMPL(loop_update_time) {
...
+  MAKE_VALGRIND_HAPPY();
+  return 0;
+}
+
+static void cb(uv_timer_t* timer) {
+  uv_close((uv_handle_t*)timer, NULL);
+}
+
+TEST_IMPL(loop_backend_timeout) {
+  uint64_t start;
```
saghul, line 42: Unused.

saghul:
That unused var aside, LGTM. Can you please squeash / rebase?

saghul:
I went ahead and landed it on 2e9d86e105f4cc7c6d8d969ab2d58d9d353679aa. Thanks @vtjnash!

PR #1387, opened 2014-08-01
fd leak with simple uv_loop_new / uv_loop_delete loop

jfirebaugh:
```
#include <uv.h>
#include <stdio.h>

int main() {
    for (int i = 0; i < 100; i++) {
        uv_loop_t* loop = uv_loop_new();

        if (!loop) {
           printf("failed to create loop!\n");
           return 0;
        }

        uv_loop_delete(loop);
    }
}
```

Compiled on OS X against current libuv HEAD (beb54fe72da0dad40dc671a3b4c1edfa3a55203f):

```
$ clang uv_loop.c -g -I ../libuv/include/ -l uv -L ../libuv/build/Release/
$ ./a.out
Abort trap: 6
```

Debugger says:

```
Process 6427 stopped
* thread #1: tid = 0x18b5bd, 0x00007fff8be2e866 libsystem_kernel.dylib`__pthread_kill + 10, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00007fff8be2e866 libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill + 10:
-> 0x7fff8be2e866:  jae    0x7fff8be2e870            ; __pthread_kill + 20
   0x7fff8be2e868:  movq   %rax, %rdi
   0x7fff8be2e86b:  jmpq   0x7fff8be2b175            ; cerror_nocancel
   0x7fff8be2e870:  ret    
(lldb) bt
* thread #1: tid = 0x18b5bd, 0x00007fff8be2e866 libsystem_kernel.dylib`__pthread_kill + 10, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007fff8be2e866 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff8cda935c libsystem_pthread.dylib`pthread_kill + 92
    frame #2: 0x00007fff8759cb1a libsystem_c.dylib`abort + 125
    frame #3: 0x00000001000070ca a.out`uv__loop_init(loop=<unavailable>, default_loop=<unavailable>) + 560 at loop.c:150
    frame #4: 0x00000001000071dc a.out`uv_loop_new [inlined] uv_loop_init + 8 at loop.c:50
    frame #5: 0x00000001000071d4 a.out`uv_loop_new + 28 at loop.c:81
    frame #6: 0x0000000100001548 a.out`main + 40 at uv_loop.c:6
```

This points to [uv_async_init](https://github.com/joyent/libuv/blob/beb54fe72da0dad40dc671a3b4c1edfa3a55203f/src/unix/loop.c#L149-150) failing, and I can see that [this uv__make_pipe](https://github.com/joyent/libuv/blob/beb54fe72da0dad40dc671a3b4c1edfa3a55203f/src/unix/async.c#L222-222) results in `EMFILE`. So it looks like there is a file descriptor leak somewhere.


jfirebaugh:
I found the issue and converted this to a pull request. @saghul as you said, `uv_loop_delete` calls `uv_loop_close` and asserts -- the issue was that it does so inside of the assert, so in release builds the whole expression is compiled out.

This was introduced in 2611ef43b15dc083ac2ffd5f16901a4ff076be03.

I considered adding a test but since they run against debug builds this would not have been caught by tests anyway.

saghul:
Great catch!

saghul:
Cheers, John! Landed in https://github.com/joyent/libuv/commit/2cd91f9768361df16505d0c2f67f48a8f1c08c88.

PR #1392, opened 2014-08-03
windows: fix buffer overflow issue on uv__getnameinfo_work()

lilohuang:
According to MSDN page (http://msdn.microsoft.com/en-us/library/windows/desktop/ms738531.aspx), "NodeBufferSize" and "ServiceBufferSize" parameters of GetNameInfoW() function are expected to fill the number of WCHAR, not the buffer size of memory layout.

Assume the value of NI_MAXHOST is 1025, and "sizeof(host)" would get 2050. A buffer overflow issue will be triggered by it.


src/win/getnameinfo.c
```diff
@@ -49,9 +49,9 @@ static void uv__getnameinfo_work(struct uv__work* w) {
   ret = GetNameInfoW((struct sockaddr*)&req->storage,
                      sizeof(req->storage),
                      host,
-                     sizeof(host),
+                     NI_MAXHOST,
                      service,
-                     sizeof(service),
+                     NI_MAXSERV,
```
saghul: Goot catch! Lets use ARRAY_SIZE(host) and service instead, though.

lilohuang: Done :+1:

saghul:
Kewl, I'll merge it later today, thanks!

JacksonTian:
cc @lilohuang , could you rebase the commit, combine 4 commits to 1.

lilohuang:
Hi @JacksonTian , it should be done, check it out, thanks :laughing:

saghul:
Thanks @lilohuang! Landed in 89fc7d80c46154976a0eeb53087f5c9539528c24

PR #1397, opened 2014-08-05
Update linux-core.c

simonmaster:
This patch allows dup2 cloned fd to work with the event handler as intended.


saghul:
This really needs a test. Also, if the operation that triggers the error is EPOLL_CTL_DEL I'm not sure we want to add it again.

@bnoordhuis it would be nice if you could take a quick peek at this

simonmaster:
This code is not triggered by the EPOLL_CTL_DEL.
It is triggered by EPOLL_CTL_MOD (see line 171-174).
The patch will rescue script from "abort()".
The event handler detected the corresponding "fd" does not EXIST in the poll chain because of the same fd was cloned by dup2 system call. The new EPOLL_CTL_ADD will allow the engine to run smoothly without introduce any new calls. Without this patch, Node.js will abort. 
Please pull this patch, it enables dup2 to work as intended.

saghul:
> This code is not triggered by the EPOLL_CTL_DEL.
> It is triggered by EPOLL_CTL_MOD (see line 171-174).

Ah, I see, sorry I missed the context.

> The patch will rescue script from "abort()".
> The event handler detected the corresponding "fd" does not EXIST in the poll chain because of the same fd was cloned by dup2 system call. The new EPOLL_CTL_ADD will allow the engine to run smoothly without introduce any new calls. Without this patch, Node.js will abort.

What script makes node abort? Since we hide file descriptors from the user, how are you dup2-ing it? Or are you using uv_poll_t?

> Please pull this patch, it enables dup2 to work as intended.

Sorry, but it needs a test :-)

simonmaster:
I wrote a special communication program which allows Server Side Java Script to accept incoming "fd" via domain socket.  This way I can transfer all fd on the same machine freely.
I will happily provide a test codes for you. Let me work on it now. I am new on this; any particular format?
 It will include a java script for node (run as server); a simple c++ addon code. And client side test c++ program.

saghul:
Thanks! The test needs to be intgerated into the libuv test suite, so plain C without any external dependencies.

src/unix/linux-core.c
```diff
@@ -177,6 +177,22 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
...
      */
     if (uv__epoll_ctl(loop->backend_fd, op, w->fd, &e)) {
+      /*
+       * Allow fd generated by dup2 to work.
+       * The original fd was monitored by the backend_fd ... dup2 close the original fd
+       * and cloned the new fd inplaced of the original. The following codes will allow
+       * cloned fd continue to work as intend.
+       */
+      if (errno == ENOENT) {
+        op = UV_EPOLL_CTL_ADD;
```
bnoordhuis, line 187: This can't possibly compile because it's `UV__EPOLL_CTL_ADD` (note the two underscores after UV).  Ditto for the call to `uv__epoll_ctl()`.

src/unix/linux-core.c
```diff
@@ -177,6 +177,22 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
      * events, skip the syscall and squelch the events after epoll_wait().
      */
     if (uv__epoll_ctl(loop->backend_fd, op, w->fd, &e)) {
+      /*
+       * Allow fd generated by dup2 to work.
+       * The original fd was monitored by the backend_fd ... dup2 close the original fd
+       * and cloned the new fd inplaced of the original. The following codes will allow
+       * cloned fd continue to work as intend.
+       */
```
bnoordhuis, line 185: The comment wasn't very elucidating to me, maybe you can rephrase it?  I think the situation you're trying to address here is:
1. This file descriptor has been watched before.
2. Something that wasn't libuv closed it.  The descriptor was removed from the epoll set.
3. Libuv still thinks it exists and tries to EPOLL_CTL_MOD its entry in the epoll set.
4. That fails with ENOENT because of 2 so try EPOLL_CTL_ADD.

That's right, isn't it?

saghul, line 185: That was my understanding too.

simonmaster, line 185: bnoordhuis, yes your rephrasing is 100% correct. Sometime it is hard to put it in writing.
Thanks

simonmaster:
saghul, I have the test package (dup2test.tga) but how do I send it for you? what is your email?
thanks

saghul:
Please make a pull request with your changes. You can check how to do it by
reading CONTRIBUTING.md.

simonmaster:
thanks will try to convert my test case according to CONTRIBUTING.md.

simonmaster:
saghul,
I did a git commit on directory node/test/addons/dup2test.
I did the git rebase upstream/v0.10  (no errors...)
Please check it out. you need to compile the broker.c and install the c++ addon for testing. Thanks

saghul:
@simonmaster There are a number of things which are not right here:
- Not sure where you committed what code, I still see the old one in this PR
- The test needs to be integrated into the libuv test suite (forget about Node), that is, it must be C only and run with the test suite, it also must be minimal

simonmaster:
I am new on this. Do you mean that I should follow the test examples on the directory: libuv/test... Is there a document on this directory. I was following the "CONTRIBUTING.md" directory which leaded me into the wrong the direction. I don't mind to re-do it. Please send me the steps to follow. Thanks for your patient.

saghul:
You should follow these instructions: https://github.com/joyent/libuv/blob/master/CONTRIBUTING.md#test That is, create your test in tests/ and register it in test-list.h. You can check any test as an example, and ask questions here, no problem.

saghul:
Ping.

PR #1402, opened 2014-08-07
Use OpenBSD implementation for uv_fs_mkdtemp on Windows

Hinidu:
As [stated](https://github.com/joyent/libuv/pull/1368/files#r15906993) by @sam-github libuv can't use glibc's modified code because glibc licensed under LGPL, but libuv under MIT. I apologize again, it was my fault because of ignorance.
So in this PR I ported [OpenBSD version](http://openbsd.cs.toronto.edu/cgi-bin/cvsweb/src/lib/libc/stdio/mktemp.c?rev=1.33&content-type=text/x-cvsweb-markup).
Main differences with original:
- `tries = TMP_MAX` instead of `tries = INT_MAX`. Though I'm not sure what is the best value.
- Used `rand` instead of `arc4random`. Though I found [portable arc4random.c](https://github.com/libevent/libevent/blob/master/arc4random.c) licensed under 3-clause BSD license. So it is possible to use it if appropriate.

### DIFF
--- a/src/win/fs.c
+++ b/src/win/fs.c
@@ -730,7 +730,9 @@ void fs__mkdtemp(uv_fs_t* req) {
   WCHAR *cp, *ep;
   unsigned int tries, i;
   size_t len;
+  HCRYPTPROV h_crypt_prov;
   uint64_t v;
+  BOOL released;
 
   len = wcslen(req->pathw);
   ep = req->pathw + len;
@@ -739,9 +741,19 @@ void fs__mkdtemp(uv_fs_t* req) {
     return;
   }
 
+  if (!CryptAcquireContext(&h_crypt_prov, NULL, NULL, PROV_RSA_FULL,
+                           CRYPT_VERIFYCONTEXT)) {
+    SET_REQ_WIN32_ERROR(req, GetLastError());
+    return;
+  }
+
   tries = TMP_MAX;
   do {
-    v = (((uint64_t) rand()) << 30) | (rand() << 15) | rand();
+    if (!CryptGenRandom(h_crypt_prov, sizeof(v), (BYTE*) &v)) {
+      SET_REQ_WIN32_ERROR(req, GetLastError());
+      break;
+    }
+
     cp = ep - num_x;
     for (i = 0; i < num_x; i++) {
       *cp++ = tempchars[v % num_chars];
@@ -752,14 +764,18 @@ void fs__mkdtemp(uv_fs_t* req) {
       len = strlen(req->path);
       wcstombs((char*) req->path + len - num_x, ep - num_x, num_x);
       SET_REQ_RESULT(req, 0);
-      return;
+      break;
     } else if (errno != EEXIST) {
       SET_REQ_RESULT(req, -1);
-      return;
+      break;
     }
   } while (--tries);
 
-  SET_REQ_RESULT(req, -1);
+  released = CryptReleaseContext(h_crypt_prov, 0);
+  assert(released);
+  if (tries == 0) {
+    SET_REQ_RESULT(req, -1);
+  }
 }
 
 
--- a/src/win/fs.c
+++ b/src/win/fs.c
@@ -721,74 +721,44 @@ void fs__mkdir(uv_fs_t* req) {
 }
 
 
-/* Some parts of the implementation were borrowed from glibc. */
+/* OpenBSD original: lib/libc/stdio/mktemp.c */
 void fs__mkdtemp(uv_fs_t* req) {
-  static const WCHAR letters[] =
+  static const WCHAR *tempchars =
     L"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+  static const size_t num_chars = 62;
+  static const size_t num_x = 6;
+  WCHAR *cp, *ep;
+  unsigned int tries, i;
   size_t len;
-  WCHAR* template_part;
-  static uint64_t value;
-  unsigned int count;
-  int fd;
-
-  /* A lower bound on the number of temporary files to attempt to
-     generate. The maximum total number of temporary file names that
-     can exist for a given template is 62**6. It should never be
-     necessary to try all these combinations. Instead if a reasonable
-     number of names is tried (we define reasonable as 62**3) fail to
-     give the system administrator the chance to remove the problems. */
-#define ATTEMPTS_MIN (62 * 62 * 62)
-
-  /* The number of times to attempt to generate a temporary file. To
-     conform to POSIX, this must be no smaller than TMP_MAX. */
-#if ATTEMPTS_MIN < TMP_MAX
-  unsigned int attempts = TMP_MAX;
-#else
-  unsigned int attempts = ATTEMPTS_MIN;
-#endif
+  uint64_t v;
 
   len = wcslen(req->pathw);
-  if (len < 6 || wcsncmp(&req->pathw[len - 6], L"XXXXXX", 6)) {
+  ep = req->pathw + len;
+  if (len < num_x || wcsncmp(ep - num_x, L"XXXXXX", num_x)) {
     SET_REQ_UV_ERROR(req, UV_EINVAL, ERROR_INVALID_PARAMETER);
     return;
   }
 
-  /* This is where the Xs start. */
-  template_part = &req->pathw[len - 6];
+  tries = TMP_MAX;
+  do {
+    v = (((uint64_t) rand()) << 30) | (rand() << 15) | rand();
+    cp = ep - num_x;
+    for (i = 0; i < num_x; i++) {
+      *cp++ = tempchars[v % num_chars];
+      v /= num_chars;
+    }
 
-  /* Get some random data. */
-  value += uv_hrtime() ^ _getpid();
-
-  for (count = 0; count < attempts; value += 7777, ++count) {
-    uint64_t v = value;
-
-    /* Fill in the random bits. */
-    template_part[0] = letters[v % 62];
-    v /= 62;
-    template_part[1] = letters[v % 62];
-    v /= 62;
-    template_part[2] = letters[v % 62];
-    v /= 62;
-    template_part[3] = letters[v % 62];
-    v /= 62;
-    template_part[4] = letters[v % 62];
-    v /= 62;
-    template_part[5] = letters[v % 62];
-
-    fd = _wmkdir(req->pathw);
-
-    if (fd >= 0) {
+    if (_wmkdir(req->pathw) == 0) {
       len = strlen(req->path);
-      wcstombs((char*) req->path + len - 6, template_part, 6);
+      wcstombs((char*) req->path + len - num_x, ep - num_x, num_x);
       SET_REQ_RESULT(req, 0);
       return;
     } else if (errno != EEXIST) {
       SET_REQ_RESULT(req, -1);
       return;
     }
-  }
+  } while (--tries);
 
-  /* We got out of the loop because we ran out of combinations to try. */
   SET_REQ_RESULT(req, -1);
 }
 
--- a/include/uv.h
+++ b/include/uv.h
@@ -1892,6 +1892,14 @@ UV_EXTERN int uv_fs_write(uv_loop_t* loop, uv_fs_t* req, uv_file file,
 UV_EXTERN int uv_fs_mkdir(uv_loop_t* loop, uv_fs_t* req, const char* path,
     int mode, uv_fs_cb cb);
 
+/*
+ * Generates a uniquely named temporary directory from tpl. The last six
+ * characters of tpl must be XXXXXX and these are replaced with a string that
+ * makes the directory name unique. On success the name of created directory
+ * will be stored in req->path.
+ * You should ensure that srand() has been called before using this function
+ * on Windows.
+ */
 UV_EXTERN int uv_fs_mkdtemp(uv_loop_t* loop, uv_fs_t* req, const char* tpl,
     uv_fs_cb cb);
 
--- a/include/uv.h
+++ b/include/uv.h
@@ -1897,8 +1897,6 @@ UV_EXTERN int uv_fs_mkdir(uv_loop_t* loop, uv_fs_t* req, const char* path,
  * characters of tpl must be XXXXXX and these are replaced with a string that
  * makes the directory name unique. On success the name of created directory
  * will be stored in req->path.
- * You should ensure that srand() has been called before using this function
- * on Windows.
  */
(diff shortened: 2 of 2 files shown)


src/win/fs.c
```diff
@@ -721,74 +721,50 @@ void fs__mkdir(uv_fs_t* req) {
...
-    template_part[1] = letters[v % 62];
-    v /= 62;
-    template_part[2] = letters[v % 62];
-    v /= 62;
-    template_part[3] = letters[v % 62];
-    v /= 62;
-    template_part[4] = letters[v % 62];
-    v /= 62;
-    template_part[5] = letters[v % 62];
+  for (i = 1; i <= num_x; ++i)
```
saghul: Replace this for with the previous "wcsncmp(&req->pathw[len - 6], L"XXXXXX", 6)", and or it in the len < num_x test.

src/win/fs.c
```diff
@@ -721,74 +721,50 @@ void fs__mkdir(uv_fs_t* req) {
...
+    for (i = 0; i < num_x; i++) {
+      *cp++ = tempchars[v % num_chars];
+      v /= num_chars;
+    }
 
-    if (fd >= 0) {
+    if (_wmkdir(req->pathw) == 0) {
       len = strlen(req->path);
-      wcstombs((char*) req->path + len - 6, template_part, 6);
+      wcstombs((char*)req->path + len - num_x, ep - num_x, 6);
```
saghul: style: (char*) req...

Also, replace the last 6 with num_x.

src/win/fs.c
```diff
@@ -721,74 +721,50 @@ void fs__mkdir(uv_fs_t* req) {
 }
 
 
-/* Some parts of the implementation were borrowed from glibc. */
+/* OPENBSD ORIGINAL: lib/libc/stdio/mktemp.c */
```
saghul: style: captitalize properly: OpenBSD original:

saghul:
I don't think we need arc4random here, there are nough number of tries to make it not collide, I guess. (famous last words)

saghul:
Made some comments, but generally LGTM/ @sam-github, @bnoordhuis, can you PTAL?

Hinidu:
@saghul thanks for review, pushed fixes.

saghul:
LGTM. I'll wait for others to chime in before merging. Thanks for handling this so quickly @Hinidu, much appreciated.

src/win/fs.c
```diff
@@ -721,74 +721,44 @@ void fs__mkdir(uv_fs_t* req) {
...
-    template_part[3] = letters[v % 62];
-    v /= 62;
-    template_part[4] = letters[v % 62];
-    v /= 62;
-    template_part[5] = letters[v % 62];
-
-    fd = _wmkdir(req->pathw);
+  tries = TMP_MAX;
+  do {
+    v = (((uint64_t)rand()) << 32) | rand();
```
Hinidu: Should I put a space between `(uint64_t)` and `rand()`?

saghul: Oh, yeah, good catch.

Hinidu:
> Thanks for handling this so quickly @Hinidu, much appreciated.

I'm very glad. It was such a shame for me, so that I actually couldn't do anything else :-)

src/win/fs.c
```diff
@@ -721,74 +721,44 @@ void fs__mkdir(uv_fs_t* req) {
...
-    template_part[3] = letters[v % 62];
-    v /= 62;
-    template_part[4] = letters[v % 62];
-    v /= 62;
-    template_part[5] = letters[v % 62];
-
-    fd = _wmkdir(req->pathw);
+  tries = TMP_MAX;
+  do {
+    v = (((uint64_t) rand()) << 32) | rand();
```
bnoordhuis: This has weak entropy.  rand() on Windows returns a value between 0 and 32768.  That's 15 bits of entropy so that means `v` will have at most 15 \* 2 = 30 bits of entropy.

bnoordhuis: I assume the original code used arc4random()?  I think the Windows equivalent is CryptGenRandom() but that API is somewhat cumbersome to use.

saghul: How about using `uv_hrtime() ^ _getpid();` as we did before?

Hinidu: The upper bound of entropy in this case is 62^6 (< 2^36). So the best and simplest implementation would be:

``` c
v = (((uint64_t) rand()) << 30) | (rand() << 15) | rand();
```

What do you think?

bnoordhuis: That looks reasonable to me.  Arguably better than time + pid because that's fairly predictable and/or externally observable.

EDIT: Assuming the PRNG has been seeded, of course.

Hinidu: > EDIT: Assuming the PRNG has been seeded, of course.

How and when I should do this?

bnoordhuis: I don't think that should be the responsibility of libuv.  Rather, I would add a note to the documentation for uv_fs_mkdtemp() that it's the responsibility of the libuv user to ensure that srand() has been called.

It's either that or:
1. Switch to a CSPRNG like CryptGenRandom(), or
2. Make the entropy the responsibility of the user, by means of a user-provided buffer or callback.

Option 1 makes the API easier to use but the user won't have much influence on the quality of the (CS)PRNG.

Option 2 makes the API more cumbersome but is useful in applications like node.js that have access to a cross-platform, high-quality entropy pool (OpenSSL's entropy pool in node's case.)

I don't have a preference, I'm just outlining the possibilities that I can think of.

saghul: > I don't think that should be the responsibility of libuv. Rather, I would add a note to the documentation for uv_fs_mkdtemp() that it's the responsibility of the libuv user to ensure that srand() has been called.

I'd go with this.

Hinidu:
@bnoordhuis @saghul pushed first solution of the problem with entropy.
In the future if appropriate it can be replaced by any solution like `CryptGenRandom` or `arc4random`...

Hinidu:
@sam-github thanks for suggestion. Updated documentation of `uv_fs_mkdtemp`.

Hinidu:
> @bnoordhuis @saghul pushed first solution of the problem with entropy.
> In the future if appropriate it can be replaced by any solution like CryptGenRandom or arc4random...

What if we use both `uv_hrtime() ^ getpid()` and `rand()`s? General scheme:

``` c
uint64_t x = uv_hrtime() ^ getpid();
while (!success) {
  uint64_t y = (((uint64_t) rand()) << 30) | (rand() << 15) | rand();
  success |= try_create_directory(x ^ y);
}
```

I think it would reduce the need to call `srand()` before using `uv_fs_mkdtemp` and accordingly simplify its usage. Though directory name will be possible to predict, I think this is not a problem in the most cases. And when it is a real problem the user can call `srand` and get what he needs.

What do you think @saghul, @bnoordhuis?

bnoordhuis:
@Hinidu  The problem with that approach is if rand() _also_ returns `hrtime() ^ getpid()`, then you have zero bits of entropy.

Real-world usage won't be that black and white, of course, but if there's some correlation between rand(), hrtime() and getpid(), then instead of strengthening the entropy, you might end up weakening it.

Hinidu:
@bnoordhuis thanks for interesting observation!
I learned a bit about CryptGenRandom. It seems not too hard to use and it seems that this function does not need initialization similiar to srand. So if it has no other problems it looks like ideal approach to me.
I can try to replace rand by CryptGenRandom in this PR in the Monday.

saghul:
Sounds good, thanks Pavel!

src/win/fs.c
```diff
@@ -721,74 +721,44 @@ void fs__mkdir(uv_fs_t* req) {
...
 
-    if (fd >= 0) {
+    if (_wmkdir(req->pathw) == 0) {
       len = strlen(req->path);
-      wcstombs((char*) req->path + len - 6, template_part, 6);
+      wcstombs((char*) req->path + len - num_x, ep - num_x, num_x);
       SET_REQ_RESULT(req, 0);
       return;
     } else if (errno != EEXIST) {
       SET_REQ_RESULT(req, -1);
```
piscisaureus, line 769: The request result is supposed to have the actual error code and not just -1

src/win/fs.c
```diff
@@ -721,74 +721,44 @@ void fs__mkdir(uv_fs_t* req) {
...
+  do {
+    v = (((uint64_t) rand()) << 30) | (rand() << 15) | rand();
+    cp = ep - num_x;
+    for (i = 0; i < num_x; i++) {
+      *cp++ = tempchars[v % num_chars];
+      v /= num_chars;
+    }
 
-    if (fd >= 0) {
+    if (_wmkdir(req->pathw) == 0) {
```
piscisaureus, line 763: I would suggest to use `CreateDirectoryW(req->pathw, NULL)`.
It returns a nonzero value on succes, on failure the error can be retrieved with `GetLastError()`.
See more here: http://msdn.microsoft.com/en-us/library/windows/desktop/aa363855%28v=vs.85%29.aspx

src/win/fs.c
Hinidu, line 769: I think this is the prefered way because SET_REQ_RESULT will handle this
https://github.com/Hinidu/libuv/blob/openbsd_mkdtemp/src/win/fs.c#L54

src/win/fs.c
saghul, line 763: Thanks for chiming in Bert! Is there an advantage over _wmkdir? As in, should we also change uv_fs_mkdir to use it?

Hinidu, line 763: I used `_wmkdir` because it is used in `uv_fs_mkdir`.

piscisaureus, line 763: @Hinidu 

The reasoning makes sense. Actually, nevermind then.
In general I have preferred to not use any I/O APIs that are implemented in msvcrt because they often do ridiculous things. However I just checked the _wmkdir source and it's a really thin wrapper around CreateDirectoryW.

saghul, line 763: I should have googled first: http://stackoverflow.com/questions/24205401/difference-between-mkdir-and-createdirectory :-)

Hinidu:
Pushed commit with CryptGenRandom().

bnoordhuis:
LGTM with the caveat that I'm not really a Windows programmer.  If I could make two suggestions:
- There are a bunch of CryptReleaseContext() calls strewn around.  From a maintenance perspective it would be neater if there was a single clean up (and exit) point.
- I would suggest assert()'ing that CryptReleaseContext() returns true (sorry, TRUE) to catch bugs where another thread clobbers the handle.

Hinidu:
> There are a bunch of CryptReleaseContext() calls strewn around. From a maintenance perspective it would be neater if there was a single clean up (and exit) point.

It bothers me too. ~~Should I use goto for that? Or some another option?~~ **Edit:** Possible solution posted below.

> I would suggest assert()'ing that CryptReleaseContext() returns true (sorry, TRUE) to catch bugs where another thread clobbers the handle.

How is it possible? The handle is the local variable. I thought that another thread should not be able to clobber it.

Hinidu:
> It bothers me too. Should I use goto for that? Or some another option?

I think I found pretty good solution:

``` diff
diff --git a/src/win/fs.c b/src/win/fs.c
index 511952f..c412db5 100644
--- a/src/win/fs.c
+++ b/src/win/fs.c
@@ -748,10 +748,9 @@ void fs__mkdtemp(uv_fs_t* req) {

   tries = TMP_MAX;
   do {
     if (!CryptGenRandom(hCryptProv, sizeof(v), (BYTE*) &v)) {
-      CryptReleaseContext(hCryptProv, 0);
       SET_REQ_WIN32_ERROR(req, GetLastError());
-      return;
+      break;
     }

     cp = ep - num_x;
@@ -763,18 +762,18 @@ void fs__mkdtemp(uv_fs_t* req) {
     if (_wmkdir(req->pathw) == 0) {
       len = strlen(req->path);
       wcstombs((char*) req->path + len - num_x, ep - num_x, num_x);
-      CryptReleaseContext(hCryptProv, 0);
       SET_REQ_RESULT(req, 0);
-      return;
+      break;
     } else if (errno != EEXIST) {
-      CryptReleaseContext(hCryptProv, 0);
       SET_REQ_RESULT(req, -1);
-      return;
+      break;
     }
   } while (--tries);

   CryptReleaseContext(hCryptProv, 0);
-  SET_REQ_RESULT(req, -1);
+  if (tries == 0) {
+    SET_REQ_RESULT(req, -1);
+  }
 }
```

saghul:
Looks good, though we need to set some error in case all tries are exhausted.

Hinidu:
> Looks good, though we need to set some error in case all tries are exhausted.

In this case `errno` will be equal to `EEXIST` because we continue loop only in this case. Do you have other candidate?

saghul:
Ah, right. No, that's OK.

Hinidu:
Force pushed cleanup extraction.
There is left only issue with asserting CryptReleaseContext. Though I think there is no errors which can occur in our simple case according to [MSDN](http://msdn.microsoft.com/en-us/library/windows/desktop/aa380268%28v=vs.85%29.aspx). So I'm not sure what to do with that.

saghul:
You can add the assert just to be deffensive :-)

Hinidu:
Ok, force-pushed assert.
~~Just curious what the result would be? Will the program crash in this case or libuv somehow handle asserts?~~

bnoordhuis:
> How is it possible? The handle is the local variable. I thought that another thread should not be able to clobber it.

Not normally, no, but a HANDLE is just a pointer to an object.  If another thread manages to get a pointer to the same object (accidentally or otherwise), then things can go badly wrong.  The assert() is a bit of defensive programming against such bugs.

Hinidu:
> Not normally, no, but a HANDLE is just a pointer to an object. If another thread manages to get a pointer to the same object (accidentally or otherwise), then things can go badly wrong. The assert() is a bit of defensive programming against such bugs.

Thanks for explanation! I've already pushed `assert` as you can see.

saghul:
Thanks @Hinidu, I'll have a final look tonight and land it if everything checks out, which looks that way.

Hinidu:
> Thanks @Hinidu, I'll have a final look tonight and land it if everything checks out, which looks that way.

Thank you and all others, your suggestions were very helpful. Will wait landing or other comments.

sam-github:
Looks OK to me, thanks for adding a doc comment.

saghul:
Thanks Pavel! Landed in https://github.com/joyent/libuv/commit/a669f21bf88fdd06f445e5507ed1f834c0ba6432.

PS: I added an include for wincrypt.h, otherwise it wouldn't work on MinGW.

Hinidu:
Thanks Saúl!

PR #1410, opened 2014-08-09
doc: s/https/http for chromium gyp git repo

rvagg:


saghul:
Why? I have no problem pulling over HTTPS.

rvagg:
Oh, right, must be these old Linux distros I'm trying stuff on, assumed https wasn't working for that repo. Will investigate further!

rvagg:
ok, I'll close this and assume that it's simply older versions of git that can't do https

PR #1411, opened 2014-08-10
win: fix building on MinGW

huxingyi:
MinGW does not have LOCALE_INVARIANT/_malloca/_freea/wmemcpy_s present on the system.

My building environment:

gcc version 4.8.1 (GCC)
GNU gdb (GDB) 7.6.1
GNU ld (GNU Binutils) 2.24
GNU windres (GNU Binutils) 2.24
GNU dlltool (GNU Binutils) 2.24
GNU Make 3.82.90


saghul:
What version of MinGW are you using? The one from mingw.org or MinGW-w64?

src/win/process.c
```diff
@@ -796,7 +796,7 @@ int make_program_env(char* env_block[], WCHAR** dst_ptr) {
       /* copy var from env_block */
       DWORD r;
       len = wcslen(*ptr_copy) + 1;
-      r = wmemcpy_s(ptr, (env_len - (ptr - dst)), *ptr_copy, len);
+      r = wmemcpy(ptr, *ptr_copy, len);
```
saghul: Why downgrade to the lowest common denominator? Here is the implementation MinGW-w64 uses: https://github.com/Alexpux/Mingw-w64/blob/master/mingw-w64-crt/secapi/wmemcpy_s.c#L36

include/uv-win.h
```diff
@@ -39,6 +39,20 @@ typedef struct pollfd {
 } WSAPOLLFD, *PWSAPOLLFD, *LPWSAPOLLFD;
 #endif
 
+#ifndef LOCALE_INVARIANT
+# define LOCALE_INVARIANT 0x007f
+#endif
+
+#ifndef _malloca
+# if defined(_DEBUG)
```
saghul, line 47: I don't really like the idea of allocating memory in one place or in another depending on how the library was compiled. I'd rather always fallback to malloc/free.

huxingyi:
I installed MinGW long time ago, and I forgot the exact source I downloaded from. sorry.

My building environment(I use this bat [windows batch file to determine MinGW version info](http://forums.codeblocks.org/index.php/topic,9054.msg65050.html)  to generate):

```
Microsoft Windows XP [Version 5.1.2600]
(C) Copyright 1985-2001 Microsoft Corp.

D:\Documents and Settings\Admin>cd /d c:\mingw

C:\MinGW>e:\tool\version-of-mingw\version-of-mingw.bat
MINGWBASEDIR=C:\MinGW
gcc version 4.8.1 (GCC)
gcc version 4.8.1 (GCC)
GNU gdb (GDB) 7.6.1
GNU ld (GNU Binutils) 2.24
GNU windres (GNU Binutils) 2.24
GNU dlltool (GNU Binutils) 2.24
GNU Make 3.82.90
#define __MINGW32_VERSION           3.20
#define __W32API_VERSION 3.17
Press any key to continue . . .
```

to question 1
I think using wmemcpy_s is not obligate in this place. Because the space needed was calculated properly and malloced before using.

to question 2
Windows does in this behavior:
http://msdn.microsoft.com/en-us/library/5471dc8s.aspx

> Unlike _alloca, which does not require or permit a call to free to free the memory so allocated, _malloca requires the use of _freea to free memory. In debug mode, _malloca always allocates memory from the heap.

saghul:
Ah, I see. You are right, LGTM.

src/win/process.c
saghul: wmemcpy returns a `wchar_t*`, so we no longer need to save the value in r. Please fix that and I'll merge it :-)

huxingyi: fixed. thanks for review.

saghul:
Thanks @huxingyi! Landed in ca4741b

PR #1412, opened 2014-08-10
unix: fix typedef of uv_buf_t

lhecker:
Typedefs of unnamed structs cannot be forward declared.
Giving the uv_buf_t struct a name makes forward declarations possible.
As a side note: The uv_buf_t struct already got a name in uv-win.h.


saghul:
Cheers, Leonard, merged!

lhecker:
Thanks for the quick merge! :smile:

PR #1414, opened 2014-08-12
windows: fix GetNameInfoW error handling

orangemocha:
GetNameInfoW return non-zero on error, and not vice versa.

### DIFF
--- a/src/win/getnameinfo.c
+++ b/src/win/getnameinfo.c
@@ -46,7 +46,7 @@ static void uv__getnameinfo_work(struct uv__work* w) {
   int ret = 0;
 
   req = container_of(w, uv_getnameinfo_t, work_req);
-  if (!GetNameInfoW((struct sockaddr*)&req->storage,
+  if (GetNameInfoW((struct sockaddr*)&req->storage,
                      sizeof(req->storage),
                      host,
                      ARRAY_SIZE(host),


saghul:
Oh my god. Thanks @orangemocha, LGTM.

saghul:
In line with this, shouldn't we handle the error the same way for getaddrinfo?

```
diff --git a/src/win/getaddrinfo.c b/src/win/getaddrinfo.c
index 6392fba..0a85f54 100644
--- a/src/win/getaddrinfo.c
+++ b/src/win/getaddrinfo.c
@@ -74,11 +74,13 @@ int uv__getaddrinfo_translate_error(int sys_err) {

 static void uv__getaddrinfo_work(struct uv__work* w) {
   uv_getaddrinfo_t* req;
-  int err;
+  int ret = 0;

   req = container_of(w, uv_getaddrinfo_t, work_req);
-  err = GetAddrInfoW(req->node, req->service, req->hints, &req->res);
-  req->retcode = uv__getaddrinfo_translate_error(err);
+  if (GetAddrInfoW(req->node, req->service, req->hints, &req->res)) {
+    ret = WSAGetLastError();
+  }
+  req->retcode = uv__getaddrinfo_translate_error(ret);
 }
```

orangemocha:
Sorry for the GetNameInfo screwup :(
This is confusing but according to MSDN and my testing GetAddrInfoW already returns Windows Socket Error Codes

saghul:
No worries, I thought it had been like that since the begining and I was wondering how it worked :-P

On MSDN (for GetAddrInfoW) I read: "Use the gai_strerror function to print error messages based on the EAI_\* codes returned by the GetAddrInfoW function. The gai_strerror function is provided for compliance with IETF recommendations, but it is not thread safe. Therefore, use of a traditional Windows Sockets function, such as WSAGetLastError, is recommended."

So, you mean it would be redundant? Any chance it's harmless? Perhaps we should do it defensively, just in case not all errors are returned?

saghul:
Landed in 7bdcf3d, thanks Alexis!

saghul:
If you can, please check if it would be good to merge the patch I just attached. Thanks!

orangemocha:
Thanks! I will triple check on GetAddrInfo and let you know.

saghul:
> I will triple check on GetAddrInfo and let you know.

Awesome, thanks!

orangemocha:
I checked and tested again. GetAddrInfo sets the last error and returns it, so the current code is correct, and calling WSAGetLastError() would be redundant, albeit harmless. 
I suspect that GetNameInfo behaves the same way. But the MSDN documentation is slightly different, and even the samples are different (calling WSAGetLastError for GetNameInfo, but using the return as error for GetAddrInfo). So I think things are fine as they are.

orangemocha:
... and as additional clarification, when MSDN says that EAI_\* error codes "map" to WSA error codes, it means that those constants are actually the same.

saghul:
Thanks for looking into it!

txdv:
The test starts failing on Ubuntu 14.04.

It doesn't recognize the -5 as an error code, which should be EAI_NODATA on my machine, but for some reason it is not defined after moving it to unix/getaddrinfo.c.

https://github.com/lattera/glibc/blob/master/resolv/netdb.h#L629

saghul:
What test? This patch only dealt with Windows.

txdv:
Wrong issue, it is this commit: https://github.com/joyent/libuv/commit/c87c44fff3f28040159db6ad08134a647e85aacf

PR #1420, opened 2014-08-14
fs: introduce uv_readdir_next() and report types

indutny:
Introduce:

```
int uv_fs_readdir_next(uv_fs_t* req, uv_dirent_t* ent);
```

`uv_fs_readdir()` is not returning a file names list in `req->ptr`
anymore, the proper way to gather them is to call `uv_fs_readdir_next()`
in a callback.

cc @saghul


include/uv-unix.h
```diff
@@ -155,6 +156,10 @@ typedef pthread_barrier_t uv_barrier_t;
 typedef gid_t uv_gid_t;
 typedef uid_t uv_uid_t;
 
+typedef struct dirent uv__dirent_t;
+
+#define UV__DT_DIR DT_DIR
```
saghul, line 161: Why don't we just use `struct dirent` on Unix, since it's available?

include/uv-win.h
```diff
@@ -289,6 +289,14 @@ typedef struct uv_once_s {
 typedef unsigned char uv_uid_t;
 typedef unsigned char uv_gid_t;
 
+typedef struct uv__dirent_s {
+  int d_type;
+  char d_name[1];
+} uv__dirent_t;
+
+#define UV__DT_DIR  UV_DIRENT_DIR
```
saghul, line 297: Why don't we just use UV_DIRENT_DIR/FILE directly? This internal type doesn't give us much, does it?

include/uv.h
```diff
@@ -1786,6 +1787,16 @@ struct uv_interface_address_s {
   } netmask;
 };
 
+typedef enum {
+  UV_DIRENT_FILE,
+  UV_DIRENT_DIR
+} uv_dirent_type_t;
```
saghul, line 1793: On Unix there are more types, do we want to support them?

include/uv.h
```diff
@@ -1931,6 +1942,15 @@ UV_EXTERN int uv_fs_rmdir(uv_loop_t* loop, uv_fs_t* req, const char* path,
 UV_EXTERN int uv_fs_readdir(uv_loop_t* loop, uv_fs_t* req,
     const char* path, int flags, uv_fs_cb cb);
 
+/*
+ * Call this after `uv_fs_readdir` cb's invocation, this function should be
+ * called until it will return `UV_EOF`.
```
saghul: "until it returns"

include/uv.h
```diff
@@ -1931,6 +1942,15 @@ UV_EXTERN int uv_fs_rmdir(uv_loop_t* loop, uv_fs_t* req, const char* path,
...
+ *
+ * The data that is put into `ent` is managed by the libuv and should not be
```
saghul: "is managed by libuv"

include/uv.h
```diff
@@ -1931,6 +1942,15 @@ UV_EXTERN int uv_fs_rmdir(uv_loop_t* loop, uv_fs_t* req, const char* path,
...
+ * deallocated by the API user.
```
saghul: "deallocated by the user"

src/uv-common.c
```diff
@@ -449,3 +449,43 @@ int uv_fs_event_getpath(uv_fs_event_t* handle, char* buf, size_t* len) {
...
+  for (; req->nbufs < (unsigned int) req->result; req->nbufs++)
+    free(dents[req->nbufs]);
+}
+
+
+int uv_fs_readdir_next(uv_fs_t* req, uv_dirent_t* ent) {
+  uv__dirent_t** dents;
+  uv__dirent_t* dent;
+
+  dents = req->ptr;
```
saghul, line 469: Shrot circuit in case dents is NULL? In case the user called the function again, after it returned EOF.

include/uv-unix.h
saghul, line 161: Ah, I see, because we use it in common. Nevermind.

src/win/fs.c
```diff
@@ -822,71 +825,72 @@ void fs__readdir(uv_fs_t* req) {
...
-      }
+    if (!(name[0] != L'.' || (name[1] && (name[1] != L'.' || name[2]))))
+      continue;
 
-      wcscpy(ptr, name);
-      ptr += len + 1;
-      result++;
+    /* Grow dents buffer, if needed */
+    if (result >= dent_size) {
+      dent_size += UV_FS_DIRENT_SLIDE;
```
saghul: Maybe inline the define here? It looks like a flag, but it isn't...

saghul:
Had a quick look and made some comments, nothing major. Will look at the windows side closer tonight.

Thanks a lot Fedor! Me likey!

include/uv-win.h
indutny, line 297: Because `uv-common.c` is using the values from `dirent.h` on unixes.

include/uv.h
indutny, line 1793: I don't know, what types are you thinking of?

src/uv-common.c
indutny, line 469: You mean `abort()`?

src/win/fs.c
indutny: Well, it is just that I don't want any magic values here.

include/uv.h
txdv, line 1793: http://www.gnu.org/software/libc/manual/html_node/Directory-Entries.html

http://msdn.microsoft.com/en-us/library/windows/desktop/gg258117(v=vs.85).aspx

The only overlapping is DIRECTORY and NOT DIRECTORY.

src/win/fs.c
saghul: I'd go for putting the value there with a comment. It's just used there, and the rest similarly named defines are flags, it's confusing.

src/uv-common.c
saghul, line 469: I would add an assert and also return EOF, so debug builds crash and release builds return the error.

include/uv.h
saghul, line 1793: We might need the unknown type, right?

indutny, line 1793: Not really :) The FILE is our unknown here :P Generally it isn't possible to detect it better.

include/uv.h
txdv: also "the libuv" => "libuv"

src/uv-common.c
indutny, line 469: I just realized that we don't put anything in `req->ptr` if there are 0 entries on unixes. Not sure if it'll work with your assert. Anyway, `UV_EOF` will be returned, even if this code is called multiple times.

indutny:
@saghul just pushed an update, hope I haven't missed anything.

src/uv-common.c
txdv, line 469: UV_EOF seems reasoanble

saghul:
I would short-circuit `uv_fs_readdir_next` in case `req->ptr` is NULL, the processing is unnecessary.

indutny:
Anything else guys?

indutny:
@saghul it won't do any processing, only a `free` call. Anyway it doesn't seem to be much profitable to anyone. Adding another branch is a bad practice, right? ;)

saghul:
@txdv but ptr would be NULL, and free(NULL) is kosher.

I still don't like it much, but it's not a big deal, leave it as is then.

txdv:
@saghul I don't understand which comment you are referencing.

saghul:
@txdv to "Free dents and return ENOMEM to the user!" but I think GitHub rendred it in the wrong place, disregard it.

saghul:
@indutny fs tests pass on Windows, other than returning ENOMEM where possible, LGTM.

indutny:
Landed in ab2c442, thank you!

PR #1423, opened 2014-08-16
backport android build for v0.8

sequoiar:
backport libuv master branch of android build to v0.8 branch

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,12 +68,17 @@ checkout GYP into the project tree manually:
 Unix users run
 
     ./gyp_uv -f make
-    make
+    make -C out
 
 Macintosh users run
 
     ./gyp_uv -f xcode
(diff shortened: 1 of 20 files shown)


saghul:
0.8 is EOLd AFAICT. This is a new feature, so even if we would still provide security fixes, we won't add this to 0.8 at this stage.

/cc @indutny

indutny:
@saghul I don't really like supporting v0.8 and backporting large code chunks to it. However, I should admit that strictly speaking this is not a new feature, it is just a new platform.

cc @tjfontaine

saghul:
@indutny well given that Android wasn't supported on v0.8 I consider it a new "feature", so I'm personally -1 on this.

saghul:
Nobody vouched strongly for this, so closing.

PR #1424, opened 2014-08-16
fix  a typo

cynron:


saghul:
Thanks! Unfortunately we are in the process of deprecating uv.h as a source for documentation, so all that will be removed. We are working on API documentation in this repo: https://github.com/libuv/api-docs, which will be merged into libuv once complete. Feel free to send pull requests there, we could really use some help :-)

txdv:
> we are in the process of deprecating uv.h as a source for documentation

wth?

saghul:
> wth?

There is no point in keeping duplicated information, which will quickly get out of sync (it already is right now) once the API documentation is merged into the main repo. The docs can be built as txt files, man pages, PDF, ePub and many other formats so that will be the one and only source of API documentation, not uv.h.

txdv:
We also could inject them into uv.h!

saghul:
Again, that would imply having the same info in 2 places, which is
redundant and unnecessary.

txdv:
No, you don't understand.

saghul:
Apparently not. Docs will be _on the repo_ already, so I don't see the need
to spend time making a way to attach them to uv.h, sorry.

PR #1428, opened 2014-08-19
Add the throw_safe variable to the uv_fs_s struct

errendir:
This variable is meant to control whether IO errors are created or not.
This enables: https://github.com/joyent/node/pull/8189/

The general idea is to throw less errors that are immediately caught.
This throw and catch behaviour is present in the module.js and can 
significantly slow down the require calls (especially in case of coffeescript).

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -1906,6 +1906,7 @@ struct uv_fs_s {
   uv_fs_type fs_type;
   uv_loop_t* loop;
   uv_fs_cb cb;
+  int throw_safe;
   ssize_t result;
   void* ptr;
   const char* path;


saghul:
Sorry, this is not the right place to add this. libuv is not Node, it's written in C and has no knowledge of what "throwing" is or who should use it. If you need extra fields just in Node you can define a struct as follows:

```
typedef struct node_fs_s {
    uv_fs_t req;
    int throw_safe;
} node_fs_t;
```

This way you can cast the `uv_fs_t` to `node_fs_t` on libuv callbacks, without having to pollute libuv.

txdv:
He just added a variable without using it?

saghul:
@txdv yes, it's supposed to be used by Node, check the link provided in the original message. That's why it has no place in libuv.

Issue #1432, opened 2014-08-19
Write request hangs in a peculiar case when connection fails


neoxic:
If a write request is initiated right after a call to `uv_connect` but before `uv_connect_cb` gets fired, it lingers indefinitely until the handle is closed.

Please consider the following test case:

``` c
#include <assert.h>
#include <stdio.h>
#include <uv.h>


uv_loop_t loop;
uv_tcp_t tcp_server;
uv_tcp_t tcp_client;
uv_connect_t crq;
uv_write_t wrq1, wrq2;
uv_buf_t buf = { "HELLO", 4 };


void connection_cb(uv_stream_t *server, int status) {
    printf("connection: %d\n", status);
}

void write_cb(uv_write_t *req, int status) {
    printf("write: %d\n", status);
}

void connect_cb(uv_connect_t *req, int status) {
    printf("connect: %d\n", status);
    //assert(uv_write(&wrq2, (uv_stream_t *)&tcp_client, &buf, 1, write_cb) == 0); /* UNCOMMENT ME */
}

int main() {
    struct sockaddr_in sa;
    assert(uv_ip4_addr("127.0.0.1", 3333, &sa) == 0);
    assert(uv_loop_init(&loop) == 0);
    assert(uv_tcp_init(&loop, &tcp_server) == 0);
    assert(uv_tcp_init(&loop, &tcp_client) == 0);

    //assert(uv_tcp_bind(&tcp_server, (const struct sockaddr *)&sa, 0) == 0); /* UNCOMMENT ME */
    //assert(uv_listen((uv_stream_t *)&tcp_server, 128, connection_cb) == 0); /* UNCOMMENT ME */

    assert(uv_tcp_connect(&crq, &tcp_client, (const struct sockaddr *)&sa, connect_cb) == 0);
    assert(uv_write(&wrq1, (uv_stream_t *)&tcp_client, &buf, 1, write_cb) == 0);

    uv_run(&loop, UV_RUN_DEFAULT);
    return 0;
}
```

The output is:

```
connect: -61
^C
```

The loop neither terminates nor executes the first write request not fails it whereas the connection request has failed.

If the "UNCOMMENT ME" lines get uncommented, the output is:

```
connect: 0
connection: 0
write: 0
write: 0
^C
```

Both writes succeed which is expected and looks fine to me.

saghul:
[...]

PR #1433, opened 2014-08-21
Tcp write after connect

txdv:
Should address #1432

### DIFF
--- a/src/unix/stream.c
+++ b/src/unix/stream.c
@@ -361,10 +361,22 @@ int uv__stream_open(uv_stream_t* stream, int fd, int flags) {
 }
 
 
-void uv__stream_destroy(uv_stream_t* stream) {
+void uv__stream_flush_write_queue(uv_stream_t* stream, int error) {
   uv_write_t* req;
   QUEUE* q;
+  while (!QUEUE_EMPTY(&stream->write_queue)) {
+    q = QUEUE_HEAD(&stream->write_queue);
+    QUEUE_REMOVE(q);
 
+    req = QUEUE_DATA(q, uv_write_t, queue);
+    req->error = error;
+
+    QUEUE_INSERT_TAIL(&stream->write_completed_queue, &req->queue);
+  }
+}
+
+
+void uv__stream_destroy(uv_stream_t* stream) {
   assert(!uv__io_active(&stream->io_watcher, UV__POLLIN | UV__POLLOUT));
   assert(stream->flags & UV_CLOSED);
 
@@ -374,16 +386,7 @@ void uv__stream_destroy(uv_stream_t* stream) {
     stream->connect_req = NULL;
   }
 
-  while (!QUEUE_EMPTY(&stream->write_queue)) {
-    q = QUEUE_HEAD(&stream->write_queue);
-    QUEUE_REMOVE(q);
-
-    req = QUEUE_DATA(q, uv_write_t, queue);
-    req->error = -ECANCELED;
-
-    QUEUE_INSERT_TAIL(&stream->write_completed_queue, &req->queue);
-  }
-
+  uv__stream_flush_write_queue(stream, -ECANCELED);
   uv__write_callbacks(stream);
 
   if (stream->shutdown_req) {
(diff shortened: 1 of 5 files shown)


src/unix/stream.c
```diff
@@ -1232,10 +1238,15 @@ static void uv__stream_connect(uv_stream_t* stream) {
 
   stream->connect_req = NULL;
   uv__req_unregister(stream->loop, req);
-  uv__io_stop(stream->loop, &stream->io_watcher, UV__POLLOUT);
 
   if (req->cb)
     req->cb(req, error);
+
+  if (error < 0) {
```
saghul, line 1249: You need to check if the fd is -1. If it is, the user has closed the handle and there is nothing to be done because the write reqs will be called with ECANCELED already.

Also, in case there was no error, if the write queue is empty you need to stop watching for POLLOUT.

src/unix/stream.c
```diff
@@ -361,6 +361,21 @@ int uv__stream_open(uv_stream_t* stream, int fd, int flags) {
 }
 
 
+void uv__stream_flush_queue(uv_stream_t* stream, int error) {
```
saghul: call this `uv__stream_flush_write_queue`

txdv:
Looks like I addressed your comments.

test/test-tcp-write-after-connect.c
```diff
@@ -0,0 +1,56 @@
...
+
+uv_loop_t loop;
+uv_tcp_t tcp_client;
+uv_connect_t connection_request;
+uv_write_t write_request;
+uv_buf_t buf = { "HELLO", 4 };
+
+
+static void write_cb(uv_write_t *req, int status) {
+  assert(status == UV_ECANCELED);
```
saghul: ASSERT.

test/test-tcp-write-after-connect.c
```diff
@@ -0,0 +1,56 @@
...
+  struct sockaddr_in sa;
+  assert(uv_ip4_addr("127.0.0.1", TEST_PORT, &sa) == 0);
+  assert(uv_loop_init(&loop) == 0);
+  assert(uv_tcp_init(&loop, &tcp_client) == 0);
+
+  assert(uv_tcp_connect(&connection_request, &tcp_client, (const struct sockaddr *)&sa, connect_cb) == 0);
+  assert(uv_write(&write_request, (uv_stream_t *)&tcp_client, &buf, 1, write_cb) == 0);
+
+  uv_run(&loop, UV_RUN_DEFAULT);
+#endif
```
saghul: USe MAVE_VALGRIND_HAPPY.

test/test-tcp-write-after-connect.c
```diff
@@ -0,0 +1,56 @@
...
+}
+
+TEST_IMPL(tcp_write_after_connect) {
+#ifdef _WIN32
+  struct sockaddr_in sa;
+  assert(uv_ip4_addr("127.0.0.1", TEST_PORT, &sa) == 0);
+  assert(uv_loop_init(&loop) == 0);
+  assert(uv_tcp_init(&loop, &tcp_client) == 0);
+
+  assert(uv_tcp_connect(&connection_request, &tcp_client, (const struct sockaddr *)&sa, connect_cb) == 0);
```
saghul: long line?

test/test-tcp-write-after-connect.c
saghul: Actually, on the entire test s/assert/ASSERT/g

uv.gyp
```diff
@@ -378,6 +378,7 @@
         'test/test-tcp-connect6-error.c',
         'test/test-tcp-open.c',
         'test/test-tcp-write-to-half-open-connection.c',
+        'test/test-tcp-writer-after-connect.c',
```
saghul: typo: write

test/test-tcp-write-after-connect.c
saghul: Close the handle here.

txdv:
addressed them again

saghul:
Looks good! Here is Jenkins doing his job: http://jenkins.nodejs.org/job/libuv-pullrequest/646/

src/unix/stream.c
```diff
@@ -374,16 +389,7 @@ void uv__stream_destroy(uv_stream_t* stream) {
     stream->connect_req = NULL;
   }
 
-  while (!QUEUE_EMPTY(&stream->write_queue)) {
```
saghul, line 377: After removing these q and req are unused.

saghul:
3 tests fail after this change: shutdown_twice, tcp_ref3 and pipe_ref3. There must be something we are missing...

txdv:
That last commit fixes it.

txdv:
It becomes kinda clear, we used to stop the IO before the callback, now we do it after. I think this is the difference.

saghul:
That's not a solution. If connect suceeds but there are no write requests pending we'd still have the fd armed watching for POLLOUT, but we shouln't.

Maybe we better stop it, call the callback and rearm if necessary.

txdv:
> It becomes kinda clear, we used to stop the IO before the callback, now we do it after. I think this is the difference.

This was the bug.

saghul:
http://jenkins.nodejs.org/job/libuv-pullrequest/647/

saghul:
Tests pass now :+1: Can you please squash?

txdv:
squashed

saghul:
Thanks Andrius! Landed! :cake:

PR #1435, opened 2014-08-22
unix, windows: add uv_fileno

saghul:
Returns the platform specific file descriptor for handles that are
backed by one. The datatype is abstracted as uv_os_fd_t, which maps to
int on Unices and HANDLE on Windows.

Users can use this function to set specific socket options, for example,
in a non portable way.

This function is essentially a shotgun, you better be careful with
whatever you do with it, don't blame me if you used it to get the fd of
a stream, close it yourself and expect things to Just Work.

### DIFF
--- a/include/uv.h
+++ b/include/uv.h
@@ -615,6 +615,8 @@ UV_EXTERN void uv_close(uv_handle_t* handle, uv_close_cb close_cb);
  */
 UV_EXTERN int uv_recv_buffer_size(uv_handle_t* handle, int* value);
 
+UV_EXTERN int uv_fileno(const uv_handle_t* handle, uv_os_fd_t* fd);
+
 /*
  * Returns or sets the size of the send buffer that the operating
  * system uses for the socket.
--- a/src/unix/core.c
+++ b/src/unix/core.c
@@ -662,6 +662,36 @@ void uv_disable_stdio_inheritance(void) {
 }
 
 
+int uv_fileno(const uv_handle_t* handle, uv_os_fd_t* fd) {
+  int fd_out;
+
+  switch (handle->type) {
+  case UV_TCP:
+  case UV_NAMED_PIPE:
+  case UV_TTY:
+    fd_out = uv__stream_fd((uv_stream_t*) handle);
+    break;
+
+  case UV_UDP:
+    fd_out = ((uv_udp_t *) handle)->io_watcher.fd;
+    break;
+
+  case UV_POLL:
+    fd_out = ((uv_poll_t *) handle)->io_watcher.fd;
+    break;
+
+  default:
+    return -EINVAL;
+  }
+
+  if (uv__is_closing(handle) || fd_out == -1)
+    return -EBADF;
+
+  *fd = fd_out;
+  return 0;
+}
+
+
 static void uv__run_pending(uv_loop_t* loop) {
   QUEUE* q;
   uv__io_t* w;
--- a/test/test-handle-fileno.c
+++ b/test/test-handle-fileno.c
@@ -1,0 +1,120 @@
+/* Copyright Joyent, Inc. and other Node contributors. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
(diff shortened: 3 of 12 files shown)


saghul:
This patch is unifinished (Windows part and tests are missing), but I wanted to start a discussion about it. I can't think of a way which would please everyone while keeping our fds hidden, so I think it might not be that bad to expose them for those with legit use cases such as setting some socket options here and there, we can't possibly map all of them.

About the Windows side, HANDLE is just a void*, which means we wouldn't be able to cast it to int because of 64bit systems. However, reading around seems like only the lower 32 bits are used because a process can only have 2^24 handles in that handle table. Anyway, I went for the typedef, since this is a platform dependent thing and feels more future-proof.

Thoughts?

Calling the cavalry! @indutny, @bnoordhuis, @piscisaureus

txdv:
IMO this is a good idea.

Let the user do whatever he wants.

txdv:
This is such an easy fix, why didn't we have it earlier?

saghul:
Just like with scotch, some things take time to become better, I guess :-P

include/uv.h
```diff
@@ -615,6 +615,9 @@ UV_EXTERN void uv_close(uv_handle_t* handle, uv_close_cb close_cb);
  */
 UV_EXTERN int uv_recv_buffer_size(uv_handle_t* handle, int* value);
 
+/* TODO */
+UV_EXTERN uv_os_fd_t uv_fileno(uv_handle_t* handle);
```
bnoordhuis: Maybe `UV_EXTERN int uv_file(const uv_handle_t* handle, uv_os_fd_t* fd);`?  Libuv can't really return a meaningful error now.

src/unix/core.c
```diff
@@ -662,6 +662,25 @@ void uv_disable_stdio_inheritance(void) {
 }
 
 
+uv_os_fd_t uv_fileno(uv_handle_t *handle) {
+  switch (handle->type) {
+  case UV_TCP:
+  case UV_NAMED_PIPE:
+  case UV_TTY:
+    return ((uv_stream_t *) handle)->io_watcher.fd;
```
bnoordhuis: I believe that on OS X, io_watcher.fd is not always the actual file descriptor.  Also, you have a style error a few lines up.  :-)

include/uv.h
saghul: That crossed my mind, but in the end I thought that I'd just return -1 when there is no fd. We could give EINVAL if you pass an unsupported handle type, I guess.

src/unix/core.c
saghul: I vaguely remember that, I'll check it out, thanks!. Based on your comments, I'll assume you are on board with this then :-)

include/uv.h
bnoordhuis: But how are you going to do that on Windows?  HANDLE is a typedef for `void*`.

saghul: I guess the same way Windows does for INVALID_SOCKET :-P On a more serious note, I do like your proposal, I'll change this.

saghul: FWIW, Windows defines INVALID_SOCKET as SOCKET(~0).

saghul:
Updated the API and Unix implementation as per Ben's suggestions. It returns EINVAL if the handle is not a stream, poll or udp and EBADF if the fd is -1.

src/unix/core.c
```diff
@@ -662,6 +662,32 @@ void uv_disable_stdio_inheritance(void) {
...
+  case UV_POLL:
+    *fd = ((uv_poll_t *) handle)->io_watcher.fd;
+    break;
+
+  default:
+    return -EINVAL;
+  }
+
+  if (*fd == -1)
+    return -EBADF;
```
bnoordhuis, line 688: It's kind of bad style to return an error code but still mutate the out parameter, IMO.

saghul, line 688: True that, will fix.

saghul:
Added tests and Windows implementation. The TTY test fails, I need to open $conin it seems, like the TTY test does, will do.

test/test-handle-fileno.c
```diff
@@ -0,0 +1,97 @@
...
+  ASSERT(r == 0);
+  r = uv_fileno((uv_handle_t*) &pipe, &fd);
+  ASSERT(r == UV_EBADF);
+  r = uv_pipe_bind(&pipe, TEST_PIPENAME);
+  ASSERT(r == 0);
+  r = uv_fileno((uv_handle_t*) &pipe, &fd);
+  ASSERT(r == 0);
+  uv_close((uv_handle_t*) &pipe, NULL);
+#ifndef _WIN32
+  /* close is not immediate on Windows */
```
bnoordhuis: Can't you filter on UV__HANDLE_CLOSING in uv_fileno()?

include/uv.h
```diff
@@ -615,6 +615,8 @@ UV_EXTERN void uv_close(uv_handle_t* handle, uv_close_cb close_cb);
  */
 UV_EXTERN int uv_recv_buffer_size(uv_handle_t* handle, int* value);
 
+UV_EXTERN int uv_fileno(uv_handle_t* handle, uv_os_fd_t* fd);
```
bnoordhuis: I think handle can be const?

test/test-handle-fileno.c
saghul: I'm not sure what guarantees we make about it. The handle (libuv wise) is closing, but the actual Windows HANDLE might need to stick around for a bit more.

include/uv.h
saghul: I think so, yes. Good call!

saghul:
Tests pass on Windows now.

saghul:
@bnoordhuis thanks, const-ified the handle now.

saghul:
Any further comments on this?

txdv:
Why do we set it to invalid?

saghul:
> Why do we set it to invalid?

Because otherwise you get it with uv_fileno after you have closed it so it might belong to another handle that reused it. This wasn't a problem before, because it was not exposed, but now that people get access to file descriptors, we cannot give them stalled ones.

Another possibility might be to return EBADF if uv_is_closing(handle). That might actually simplify things a bit...

saghul:
I'll give that a shot (I think that's what @bnoordhuis said before, but I didn't quite get it until now). I'd still leave the commit that sets handles to invalid, as a defense mechanism.

saghul:
Updated with uv_is_closing check, now we no longer need any ifdef-ery on the tests \o/.

bnoordhuis:
LGTM.  And yes, that was what I meant. :-)

saghul:
Thanks folks, landed! Load your shotguns.

PR #1436, opened 2014-08-23
windows: ensure fs_event always provides filename

nitoyon:
`uv_fs_event` didn't provide filename argument on Windows when a file is deleted or renamed, although it does on Linux.

Fix this by ensuring that fs_event always provides filename argument on Windows.

---

Node.js API document says:

> Providing `filename` argument in the callback is not supported on every platform (currently it's only supported on Linux and Windows). Even on supported platforms `filename` is not always guaranteed to be provided.
> 
> http://nodejs.org/docs/latest/api/fs.html#fs_filename_argument

But, it's better that we provide `filename` argument if we can, isn't it?
# How to reproduce this bug

Use following script:

``` javascript
var fs = require('fs');

var monitor = fs.watch('./', function(changes, fn) {
  console.log("%s :  %s", changes, fn);
});

fs.writeFileSync('foo.txt', 'new contents');
fs.renameSync('foo.txt', 'bar.txt');
fs.unlinkSync('bar.txt');

setTimeout(function() {
  monitor.close();
}, 1000);
```
## Linux

On CentOS 6.4 x64 (Node.js v0.10.29)

``` console
$ node test.js
rename :  foo.txt  # writeFile
change :  foo.txt  # writeFile
rename :  foo.txt  # rename (src)
rename :  bar.txt  # rename (dst)
rename :  bar.txt  # unlink
```

On Windows 8.1 x64 (Node.j v0.10.21)

``` console
> node test.js
rename :  foo.txt  # writeFile
change :  foo.txt  # writeFile
rename :  null     # rename (src) <--- FIX THIS
rename :  bar.txt  # rename (dst)
rename :  null     # unlink       <--- FIX THIS
```

### DIFF
--- a/src/win/fs-event.c
+++ b/src/win/fs-event.c
@@ -369,7 +369,8 @@ void uv_process_fs_event_req(uv_loop_t* loop, uv_req_t* req,
             /*
              * We attempt to convert the file name to its long form for
              * events that still point to valid files on disk.
-             * For removed and renamed events, we do not provide the file name.
+             * For removed and renamed from events, we do not provide the
+             * long file name.
              */
             if (file_info->Action != FILE_ACTION_REMOVED &&
               file_info->Action != FILE_ACTION_RENAMED_OLD_NAME) {
@@ -420,18 +421,15 @@ void uv_process_fs_event_req(uv_loop_t* loop, uv_req_t* req,
                   long_filenamew = NULL;
                 }
               }
+            }
 
-              /*
-               * If we couldn't get the long name - just use the name
-               * provided by ReadDirectoryChangesW.
-               */
-              if (!long_filenamew) {
-                filenamew = file_info->FileName;
-                sizew = file_info->FileNameLength / sizeof(WCHAR);
-              }
-            } else {
-              /* Removed or renamed callbacks don't provide filename. */
-              filenamew = NULL;
+            /*
+             * If we couldn't get the long name - just use the name
+             * provided by ReadDirectoryChangesW.
+             */
+            if (!long_filenamew) {
+              filenamew = file_info->FileName;
+              sizew = file_info->FileNameLength / sizeof(WCHAR);
             }
           } else {
             /* We already have the long name of the file, so just use it. */


saghul:
@nitoyon is this related to any issue open in Node? I remember something similar to this being discussed there.

nitoyon:
@saghul this PR fixes joyent/node#8372 and joyent/libuv#1479.

saghul:
@nitoyon thanks for the heads up. Could you please add a test, or modify an existing one, which validates the fix?

piscisaureus:
It's not that I necessarily disagree.
But this feels like a behavioral change dressed up as a bug fix.
Also if we are going to give the user short filenames, we should at least be consistent about it and also do it for the create and rename-newname events.

saghul:
Thanks for chiming in @piscisaureus :-) On top of that, we also need to be consistent with what we do in other platforms (which I don't remember from the top of my head).

nitoyon:
@saghul It looks hard to add a test because there is no test which assert that `filename` should be given (e.g. `ASSERT(filename == NULL || strcmp(filename, "file1") == 0);` in [test-fs-event.c L117](https://github.com/joyent/libuv/blob/v1.0.0-rc1/test/test-fs-event.c#L117)).

And the [document for fs.watch() in Node.js](http://nodejs.org/api/fs.html#fs_filename_argument) says:

> Providing `filename` argument in the callback is not supported on every platform (currently it's only supported on Linux and Windows). Even on supported platforms `filename` is not always guaranteed to be provided. 

It would break unit test if we add a test which asserts `filename` should be given.

@piscisaureus We have already provided short filenames if `Action` isn't `REMOVE` or `RENAMED_OLD_NAME`, and `GetLongPathNameW` fails. For details, please read [my table](https://github.com/joyent/libuv/pull/1436#commitcomment-7503088).

jorangreef:
fs.watch is different to most Node functionality in that it does not promise to be the same cross-platform.

But this promise is still very helpful. It makes it much easier to watch for filesystem events without resorting to any native dependencies. And for anyone using fs.watch, they immediately need to start out by considering cross-platform differences anyway. They are probably going to already be aware of things like 8.3 short names and unicode and casing differences.

So it would be helpful to me if fs.watch gave as much information as possible according to each platform. In this specific case, even the 8.3 short name would be very useful. And if the callback could always provide the 8.3 short name as extra info on Windows for the other events that would be even better.

PR #1438, opened 2014-08-24
build: use same CFLAGS in autotools build as in gyp

### LINKED ISSUE #1362: unused variables warning
How about we turn on the unused variable flag for the C compiler.

I noticed that when reviewing my patch, saghul found some unused variables. Should be the compilers jobs to check for those little buggers.

I bet it would reveal a lot of those unused variables.

saghul:
Closes #1362


txdv:
Does return a warning now, looks good to me.

saghul:
Landed, thanks for checking it out Andrius.

Issue #1426, opened 2014-08-18
UDP memory leaking when sending to inactive local port on Windows


k0dek0mmand0:
I'm using libuv version shipped with node 0.10.30 on Windows platform.

I think there's a problem with the UDP implementation, similar to the one discussed in issue #1194

If a datagram is sent from 127.0.0.1:10000 to 127.0.0.1:12345 and no-one is listening on port 12345, an alloc callback is triggered on the sending handle, but the allocated memory is never used anywhere (no read callback comes with this pointer).

I'm assuming an ICMP error message (destination unreachable?) is being received, but not handled.

If I add a listener on port 12345 (e.g. ffplay), everything works as advertised.

Here's some c++ code to reproduce the issue. In the output on my machine I see 10x SENT (status 0), followed by 10x ALLOC, then nothing else. Startup code not included.

``` c++
void _readCb(uv_udp_t*, ssize_t, uv_buf_t buf, struct sockaddr*, unsigned)
{
   printf( "READ %p\n", buf.base );
   free( buf.base );
}

uv_buf_t _allocCb(uv_handle_t*, size_t size)
{
   char* ptr = (char*)malloc( size );
   //
   printf( "ALLOC %p\n", ptr );
   //
   return uv_buf_init( ptr, (unsigned)size );
}

void _sendCb(uv_udp_send_t* req, int)
{
   printf( "SENT %p\n", req->data );
   free( req->data );
   free( req );
}

void testCase()
{
   uv_udp_t* h = (uv_udp_t*)malloc( sizeof( uv_udp_t ) );
   //
   sockaddr_in addrSrc = uv_ip4_addr( "127.0.0.1", 10000 );
   sockaddr_in addrDst = uv_ip4_addr( "127.0.0.1", 12345 );
   //
   uv_udp_init( uv_default_loop(), h );
   uv_udp_bind( h, addrSrc, 0 );
   uv_udp_recv_start( h, _allocCb, _readCb );
   //
   for( int i = 0; i < 10; i++ )
   {
      uv_buf_t buf = uv_buf_init( (char*)malloc( 10 ), 10 );
      //
      uv_udp_send_t *req = (uv_udp_send_t*)malloc( sizeof( uv_udp_send_t ) );
      //
      req->data = buf.base;
      //
      uv_udp_send( req, h, &buf, 1, addrDst, _sendCb );
   }
}
```

saghul:
Can you make this a proper libuv test case?

[...]

PR #1441, opened 2014-08-27
windows: fix buffer leak after failed udp send

piscisaureus:
Ref: #1426


k0dek0mmand0:
Other than a missing copyright notice at the top of `test-udp-send-unreachable.c` I can't see any problems. I've run it and the test case passes.
I'd personally not remove from the test case the comment about a 'missing' server, but it's clear from the context what's being tested.
I'd also not remove `ASSERT(timer_cb_called == 1);` after `uv_run`.

Thanks for the fix.

saghul:
@k0dek0mmand0 thanks! In the begining I thought we could get away without the timer, but it turns out it's not the case, but I forgot to re-add all the code. Will do.

saghul:
Amended and force pushed.

saghul:
Thanks guys, landed on master!

PR #1446, opened 2014-09-01
unix: Prefer UV_EADDRINFO to UV_UNKNOWN as uv_getaddrinfo error code

rolandyoung:
Commit dd893814a assumes that errno will always be successfully interpreted
by uv__set_sys_error, but in some cases loop->last_err.code now gets set to
UV_UNKNOWN.
In such cases, this change restores the previous default code, UV_EADDRINFO.

### DIFF
--- a/src/unix/getaddrinfo.c
+++ b/src/unix/getaddrinfo.c
@@ -72,6 +72,8 @@ static void uv__getaddrinfo_done(struct uv__work* w, int status) {
   if (req->retcode < 0) {
     /* EAI_SYSTEM error */
     uv__set_sys_error(req->loop, -req->retcode);
+    if (req->loop->last_err.code == UV_UNKNOWN)
+      req->loop->last_err.code = UV_EADDRINFO;
   } else if (req->retcode == 0) {
     /* OK */
 #if defined(EAI_NODATA) /* FreeBSD deprecated EAI_NODATA */


saghul:
What error are you getting, which is not mapped? We should have a mapping for it, not workaround it.

rolandyoung:
In my test scenario, errno was 8, ENOEXEC. This was a test of error reporting in our application; the scenario involved calling uv_getaddrinfo with host = "localhost" and service = "xyzzy". I have no idea why ENOEXEC would be the return code in that case and I'm not sure that getting a more specific error than UV_EADDRINFO is important in this case.

Although I'm a little surprised that there's no mapping in uv.h for ENOEXEC, I imagine that the reason that EAI_SYSTEM exists as a return code for getaddrinfo is that it is very tedious and not necessarily helpful to enumerate all the possible return codes on all platforms. I could have fixed the behaviour for this test case on this platform (Linux) by adding a mapping to uv.h, but I thought I was likely to run into a similar issue with a different unmapped error on one of the other platforms we target.

Given that there was already an 'else' clause to catch unanticipated return codes from getaddrinfo: it seemed reasonable that unanticipated errno values should be treated similarly. However I can see that the revised API in the master branch will make this obsolete. Maybe I should code the fallback in the code that calls libuv.

Summary: ENOEXEC ought to have a mapping, but that's a separate issue. My patch is the best for my use case, I considered it was general enough to submit it back to the project, but I don't mind if you would rather not take it.

saghul:
> Summary: ENOEXEC ought to have a mapping, but that's a separate issue. My patch is the best for my use case, I considered it was general enough to submit it back to the project, but I don't mind if you would rather not take it.

It's actually the other way around :-) Mapping ENOEXEC is the proper fix for this. When getaddrinfo return EAI_SYSTEM the actual errno is in 'errno', that's the one we need to map. UV_EADDRINFO means nothing, it's not even a valid POSIX errno.

Here is how you map an errno in the v0.10 branch: https://github.com/joyent/libuv/commit/6f98f4efd112b21414ed92a048cda9f6f5978eb7

Now, I see we don't map it on master either, but if I try to do a getaddrinfo to a non-existing service I get UV_EAI_SERVICE.

rolandyoung:
UV_EAI_SERVICE is what I would have expected. But I have ceased being surprised at the behaviour of getaddrinfo. The fact that the underlying errno is ENOEXEC on my test machine for this case illustrates my point that you could get just about anything back.

Our code (yet another Lua binding for libuv - we hope to publish it in due course but have to get agreement from our employer) can't report 'unknown error' in this or any as yet undiscovered case of a missing code mapping. Up to v.0.10.25, we could rely on 'getaddrinfo error' at worst, I can work around this in our code and will have to in v1.x, but my proposed patch seemed useful enough to submit.

I will see if I have time to add the missing mapping to uv.h in our project (doing so correctly involves three repos and some subtree stuff) and regression test it in our CI. Otherwise, I will have to leave this as it is for now.

saghul:
> I will see if I have time to add the missing mapping to uv.h in our project (doing so correctly involves three repos and some subtree stuff) and regression test it in our CI. Otherwise, I will have to leave this as it is for now.

This fix could go into v0.10, as other missing errno mappings have, and we don't really need a regression test here, it's just a missing errno mapping.

saghul:
I'm closing this for the reasons mentioned above. Feel free to open a PR mapping ENOEXEC.

PR #1448, opened 2014-09-01
darwin: do not block in uv_setproctitle

indutny:
@saghul: this is the most awful thing that I ever did in `darwin` parts, but it should fix https://github.com/joyent/node/issues/8288#issuecomment-54038095


indutny:
Nah, this is a terrible hack.

saghul:
Wow, I don't have the slightest clue what that code does :-S

PR #1455, opened 2014-09-04
Don't allow NULL callback on timers

saghul:


saghul:
Quick review @indutny? Also @trevnorris, since discussing with him on IRC lead to discovering this.

trevnorris:
Not sure about the windows timer simplification (e.g. why was it was so complicated before, and why you could simplify it so much). I can spin up my win vm and test it, but I'm sure you've already done that.

Not being allowed to pass `NULL` to `uv_timer_start()` LGTM.

saghul:
> Not sure about the windows timer simplification (e.g. why was it was so complicated before, and why you could simplify it so much). I can spin up my win vm and test it, but I'm sure you've already done that.

Basically the code removed is already tesed for in the functions we call now, that's why I could simplify.

trevnorris:
Awesome. Then LGTM.

txdv:
I feel like the first commit message could have a better description.

saghul:
> I feel like the first commit message could have a better description.

What do you suggest? I didn't really think about it much since it doesn't change functionality. How about "windows: remove duplicated code in uv_timer_again"?

saghul:
Updated commit message, @txdv.

txdv:
well done

saghul:
Thanks for double-checking! Landed on v1.x.

PR #1457, opened 2014-09-05
windows: further simplify the code for timers

saghul:
- Remove the UV_HANDLE_ACTIVE flag. It's a duplicate from
  UV__HANDLE_ACTIVE, which was used solely on timers and loop watchers.
- Avoid duplicated code when running timers by stopping the handle and
  rearming it with the repeat time, thus having a single place where the
  timers are added and removed to and from the RB tree, respectively.


saghul:
Change is pretty straighforward and tests pass, anyone want to review?

saghul:
Anyone?

PR #1462, opened 2014-09-07
stream: allocate enough space for select() hack

indutny:
`fd_set`s are way too small for `select()` hack when stream's fd is
bigger than 1023. Make `fd_set`s a part of `uv__stream_select_t`
structure.

Should be backported to all branches up to (and including) v0.10


indutny:
cc @saghul

src/unix/stream.c
```diff
@@ -318,26 +341,36 @@ int uv__stream_try_select(uv_stream_t* stream, int* fd) {
...
+failed_async_sem_init:
   uv_sem_destroy(&s->close_sem);
-fatal1:
+
+failed_close_sem_init:
+  uv__close(fds[0]);
+  uv__close(fds[1]);
   uv_close((uv_handle_t*) &s->async, uv__stream_osx_cb_close);
-  return -errno;
+  return err;
```
saghul, line 362: 's' needs to be freed here as well.

src/unix/stream.c
```diff
@@ -318,26 +341,36 @@ int uv__stream_try_select(uv_stream_t* stream, int* fd) {
...
+
+failed_async_init:
+  free(s);
```
saghul, line 365: s is already bein freed before you jump here.

src/unix/stream.c
```diff
@@ -283,32 +288,50 @@ int uv__stream_try_select(uv_stream_t* stream, int* fd) {
...
+    max_fd = fds[1];
+
+  sread_sz = (max_fd + NBBY) / NBBY;
+  swrite_sz = sread_sz;
+
   /* At this point we definitely know that this fd won't work with kqueue */
-  s = malloc(sizeof(*s));
-  if (s == NULL)
-    return -ENOMEM;
+  s = malloc(sizeof(*s) + sread_sz + swrite_sz);
```
saghul, line 306: Mode the comment above this line to the right place, that is, before creating the socketpair.

saghul:
Left some comments. As for the commit log, shouldn't it be "darwin:"? This hack doesn't really apply to other platforms.

Test passes here, so I think we got this :-)

saghul:
Note: we need to merge this into v1.x, not master.

src/unix/stream.c
indutny, line 362: It is freed in `uv__stream_osx_cb_close`, right?

saghul, line 362: Ah, missed that, you're right.

indutny:
Landed in 145b76b894c6ed6406d7d3d47abedea5b29d942c, thank you!

saghul:
Awesome!

indutny:
Backported to v0.10 in c60d6af

PR #1463, opened 2014-09-08
Properly cleanup  Windows loops

saghul:
- Close peer sockets also for the default loop
- Close the IOCP handle
- Reset the default loop pointer when it's closed

Quick review @piscisaureus ?


saghul:
I feel good about this one so I went ahead and landed it. Tests pass and all :-)

PR #1464, opened 2014-09-09
Add uv_try_write on windows

mattn:


saghul:
Thanks @mattn! We need to wait for https://github.com/joyent/libuv/pull/1439 in order to have proper write queuing support on Windows. `uv_try_write` must fail if there are ourstanding write requests.

src/win/stream.c
```diff
@@ -184,8 +184,13 @@ int uv_write2(uv_write_t* req,
...
                  const uv_buf_t bufs[],
                  unsigned int nbufs) {
-  /* NOTE: Won't work with overlapped writes */
-  return UV_ENOSYS;
+  switch (stream->type) {
+    case UV_TCP:
+      return uv_tcp_try_write((uv_tcp_t*) stream, bufs, nbufs);
+      break;
+    default:
+      assert(0);
```
indutny: Let's do `UV_ENOSYS` trick here, not just fail :)

src/win/tcp.c
```diff
@@ -876,6 +876,22 @@ int uv_tcp_write(uv_loop_t* loop,
...
+  DWORD bytes = 0;
+  if (WSASend(((uv_tcp_t*) handle)->socket,
+               (WSABUF*) bufs,
+               nbufs,
+               &bytes,
+               0,
+               NULL,
+               NULL) == 0)
+    return (int) bytes;
+  return uv_translate_sys_error(WSAGetLastError());
```
indutny, line 891: I think it should be returning some sort of EAGAIN if `WSAGetLastError()` is `WSAEWOULDBLOCK`.

saghul, line 891: Not only, we also need to return it if there are other write requests pending. That's why I we need to wait for #1439.

src/win/stream.c
mattn: I don't know what is the spec.

See https://github.com/joyent/libuv/blob/master/src/win/stream.c#L42

saghul: There is no written spec per se, but we try to not fail so abuptly. Please return an error.

mattn: Done c20bb226e8b72f9150e5de69504bddf71aac52cd

src/win/tcp.c
piscisaureus, line 891: > I think it should be returning some sort of EAGAIN if WSAGetLastError() is WSAEWOULDBLOCK.

Does uv_translate_sys_error not do this?

> Not only, we also need to return it if there are other write requests pending. That's why I we need to wait for #1439.

It would be possible to just look at the `write_reqs_pending` field and check if there are pending overlapped requests. That seems like a simple addition to this patch which would make it correct.

https://github.com/joyent/libuv/blob/49cb40c47902dd04c7787e6543e9e13368dc1840/include/uv-win.h#L412

txdv, line 891: > It would be possible to just look at the write_reqs_pending

Just like we do on linux.

PR #1465, opened 2014-09-09
On Windows, don't indicate that the pipe is invalid when it is read only.

devinsmith:
....

### DIFF
--- a/src/win/pipe.c
+++ b/src/win/pipe.c
@@ -1883,8 +1883,8 @@ int uv_pipe_open(uv_pipe_t* pipe, uv_file file) {
     return UV_EINVAL;
 
   if (pipe->ipc) {
-    if (!(access.AccessFlags & FILE_WRITE_DATA) ||
-        !(access.AccessFlags & FILE_READ_DATA)) {
+    if (!((access.AccessFlags & FILE_WRITE_DATA) ||
+          (access.AccessFlags & FILE_READ_DATA))) {
       return UV_EINVAL;
     }
   }


saghul:
IPC pipes are meant to send handles over to another process. If an IPC pipe is not writable it doesn't serve much purpose, does it? What usecase / problem did you run into, which motivated this patch?

devinsmith:
In the parent process, the pipe is writable, but when spawning my child process I've only set the pipe to be READABLE via UV_READABLE_PIPE.

I think, in addition, if you look down in the code libuv also checks the access flags again to properly set the duplex_flags. If the requirement is now to have readable and writable pipes, then we could just set the duplex_flags back to UV_HANDLE_READABLE | UV_HANDLE_WRITABLE like it was previously (ebafb90acd6de8f2bf766ac58dc1091ddfacfad4).

saghul:
So, you created a pipe and initialized it with `uv_pipe_init` by setting ipc to 1 and the you tried to share it by using `UV_INHERIT_STREAM`?

Can you share a test case perhaps?

devinsmith:
See below for parent.c and child.c  (these should compile to parent.exe and child.exe)  (parent.c will look for child.exe in the same directory)

Please also note the WORKAROUND in parent.c which creates a READ/WRITE pipe instead of a READ only pipe for the child.

Parent creates a tcp server listening on port 8000 that you can telnet to to test on win32.

--- parent.c CUT ------

``` c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <uv.h>

uv_loop_t *loop; /* Main event loop */

int round_robin_counter;
int child_worker_count;

struct child_worker {
    uv_process_t req;
    uv_process_options_t options;
    uv_pipe_t pipe;
} *workers;

static void
close_process_handle(uv_process_t *req, int64_t exit_status, int term_signal)
{
    fprintf(stderr, "Process exited with status %d, signal %d\n",
        (int)exit_status, term_signal);
    uv_close((uv_handle_t *)req, NULL);
}

static void
spawn_processors(void)
{
    char processor_path[500];
    size_t path_size = sizeof(processor_path);
    char *p;
    char *args[2];
    int cpu_count;
    uv_cpu_info_t *info;

    /* uv_exepath determines the full path (containing the executable) of the
     * currently running process. */
    uv_exepath(processor_path, &path_size);

    /* We make the assumption here that our processors are in the same
     * directory. */
    fprintf(stderr, "Parent path: %s\n", processor_path);
#ifdef WIN32
    p = strrchr(processor_path, '\\');
#else
    p = strrchr(processor_path, '/');
#endif
    if (p) {
        p[1] = '\0';
#ifdef WIN32
        strncat(processor_path, "child.exe",
            sizeof(processor_path) - 1 - strlen("child.exe"));
#else
        strncat(processor_path, "child",
            sizeof(processor_path) - 1 - strlen("child"));
#endif
    }
    fprintf(stderr, "Worker path: %s\n", processor_path);

    /* Setup argv for spawned processors */
    args[0] = processor_path;
    args[1] = NULL; /* no arguments to executable */

    round_robin_counter = 0;

    /* launch same number of workers as number of CPUs */
    uv_cpu_info(&info, &cpu_count);
    uv_free_cpu_info(info, cpu_count);

    child_worker_count = cpu_count;

    workers = calloc(sizeof(struct child_worker), cpu_count);
    while (cpu_count--) {
        struct child_worker *worker = &workers[cpu_count];

        uv_pipe_init(loop, &worker->pipe, 1);

        uv_stdio_container_t child_stdio[3];
        /* WORKAROUND
        child_stdio[0].flags = UV_CREATE_PIPE | UV_READABLE_PIPE | UV_WRITABLE_PIPE;
        * END_WORKAROUND */
        child_stdio[0].flags = UV_CREATE_PIPE | UV_READABLE_PIPE;
        child_stdio[0].data.stream = (uv_stream_t*) &worker->pipe;
        child_stdio[1].flags = UV_IGNORE;
        child_stdio[2].flags = UV_INHERIT_FD;
        child_stdio[2].data.fd = 2;

        worker->options.stdio = child_stdio;
        worker->options.stdio_count = 3;

        worker->options.exit_cb = close_process_handle;
        worker->options.file = args[0];
        worker->options.args = args;

        uv_spawn(loop, &worker->req, &worker->options); 
        fprintf(stderr, "Started worker %d\n", worker->req.pid);
    }
}

static void
on_new_connection(uv_stream_t *server, int status)
{
    if (status == -1) {
        // error!
        return;
    }

    uv_tcp_t *client = (uv_tcp_t *) malloc(sizeof(uv_tcp_t));
    uv_tcp_init(loop, client);

    if (uv_accept(server, (uv_stream_t*) client) == 0) {
        uv_buf_t dummy_buf;
        uv_write_t *write_req = (uv_write_t*) malloc(sizeof(uv_write_t));
        dummy_buf = uv_buf_init(".", 1);
        /* Send accepted TCP socket over pipe. */
        struct child_worker *worker = &workers[round_robin_counter];
        uv_write2(write_req, (uv_stream_t*) &worker->pipe, &dummy_buf, 1, (uv_stream_t*) client, NULL);
        round_robin_counter = (round_robin_counter + 1) % child_worker_count;
    }
    /* The handle is duplicated when it's sent to the child process and we don't
     * need to use it here, so close it */
    uv_close((uv_handle_t*) client, NULL);
}

int
main(int args, char **argsv)
{
    struct sockaddr_in listen_address;
    uv_tcp_t server;

    /* TODO: Handle arguments / configuration files */
    loop = uv_default_loop();

    spawn_processors();

    uv_tcp_init(loop, &server);
    uv_ip4_addr("0.0.0.0", 8000, &listen_address);

    uv_tcp_bind(&server, (const struct sockaddr *)&listen_address, 0);
    if (uv_listen((uv_stream_t*) &server, 128, on_new_connection)) {
        fprintf(stderr, "uv_listen: Error couldn't listen\n");
        return 2;
    }

    uv_run(loop, UV_RUN_DEFAULT);
    return 0;
}
```

---- child.c CUT ---

``` c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <uv.h>

uv_loop_t* uv_loop;
uv_pipe_t queue;
uv_tcp_t server;

void
on_alloc(uv_handle_t *client, size_t suggested_size, uv_buf_t *buf)
{
    buf->base = malloc(suggested_size);
    buf->len = suggested_size;
}

void
stream_on_close(uv_handle_t *handle)
{
    free(handle);
}

void
stream_on_read(uv_stream_t *tcp, ssize_t nread, const uv_buf_t *buf)
{
    fprintf(stderr, "Read buffer, closing...\n");
    uv_close((uv_handle_t *)tcp, stream_on_close);
}

void
on_new_connection(uv_stream_t *q, ssize_t nread, const uv_buf_t *buf)
{
    uv_tcp_t *stream = (uv_tcp_t *)malloc(sizeof(uv_tcp_t));
    uv_tcp_init(uv_loop, stream);

    if (uv_accept((uv_stream_t *)q, (uv_stream_t*)stream) == 0) {
        uv_read_start((uv_stream_t*)stream, on_alloc, stream_on_read);
    }
}

int
main(int args, char **argsv)
{
    uv_loop = uv_default_loop();
    uv_tcp_init(uv_loop, &server);

    uv_pipe_init(uv_loop, &queue, 1);
    uv_pipe_open(&queue, 0);

    uv_read_start((uv_stream_t *)&queue, on_alloc,
        on_new_connection);

    uv_run(uv_loop, UV_RUN_DEFAULT);

    return 0;
}
```

saghul:
In the child you initialize the pipe as an IPC pipe. IPC pipes are used to send handles over them and thus they need to be writable. You should be able to get your script working just by initializing the pipe as a normal one:

```
uv_pipe_init(uv_loop, &queue, 0);
```

devinsmith:
When I use

```
uv_pipe_init(uv_loop, &queue, 0);
```

The child process does not exit, but the stream_on_read function never gets called. I think the solution for now is to use uv_pipe_init(uv_loop, &queue, 1); but have the parent create the writable pipe (see workaround comment in parent.c above). I just wasn't sure if this was a regression.

saghul:
Hum, I see what you are trying to do. Can you try to use another fd instead of one which doubles with stdio (0-2)? You can set the first 3 elements of the stdio array to UV_INGORE and use the 4th one for the communication channel. IIRC this is how node does it and it also works for me on pyuv. 

There might be some limitation on using 0-2 fds for transferring handles, which we need to document, if that is the case.

saghul:
No follow up, closing.

PR #1466, opened 2014-09-10
test: don't close connection on write error

trevnorris:
The echo server shouldn't close the connection when there's an error on
write. Instead simply echo the error message and allow the other side to
close the connection.


saghul:
I managed to reproduce the issue on OSX by running the test in a loop like crazy. Applying this diff on top of your patch seems to finally fix it: https://gist.github.com/saghul/03641437563268ab04f2

Could you apply it, and rebase?

Thanks Trevor!

trevnorris:
@saghul Forced pushed.

saghul:
Thanks Trevor! Landed in :sparkles: 3d38eae :sparkles:

Issue #1447, opened 2014-09-01
Lot of warnings compiling libuv 1.0.0 in OSX with clang


ibc:
```
gcc -v
Configured with: --prefix=/Library/Developer/CommandLineTools/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 5.1 (clang-503.0.40) (based on LLVM 3.4svn)
Target: x86_64-apple-darwin13.3.0
Thread model: posix
```

I call `make clean && ./configure && make` on today's master branch and get lot of warnings:

```
./include/uv-dtrace.h:43:45: warning: '$' in identifier [-Wdollar-in-identifier-extension]
extern int __dtrace_isenabled$uv$tick__start$v1(void);
                                            ^
./include/uv-dtrace.h:44:27: warning: '$' in identifier [-Wdollar-in-identifier-extension]
extern void __dtrace_probe$uv$tick__stop$v1$766f6964202a$696e74(const void *, int);
                          ^
./include/uv-dtrace.h:44:30: warning: '$' in identifier [-Wdollar-in-identifier-extension]
extern void __dtrace_probe$uv$tick__stop$v1$766f6964202a$696e74(const void *, int);
                             ^
./include/uv-dtrace.h:44:41: warning: '$' in identifier [-Wdollar-in-identifier-extension]
extern void __dtrace_probe$uv$tick__stop$v1$766f6964202a$696e74(const void *, int);
                                        ^
./include/uv-dtrace.h:44:44: warning: '$' in identifier [-Wdollar-in-identifier-extension]
extern void __dtrace_probe$uv$tick__stop$v1$766f6964202a$696e74(const void *, int);
                                           ^
./include/uv-dtrace.h:44:57: warning: '$' in identifier [-Wdollar-in-identifier-extension]
extern void __dtrace_probe$uv$tick__stop$v1$766f6964202a$696e74(const void *, int);
                                                        ^
./include/uv-dtrace.h:45:30: warning: '$' in identifier [-Wdollar-in-identifier-extension]
extern int __dtrace_isenabled$uv$tick__stop$v1(void);
                             ^
./include/uv-dtrace.h:45:33: warning: '$' in identifier [-Wdollar-in-identifier-extension]
extern int __dtrace_isenabled$uv$tick__stop$v1(void);
                                ^
./include/uv-dtrace.h:45:44: warning: '$' in identifier [-Wdollar-in-identifier-extension]
[...]

Issue #1444, opened 2014-08-29
test: spawn_reads_child_path fails


recp:
I have just redownloaded latest version of libuv even I've already have it. Because I was wanted to compare test results of the current libuv with my forked version. 

There are no errors about make but there are too many warnings about void \* such as:

```
warning: assigning to 'void (*)(void)' from 'void *' converts between void pointer and function
      pointer [-Wpedantic]

...

warning: assigning to 'CFBundleRef (*)(CFStringRef)' from 'void *' converts between void
      pointer and function pointer [-Wpedantic]
```

The test (-spawn_reads_child_path-) fails always on OS X 10.10 (Darwin 14.0.0).

```
`spawn_reads_child_path` failed: exit code 6
Output from process `spawn_reads_child_path`:
exit_cb
Assertion failed in test/test-spawn.c on line 66: exit_status == 1

```

libuv has already placed (forked version -unstable-) in /usr/local/lib and /usr/local/include

saghul:
Hi!

I just compiled libuv on my OSX 10.10 machine and I got no warnings when compiling nor errors when running the test suite. Can you please detail the steps you followed?

recp:
Hi @saghul !
I have just redownloaded libuv and I followed the steps (again):
1. Download libuv-master via “Download ZIP”
2. Open Terminal.app and enter  ~/Downloads/libuv-master 
3. Run the following commands 

```
sh ./autogen.sh [OK]
./configure [OK]
make [OK with warnings]
make check [FAIL: spawn_reads_child_path]
```

GCC Version:

```
recp:libuv-master RCp$ gcc -v
Configured with: --prefix=/Applications/Xcode6-Beta6.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 6.0 (clang-600.0.47) (based on LLVM 3.5svn)
Target: x86_64-apple-darwin14.0.0
Thread model: posix
```

I got same warnings for the GCC Version (Xcode5.1.1):

```
recp:libuv-master RCp$ gcc -v
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 5.1 (clang-503.0.40) (based on LLVM 3.4svn)
Target: x86_64-apple-darwin14.0.0
Thread model: posix
```

I got the following warnings for each object when make compiling:

```
recp:libuv-master RCp$ make
[...]

PR #1467, opened 2014-09-11
Fix warnings about dollar and dlsym

recp:
We were getting some warnings about dlsym when compiling like warnings in the issues #1447 and #1444

C does not support casts between function pointers and data pointers (yet) because they sizes can be different.... Maybe there are some optimizations such as implicit casts when compiling/linking for dlsym. This can be used until found better solution

There are some solutions here: http://christopherpoole.github.io/dynamically-linking-shared-objects-in-c++/ for avoiding void \* to fn pointer.

There is a comment on this man page http://linux.die.net/man/3/dlsym like this:

``` C
/* Writing: cosine = (double (*)(double)) dlsym(handle, "cos");
       would seem more natural, but the C99 standard leaves
       casting from "void *" to a function pointer undefined.
       The assignment used below is the POSIX.1-2003 (Technical
       Corrigendum 1) workaround; see the Rationale for the
       POSIX specification of dlsym(). */

   *(void **) (&cosine) = dlsym(handle, "cos");
```

I applied the approach/solution for avoiding warnings.
- Fix warnings about dlsym and others
- Fix warnings about dollar


saghul:
Thanks! Please rebase your PR on top of the v1.x branch, not master.

Makefile.am
```diff
@@ -81,6 +81,7 @@ else  # WINNT
 
 include_HEADERS += include/uv-unix.h
 AM_CPPFLAGS += -I$(top_srcdir)/src/unix
+libuv_la_CFLAGS += -g --std=gnu89 -pedantic -Wall -Wextra -Wno-unused-parameter -Wno-dollar-in-identifier-extension
```
saghul, line 85: AFAIK not all compilers understand "-Wno-dollar-in-identifier-extension", only clang does. See what Ben said on #1447.

saghul:
Superseded by #1468

PR #1468, opened 2014-09-11
Fix warnings about dollar and dlsym 

recp:
We were getting some warnings about dlsym when compiling like warnings in the issues #1447 and #1444

C does not support casts between function pointers and data pointers (yet) because they sizes can be different.... Maybe there are some optimizations such as implicit casts when compiling/linking for dlsym. This can be used until found better solution

There are some solutions here: http://christopherpoole.github.io/dynamically-linking-shared-objects-in-c++/ for avoiding void \* to fn pointer.

There is a comment on this man page http://linux.die.net/man/3/dlsym like this:

``` C
/* Writing: cosine = (double (*)(double)) dlsym(handle, "cos");
       would seem more natural, but the C99 standard leaves
       casting from "void *" to a function pointer undefined.
       The assignment used below is the POSIX.1-2003 (Technical
       Corrigendum 1) workaround; see the Rationale for the
       POSIX specification of dlsym(). */

   *(void **) (&cosine) = dlsym(handle, "cos");
```

I applied the approach/solution for avoiding warnings.
- Fix warnings about dlsym and others
- Fix warnings about dollar

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,6 @@ ipch
 
 # sphinx generated files
 /docs/build/
+
+*.xcodeproj
+*.xcworkspace
--- a/src/unix/darwin-proctitle.c
+++ b/src/unix/darwin-proctitle.c
@@ -36,7 +36,7 @@ static int uv__pthread_setname_np(const char* name) {
   int err;
 
   /* pthread_setname_np() first appeared in OS X 10.6 and iOS 3.2. */
-  dynamic_pthread_setname_np = dlsym(RTLD_DEFAULT, "pthread_setname_np");
+  *(void **)(&dynamic_pthread_setname_np) = dlsym(RTLD_DEFAULT, "pthread_setname_np");
   if (dynamic_pthread_setname_np == NULL)
     return -ENOSYS;
 
@@ -94,13 +94,13 @@ int uv__set_process_title(const char* title) {
   if (application_services_handle == NULL || core_foundation_handle == NULL)
     goto out;
 
-  pCFStringCreateWithCString =
+  *(void **)(&pCFStringCreateWithCString) =
       dlsym(core_foundation_handle, "CFStringCreateWithCString");
-  pCFBundleGetBundleWithIdentifier =
+  *(void **)(&pCFBundleGetBundleWithIdentifier) =
       dlsym(core_foundation_handle, "CFBundleGetBundleWithIdentifier");
-  pCFBundleGetDataPointerForName =
+  *(void **)(&pCFBundleGetDataPointerForName) =
       dlsym(core_foundation_handle, "CFBundleGetDataPointerForName");
-  pCFBundleGetFunctionPointerForName =
+  *(void **)(&pCFBundleGetFunctionPointerForName) =
       dlsym(core_foundation_handle, "CFBundleGetFunctionPointerForName");
 
   if (pCFStringCreateWithCString == NULL ||
@@ -118,14 +118,14 @@ int uv__set_process_title(const char* title) {
   if (launch_services_bundle == NULL)
     goto out;
 
-  pLSGetCurrentApplicationASN =
+  *(void **)(&pLSGetCurrentApplicationASN) =
       pCFBundleGetFunctionPointerForName(launch_services_bundle,
                                          S("_LSGetCurrentApplicationASN"));
 
   if (pLSGetCurrentApplicationASN == NULL)
     goto out;
 
-  pLSSetApplicationInformationItem =
+  *(void **)(&pLSSetApplicationInformationItem) =
       pCFBundleGetFunctionPointerForName(launch_services_bundle,
                                          S("_LSSetApplicationInformationItem"));
 
@@ -138,9 +138,9 @@ int uv__set_process_title(const char* title) {
   if (display_name_key == NULL || *display_name_key == NULL)
     goto out;
 
-  pCFBundleGetInfoDictionary = dlsym(core_foundation_handle,
+  *(void **)(&pCFBundleGetInfoDictionary) = dlsym(core_foundation_handle,
                                      "CFBundleGetInfoDictionary");
-  pCFBundleGetMainBundle = dlsym(core_foundation_handle,
+  *(void **)(&pCFBundleGetMainBundle) = dlsym(core_foundation_handle,
                                  "CFBundleGetMainBundle");
   if (pCFBundleGetInfoDictionary == NULL || pCFBundleGetMainBundle == NULL)
     goto out;
@@ -152,13 +152,13 @@ int uv__set_process_title(const char* title) {
   if (hi_services_bundle == NULL)
     goto out;
 
-  pSetApplicationIsDaemon = pCFBundleGetFunctionPointerForName(
+  *(void **)(&pSetApplicationIsDaemon) = pCFBundleGetFunctionPointerForName(
       hi_services_bundle,
       S("SetApplicationIsDaemon"));
-  pLSApplicationCheckIn = pCFBundleGetFunctionPointerForName(
+  *(void **)(&pLSApplicationCheckIn) = pCFBundleGetFunctionPointerForName(
       launch_services_bundle,
       S("_LSApplicationCheckIn"));
-  pLSSetApplicationLaunchServicesServerConnectionStatus =
+  *(void **)(&pLSSetApplicationLaunchServicesServerConnectionStatus) =
       pCFBundleGetFunctionPointerForName(
           launch_services_bundle,
           S("_LSSetApplicationLaunchServicesServerConnectionStatus"));
--- a/src/unix/fsevents.c
+++ b/src/unix/fsevents.c
@@ -525,7 +525,7 @@ static int uv__fsevents_global_init(void) {
   err = -ENOENT;
 #define V(handle, symbol)                                                     \
   do {                                                                        \
-    p ## symbol = dlsym((handle), #symbol);                                   \
+    *(void **)(&p ## symbol) = dlsym((handle), #symbol);                      \
     if (p ## symbol == NULL)                                                  \
       goto out;                                                               \
   }                                                                           \
--- a/uv.gyp
+++ b/uv.gyp
@@ -26,6 +26,30 @@
         ],
       }],
     ],
+    'xcode_settings': {
+        'conditions': [
+          [ 'clang==1', {
+            'WARNING_CFLAGS': [
+              '-Wall',
+              '-Wextra',
+              '-Wno-unused-parameter',
+              '-Wno-dollar-in-identifier-extension'
+            ]}, {
+           'WARNING_CFLAGS': [
+             '-Wall',
+             '-Wextra',
+             '-Wno-unused-parameter'
+          ]}
+        ]
+      ],
+      'OTHER_LDFLAGS': [
+      ],
+      'OTHER_CFLAGS': [
+        '-g',
+        '--std=gnu89',
+        '-pedantic'
+      ],
+    }
   },
 
   'targets': [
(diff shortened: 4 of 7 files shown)


txdv:
You mean the official C89 standard?

.gitignore
```diff
@@ -64,3 +64,6 @@ ipch
 
 # sphinx generated files
 /docs/build/
+
+*.xcodeproj
+*.xcworkspace
```
saghul: Please make this a different commit, it's unrelated change.

Makefile.am
```diff
@@ -81,7 +81,9 @@ else  # WINNT
 
 include_HEADERS += include/uv-unix.h
 AM_CPPFLAGS += -I$(top_srcdir)/src/unix
-libuv_la_CFLAGS += -g --std=gnu89 -pedantic -Wall -Wextra -Wno-unused-parameter
+
+libuv_la_CFLAGS += -g --std=gnu89 -pedantic -Wall -Wextra -Wno-unused-parameter -Wno-dollar-in-identifier-extension
```
saghul: "-Wno-dollar-in-identifier-extension" won't work with compiler other than Clang, AFAIK. Check the comment by Ben in #1447.

saghul:
Please squash the last 2 commits into a single one.

Makefile.am
```diff
@@ -81,7 +81,9 @@ else  # WINNT
 
 include_HEADERS += include/uv-unix.h
 AM_CPPFLAGS += -I$(top_srcdir)/src/unix
-libuv_la_CFLAGS += -g --std=gnu89 -pedantic -Wall -Wextra -Wno-unused-parameter
+
```
saghul: Extraneous newline.

Makefile.am
saghul: The flag would also need to be set on GYP builds. Unfortunately I don't know how to check for the compiler name in a GYP file.

recp:
Hi @txdv actually I didn't mean to C89 only. There are some reasons here about why we are getting the warnings about dlsym (data pointers -> fn-ptrs)

[Why are function pointers and data pointers incompatible in C/C++?](http://stackoverflow.com/questions/12358843/why-are-function-pointers-and-data-pointers-incompatible-in-c-c)

.gitignore
recp: done

recp:
> Please squash the last 2 commits into a single one.

Squashed!

saghul:
GitHub doesn't show my comment anymore, but the "-Wno-dollar-in-identifier-extension" only works with Clang, not GCC. You'll need to detect if the compiler supports it, see Ben's comment in #1447.

Also, that fix would need to be applied for GYP builds.

recp:
> GitHub doesn't show my comment anymore, but the "-Wno-dollar-in-identifier-extension" only works with Clang, not GCC. You'll need to detect if the compiler supports it, see Ben's comment in #1447.

Are we getting warnings about dollar on GCC? I've just compiled libuv-master on Fedora I got no warnings. How can I know/monitor warnings on GCC then (Maybe its related Fedora or Linux)?

saghul:
> Are we getting warnings about dollar on GCC? I've just compiled libuv-master on Fedora I got no warnings. How can I know/monitor warnings on GCC then (Maybe its related Fedora or Linux)?

Those symbols only show up when building with dtrace, so just on OSX. Also, looks like recent GCC may support it, but it wasn't in the past: https://github.com/joyent/libuv/compare/633d33a92a40...0ff06b4ac38a

So we'd need to feature-test it or compilation could fail. Unfortunately I no longer have a OSX 10.6 machine with had GCC 4.2.1.

saghul:
Managed to install the old GCC on OSX Mavericks:

```
(git|pr/1468)saghul@echo.local:~/src/libuv
$ make
  GEN      include/uv-dtrace.h
/Applications/Xcode.app/Contents/Developer/usr/bin/make  all-am
  CC       src/libuv_la-fs-poll.lo
cc1: error: unrecognized command line option "-Wno-dollar-in-identifier-extension"
make[1]: *** [src/libuv_la-fs-poll.lo] Error 1
make: *** [all] Error 2
(git|pr/1468)saghul@echo.local:~/src/libuv
```

recp:
> So we'd need to feature-test it or compilation could fail. Unfortunately I no longer have a OSX 10.6 machine with had GCC 4.2.1.

@saghul can we check the flag in configure.ac like Ben's commit f75bb7c3379d4398de8ed588a110aa7ab8708a91 :

configure.ac

```
CC_CHECK_CFLAGS_APPEND([-Wno-dollar-in-identifier-extension])
```

Ben was created a m4 file for check flags m4/libuv-check-flags.m4 but the commit was reverted by the commit 1d42b3ce7c184696960a96f129ac586749b76a84 :/

Maybe we can re-create this file for checking the flag for avoiding dollar warnings?

/cc @bnoordhuis @indutny

saghul:
That's exactly what we need. That's why I mentioned the note Ben added on
the other issue.

Reminder: gyp :-)

saghul:
I forgot to mention, that while Ben's approach is what we need, we don't want to enable all warnings at the moment, since we need to analyse them on a case by case basis. Please just enable the relevant flag in this case. Thanks!

recp:
I thought so according to Fedor's comments on revert commit: 1d42b3ce7c184696960a96f129ac586749b76a84 ;)

recp:
@saghul I added the flag -Wno-dollar-in-identifier-extension to gyp by checking clang. Im not sure that is working correctly or not, I built on my OS X with gyp and it seems ok

Can you review it?

uv.gyp
```diff
@@ -120,13 +120,28 @@
...
           },
         }, { # Not Windows i.e. POSIX
-          'cflags': [
-            '-g',
-            '--std=gnu89',
-            '-pedantic',
-            '-Wall',
-            '-Wextra',
-            '-Wno-unused-parameter',
+		  'conditions': [
```
saghul: Please use 2 spaces for indentation, like the rest of the file.

recp: Its done. The spaces seems different when working in Xcode and out :/ I re-indented spaces in TextMate.

saghul:
Looks good! I'll check it out a bit later.

saghul:
There seems to be something wrong with the syntax in uv.gyp. None of the flags in the edited block are applied now.

saghul:
This seems like the proper fix:

```
diff --git a/uv.gyp b/uv.gyp
index f121898..3213a05 100644
--- a/uv.gyp
+++ b/uv.gyp
@@ -195,6 +195,10 @@
             '_DARWIN_UNLIMITED_SELECT=1',
           ]
         }],
+        [ 'OS!="win" and clang==1', {
+          # Not all versions of GCC support it, so just enable it on clang
+          'cflags': [ '-Wno-dollar-in-identifier-extension' ],
+        }],
         [ 'OS!="mac"', {
           # Enable on all platforms except OS X. The antique gcc/clang that
           # ships with Xcode emits waaaay too many false positives.
```

saghul:
Hum, I don't get it. None of the flags seem to affect OSX, they are just not there :-O The simplified fix I proposed works on Linux, but with or without the last patch, cflags are not put into the generated makefile :-(

recp:
> ```
>    [ 'OS!="win" and clang==1', {
>       # Not all versions of GCC support it, so just enable it on clang
>       'cflags': [ '-Wno-dollar-in-identifier-extension' ],
>     }],
> ```

You mean its wrong? The gyp puzzled me :-S I thought clang doesnt work on win.

saghul:
The symbols are only present on Unix source files, hence the check.

bnoordhuis:
> None of the flags seem to affect OSX, they are just not there

That's GYP's Xcode emulation mode at work.  You enable the flags like this:

```
xcode_settings': {
  'OTHER_CFLAGS': [ ... ]
}
```

EDIT: Warnings should go in WARNING_CFLAGS, linker flags in OTHER_LDFLAGS and everything else in OTHER_CFLAGS.

saghul:
@bnoordhuis Oh! Does that need to be used in even when makefiles or ninja files are generated instead of xcode project files?

bnoordhuis:
@saghul Yes.

saghul:
How fun, uv.gyp has been fucked up for a while then :-S Thanks a lot for the pointer Ben!

recp:
@bnoordhuis @saghul I found the [tutorial](http://www.rioki.org/2013/02/07/experimenting-with-gyp.html) and there are some xcode_settings like Ben's comment.

if I'll replace prev commit like below will it be correct? I think there is no req OS="mac" anymore :)

``` javascript
{
  'target_defaults': {
    'default_configuration': 'Release',
    'configurations': {
      'Debug': {
         'xcode_settings': {
           'WARNING_CFLAGS': [
              '-Wno-dollar-in-identifier-extension'
           ]
        },

       'Release': {
         'xcode_settings': {
           'WARNING_CFLAGS': [
              '-Wno-dollar-in-identifier-extension'
           ]
        }
```

saghul:
@recp the other flags also need to be set (though you can do that in a separate commit).

recp:
@saghul I updated the gyp to use `xcode_settings` 

It seems work. I checked Clang for dollar-sign for avoiding errors/crashes when building with older Xcode's which use GCC.  
The flags which we set on gyp seems on output when compiling with `xcodebuild` but not with `make` ($ ./gyp_uv.py -f make && make -C out). Also I think the commit can be revert 7e3403e8992ec0970fd160641b8fbd804a5b9521 ?

saghul:
@recp thanks! Yes, that commit should go away. I'll test it throroughly tomorrow, can you please squash your commits?

recp:
I'll squash them which have same title or all into one?

saghul:
Not all, just squash those which are intermediate states.

recp:
Got it, I'll do that tomorrow. Thanks

recp:
@saghul squashed! Can you review it?

saghul:
Will do! Thanks!

uv.gyp
```diff
@@ -26,6 +26,32 @@
...
+          ]}
+        ]
+      ],
+      'OTHER_LDFLAGS': [
+      ],
+      'OTHER_CFLAGS': [
+        '-g',
+        '--std=gnu89',
+        '-pedantic',
+        '-D_DARWIN_USE_64_BIT_INODE=1',
```
saghul: These last 2 (-D_DARWIN...) are not needed, as gyp properly understands them from above, please remove them.

src/unix/fsevents.c
```diff
@@ -525,7 +525,7 @@ static int uv__fsevents_global_init(void) {
   err = -ENOENT;
 #define V(handle, symbol)                                                     \
   do {                                                                        \
-    p ## symbol = dlsym((handle), #symbol);                                   \
+    *(void **)(&p ## symbol) = dlsym((handle), #symbol);                                   \
```
saghul: style: please align the \

src/unix/darwin-proctitle.c
```diff
@@ -118,14 +118,14 @@ int uv__set_process_title(const char* title) {
   if (launch_services_bundle == NULL)
     goto out;
 
-  pLSGetCurrentApplicationASN =
-      pCFBundleGetFunctionPointerForName(launch_services_bundle,
+  *(void **)(&pLSGetCurrentApplicationASN) =
+       pCFBundleGetFunctionPointerForName(launch_services_bundle,
```
saghul: extra leading space?

saghul:
Great work @recp! Left some last comments, address those, squash, and we are game!

recp:
@saghul it's done :+1:

saghul:
Kewl looks great, will merge later today!

saghul:
All commits landed on v1.x, thanks a lot for working on this @recp, much appreciated!

recp:
Thanks!

PR #1469, opened 2014-09-11
version: add ability to specify a version suffix

saghul:
Since we will no longer have "stable" and "unstable" releases after 1.0.0, use a suffix for pre-releases, such as "rc1", "rc2", ... so the full version string reads: 1.0.0-rc1, for example.

While in development the suffix shall be "dev".


bnoordhuis:
LGTM.  This probably also requires an update to the release script?

saghul:
Yes, indeed. The tool cannot increase the version anymore since we will need to bump minor or patch depending on what changed. I hope @piscisaureus can help me get the tool in shape :-)

PR #1473, opened 2014-09-13
Implement recursive fs watch on Windows

ghostoy:


include/uv-win.h
```diff
@@ -626,6 +626,7 @@ RB_HEAD(uv_timer_tree_s, uv_timer_s);
   HANDLE dir_handle;                                                          \
   int req_pending;                                                            \
   uv_fs_event_cb cb;                                                          \
+  unsigned int win_flags;                                                     \
```
saghul: Call it `fs_flags`, this code in on the Windows side anyway.

src/win/fs-event.c
```diff
@@ -137,6 +152,8 @@ int uv_fs_event_start(uv_fs_event_t* handle,
     return UV_EINVAL;
 
   handle->cb = cb;
+  // Added win_flags for recursive watcher
```
saghul: Superfluous comment.

saghul:
Thanks! Could you please add a test for this?

ghostoy:
@saghul Already done.

saghul:
Thanks! I'll take a look.

jorangreef:
Thanks, this would be really helpful.

nitoyon:
:+1: 
Tested on Windows 8.1 x64. It works fine:sparkles:

ghostoy:
Any plan to merge this PR?

saghul:
Plans, yes, time, not so much lately, but it will! :-)

On Thu, Nov 6, 2014 at 9:26 AM, Cong Liu notifications@github.com wrote:

> Any plan to merge this PR?
> 
> —
> https://github.com/joyent/libuv/pull/1473#issuecomment-61985755.

## 

/Saúl
bettercallsaghul.com

jorangreef:
Would be great to have this landed soon. There is already recursive watch support on OS X and Linux can use per-directory watchers. But Windows can't just use multiple watchers as this would lock each watched directory (i.e. preventing the user from renaming those directories etc.). This is the last thing stopping a decent cross-platform watch implementation.

saghul:
Ouch, looks like this fell through the cracks :-(

It will need a bit of adjustment, because as it stands it breaks the ABI, but we can probably use one of the reserved fields.

saghul:
Superseded by https://github.com/libuv/libuv/pull/421

PR #1474, opened 2014-09-15
uv__epoll_wait: fall back to uv__epoll_pwait

mwhudson:
It seems that epoll_wait is implemented in glibc in terms of epoll_pwait and new architectures (like arm64) do not implement the epoll_wait syscall at all.  So if __NR_epoll_wait is undefined, we should just call uv__epoll_pwait.

### DIFF
--- a/src/unix/linux-syscalls.c
+++ b/src/unix/linux-syscalls.c
@@ -311,7 +311,7 @@ int uv__epoll_wait(int epfd,
 #if defined(__NR_epoll_wait)
   return syscall(__NR_epoll_wait, epfd, events, nevents, timeout);
 #else
-  return errno = ENOSYS, -1;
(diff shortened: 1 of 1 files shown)


saghul:
Hi! The code in linux-syscalls should just expose the syscall. The calling functions are the ones which need to pick bettween the wait / pwait variants. This is the convention we use.

Please use a static variable in the `uv__io_poll` function to "remember" if pwait needs to be used. As an example, look at what `uv__dup2_cloexec` does to detect `dup3` support.

mwhudson:
Ah, OK, so like https://github.com/joyent/libuv/pull/1477 ?  I'll close this.

saghul:
FTR, you can iterate on a pull request by amending your changes and force pushing.

PR #1477, opened 2014-09-15
uv__io_poll: try epoll_pwait if epoll_wait is missing

mwhudson:
It seems that epoll_wait is implemented in glibc in terms of epoll_pwait and
new architectures (like arm64) do not implement the epoll_wait syscall at all.
So if __NR_epoll_wait is undefined, we should just call uv__epoll_pwait.

An alternative to https://github.com/joyent/libuv/pull/1474.


src/unix/linux-core.c
```diff
@@ -195,10 +196,18 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
...
-                          events,
-                          ARRAY_SIZE(events),
-                          timeout);
+    if (!no_epoll_wait) {
+      nfds = uv__epoll_wait(loop->backend_fd,
+                            events,
+                            ARRAY_SIZE(events),
+                            timeout);
+    } else {
+      nfds = uv__epoll_wait(loop->backend_fd,
```
saghul: I guess you mean pwait here?

src/unix/linux-core.c
```diff
@@ -212,8 +221,13 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
     }
 
     if (nfds == -1) {
-      if (errno != EINTR)
+      if (errno != EINTR) {
+	if (errno == ENOSYS && !no_epoll_wait) {
+          no_epoll_wait = 1;
```
saghul: Please move this test above, right after doing the epoll_wait.

mwhudson:
Ooops.  Fixed, and amended as asked.

src/unix/linux-core.c
```diff
@@ -195,10 +196,22 @@ void uv__io_poll(uv_loop_t* loop, int timeout) {
...
+      nfds = uv__epoll_wait(loop->backend_fd,
+                            events,
+                            ARRAY_SIZE(events),
+                            timeout);
+      if (nfds == -1 && errno == ENOSYS) {
+        no_epoll_wait = 1;
+        continue;
+      }
+    } else {
+      nfds = uv__epoll_wait(loop->backend_fd,
```
saghul: This needs to be pwait.

mwhudson:
Aaargh git fail.  This time for sure?

saghul:
Yeah, landed in 861de3d, thanks!

mwhudson:
Yay, thanks!

Just curious really but I originally found this problem in node; will this fix appear there reasonably soon?

saghul:
I'd say so, yes. I'm also adding it to the list of backports that will go on 0.10 as well.

PR #1481, opened 2014-09-16
Use epoll_pwait if epoll_wait doesn't exist

trevnorris:
Figured this is the less intrusive approach.


saghul:
That is not the correct approach. The user actually proposed that first and I rejected it in favor of this one. The linux syscalls we expose are just the raw syscalls, the user needs to take care of dealing with the ones that are unsupported. There are multiple examples of this, one being `uv__dup2_cloexec`.

trevnorris:
Though I feel like the approach of adding a conditional check to every loop of `uv__io_poll()` is also incorrect. The example of `uv__dup2_cloexec()` also uses `#ifdef`'s instead of a conditional. So I'm not sure how that supports your point of dealing with `syscall()`.

saghul:
See this: https://github.com/joyent/libuv/blob/master/src/unix/core.c#L944

trevnorris:
@saghul Seems the PR isn't being updated because it was closed. Here's the new commit: https://github.com/trevnorris/libuv/commit/8d1c06e1fd3abf9278356f31b41f507c35489c51

I really have a hard time accepting that type of one time conditional checking is the correct solution when that information can be determined at compile time. Is it correct because there could have been a change to the system between runs that allows one or the other to be used?

saghul:
You could have different kernel headers than the running kernel, that's why we do syscalls like that, instead of detecting what's available at compile time. So, it cannot be determined at compile time.

saghul:
Grep for "no_" in unix/core.c, I'm not making rules up on the go :-)

trevnorris:
Heh. Not that I don't believe you. :-)

I just hate the idea of needing to have these one-off conditional checks every time the function runs. I know it's not possible to do in C (well, AFAIK) but for example when I did JS for the browser back in the IE 6 days conditional features were rampant. So to get around this type of conditional check there is the flexibility of just reassigning the function.

Oh well. If this is a normal C-ism then I'll learn to live with it. :-P

PR #1482, opened 2014-09-18
Windows: fix netmask detection

orangemocha:
uv_interface_addresses was using the linked list pointed to by
the FirstPrefix member of IP_ADAPTER_ADDRESSES in order to compute the
network prefix / network mask.

This was flawed in several ways:
- FirstPrefix can be NULL, and we would crash.
- On Windows Vista and later, the prefix list includes three IP adapter prefixes
  for each IP address assigned to the adapter. We were assuming a 1:1 mapping
  with the unicast address list.
- Even on Windows versions (i.e. XP) where the prefix list is supposed to have
  one and only one element for each unicast address, the order of the two lists
  is not guaranteed to be the same.

This fix was inspired and adapted from a commit in the Chromium project:
https://codereview.chromium.org/25167002/diff/6001/net/base/net_util_win.cc

See MSDN article for reference:
http://msdn.microsoft.com/en-us/library/windows/desktop/aa366058(v=vs.85).aspx

Excerpt from MSDN below:

In addition, the linked IP_ADAPTER_UNICAST_ADDRESS structures pointed to
by the FirstUnicastAddress member and the linked IP_ADAPTER_PREFIX
structures pointed to by the FirstPrefix member are maintained as separate
internal linked lists by the operating system. As a result, the order of
linked IP_ADAPTER_UNICAST_ADDRESS structures pointed to by the
FirstUnicastAddress member does not have any relationship with the order
of linked IP_ADAPTER_PREFIX structures pointed to by the FirstPrefix member.

On Windows Vista and later, the linked IP_ADAPTER_PREFIX structures pointed to
by the FirstPrefix member include three IP adapter prefixes for each IP address
assigned to the adapter. These include the host IP address prefix, the subnet
IP address prefix, and the subnet broadcast IP address prefix. In addition, for
each adapter there is a multicast address prefix and a broadcast address prefix.

### DIFF
--- a/src/win/util.c
+++ b/src/win/util.c
@@ -777,12 +777,71 @@ void uv_free_cpu_info(uv_cpu_info_t* cpu_infos, int count) {
   free(cpu_infos);
 }
 
+int is_windows_version_or_greater(DWORD os_major, DWORD os_minor,
+    WORD service_pack_major, WORD service_pack_minor) {
+  OSVERSIONINFOEX osvi;
+  DWORDLONG condition_mask = 0;
+  int op = VER_GREATER_EQUAL;
+
+  /* Initialize the OSVERSIONINFOEX structure. */
+  ZeroMemory(&osvi, sizeof(OSVERSIONINFOEX));
+  osvi.dwOSVersionInfoSize = sizeof(OSVERSIONINFOEX);
+  osvi.dwMajorVersion = os_major;
+  osvi.dwMinorVersion = os_minor;
+  osvi.wServicePackMajor = service_pack_major;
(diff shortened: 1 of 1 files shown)


saghul:
Thanks Alexis! I'll take a look as soon as I can. /cc @piscisaureus

saghul:
Change looks solid. Tested and it works. I'll adjust style a bit and land tonight if nobody objects.

saghul:
Cheers @orangemocha! Fixed style a bit (didn't want to bother you as I took so long to review) and landed it in 68ac0a6.

orangemocha:
Thank you!

PR #1485, opened 2014-09-20
Add missing test data files with EXTRA_DIST

readams:
Ensures that tests pass from a dist tarball.

Run make dist, then untar the resulting file.  Running make check will fail because of missing test data files.  With this patch, things will work again.  Hooray!


saghul:
Thanks Rob! Landed in 1cc25170ed6d4c5d2ebc60d63776c33ce2210fcc.

PR #1492, opened 2014-09-23
add error code for success

txdv:
I feel like we should have an error code for success, because 0 is a magic number and checking against a magic number is bad.

``` c
assert(status == UV_SAGHUL);
```

looks way better.


edef1c:
:+1:

saghul:
:-1:

saghul:
Sorry, not happening.

PR #1494, opened 2014-09-23
fs: fix readdir on empty directory

indutny:
`uv_fd_readdir()` should initialize all fields even if there are no
files. Fix the implementation and fix the test that should have been
covered it.


indutny:
cc @saghul: this is a blocker for v0.11 release

indutny:
I meant node.js

tjfontaine:
I can note that this does fix the failure on node.js

saghul:
LGTM

indutny:
Landed in 2f54947.

PR #1496, opened 2014-09-25
fs: rename uv_fs_readdir to uv_fs_scandir

### LINKED ISSUE #1431: add uv_fs_scandir
With the advent of `uv_fs_readdir` we should provide a migration path for old code who rely on the old behavior and add `uv_fs_scandir`

saghul:
Closes #1431


indutny:
LGTM

PR #1497, opened 2014-09-25
abstract socket support

### LINKED ISSUE #1486: linux abstract sockets support in connect / listen
libuv already [supports](https://github.com/joyent/libuv/commit/5ac214c712b99a94bc0e37e6544a53b7329d3ab1) abstract sockets in `uv_pipe_getsockname`

Unfortunately, there is no way to connect to abstract socket with `uv_pipe_connect`. 

I'm currently wrapping socat binary to be able to connect to abstract socket address from node-dbus client but it would make life so much easier if abstract socket would just work from core node.

Some my notes here: https://github.com/sidorares/abstractsocket/issues/3

Few problems here:
Afaik python and ruby connect to abstract sockets using "\0name". At uv level allowing zero in socket name would require changing `uv_pipe_connect` signature (or adding very similar one like `uv_pipe_connect2(req, handle, name, name_length, cb)`. Possible solution: encode leading zero somehow ( "@" prefix? do we want to allow non-abstract sockets to connect to `"@name"`? If so, how to escape it - `"@@name"`? ).

I have node version which is able to connect to `socat abstract-listen:somename -` server by doing `net.connect('\0somename')`.  Would you be interested in PR? Is yes, which is preferred way - add `name_length` parameters where required or encode "\0" at js side and decode back at uv side ( by prefixing with @ for example )?

txdv:
fix #1486


saghul:
This cannot be fixed without accepting a length parameter: https://github.com/joyent/libuv/issues/1486#issuecomment-56819597

PR #1501, opened 2014-09-26
build: add DragonFly BSD support in autotools

rolinh:


Makefile.am
```diff
@@ -271,6 +271,11 @@ libuv_la_SOURCES += src/unix/darwin.c \
                     src/unix/proctitle.c
 endif
 
+if DRAGONFLY
+include_HEADERS += include/uv-bsd.h
+libuv_la_SOURCES += src/unix/kqueue.c src/unix/freebsd.c
```
saghul, line 276: Mind reversing thr order of the files? Just like the freebsd section below.

rolinh, line 276: I can, of course, but note that only FreeBSD has it the other way around (OpenBSD and NetBSD are ordered this way too).

saghul, line 276: Ah, leave it then :-)

saghul:
Thanks Robin! Landed in 3193b67

PR #1502, opened 2014-09-26
Fix issues pointed out by clang-analyzer

mmalecki:
This fixes all clang-analyzer warnings besides `tree.h`-related ones and the one in `tty.c`, which would break Darwin.

We went down from this:
2014-09-27-005350_1920x1080_scrot

To this:
2014-09-27-005347_1920x1080_scrot


src/unix/linux-core.c
```diff
@@ -743,6 +743,9 @@ int uv_interface_addresses(uv_interface_address_t** addresses,
     (*count)++;
   }
 
+  if (*count == 0)
+    return 0;
```
saghul, line 748: Shouldn't we set *addresses to NULL before returning? Otherwise `uv_free_interface_addresses` would probably crash.

mmalecki, line 748: Good point. We should also indicate that `uv_interface_addresses` will set `addresses` to `NULL` if we have 0 interfaces in our documentation.

mmalecki, line 748: @saghul force pushed that commit

mmalecki:
@saghul I just realized that I opened this PR against `master`, should I reopen against `v1.x`?

saghul:
Please do :-)

saghul:
Closing, if you still want to fix these, please open a PR against the 1,x branch in libuv/libuv. Thanks!

mmalecki:
Ah, damn, completely forgot about this. PR'ing now.

PR #1506, opened 2014-09-28
android: use lowres stat fields in uv__to_stat

ghost:
With this patch, libuv can be built for an Android x86 target [1].
No Android toolchain seems to provide stat->st_atim/ctim/atm,
although ARM targets do build without this patch -- I don't know.

The Dart language had the same problem and reverted back to
second-precision timestamps. [2]

The manpage for stat(2) [3] notes that glibc implemented these fields
in accordance with POSIX.1-2008 [4]. Android's bionic libc apparantly
didn't, none of their sys/stat.h headers contains it.

[1] ./gyp_uv.py -Dtarget_arch=x64 -DOS=android
[2] https://code.google.com/p/dart/issues/detail?id=18524
[3] http://linux.die.net/man/2/stat

> Glibc exposes the nanosecond component of each field using names
> of the form st_atim.tv_nsec if the _BSD_SOURCE or _SVID_SOURCE
> feature test macro is defined. These fields are specified in
> POSIX.1-2008, and, starting with version 2.12, glibc also exposes
> these field names if _POSIX_C_SOURCE is defined with the value
> 200809L or greater, or _XOPEN_SOURCE is defined with the value 700
> or greater. If none of the aforementioned macros are defined, then
> the nanosecond values are exposed with names of the form st_atimensec.

[4] http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html


test/test-fs.c
```diff
@@ -1059,10 +1059,10 @@ TEST_IMPL(fs_fstat) {
   ASSERT(s->st_mtim.tv_nsec == 0);
   ASSERT(s->st_ctim.tv_sec == t.st_ctime);
   ASSERT(s->st_ctim.tv_nsec == 0);
-#elif defined(__sun) || \
+#elif !defined(__ANDROID) && (defined(__sun) || \
```
saghul, line 1062: Shouldn't this be `__ANDROID__` ?

saghul:
This patch disables those fields for all Android toolchains, not just x86. We'd need to better understand why this is not a problem on ARM before moving forward with it.

test/test-fs.c
ghost, line 1062: You're right, I was a bit sloppy there...

ghost:
> We'd need to better understand why this is not a problem on ARM before moving forward with it.

Could you give me a hint how to check whether the ifdefs trigger on ARM? I've looked through all the Android NDK header files and there is definitely no st_atim/ctim/mtim.

I will also try to build for MIPS, the third existing Android target.

saghul:
> Could you give me a hint how to check whether the ifdefs trigger on ARM? I've looked through all the Android NDK header files and there is definitely no st_atim/ctim/mtim.

Not sure, the easiest would be to stick an `#error aouch` in the code which tries to assign then, and see if compilation fails. Then you'd need to figure out why the x86 variant does take that path.

ghost:
Just wanted to say I'm still planning to get to the root of this.

saghul:
@lgierth when you do, please open a PR at libuv/libuv, we moved development over there. Closing.

PR #1513, opened 2014-10-03
doc: fix typo in README

helino:
I'm not a native English speaker, so this change might be wrong, but I guess the author intended to use "keep" instead of "keel" in the sentence "libuv will _keel_ a stable ABI across major releases".


saghul:
Thanks! This has been fixed in the v1.x branch already though.

helino:
Ok, thanks!

Issue #1480, opened 2014-09-16
adding uv_compare


deepak1556:
Following this thread https://groups.google.com/forum/#!topic/libuv/hrVIAyurYm0 , providing pthread#equal like helper would be nice. I am currently requring such an helper here https://github.com/deepak1556/node-ffi/blob/nan_upgrade/src/callback_info.cc#L153,   Any suggestions ?

saghul:
I think I would entertain such a patch. But please don't call it `uv_compare`, call it `uv_thread_equal`, which would map to `pthread_equal` on Unix, and an arithmetic equal on Windows. Thread ids would also need to be typedef'd to their respective types an have some `uv_thread_id_t` type.

saghul:
Superseded by #1516

deepak1556:
:+1:

PR #1516, opened 2014-10-06
unix, windows: Added uv_thread_equal.

neojski:
Implementing uv_thread_equal that compare thread just like pthread_equals.
Changing the return type of uv_thread_self to more proper one uv_thread_t.
Adding simple test.

Fixes issue #1480.


saghul:
Thanks! Please add the test also to Makefile.am so that it's also built in autoconf builds.

neojski:
I just added test to Makefile.am. Let me know if there are other issues.

saghul:
Any objection @bnoordhuis, @indutny?

include/uv.h
```diff
@@ -2376,8 +2376,9 @@ UV_EXTERN void uv_key_set(uv_key_t* key, void* value);
 typedef void (*uv_thread_cb)(void* arg);
 
 UV_EXTERN int uv_thread_create(uv_thread_t* tid, uv_thread_cb entry, void* arg);
-UV_EXTERN unsigned long uv_thread_self(void);
+UV_EXTERN uv_thread_t uv_thread_self(void);
 UV_EXTERN int uv_thread_join(uv_thread_t *tid);
+UV_EXTERN int uv_thread_equal(uv_thread_t* t1, uv_thread_t* t2);
```
bnoordhuis: The pointers should be const to signal that the inputs are not modified.

src/uv-common.c
```diff
@@ -306,11 +306,11 @@ int uv_thread_create(uv_thread_t *tid, void (*entry)(void *arg), void *arg) {
...
 
 
-unsigned long uv_thread_self(void) {
+uv_thread_t uv_thread_self(void) {
 #ifdef _WIN32
-  return (unsigned long) GetCurrentThreadId();
+  return (uv_thread_t) GetCurrentThreadId();
 #else
-  return (unsigned long) pthread_self();
+  return (uv_thread_t) pthread_self();
```
bnoordhuis: The casts aren't necessary.

neojski:
Good points @bnoordhuis. I updated the patch!

test/test-thread-equal.c
```diff
@@ -0,0 +1,49 @@
...
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ */
+
+#include "uv.h"
+#include "task.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h> /* memset */
```
saghul: Unneeded?

neojski: Yeah, copy-paste artifact. Will fix.

By the way:
test-tcp-close.c and test-thread-equal.c has the same line as 27 above but they don't use memset.

neojski:
Unnecessary includes removed.

saghul:
Thanks @neojski! Landed in 59658a8

PR #1519, opened 2014-10-07
unix: don't abort when out of FDs

### LINKED ISSUE #1510: Out of file descriptors should be handled more gracefully
If you call uv_loop_new() and are out of file descriptors there is a chance it will either return NULL or abort().  It should return NULL every time (preferably have some kind of error to handle too) and let the application deal with the problem.

You should be able to reproduce this by calling uv_loop_new() in an endless loop, but there is a chance you could run an application using libuv when you are near the FD limit.

LinuxJedi:
When using uv_loop_new() and the system is low on FDs an abort will be
triggered due to uv_async_init() failing.

To fix this uv_async_init() has been moved up uv__loop_init() so that
allocations are not done before the FD creation and if the FD creation
fails we return -1.  This will gracefully cause uv_loop_new() to fail so
the application can handle it.

Fixes #1510

### DIFF
--- a/src/unix/loop.c
+++ b/src/unix/loop.c
@@ -62,6 +62,9 @@ int uv__loop_init(uv_loop_t* loop, int default_loop) {
   if (uv__platform_loop_init(loop, default_loop))
     return -1;
 
+  if (uv_async_init(loop, &loop->wq_async, uv__work_done))
+    return -1;
+
   uv_signal_init(loop, &loop->child_watcher);
   uv__handle_unref(&loop->child_watcher);
   loop->child_watcher.flags |= UV__HANDLE_INTERNAL;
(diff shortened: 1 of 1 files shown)


saghul:
Thanks! This needs to be applied to the v1.x branch, not v0.10. Then we can think about backporting it.

Also, returning -1 might not be enough, `uv__platform_loop_init` was already called and we'd need to undo what was done there. If `uv_loop_init` fails the user is not expected to call `uv_loop_close`.

LinuxJedi:
ok, no problem, will resubmit in the next 24 hours to the correct branch.

Sorry, I was trying to do this before allocs but must have missed that.  I'll get it to clean up appropriately.

saghul:
Kewl, thanks!

LinuxJedi:
Closing this pull request to open a new one against v1.x branch as requested.

PR #1520, opened 2014-10-08
win: return libuv error codes in uv_poll_init()

### LINKED ISSUE #1518: Windows. uv_poll_init returns windows error codes not libuv error codes.

cjihrig:
Currently, `uv_poll_init()` returns the result of `WSAGetLastError()` on error. This commit wraps the error with a call to `uv_translate_sys_error()`.

Closes #1518


trevnorris:
Seems there are several locations where `WSAGetLastError()` is returned. Think those should be wrapped as well?

saghul:
@trevnorris all functions that return values to the user should have the value translated, but on Windows not all functions return directly, so translation happens later on.

saghul:
Thanks Colin! Landed in 416b733

cjihrig:
Thanks. #1518 can probably be closed too I think.

PR #1522, opened 2014-10-09
unix: don't abort loop initalization

LinuxJedi:
There are several conditions which could cause uv_loop_init() (and by
extension uv_loop_new()) to trigger an SIGABRT.  This includes mutex
initialization failure or running out of FDs.

This patch removes the abort()s and returns error codes instead.
Cleaning up prior initialization if something has actually failed.

Fixes #1519

### DIFF
--- a/src/unix/loop.c
+++ b/src/unix/loop.c
@@ -139,14 +139,26 @@ static int uv__loop_init(uv_loop_t* loop, int default_loop) {
   loop->child_watcher.flags |= UV__HANDLE_INTERNAL;
   QUEUE_INIT(&loop->process_handles);
 
-  if (uv_rwlock_init(&loop->cloexec_lock))
-    abort();
+  err = uv_rwlock_init(&loop->cloexec_lock);
+  if (err) {
+    uv__platform_loop_delete(loop);
+    return err;
+  }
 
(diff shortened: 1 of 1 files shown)


txdv:
Can you make a separate commit/pull request for the unix/thread.c changes?

LinuxJedi:
Changed this pull request to just the loop.c file.  Will submit the other shortly

saghul:
There is also a signal watcher which is started for monitoring child processes which we should also check for failure.

LinuxJedi:
almost makes you wish C had exceptions :)  I'll take a look ASAP.

saghul:
Also, despite many hating goto, it would help here: do all cleanup at the end of the function, using different labels.

LinuxJedi:
well, I was thinking initially of calling uv__loop_close() for cleanup, but that function assumes that init was good,  Maybe a cleanup function instead of goto?

saghul:
There is not much to cleanup, the key thing is _what_ needs to be cleaned up, depending on where the failure occurs.

saghul:
Ping?

LinuxJedi:
Sorry!  Between conference talks and project deadlines I haven't had a lot of time to revisit this.  I'll aim to take another look in the next week.

PR #1524, opened 2014-10-09
unix: don't abort if mutex init fails

LinuxJedi:
Instead of SIGABRT when mutex init fails gracefully return an error code
and cleanup.

Related to pull request #1522

### DIFF
--- a/src/unix/thread.c
+++ b/src/unix/thread.c
@@ -44,9 +44,8 @@ int uv_mutex_init(uv_mutex_t* mutex) {
   int err;
 
   err = pthread_mutexattr_init(&attr);
-  if (err) {
+  if (err)
     return -err;
-  }
 
   err = pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_ERRORCHECK);
   if (err) {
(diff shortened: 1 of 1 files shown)


saghul:
Please drop braces for single statement conditionals.

LinuxJedi:
sorry, every other project I work on requires them due to things like the Apple double-goto SSL bug :)

Has been fixed

src/unix/thread.c
```diff
@@ -43,16 +43,23 @@ int uv_mutex_init(uv_mutex_t* mutex) {
...
 
   err = pthread_mutex_init(mutex, &attr);
+  if (err) {
+    phread_mutexattr_destroy(&attr);
+    return -err;
+  }
 
-  if (pthread_mutexattr_destroy(&attr))
-    abort();
+  err = pthread_mutexattr_destroy(&attr);
```
saghul, line 62: return -... instead.

saghul:
The whole mutex / semaphore / condvar code is filled with aborts, I'd rather handle them all in a single go than have some functions abort and some return proper errors. As a result, some APIs might need to return int, since now they return void.

LinuxJedi:
hmm... so maybe I should abandon this until a 2.x version?

saghul:
Just send the PR against master, work on it, and we'll merge it when ready. No rush :-)

PR #1526, opened 2014-10-14
Windows: update loop time lazily

orangemocha:
We call uv_update_time() whenever we think the loop time should be updated.
But we might not need to read the time after every update, and the update
operation has non-negligible cost.
So I am making uv_update_time() invalidate the loop time. The time will be
actually updated only on the first call to uv_now() after it's been
invalidated.

### DIFF
--- a/include/uv-win.h
+++ b/include/uv-win.h
@@ -315,7 +315,9 @@ RB_HEAD(uv_timer_tree_s, uv_timer_s);
     /* The loop's I/O completion port */                                      \
   HANDLE iocp;                                                                \
   /* The current time according to the event loop. in msecs. */               \
-  uint64_t time;                                                              \
+  /* Don't access these fields directly. Use uv__loop_time() instead. */      \
+  uint64_t cached_time;                                                       \
+  int cached_time_is_valid;                                                   \
   /* Tail of a single-linked circular queue of pending reqs. If the queue */  \
   /* is empty, tail_ is NULL. If there is only one item, */                   \
   /* tail_->next_req == tail_ */                                              \
--- a/src/win/core.c
+++ b/src/win/core.c
@@ -133,11 +133,8 @@ int uv_loop_init(uv_loop_t* loop) {
   if (loop->iocp == NULL)
     return uv_translate_sys_error(GetLastError());
 
-  /* To prevent uninitialized memory access, loop->time must be intialized
-   * to zero before calling uv_update_time for the first time.
-   */
-  loop->time = 0;
-  uv_update_time(loop);
+  loop->cached_time = 0;
+  loop->cached_time_is_valid = 0;
 
   QUEUE_INIT(&loop->wq);
   QUEUE_INIT(&loop->handle_queue);
--- a/src/win/timer.c
+++ b/src/win/timer.c
@@ -33,14 +33,34 @@
 
 
 void uv_update_time(uv_loop_t* loop) {
-  uint64_t new_time = uv__hrtime(UV__MILLISEC);
-  if (new_time > loop->time) {
-    loop->time = new_time;
-  }
+  /* Invalidate the loop time, so that the next invocation of uv__loop_time()
+   * will actually update it.
+   */
+  loop->cached_time_is_valid = 0;
 }
 
 void uv__time_forward(uv_loop_t* loop, uint64_t msecs) {
-  loop->time += msecs;
+  loop->cached_time += msecs;
+  loop->cached_time_is_valid = 0;
+}
+
+uint64_t uv__loop_time(const uv_loop_t* loop) {
+  uv_loop_t* mutable_loop;
+  uint64_t new_time;
+  if (!loop->cached_time_is_valid) {
+    new_time = uv__hrtime(UV__MILLISEC);
+    mutable_loop = (uv_loop_t*) loop;
+    if (new_time > mutable_loop->cached_time) {
+      mutable_loop->cached_time = new_time;
+    }
+    mutable_loop->cached_time_is_valid = 1;
+  }
+
+  return loop->cached_time;
+}
+
+uint64_t uv_now(const uv_loop_t* loop) {
+  return uv__loop_time(loop);
 }
 
 
@@ -104,7 +124,7 @@ int uv_timer_start(uv_timer_t* handle, uv_timer_cb timer_cb, uint64_t timeout,
     uv_timer_stop(handle);
 
   handle->timer_cb = timer_cb;
-  handle->due = get_clamped_due_time(loop->time, timeout);
+  handle->due = get_clamped_due_time(uv__loop_time(loop), timeout);
   handle->repeat = repeat;
   uv__handle_start(handle);
 
@@ -168,7 +188,7 @@ DWORD uv__next_timeout(const uv_loop_t* loop) {
    */
   timer = RB_MIN(uv_timer_tree_s, &((uv_loop_t*)loop)->timers);
   if (timer) {
-    delta = timer->due - loop->time;
+    delta = timer->due - uv__loop_time(loop);
     if (delta >= UINT_MAX - 1) {
       /* A timeout value of UINT_MAX means infinite, so that's no good. */
       return UINT_MAX - 1;
@@ -190,7 +210,7 @@ void uv_process_timers(uv_loop_t* loop) {
 
   /* Call timer callbacks */
   for (timer = RB_MIN(uv_timer_tree_s, &loop->timers);
-       timer != NULL && timer->due <= loop->time;
+       timer != NULL && timer->due <= uv__loop_time(loop);
        timer = RB_MIN(uv_timer_tree_s, &loop->timers)) {
 
     uv_timer_stop(timer);
--- a/src/unix/timer.c
+++ b/src/unix/timer.c
@@ -26,6 +26,11 @@
 #include <limits.h>
 
 
+uint64_t uv_now(const uv_loop_t* loop) {
+  return loop->time;
+}
+
+
 static int timer_less_than(const struct heap_node* ha,
                            const struct heap_node* hb) {
   const uv_timer_t* a;
--- a/src/uv-common.c
+++ b/src/uv-common.c
@@ -391,12 +391,6 @@ void uv_stop(uv_loop_t* loop) {
 }
 
 
-uint64_t uv_now(const uv_loop_t* loop) {
-  return loop->time;
-}
-
-
-
 size_t uv__count_bufs(const uv_buf_t bufs[], unsigned int nbufs) {
   unsigned int i;
   size_t bytes;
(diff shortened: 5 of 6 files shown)


orangemocha:
cc @saghul , @piscisaureus 
This is a follow up to https://github.com/joyent/libuv/pull/1165
If you don't like the use of uv_now() I could add an internal uv__time(loop_t*) and have uv_now() call that.

saghul:
> If you don't like the use of uv_now() I could add an internal uv__time(loop_t*) and have uv_now() call that.

I think I'd like that, actually :-)

orangemocha:
@txdv : to signify that it's private

saghul:
@orangemocha we don't really use that convention in libuv, basically everything except the data field in a loop is considered private. "time" is fine.

txdv:
Also, note the "txdv commented on 9c100d5 include/uv-win.h:L319 3 hours ago", you can derive by that that I commented on a specific line of code.

You can click on that to get to that line and with comment on my comment on my comment directly, this way the conversation stays nicely formatted.

orangemocha:
@saghul That seems a little bug-prone. The unix side is using loop->time directly. I am worried about people adding some common code that accesses loop->time directly. I'd like them to get a compile error on Windows.
If you don't like the underscore, can I change it to something like 'cached_time'?

txdv:
Will you provide a unix patch?

orangemocha:
Currently, I am leaving the unix side as is. I am debating whether I should introduce uv__time() for unix as well. Let me publish the current change so you can see what I mean. Thanks!

orangemocha:
updated

txdv:
@saghul Didn't we agree to prefix internal functions with uv__? (double _)

uv_update_time would be a perfect target for this.

src/win/timer.c
```diff
@@ -33,14 +33,32 @@
...
+  loop->cached_time_is_valid = 0;
 }
 
 void uv__time_forward(uv_loop_t* loop, uint64_t msecs) {
-  loop->time += msecs;
+  loop->cached_time += msecs;
+  loop->cached_time_is_valid = 0;
+}
+
+uint64_t uv__time(const uv_loop_t* loop) {
```
saghul: `uv__loop_time` maybe?

src/win/timer.c
```diff
@@ -33,14 +33,32 @@
...
+      mutable_loop->cached_time = new_time;
+    }
+    mutable_loop->cached_time_is_valid = 1;
+  }
+
+  return loop->cached_time;
+}
+
+uint64_t uv_now(const uv_loop_t* loop) {
+  return uv__time(loop);
```
saghul: Cast to uv_loop_t\* before passing it perhaps?

src/win/timer.c
```diff
@@ -33,14 +33,32 @@
...
-  if (new_time > loop->time) {
-    loop->time = new_time;
-  }
+  /* Invalidate the loop time, so that uv__time() will actually update it. */
+  loop->cached_time_is_valid = 0;
 }
 
 void uv__time_forward(uv_loop_t* loop, uint64_t msecs) {
-  loop->time += msecs;
+  loop->cached_time += msecs;
```
saghul, line 43: Isn't this a bit dangerous? We's need to make sure uv__time_forward is called always _after_ updating the time, right?

src/win/timer.c
```diff
@@ -33,14 +33,32 @@
...
+  loop->cached_time_is_valid = 0;
+}
+
+uint64_t uv__time(const uv_loop_t* loop) {
+  uv_loop_t* mutable_loop;
+  uint64_t new_time;
+  if (!loop->cached_time_is_valid) {
+    new_time = uv__hrtime(UV__MILLISEC);
+    mutable_loop = (uv_loop_t*) loop;
+    if (new_time > mutable_loop->cached_time) {
```
saghul, line 53: How could this happen?

src/win/core.c
```diff
@@ -133,11 +133,11 @@ int uv_loop_init(uv_loop_t* loop) {
   if (loop->iocp == NULL)
     return uv_translate_sys_error(GetLastError());
 
-  /* To prevent uninitialized memory access, loop->time must be intialized
+  /* To prevent uninitialized memory access, cached_time must be intialized
```
saghul: Does this comment still apply since we don't call uv_update_time here anymore?

include/uv-win.h
```diff
@@ -315,7 +315,9 @@ RB_HEAD(uv_timer_tree_s, uv_timer_s);
     /* The loop's I/O completion port */                                      \
   HANDLE iocp;                                                                \
   /* The current time according to the event loop. in msecs. */               \
-  uint64_t time;                                                              \
+  /* Don't access these fields directly. Use uv__time() instead. */           \
+  uint64_t cached_time;                                                       \
+  int cached_time_is_valid;                                                   \
```
saghul, line 320: Why did you go with 2 attributes instead of a single one like on the first patch?

saghul:
> @saghul Didn't we agree to prefix internal functions with uv__? (double _)
> 
> uv_update_time would be a perfect target for this.

Yes.

include/uv-win.h
txdv, line 320: He changed the name of the first attribute so it would reflect its purpose better.

cached_time -> time.

saghul, line 320: No, before that. The original patch used loop->time and it invalidated it by setting it to 0.

txdv:
uv_update_time is not internal, I assumed for a moment that it was. Forget my stupid comment.

saghul:
Quick (perhaps stupid) idea: get rid of uv_update_time and modify uv_now (remove constness + docs) to do it. Also on Unix.

txdv:
To do what exactly?

saghul:
> To do what exactly?

To do what uv_update_time currently does.

txdv:
Right now the time is being updated at the beginning of the event loop. Then you can call uv_now to get that 'cached' time.

Now if you would have long(er) operation, you might want to update the time manually, which is why uv_update_time exists. You call it and then uv_now returns the new most currently cached time.

Now if you merge uv_update_time and uv_now, and call uv_now a lot of times in a row, it will be reduce performance.

The existence of uv_update_time lets us be as fast as possible in most situations while letting us be as accurate as possible when the developer invokes uv_update_time. I would like to merge that into one function as well, but I don't think that is possible.

This makes me just realize that making uv_now lazy is kinda useless, because when the developer invokes uv_update_time he really means to update it now.

saghul:
`uv_now` is pretty useless, since it returns the time perception of the loop, if you want high precission time there is `uv_hrtime` for that. I seriously doubt anybody uses `uv_now` and would be hit by this, though I could be wrong.

src/win/timer.c
orangemocha: sure

src/win/timer.c
orangemocha, line 43: The purpose of uv__time_forward is to ensure that when we finish polling with a timeout, the timeout value is reflected in the loop time (because as we saw, GetQueuedCompletionStatus can return a tad early). "the timeout value is reflected in the loop time" is equivalent to saying that the difference between the next time the time is read (with uv__loop_time()) and the last time it was read must be at least 'msecs'. This code satisfies this requirement, because cached_time is the loop time as it was read last.

src/win/timer.c
orangemocha, line 53: Same logic as before this change: because of uv__time_forward(). the current time (new_time) could actually be lower than cached_time, if GetQueuedCompletionStatus returned early.

src/win/core.c
orangemocha: It should be initialized, but probably not for the same reasons as pointed out in this comment. I'll just remove the comment, and leave the initialization, which is pretty uncontroversial.

include/uv-win.h
orangemocha, line 320: That's because in the previous patch we had nuked uv__time_forward, so it was sufficient to set it to 0 to mean 'invalidated'. Then we realized that we needed to bring uv__time_forward back. So we need to store the minimum value for the next readout, even while the cachec time is invalidated.

orangemocha:
I agree with @txdv:

> The existence of uv_update_time lets us be as fast as possible in most situations while letting us be as accurate as possible when the developer invokes uv_update_time.

uv_update_time is cheap to call and we call it any time we know the time might have changed. If we didn't use it, we would have to update the time every time we read it, which would be less performant and also change the behavior. It's not uv_now() that we need to worry about, but uv__loop_time(), which is used for timer manipulation.

src/win/timer.c
orangemocha: I think it's better to encapsulate the cast inside uv__loop_time (formerly uv__time) rather than having a cast in all the sites that call uv__loop_time. Conceptually, uv__loop_time is a read operation and uv_update_time modifies the state of the loop. The fact that the time value is cached and updated at read time is an implementation detail and should not affect the function's signature. 
BTW caching is the most typical example of 'conceptual constness', for which C++ introduced the mutable keyword. That would have been appropriate here, but since we are using C we are stuck with the cast.

orangemocha:
Pushed a new commit with changes from code review feedback above.

I am thinking it might be better to introduce uv__loop_time on Unix as well, to keep a consistent internal interface, even if doesn't do the caching there. What do you think?

txdv:
Why would someone not call uv_now after he called uv_update_time?

uv_update_time is used internally though ... maybe it might make sense.

orangemocha:
Correct, uv_update_time is used internally, it seems mostly to support timer operations.

txdv:
Not is not, it is an exposed function in the uv.h header

orangemocha:
I meant it's used internally, in addition to being part of the public API. The internal use is for timers and it doesn't imply that uv_now() will be called afterwards.

txdv:
Yeah, that makes sense in that case. Saves maybe a few time queries.

saghul:
> I meant it's used internally, in addition to being part of the public API. The internal use is for timers and it doesn't imply that uv_now() will be called afterwards.

My point is that `uv_now` is not designed to be precise. The fact that on Windows is precise is a coincidence. This becomes clearer when looking at the Unix implementation, because it uses different clock sources for `uv_now` and `uv_hrtime`: https://github.com/joyent/libuv/blob/master/src/unix/internal.h#L293-L297 and https://github.com/joyent/libuv/blob/master/src/unix/core.c#L89-L91

That said, what I proposed would make `uv_now` update the loop time + return it, thus "deprecating" `uv_update_time`. Those who were manually updating the time by calling `uv_update_time` would just call `uv_now`. Alternatively we could remove `uv_now` completely since, as I said earlier, it's pretty useless. Then we'd just leave `uv_update_time` and there would be no way to fetch the loop's time, which I'd say it's ok.

saghul:
@piscisaureus mind throwing your 2 cents here?

PR #1528, opened 2014-10-16
unix, windows: don't include null byte in uv_cwd size

saghul:
Make it consistent with uv_exepath


saghul:
@creationix, can you please test this out? Also, @indutny, this will need a change in Node, I'll send a PR.

creationix:
Tested in linux (arch chroot on chromebook) while waiting for windows machine to boot and build.
Works in windows too.

+1

Issue #829, opened 2013-06-06
unix, windows: always defer reporting runtime errors to callback


bnoordhuis:
Right now, we're inconsistent. Most functions always defer
while others sometimes return an error code. Examples:
- uv_pipe_connect() - errors are always passed to the callback. (Good.)
- uv_udp_send() - returns an error code when it fails to allocate a file descriptor. (Bad.)

Only return errors for bad arguments (EINVAL) with a compile-time option to turn those into asserts for easier debugging.

Fix before v0.12, it lets us simplify error handling in node.

txdv:
Why do you think it is bad to return an error code immediately?

The negative stuff I can come up:
1. From a C perspective it means double checking, immediately when the function is called and in the callback. A bit tedious, but it is C, callback code with errors is already tedious. Can be countered by putting error checking code in a function (no separate locations of error checking), but it is even more verbose (need for another function, which is called just basically in two different places).
2. If one OS returns the same error in the callback and another directly, it will be inconsistent and a program handling exceptions correctly on one operating system might not handle them correctly unless the dev wrote correct code (needs more detail to attention by the dev). We can circumvent this in libuv directly by covering all the corner cases and queue the error code and make it return in the callback.

The positive stuff about immediately returning the error:
1. It is faster. Way faster. We are not going through the entire queueing process, it is one ret assembly as opposed to the entire pending request queue code. That being said functions returning immediately error codes should be a rather rare event. Can't create a socket, not enough memory, invalid arguments.

A rewrite of ALL tests would be needed, so this goes against making it return immediately I guess.

I feel like the overhead is against the very grain of C, but it is not thaaaaat much overhead, because erroneous situations are not common.

The change would make it definitely easier to write libuv C code, since all the errors checking code would be in the callback (only one place instead of two).

I'm actually for this, maybe people would write more C code instead of javascript code.

As far as it goes for binding simplicity, I have this: https://github.com/txdv/LibuvSharp/blob/master/LibuvSharp/Tcp.cs#L108-L130 its C#, but I guess you can still understand it...  new ConnectRequest() might throw an exception, so I need to wrap it myself in a try { } and call the appropriate callback, not too tedious though.

If I get an OK from the core team, I will write this patch. 

Any comments @bnoordhuis @saghul?

bnoordhuis:
> Why do you think it is bad to return an error code immediately?

[...]

PR #1538, opened 2014-10-22
API change: defer everything to the callback

txdv:
Makes it easier for use for the user to use the library, because the user now has only to do error checking in one place instead of two. 

Discussion: #829


saghul:
I don't like the idea of deferring bogus parameter errors to the callback.

Ben's original proposal (which I like too) contained that last bit about bad arguments returning EINVAL on debug builds and triggering an assert on release builds. I find that very useful. Moreover, What do you plan to do with functions such as `uv_timer_start`, which cannot get a NULL callback? Return EINVAL, of course.

You said you wanted write an RFC about error handling, I'd rather see that first before jumping into writing any code.

huxingyi:
+1 for no return value. (Some apis have no callback, then could add a callback)
It will ease the user level coding and reduce bugs which was caused by forgetting to process the return value.

txdv:
The entire discussion is here https://github.com/joyent/libuv/issues/829. We should talk about this in bnoordhuis issue and use this pull request to comment on the code strictly. So I'll post your questions there and then add my answers.

piscisaureus:
I agree that errors should be deferred to the callback as much as possible.

In principle I agree with @bnoordhuis's suggestion that parameter errors should either cause crashes or be reported synchronously. In node, for example, we'll throw if the user doesn't specify a callback, or if the user calls fs.open without specifying a filename. It's simply impossible to interpret what the user might have wanted, and if there's no callback node can't even defer the callback.

In libuv this type of error typically causes a crash. That's because we can't verify that the user specified a valid loop pointer to uv_tcp_init(), or whether a callback pointer is valid or not. In these situations crashing is the only thing we can do.

However @txdv is right about the fact that if parameter verification is taken a step further, it becomes an extremely fine line. Some examples:
- attempt to open a file but the filename isn't well-formed
- attempt to write to a shutdown()-ed socket
- attempt to write to a closing socket
- attempt to uv_connect() a stream handle that's identified to be a server (and not a connection)
- attempt to write an empty buffer

In my opinion, all of these errors would ideally deferred.

This is easy enough to define for functions like uv_write() that use the "request-callback" pattern.
But libuv currently doesn't always do that. Functions like `uv_read_start(stream, cb)` and `uv_listen(stream, cb)` put the handle in a certain "state" and then fire callbacks repeatedly.
To define what should happen here is a bit more difficult - e.g. should `uv_read_start()` on a server handle cause a one-time invocation of the read callback, or do we report the failure synchronously?

 Note I think that these APIs will be replaced by "request-callback" like APIs in the future.

txdv:
Yours pointed out should result in EINVAL too? And you think they should be deferred?

You forgot ENOMEM. everyone forgets ENOMEM, but that one is easy to defer.

As for for uv_listen and uv_read_start, YES, they will become request callback APIs.

txdv:
no resolution

saghul:
I've kept this one on the back of my head, and the more I thought of it, the messier it becomes. IMHO, we just cannot defer everything to the callback because the close callback has to be last one to be called, so if the user calls `uv_write` after that it needs to fail synchronously. I couldn't think of a way around this.

Also, we have functions like `uv_try_write` which don't take a callback but work similarly to `uv_write`, so making one report some errors inline and the other defer them could be seen as an inconsistency.

So, my current state of mind is that I find trouble wording how the behavior should be, without having tons of exceptions.

I think, however, that a refined version of what we have could work better and could potentially be easier to explain: "if the function returns 0 it's guaranteed that the callback will be called, if it returned < 0 then the callback won't be called.". Then, we'd only defer errors to the callback under special circumstances:
- platform differences, as in the bind / listen EADDRINUSE (IIRC)
- in case we try to perform a blocking operation (as in uv_write)
- ?

Even if we won't use the code in this PR, I'll open an issue about this on libuv/libuv in order to further investigate this.

txdv:
pipe_connect is deffering EINVAL has currently inconsistent API

saghul:
We are currently inconsistent in a few places, yes, that's why we need to
find the best way forward.

PR #1541, opened 2014-10-23
unix: remove overzealous assert in uv_read_stop

saghul:
Also make it a noop if we weren't reading at all.


txdv:
I tried to reason what that monster assert does but I failed.

The entire statement has to be true and since it is ORed together.

One of these statements has to be true:
- UV_POLLOUT has to be not set (we shouldn't be writing)
- The write queue should not be empty
- The write completed queue should not bet empty
- There should be a shutdown request
- There should be a connect request

I am not sure about the last 2, because it might be that 'there should have been a shutdown request' holds, if the fields are not set appropriately to NULL after the request have been issued.

@bnoordhuis has written this thing, as far as I can tell from the merge history (since he merged one half on his name from multiple commits).

Could you take a look at this?

imzyxwvu:
The following code checks a cond appeared in the assert:

```
 if (!uv__io_active(&stream->io_watcher, UV__POLLOUT))
  uv__handle_stop(stream);
```

This is uv_**read**_stop, why does it stop the entire handle? If later I want to write to it, will it cause problems? So the assertion is to make sure it won't be writable any more if it isn't writing currently? I am new to libuv internal so I want to know about what uv__handle_stop does. Is it just like uv_close?

I don't know what does uv_read_stop do, and I think the document should be completed.

txdv:
You can stop/start as much as you want.

bnoordhuis:
@txdv Yes, I wrote that.  It tests that UV_POLLOUT is clear OR there is an action in progress that warrants setting UV_POLLOUT.

imzyxwvu:
@bnoordhuis , is there an better way to test the flags? returning error or doing nothing?

bnoordhuis:
It's an internal consistency check.  If you hit that assert, it means there is a bug in libuv or your code.  (Assuming the check itself isn't buggy, of course.)

saghul:
I'll reopen this at libuv/libuv.

PR #1542, opened 2014-10-23
win: fix lost of OnLinkPrefixLength on xp.

huxingyi:
Use OnLinkPrefixLength will cause compile error on windows xp.
sizeof(IP_ADAPTER_UNICAST_ADDRESS_XP) = 48, is aligned.


src/win/util.c
```diff
@@ -1037,7 +1037,8 @@ int uv_interface_addresses(uv_interface_address_t** addresses_ptr,
 
       /* XP has no OnLinkPrefixLength field. */
       if (is_vista_or_greater) {
-        prefix_len = unicast_address->OnLinkPrefixLength;
+        prefix_len = *(UINT8 *)((IP_ADAPTER_UNICAST_ADDRESS_XP *)unicast_address
```
orangemocha, line 1040: Well, that isn't exactly the most portable way of doing it. Can you do something like:

`prefix_len = ((_IP_ADAPTER_UNICAST_ADDRESS_LH *) unicast_address)->OnLinkPrefixLength;
`

huxingyi, line 1040: OnLinkPrefixLength is not defined on old windows. This is why I made such a PR.

`if (is_vista_or_greater) {` is runtime check, not compile time. Compiling `unicast_address->OnLinkPrefixLength;` in mingw on windows xp will caused compile error.

orangemocha, line 1040: _IP_ADAPTER_UNICAST_ADDRESS_LH should be defined in the Windows SDK for all versions of Visual Studio we support.
For mingw, you can defined it in winsock.h like we do for _IP_ADAPTER_UNICAST_ADDRESS_XP: https://github.com/joyent/libuv/blob/master/src/win/winsock.h#L150

huxingyi, line 1040: Ok. I will make a new PR and close this.

PR #1564, opened 2014-11-07
win: fix uv_thread_self()

orangemocha:
59658a8de7cc05a58327a164fd2ed4b050f8b4f4 changed uv_thread_self()
to return uv_thread_t, but uv_thread_t is a thread's HANDLE while
uv_thread_self() returns the current thread's id.
This means that uv_thread_equal() is also broken, as we are
potentially comparing HANDLES to ids.

Changed uv_thread_t to DWORD.
Fixed small doc issue.

### DIFF
--- a/docs/src/threading.rst
+++ b/docs/src/threading.rst
@@ -56,7 +56,7 @@ Threads
 ^^^^^^^
 
 .. c:function:: int uv_thread_create(uv_thread_t* tid, uv_thread_cb entry, void* arg)
-.. c:function:: unsigned long uv_thread_self(void)
+.. c:function:: uv_thread_t uv_thread_self(void)
 .. c:function:: int uv_thread_join(uv_thread_t *tid)
 .. c:function:: int uv_thread_equal(const uv_thread_t* t1, const uv_thread_t* t2)
 
--- a/include/uv-win.h
+++ b/include/uv-win.h
@@ -232,7 +232,7 @@ typedef int uv_file;
 typedef SOCKET uv_os_sock_t;
 typedef HANDLE uv_os_fd_t;
 
-typedef HANDLE uv_thread_t;
+typedef DWORD uv_thread_t;
 
 typedef HANDLE uv_sem_t;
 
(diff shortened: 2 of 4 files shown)


orangemocha:
This will need some follow up because now we are leaking the HANDLE return by _beginthreadex

piscisaureus:
@orangemocha @saghul 
My suggestion would be to partially revert https://github.com/joyent/libuv/commit/59658a8de7cc05a58327a164fd2ed4b050f8b4f4

uv_thread_self() should not return a HANDLE but just the thread id.
uv_thread_equal() should just compare two unsigned longs on windows.

saghul:
I'm ok with that. In the end we treat uv_thread_t as opaque, so it
shouldn't matter much.

On Friday, November 7, 2014, Bert Belder notifications@github.com wrote:

> @orangemocha https://github.com/orangemocha @saghul
> https://github.com/saghul
> My suggestion would be to partially revert 59658a8
> https://github.com/joyent/libuv/commit/59658a8de7cc05a58327a164fd2ed4b050f8b4f4
> 
> uv_thread_self() should not return a HANDLE but just the thread id.
> uv_thread_equal() should just compare two unsigned longs on windows.
> 
> —
> https://github.com/joyent/libuv/pull/1564#issuecomment-62197647.

## 

/Saúl
bettercallsaghul.com

orangemocha:
> uv_thread_self() should not return a HANDLE but just the thread id.
> uv_thread_equal() should just compare two unsigned longs on windows.

The issue with that is that uv_thread_start() returns uv_thread_t and you wouldn't be able to compare that with the return from uv_thread_self(). In fact, you would only be able to compare the return of uv_thread_self() with another value returned by the same function.

Per discussion with @piscisaureus , the viable options seem either a) to either extend the API to distinguish the thread object type the thread id type, or b) to implement the current API correctly on Windows, which would require being able to retrieve the creation handle for the current thread (using TLS). I am investigating further.

orangemocha:
Better solution at https://github.com/joyent/libuv/pull/1570. Closing this one.

Issue #1559, opened 2014-11-04
`make check` failed: get_currentexe; but, succeeded with gyp ## get_currentexe is not support symlink yet


usagi:
version: current HEAD a4f88760be1838603fe2eae89a651066cc42eedd
environment: ubuntu-14.04/x86_64, g++-4.8.2

`make` is succeeded, but `make check` is failed.

```
./autogen.sh
./configure
make
make check
```

```
=============================================================
[%  58|+ 142|-   0|T   0|S   0]: get_currentexe
`get_currentexe` failed: exit code 134
Output from process `get_currentexe`:
Assertion failed in test/test-get-currentexe.c on line 53: match && !strcmp(match, path)
=============================================================
[% 100|+ 242|-   1|T   0|S   0]: Done.
FAIL: test/run-tests
=======================================================
1 of 1 test failed
Please report to https://github.com/joyent/libuv/issues
=======================================================
```

---

But, it is succeeded use gyp.

```
./gyp_uv.py -f make
make -C out
./out/Debug/run-tests
```

```
=============================================================
[% 100|+ 243|-   0|T   0|S   0]: Done.
```

usagi:
I got the cause and the answer for the issue.

I was injected the debugging code into the test code:

```
FILE* f = fopen("/tmp/log", "w");
fprintf(f, "%s\n", executable_path);
fprintf(f, "%s\n", path);
fprintf(f, "%s\n", buffer);
fclose(f);
```

I got the result with `./autogen && ./configure && make && make check`:

```
/home/usagi/repos/libuv/test/.libs/lt-run-tests
/home/usagi/repos/libuv/test/.libs/lt-run-tests
/home/usagi/Dropbox/repos/libuv/test/.libs/lt-run-tests
```

And, `./gyp_uv.py -f make && make -C out && ./out/Debug/run-tests`:

```
./out/Debug/run-tests
out/Debug/run-tests
/home/usagi/Dropbox/repos/libuv/out/Debug/run-tests
```

The cause is current implement of `executable_path`. `executable_path` is not support symlinks. "/home/usagi/repos" in my environment is symlink, it is link to "Dropbox/repos".

Then, I propose a fixup to the test code `test-get-currentexe.c`:

The source of `executable_path` change from `argv[0]` to `/proc/self/exe` in GNU/Linux or `/proc/curproc/file` in BSD(and Darwin) and read it with support symlink resolver as `readlink` by `unistd.h`. But, I don't know how to solve in Windows.

[...]

PR #1565, opened 2014-11-08
test: support unix symlink

usagi:
The pull-request by:
https://github.com/joyent/libuv/issues/1559#issuecomment-62137079

It solve the problem that is failure to the test as `test-get-currentexe` unexpected if run with a path included symlinks.


saghul:
Should be fixed by https://github.com/libuv/libuv/commit/ff29322b99533fde798ad4c4ee5344836aa2a11a

PR #1567, opened 2014-11-09
Fix uv_tty_set_mode / use cfmakeraw() when available.

wavexx:
The uv_tty_set_mode(tty, 1) should switch to raw mode, but currently doesn't set all the required flags. The resulting TTY is still not suitable for binary I/O.

In this patch we make use of cfmakeraw() (if available), or set the same flags that cfmakeraw() would use (taken from the glibc implementation).

I'm trying to push this change upstream for the JuliaLang fork: https://github.com/JuliaLang/libuv/pull/27


bnoordhuis:
LGTM but is there a reason to call cfmakeraw() conditionally?  As far as I can tell, all supported platforms support it.

wavexx:
On 11/10/2014 04:39 PM, Ben Noordhuis wrote:

> LGTM but is there a reason to call cfmakeraw() conditionally?  As far as I can tell, all supported platforms support it.

On glibc and BSDs it's supported.

I don't have access to all configurations (uclibc maybe?) to tell if
this is always the case.

wavexx:
On 11/10/2014 07:27 PM, Yuri D'Elia wrote:

> On 11/10/2014 04:39 PM, Ben Noordhuis wrote:
> 
> > LGTM but is there a reason to call cfmakeraw() conditionally?  As far as I can tell, all supported platforms support it.
> 
> On glibc and BSDs it's supported.
> 
> I don't have access to all configurations (uclibc maybe?) to tell if
> this is always the case.

Actually, just checked, it's supported also in uclibc.
Should I remove the conditional?

bnoordhuis:
Yes, please.

wavexx:
Done.

saghul:
Cheers @wavexx, landed in 0f25560 :sparkles:

cjihrig:
This change (related to `c_oflag`) caused https://github.com/joyent/node/issues/8817. Is this something that should be fixed in libuv or node/iojs?

bnoordhuis:
@cjihrig If reverting just this change fixes it, then it's likely a libuv issue.  If you file an issue at libuv/libuv and cc me, I'll look into it.

(EDIT: FWIW, I can't reproduce it with x86_64 io.js HEAD on Linux.)

cjihrig:
Thanks @bnoordhuis. Yes, just reverting this change fixes it. Technically, I'm able to fix the problem with joyent/node by just adding `raw.c_oflag = OPOST|ONLCR;` after the call to `cfmakeraw()`.

It's not an issue on io.js yet because this commit [hasn't landed](https://github.com/iojs/io.js/blob/v0.12/deps/uv/src/unix/tty.c#L126) there yet.

I'll open an issue on libuv/libuv.

bnoordhuis:
@wavexx Heads up, this change had to be reverted.  If you want to pursue this further, please file an issue at libuv/libuv.  I think we'll need to consider a flag or function for selecting between mostly raw and really raw mode.

wavexx:
On 12/04/2014 06:46 PM, Colin Ihrig wrote:

> This change (related to `c_oflag`) caused
> https://github.com/joyent/node/issues/8817. Is this something that
> should be fixed in libuv or node/iojs?

Without looking at the code, my take would be an issue in iojs though.
What is iojs expecting to achieve by requesting raw i/o and still using
ONLCR?

bnoordhuis:
It's possible but I don't want to introduce a regression in the stable branch.  If you file an issue, we'll see what we can come up with.

wavexx:
On 12/05/2014 05:32 PM, Ben Noordhuis wrote:

> It's possible but I don't want to introduce a regression in the
> stable branch.  If you file an issue, we'll see what we can come up
> with.

File an issue, where (iojs)?

The thing is, I need clean I/O to communicate with subprocesses.
I could handle another raw state (0/1/2 in the mode flag), but that
feels backwards to me. The current "raw" state is a hybrid which is
currently abused by Julia as well (incorrectly!).

I will fix Julia's, but I don't really want to mess with another library
of which I have no clue.

If I can push this change forward with a different mode flag, I will.
But this thing needs a mention in the README somewhere, stating that the
previous mode is broken.

bnoordhuis:
> File an issue, where (iojs)?

https://github.com/libuv/libuv/issues - I'll take care of propagating things to io.js.

wavexx:
On 12/05/2014 10:27 PM, Ben Noordhuis wrote:

> > File an issue, where (iojs)?
> 
> https://github.com/libuv/libuv/issues - I'll take care of propagating things to io.js.

In this case, should I be filing the same report as this one, or ...?

PR #1569, opened 2014-11-10
Update misc.rst

shadowmanos:
The link was broken because it contains an underscore which is a markup character.


saghul:
Thanks Manos! Landed in 0fc24f6

PR #1570, opened 2014-11-10
win: fix uv_thread_self()

orangemocha:
59658a8de7cc05a58327a164fd2ed4b050f8b4f4 changed uv_thread_self()
to return uv_thread_t, but uv_thread_t is a thread's HANDLE while
uv_thread_self() returns the current thread's id.
This means that uv_thread_equal() is also broken, as we are
potentially comparing HANDLES to ids.

Changed uv_thread_self() to return the current thread's creation handle.
Fixed small doc issue.


piscisaureus:
LGTM, nice work @orangemocha !

saghul:
Thanks Alexis (and Bert for reviewing)! Landed in 6591d00 :sparkles:

orangemocha:
Thanks guys!

PR #1572, opened 2014-11-10
Remove Rust from users

elijahandrews:
As of version 0.12.0, Rust no longer uses libuv.

https://mail.mozilla.org/pipermail/rust-dev/2014-October/011267.html
https://github.com/rust-lang/rfcs/pull/230


txdv:
Have you shipped it off to an external package or such?

samlh:
https://github.com/alexcrichton/green-rs

PR #1555, opened 2014-10-27
win: add missing IP_ADAPTER_UNICAST_ADDRESS_LH definition for MinGW

huxingyi:
fix https://github.com/joyent/libuv/pull/1542

PR #1575, opened 2014-11-11
Fix mingw32 build

mattn:
mingw32 doesn't have OnLinkPrefixLength

### DIFF
--- a/src/win/util.c
+++ b/src/win/util.c
@@ -1036,9 +1036,12 @@ int uv_interface_addresses(uv_interface_address_t** addresses_ptr,
       sa = unicast_address->Address.lpSockaddr;
 
       /* XP has no OnLinkPrefixLength field. */
+#ifndef __MINGW32__
       if (is_vista_or_greater) {
         prefix_len = unicast_address->OnLinkPrefixLength;
-      } else {
+      } else
+#endif
+      {
         /* Prior to Windows Vista the FirstPrefix pointed to the list with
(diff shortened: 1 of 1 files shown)


saghul:
Does MinGW64 also not have it?

mattn:
It seems OK to me 

https://gitorious.org/w64/w64/source/389d24fb1dc74d849fcf3794989924c39d6c0a18:mingw-w64-headers/include/iptypes.h#L96-105

piscisaureus:
I disagree with this patch; on mingw builds `prefix_len` will simply be never assigned a value, leading to undefined behavior further down. 

This issue isn't caused by mingw not supporting this API; it's simply lacking the right definitions in one of it's header files. Libuv typically carries definitions that certain toolchains might lack; take a look at `src/win/winapi.h`, that's where a solution should go.

mattn:
@piscisaureus Ah, I notice it just now. I'll fix tommorow

huxingyi:
already fixed in https://github.com/joyent/libuv/pull/1555

mattn:
That is in 1.X branch. please backport to master

saghul:
1.x is the stable branch.

mattn:
Yes, I know. When I will be able to build master branch on mingw32 ?

saghul:
I'll merge 1.x into master after the next release. After we changed the
versioning scheme I'd suggest to stick to the 1.x branch. master will
eventually introduce breaking changes without notice, since that's where
new development will happen.

PR #1577, opened 2014-11-12
Add uv_device_t as stream on windows try to handle device io

zhaozg:
Please help me to check right or wrong.


txdv:
Why are there commits by other people which are already in the mainline?

Open up a pull request against 1.x, not against master

zhaozg:
Some mistake

saghul:
If you plan on working on a big feature like this I suggest you comment on the rationale. Also, I can already see someone asking for baud rates and such, for serial devices. I'm not sure we want to have any of that in libuv.

zhaozg:
I will make a new pr,and with more comments

PR #1580, opened 2014-11-13
Add uv_device_t as stream on windows and linux, handle device I/O asyncly.

zhaozg:
I'm a multi-language programmer, but not family with english, so maybe trouble with options. libuv is a good i/o framework, and easy to extent, I hope persist on it to do my work.

When I want to do r/w device on windows, I try filesystem api, it works for device handle, but not fit my needs. I need a real stream, I try wrap handle as pipe, and fail to work. When I study TTY and TCP, I realize that it not hard to implication a device type stream, because just file operation with overlapped,  even simpler than other stream types.
When I finish my job on windows, i realize libUV is born for linux, simple to add a device type, so i did it.

There are only three type related api.

``` c
UV_EXTERN int uv_device_init(uv_loop_t* handle, uv_device_t* device,  const char* path, int flags);
UV_EXTERN int uv_device_ioctl(uv_device_t* device, int cmd, uv_ioargs_t* args);
```

uv_device_init will init and create a uv_deive_t, path should be '/dev/??' on linux or "\.\Global\???" on windows.
flags should be  O_RDONLY,O_WRONLY  or O_RDWR.

uv_device_ioctl will set or get device paramater,  arg is relative with OS platform, return >=0 for success, detail please see header file.

because not test widely, so maybe have bugs, please attention.

Happy for do something for libUV.


src/win/device.c
```diff
@@ -0,0 +1,383 @@
...
+  req->handle = (uv_stream_t*) handle;
+  req->cb = cb;
+  req->queued_bytes = 0;
+  memset(&(req->overlapped), 0, sizeof(req->overlapped));
+
+  result = WriteFile(handle->handle,
+    bufs[0].base,
+    bufs[0].len,
+    &bytes,
+    &req->overlapped);
```
cjdelisle: My experience with windows has been that you get called back even if the write is synchronous so you can safely ignore "returned immediately" scenarios.

cjdelisle:
In general it looks like it will work for my use case, I'm still (of course) favorable to my own patch, in this case because it's fewer lines of code, but if this is code which I do not maintain, I'm happy with that so +1 in theory. If you feel like patching https://github.com/cjdelisle/cjdns/blob/master/interface/tuntap/windows/TAPInterface.c to work with your version then I'll really love you :)

saghul:
Before getting into reviewing any code, I'd like to hear some general thoughts from @indutny, @bnoordhuis and @piscisaureus.

The approach taken here looks clean and nice to me, if that's enough for people wanting to use serial devices (for example) not to (ab)use uv_fs_open, I'm all in :-)

src/win/device.c
piscisaureus: It could be, if SetFileCompletionNotificationModes had been used.

But yes by default IOCP reports all succesful operations, even if they completed synchronously. So that looks like it needs to be fixed here.

zhaozg: Fixed in https://github.com/zhaozg/libuv/commit/ee6d4c4b883a348e333d9f3f49e06c9518dd3b19

include/uv-win.h
```diff
@@ -453,6 +453,10 @@ RB_HEAD(uv_timer_tree_s, uv_timer_s);
   LPFN_WSARECV func_wsarecv;                                                  \
   LPFN_WSARECVFROM func_wsarecvfrom;
 
+#define UV_DEVICE_PRIVATE_FIELDS                                              \
+  HANDLE handle;                                                              \
+  uv_buf_t buf;
```
saghul: why is this needed?

include/uv.h
```diff
@@ -614,6 +616,20 @@ UV_EXTERN int uv_udp_recv_stop(uv_udp_t* handle);
...
+ *
+ * Represents a device stream.
+ */
+struct uv_device_s {
+  UV_HANDLE_FIELDS
+  UV_STREAM_FIELDS
+  UV_DEVICE_PRIVATE_FIELDS
+};
+
+UV_EXTERN int uv_device_open(uv_loop_t* handle, uv_device_t* device, const char*path, int flags);
```
saghul: Handles are initialized with uv_*_init, this is a convention we use in libuv, so this should be uv_device_init.

include/uv-win.h
```diff
@@ -479,6 +483,18 @@ RB_HEAD(uv_timer_tree_s, uv_timer_s);
     struct { uv_pipe_connection_fields };                                     \
   };
 
+/* should change to *nux style */
```
saghul: Please do make it more Unix style :-)

src/unix/device.c
```diff
@@ -0,0 +1,56 @@
...
+#include "internal.h"
+
+#include <stdlib.h>
+#include <unistd.h>
+#include <sys/ioctl.h>
+
+int uv_device_open(uv_loop_t* loop, uv_device_t* device, const char*path, int flags) {
+  int fd;
+  int stream_flags;
+  if( (fd = open(path, flags)) < 0 ) {
```
saghul: Please split this in 2 lines.

src/unix/device.c
```diff
@@ -0,0 +1,56 @@
...
+    return fd;
```
saghul: You need to return -errno.

src/unix/device.c
```diff
@@ -0,0 +1,56 @@
...
+    stream_flags |= UV_STREAM_WRITABLE;
+  if ( flags & O_RDWR )
+    stream_flags |= UV_STREAM_READABLE | UV_STREAM_WRITABLE;
+
+  uv__nonblock(fd, 1);
+  return uv__stream_open((uv_stream_t*)device, fd, stream_flags);
+}
+
+int uv_device_ioctl(uv_device_t* device, int cmd, void* arg) {
+  return ioctl(uv__stream_fd((uv_stream_t*)device), cmd, arg);
```
saghul: Check if the return value is -1 and return -errno in that case.

src/unix/device.c
```diff
@@ -0,0 +1,56 @@
...
+
+  stream_flags = 0;
+  if ( flags & O_RDONLY ) 
+    stream_flags |= UV_STREAM_READABLE;
+  if ( flags & O_WRONLY )
+    stream_flags |= UV_STREAM_WRITABLE;
+  if ( flags & O_RDWR )
+    stream_flags |= UV_STREAM_READABLE | UV_STREAM_WRITABLE;
+
+  uv__nonblock(fd, 1);
```
saghul: Check the return value, it could fail.

src/unix/device.c
```diff
@@ -0,0 +1,56 @@
...
+  return uv__stream_open((uv_stream_t*)device, fd, stream_flags);
```
saghul: Check the return value, if it's < 0 then close the fd and propagate.

src/unix/stream.c
```diff
@@ -1296,7 +1297,8 @@ int uv_write2(uv_write_t* req,
   assert(nbufs > 0);
   assert((stream->type == UV_TCP ||
           stream->type == UV_NAMED_PIPE ||
-          stream->type == UV_TTY) &&
+          stream->type == UV_TTY || 
+          stream->type == UV_DEVICE) &&
```
saghul: Does write2 really work on uv_device?

src/uv-common.c
```diff
@@ -308,7 +308,7 @@ int uv_thread_create(uv_thread_t *tid, void (*entry)(void *arg), void *arg) {
 
 uv_thread_t uv_thread_self(void) {
 #ifdef _WIN32
-  return GetCurrentThreadId();
+  return (uv_thread_t)GetCurrentThreadId();
```
saghul: Unrelated change.

src/win/device.c
```diff
@@ -0,0 +1,371 @@
...
+/* A zero-size buffer for use by uv_device_read */
+static char uv_zero_[] = "";
+
+int uv_device_open(uv_loop_t* loop, uv_device_t* handle, const char* path, int flags) {
+  DWORD dwCreationDisposition;
+  memset(handle, 0, sizeof(uv_device_t));
+  uv_stream_init(loop, (uv_stream_t*) handle, UV_DEVICE);
+  uv_connection_init((uv_stream_t*) handle);
+  handle->buf.len = 64 * 1024;
+  handle->buf.base = malloc(handle->buf.len);
```
saghul: Why do you preallocate this buffer?

src/win/device.c
```diff
@@ -0,0 +1,371 @@
...
+    OPEN_EXISTING,
+    FILE_ATTRIBUTE_SYSTEM|FILE_FLAG_OVERLAPPED,
+    0);
+  if (handle->handle==INVALID_HANDLE_VALUE) {
+    return GetLastError();
+  }
+  /* Try to associate with IOCP. */
+
+  if (!CreateIoCompletionPort(handle->handle, loop->iocp, (ULONG_PTR)handle, 0)) {
+    if (ERROR_INVALID_PARAMETER == GetLastError()) {
```
saghul: Call GetLastError once an use it.

src/win/device.c
```diff
@@ -0,0 +1,371 @@
...
+  return 0;
+}
+
+int uv_device_ioctl(uv_device_t* device, int cmd, void* arg) {
+  uv_ioargs_t *args = (uv_ioargs_t*) arg;
+  BOOL r;
+
+  assert(device && device->handle != INVALID_HANDLE_VALUE);
+  assert(args && args->nInBufferSize && args->nOutBufferSize);
+  r = DeviceIoControl(device->handle, (DWORD) cmd,
```
saghul: Style: each arg in a line, if the all don't fit.

src/win/device.c
```diff
@@ -0,0 +1,371 @@
...
+    args->lpInBuffer,  args->nInBufferSize,
+    args->lpOutBuffer, args->nOutBufferSize,
+    &args->lpBytesReturned,
+    NULL);
+  if (r) {
```
saghul: style: no braces for single statement ifs.

src/win/device.c
```diff
@@ -0,0 +1,371 @@
...
+  DWORD err;
+
+  assert(handle->flags & UV_HANDLE_READING);
+  assert(!(handle->flags & UV_HANDLE_READ_PENDING));
+
+  assert(handle->handle && handle->handle != INVALID_HANDLE_VALUE);
+  req = &handle->read_req;
+  memset(&req->overlapped, 0, sizeof(req->overlapped));
+
+  r = ReadFile(handle->handle,
```
saghul, line 133: ditto, style.

src/win/device.c
```diff
@@ -0,0 +1,371 @@
...
+    handle->buf.base,
+    handle->buf.len,
+    NULL,
+    &req->overlapped);
+  if(r==TRUE) {
```
saghul: if (r) ?

src/win/device.c
```diff
@@ -0,0 +1,371 @@
...
+    handle->flags |= UV_HANDLE_READ_PENDING;
+    handle->reqs_pending++;
+    uv_insert_pending_req(loop, (uv_req_t*)req);
+  }else {
```
saghul: space.

src/win/device.c
```diff
@@ -0,0 +1,371 @@
...
+  if (device->reqs_pending == 0) {
+    uv_want_endgame(device->loop, (uv_handle_t*)device);
+  }
+}
+
+
+/* This function is an egress point, i.e. it returns libuv errors rather than
+ * system errors.
+ */
+int uv__device_bind(uv_device_t* handle,
```
saghul: Not needed.

src/win/device.c
```diff
@@ -0,0 +1,371 @@
...
+/* This function is an egress point, i.e. it returns libuv errors rather than
+ * system errors.
+ */
+int uv__device_connect(uv_connect_t* req,
+                    uv_device_t* handle,
+                    const struct sockaddr* addr,
+                    unsigned int addrlen,
+                    uv_connect_cb cb) {
+  abort();
+  return 0;
```
saghul: Not needed.

src/win/device.c
```diff
@@ -0,0 +1,371 @@
...
+  uv_req_t* raw_req) {
+    abort();
+}
+
+
+/* TODO: remove me */
+void uv_process_device_connect_req(uv_loop_t* loop, uv_device_t* handle,
+  uv_connect_t* req) {
+    abort();
+}
```
saghul, line 344: Ditto for these 2.

test/test-device-tun-echo.c
```diff
@@ -0,0 +1,198 @@
...
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <arpa/inet.h> 
+#include <sys/select.h>
+#include <sys/time.h>
+#include <errno.h>
+#include <stdarg.h> 
+#include <assert.h>
+
+#define ASSERT assert
```
saghul: No way.

test/test-device-tun-echo.c
```diff
@@ -0,0 +1,198 @@
...
+
+#ifdef WIN32
```
saghul, line 38: This test seems to be Linux only, which is fine, but please have a look at how we deal with that in other tests.

saghul:
Thanks @zhaozg! I did an initial review of the code (I didn't cover it all yet though).

Makefile.am
```diff
@@ -70,6 +70,7 @@ libuv_la_SOURCES += src/win/async.c \
                     src/win/thread.c \
                     src/win/timer.c \
                     src/win/tty.c \
+                    src/win/device.c \
```
saghul: You also need to add this to uv.gyp.

test/test-device-tun-echo.c
```diff
@@ -0,0 +1,198 @@
+/* Copyright Joyent, Inc. and other Node contributors. All rights reserved.
```
saghul, line 1: The test needs to be added to the build system and to test-list.h so that its properly compiled and ran.

src/unix/device.c
txdv: no comment on the codestyle?

include/uv-win.h
zhaozg: This define device private field, for device handle, and read data buffer.

include/uv.h
zhaozg: I will add it.

src/unix/device.c
zhaozg: Done

src/unix/device.c
zhaozg: Done

src/unix/device.c
zhaozg: Done,uv__stream_open handle it, only check uv__nonblock

src/unix/device.c
zhaozg: Done

src/unix/device.c
zhaozg: sure

src/unix/stream.c
zhaozg: Yes, because uv_write call uv_write2

src/uv-common.c
zhaozg: This fix warning on windows, realy unrelated with uv_device_t, should I revert it?

src/win/device.c
zhaozg: preallocate will make buffer manage simplest, and  we not know data size real need.

src/win/device.c
zhaozg: Done

src/win/device.c
zhaozg: OK

src/win/device.c
zhaozg: I need take more attention on this, but I see more than one like 

``` c
void uv_fs_event_close(uv_loop_t* loop, uv_fs_event_t* handle) {
  uv_fs_event_stop(handle);

  uv__handle_closing(handle);

  if (!handle->req_pending) {
    uv_want_endgame(loop, (uv_handle_t*)handle);
  }

}
```

different for linux and win?

src/win/device.c
zhaozg, line 133: ok

src/win/device.c
zhaozg: ok

src/win/device.c
zhaozg: ok

src/win/device.c
zhaozg: Done

src/win/device.c
zhaozg: ok

src/win/device.c
zhaozg, line 344: this two need, because uv_process_reqs link it, but never call

test/test-device-tun-echo.c
zhaozg, line 38: I only have win and linux, so other platform need help to finished.

Makefile.am
zhaozg: Done

test/test-device-tun-echo.c
saghul, line 38: Other Unixes should Just Work (TM), but once you are done with Linux and Windows I can give it a try on some other systems.

src/win/device.c
saghul, line 344: ok, fix the parameter style then :-) and a comment saying "this should never be called".

src/win/device.c
saghul: Yeah, the code is a bit inconsistent, but the current convention is to drop braces for single statement conditionals.

src/win/device.c
saghul: I don't think so. You preallocate 64K, but when alloc_cb is called the user might have only given you a 32K buffer. Please follow the same pattern we use for all streams: don't preallocate any buffer, just use what the user provides in alloc_cb.

src/uv-common.c
saghul: Yes.

src/unix/device.c
```diff
@@ -0,0 +1,56 @@
...
+int uv_device_open(uv_loop_t* loop, uv_device_t* device, const char*path, int flags) {
+  int fd;
+  int stream_flags;
+  if( (fd = open(path, flags)) < 0 ) {
+    return fd;
+  } 
+  uv__stream_init(loop, (uv_stream_t*)device, UV_DEVICE);
+
+  stream_flags = 0;
+  if ( flags & O_RDONLY )
```
saghul: Aren't these flags mutually exclusive? Also, we need to sanity check them, in case someone passes 420 or something.

src/uv-common.c
zhaozg: OK

src/unix/device.c
zhaozg: Done

test/test-device-tun-echo.c
zhaozg, line 1: Done

test/test-device-tun-echo.c
zhaozg, line 38: thanks

test/test-device-tun-echo.c
zhaozg: use macro to support linux and windows.

src/win/device.c
zhaozg, line 344: fixed all parameter style

src/win/device.c
zhaozg: OK

zhaozg:
OK, now I update finished according to libvu style, and thanks @saghul  very much.

include/uv-win.h
zhaozg: Done

test/test-device-tun-echo.c
saghul: That macro is defined by us somewhere, just use the uppercase variant :-)

src/win/device.c
zhaozg: Done

sequoiar:
@zhaozg amazing. And, are you going integrate device stream to node.js ?

zhaozg:
not node.js, but luvit https://github.com/luvit/luvit

include/uv.h
```diff
@@ -152,6 +152,7 @@ extern "C" {
   XX(PROCESS, process)                                                        \
   XX(STREAM, stream)                                                          \
   XX(TCP, tcp)                                                                \
+  XX(DEVICE,device)                                                           \
```
txdv: space between , and device

Makefile.am
```diff
@@ -214,6 +216,7 @@ test_run_tests_SOURCES = test/blackhole-server.c \
                          test/test-timer-from-check.c \
                          test/test-timer.c \
                          test/test-tty.c \
+                         test/test-device-tun-echo.c \
```
saghul: Please sort alphabetically.

zhaozg: I sort in Makefile.am and uv.gyp

src/unix/device.c
```diff
@@ -0,0 +1,76 @@
...
+
+#include "uv.h"
+#include "internal.h"
+
+#include <stdlib.h>
+#include <unistd.h>
+#include <sys/ioctl.h>
+
+int uv_device_init(uv_loop_t* loop, uv_device_t* handle) {
+  memset(handle, 0, sizeof(uv_device_t));
```
bnoordhuis: Two things:
1. This should preserve the contents of `handle->data`.
2. It's safer and more idiomatic to use `sizeof(*handle)`.

src/unix/device.c
```diff
@@ -0,0 +1,76 @@
...
+};
+
+int uv_device_open(uv_loop_t* loop,
+                   uv_device_t* device,
+                   const char*path,
+                   int flags) {
+  int fd, err;
+  int stream_flags;
+
+  assert(flags & O_RDONLY || flags & O_WRONLY || flags & O_RDWR);
```
bnoordhuis: Doesn't gcc complain about the lack of parens?

src/unix/device.c
```diff
@@ -0,0 +1,76 @@
...
+
+  if((fd = open(path, flags)) < 0) {
```
bnoordhuis: Style: don't assign in an if statement.  Space before opening paren.

src/unix/device.c
```diff
@@ -0,0 +1,76 @@
...
+  if (err) {
+    close(fd);
+    return err;
+  }
+
+  return uv__stream_open((uv_stream_t*)device, fd, stream_flags);
+}
+
+int uv_device_ioctl(uv_device_t* device, int cmd, void* arg) {
+  int err = ioctl(uv__stream_fd((uv_stream_t*) device), cmd, arg);
```
bnoordhuis: Style: libuv doesn't normally use single-line definition and assignment.

src/unix/device.c
```diff
@@ -0,0 +1,76 @@
...
+  if (err)
```
bnoordhuis: A non-zero return code doesn't necessarily indicate error, it depends on the ioctl.

src/unix/stream.c
```diff
@@ -1509,6 +1512,7 @@ int uv___stream_fd(const uv_stream_t* handle) {
 
   assert(handle->type == UV_TCP ||
          handle->type == UV_TTY ||
+         handle->type == UV_DEVICE ||
```
bnoordhuis: Maybe this merits replacement with a  `uv__is_stream(handle)` call.

src/win/device.c
```diff
@@ -0,0 +1,350 @@
...
+#include "internal.h"
+#include "handle-inl.h"
+#include "stream-inl.h"
+#include "req-inl.h"
+
+/* A zero-size buffer for use by uv_device_read */
+static char uv_zero_[] = "";
+
+int uv_device_init(uv_loop_t* loop, uv_device_t* handle) {
+  memset(handle, 0, sizeof(uv_device_t));
```
bnoordhuis: Same conditions apply here: prefer `sizeof(*handle)` and preserve `handle->data`.

src/win/device.c
```diff
@@ -0,0 +1,350 @@
...
+  if (!CreateIoCompletionPort(handle->handle, loop->iocp, (ULONG_PTR)handle, 0)) {
+    DWORD err = GetLastError();
+    if (ERROR_INVALID_PARAMETER == err) {
+       /* Already associated. */
+    } else {
+      return uv_translate_sys_error(err);
+    }
+  }
+
+  handle->flags |= UV_HANDLE_READABLE | UV_HANDLE_WRITABLE;
```
bnoordhuis: Is there a reason it's always RW here, while on UNIX, it can be RO, WR or RW?

src/win/device.c
```diff
@@ -0,0 +1,350 @@
...
+  return 0;
+}
+
+int uv_device_ioctl(uv_device_t* device, int cmd, void* arg) {
+  uv_ioargs_t *args = (uv_ioargs_t*) arg;
```
bnoordhuis: Style issue: `uv_ioargs_t* args`

src/win/device.c
```diff
@@ -0,0 +1,350 @@
...
+  assert(device && device->handle != INVALID_HANDLE_VALUE);
+  assert(args && args->input_len && args->output_len);
+  r = DeviceIoControl(device->handle, 
+    (DWORD) cmd,
+    args->input,
+    args->input_len,
+    args->output,
+    args->output_len,
+    &args->size,
+    NULL);
```
bnoordhuis: Style: arguments should line up with first argument.

src/win/device.c
```diff
@@ -0,0 +1,350 @@
...
+  if (r) 
+    return 0;
+
+  return uv_translate_sys_error(GetLastError());
+}
+
+void uv_device_endgame(uv_loop_t* loop, uv_device_t* handle) {
+  if (handle->buf.base)
+    free(handle->buf.base);
```
bnoordhuis: Set `handle->buf.base` to NULL here.

src/win/device.c
```diff
@@ -0,0 +1,350 @@
...
+  if (handle->buf.len == 0) {
+    handle->read_cb((uv_stream_t*) handle, UV_ENOBUFS, &handle->buf);
+    return;
+  }
+
+  r = ReadFile(handle->handle,
+    handle->buf.base,
+    handle->buf.len,
+    NULL,
+    &req->overlapped);
```
bnoordhuis: Style: arguments should line up.

src/win/device.c
```diff
@@ -0,0 +1,350 @@
...
+  req->handle = (uv_stream_t*) handle;
+  req->cb = cb;
+  req->queued_bytes = 0;
+  memset(&(req->overlapped), 0, sizeof(req->overlapped));
+
+  result = WriteFile(handle->handle,
+    bufs[0].base,
+    bufs[0].len,
+    NULL,
+    &req->overlapped);
```
bnoordhuis: Style: arguments should line up.

src/win/device.c
```diff
@@ -0,0 +1,350 @@
...
+      handle->read_cb((uv_stream_t*) handle,
+                      uv_translate_sys_error(err),
+                      &handle->buf);
+    }
+  } else {
+    if (req->overlapped.InternalHigh > 0) {
+      /* Successful read */
+      handle->read_cb((uv_stream_t*) handle,
+        req->overlapped.InternalHigh,
+        &handle->buf);
```
bnoordhuis: Style: arguments should line up.

src/win/device.c
```diff
@@ -0,0 +1,350 @@
...
+      goto done;
+    }
+  }
+  /* 
+    may be we need more do read operation 
+  */
+done:
+  /* Post another read if still reading and not closing. */
+  if ((handle->flags & UV_HANDLE_READING) &&
+    !(handle->flags & UV_HANDLE_READ_PENDING)) {
```
bnoordhuis: Style: condition should line up with the first one.

src/unix/device.c
zhaozg: OK，
handle->data should be set after init or before?
Why more safer, I think same

src/win/device.c
```diff
@@ -0,0 +1,350 @@
...
+
+  if (req->cb) {
+    err = GET_REQ_ERROR(req);
+    req->cb(req, uv_translate_sys_error(err));
+  }
+
+  handle->write_reqs_pending--;
+
+  if (handle->shutdown_req != NULL &&
+    handle->write_reqs_pending == 0) {
```
bnoordhuis: Style: condition should line up with the first one.

src/win/internal.h
```diff
@@ -165,6 +165,24 @@ void uv_process_udp_send_req(uv_loop_t* loop, uv_udp_t* handle,
...
+ */
+
+int uv_device_read_start(uv_device_t* handle, uv_alloc_cb alloc_cb,
+    uv_read_cb read_cb);
+int uv_device_write(uv_loop_t* loop, uv_write_t* req, uv_device_t* handle,
+    const uv_buf_t bufs[], unsigned int nbufs, uv_write_cb cb);
+
+void uv_process_device_read_req(uv_loop_t* loop, uv_device_t* handle, uv_req_t* req);
+void uv_process_device_write_req(uv_loop_t* loop, uv_device_t* handle,
+    uv_write_t* req);
```
bnoordhuis, line 180: Style: argument should line up here and a few lines up.

test/test-device-tun-echo.c
```diff
@@ -0,0 +1,523 @@
...
+#define TAP_IOCTL_GET_MTU               TAP_CONTROL_CODE (3, METHOD_BUFFERED)
+#define TAP_IOCTL_GET_INFO              TAP_CONTROL_CODE (4, METHOD_BUFFERED)
+#define TAP_IOCTL_CONFIG_POINT_TO_POINT TAP_CONTROL_CODE (5, METHOD_BUFFERED)
+#define TAP_IOCTL_SET_MEDIA_STATUS      TAP_CONTROL_CODE (6, METHOD_BUFFERED)
+#define TAP_IOCTL_CONFIG_DHCP_MASQ      TAP_CONTROL_CODE (7, METHOD_BUFFERED)
+#define TAP_IOCTL_GET_LOG_LINE          TAP_CONTROL_CODE (8, METHOD_BUFFERED)
+#define TAP_IOCTL_CONFIG_DHCP_SET_OPT   TAP_CONTROL_CODE (9, METHOD_BUFFERED)
+
+static int is_tap_win32_dev(const char *guid)
+{
```
bnoordhuis: Style: brace on same line as definition.

src/unix/device.c
saghul: > This should preserve the contents of handle->data.

@bnoordhuis: uv_loop_init does not preserve it, FWIW. I don't mind either way, but if the handle initialization fails what's the point of storing anything in its data field? Whichever we choose we should add some explanation in the docs, I think.

bnoordhuis: It's how other init functions work so, for consistency, this one should as well.  If uv_loop_init() doesn't, then that's a bug and you should go fix it. :-)

bnoordhuis: > handle->data should be set after init or before?

@zhaozg Code like this should work:

``` c
uv_device_t dev;
dev.handle = &dev;
uv_device_init(loop, &dev);
assert(dev.handle == (void*) &dev);
```

bnoordhuis:
Sorry, have to cut the review short.  At a quick glance, I notice there are a bunch of style errors in the test but nothing worse.

src/unix/device.c
zhaozg: thanks, Done that.

src/unix/device.c
zhaozg: I think that's ok, because & higher than ||, but better to use parens, should I @saghul

test/test-device-tun-echo.c
zhaozg: Fixed style in test code.

src/unix/device.c
zhaozg: Done

src/unix/device.c
zhaozg: Done

src/unix/device.c
zhaozg: Important bug, Fixed

src/unix/stream.c
zhaozg: I try keep more change of libuv source code.
So I will not change this.

src/win/device.c
zhaozg: Done

src/win/device.c
zhaozg: have changed it according flags paramater

src/win/device.c
zhaozg: Done

src/win/device.c
zhaozg: Done

src/win/device.c
zhaozg: Done

src/win/device.c
zhaozg: Done

src/win/device.c
zhaozg: Done

src/win/device.c
zhaozg: Done

src/win/device.c
zhaozg: Done

src/win/device.c
zhaozg: Done

src/win/internal.h
zhaozg, line 180: fixed function declare like other stream types.

zhaozg:
thanks @bnoordhuis, code updated.

include/uv-win.h
```diff
@@ -479,6 +483,18 @@ RB_HEAD(uv_timer_tree_s, uv_timer_s);
...
 
+/* struct parameters for DeviceIoControl */
+typedef struct uv_ioargs_s {
+  void* input;
+  uint32_t input_len;
+  void* output;
+  uint32_t output_len;
+  uint32_t size;          /* ouput returned bytes */
+  /*
+  _Inout_opt_  LPOVERLAPPED lpOverlapped;
```
saghul: Unused, please remove.

include/uv.h
```diff
@@ -614,6 +616,24 @@ UV_EXTERN int uv_udp_recv_stop(uv_udp_t* handle);
...
+  UV_STREAM_FIELDS
+  UV_DEVICE_PRIVATE_FIELDS
+};
+
+UV_EXTERN int uv_device_init(uv_loop_t* loop, uv_device_t* device);
+UV_EXTERN int uv_device_open(uv_loop_t* handle,
+                             uv_device_t* device,
+                             const char* path,
+                             int flags);
+UV_EXTERN int uv_device_ioctl(uv_device_t* device, int cmd, void* arg);
```
saghul: I wonder if we should always take a uv_ioargs_t as the argument, which is typedef-d to void\* on Unix and it's that structure you defined on Windows... @bnoordhuis thoughts?

include/uv.h
```diff
@@ -614,6 +616,24 @@ UV_EXTERN int uv_udp_recv_stop(uv_udp_t* handle);
...
+ * Represents a device stream.
+ */
+struct uv_device_s {
+  UV_HANDLE_FIELDS
+  UV_STREAM_FIELDS
+  UV_DEVICE_PRIVATE_FIELDS
+};
+
+UV_EXTERN int uv_device_init(uv_loop_t* loop, uv_device_t* device);
+UV_EXTERN int uv_device_open(uv_loop_t* handle,
```
saghul: This one is inconsistent with the other _open functions. Maybe we should remove it and open the device in _init? That's what uv_tty_init does, for example.

src/unix/device.c
```diff
@@ -0,0 +1,78 @@
...
+  int stream_flags;
+
+  assert(flags & O_RDONLY || flags & O_WRONLY || flags & O_RDWR);
+
+  fd = open(path, flags); 
+  if (fd < 0)
+    return -errno;
+
+  stream_flags = 0;
+  if (flags & O_RDONLY)
```
saghul, line 46: Aren't these flags mutually exclusive? If so, use a switch statement or if-else-if.

src/unix/device.c
```diff
@@ -0,0 +1,78 @@
...
+  if (flags & O_RDWR)
+    stream_flags |= UV_STREAM_READABLE | UV_STREAM_WRITABLE;
+
+  err = uv__nonblock(fd, 1);
+  if (err) {
+    close(fd);
+    return err;
+  }
+
+  return uv__stream_open((uv_stream_t*)device, fd, stream_flags);
```
saghul: Handle error, if this function fails you need to close the fd before returning.

src/unix/stream.c
```diff
@@ -1442,6 +1444,7 @@ int uv_read_start(uv_stream_t* stream,
                   uv_alloc_cb alloc_cb,
                   uv_read_cb read_cb) {
   assert(stream->type == UV_TCP || stream->type == UV_NAMED_PIPE ||
+      stream->type == UV_DEVICE ||
```
saghul: put it on the same line as the UV_TTY check.

src/unix/stream.c
saghul: It's ok to make changes that are for the better! :-)

src/win/device.c
```diff
@@ -0,0 +1,365 @@
...
+                   int flags) {
+  DWORD dwCreationDisposition;
+  int uvflags;
+
+  assert(handle && handle->handle == INVALID_HANDLE_VALUE);
+  uv_connection_init((uv_stream_t*) handle);
+
+  dwCreationDisposition = 0;
+  uvflags = 0; 
+  if (flags & O_RDONLY) {
```
saghul, line 54: Ditto, aren't the flags mutually exclusive?

src/win/device.c
```diff
@@ -0,0 +1,365 @@
...
+
+  handle->handle = CreateFile(path,
+                              dwCreationDisposition,
+                              0,
+                              NULL,
+                              OPEN_EXISTING,
+                              FILE_ATTRIBUTE_SYSTEM | FILE_FLAG_OVERLAPPED,
+                              0);
+  if (handle->handle == INVALID_HANDLE_VALUE)
+    return GetLastError();
```
saghul: Translate the error.

src/win/device.c
```diff
@@ -0,0 +1,365 @@
...
+
+  /* Try to associate with IOCP. */
+  if (!CreateIoCompletionPort(handle->handle, loop->iocp, (ULONG_PTR)handle, 0)) {
+    DWORD err = GetLastError();
+    if (ERROR_INVALID_PARAMETER == err) {
+       /* Already associated. */
```
saghul: Really? What error do you get if the passes handle is actually invalid / NULL?

src/win/device.c
```diff
@@ -0,0 +1,365 @@
...
+                    unsigned int nbufs,
+                    uv_write_cb cb) {
+  int result;
+  DWORD err = 0;
+  
+  if (nbufs != 1 && (nbufs != 0 || !handle)) {
+    return ERROR_NOT_SUPPORTED;
+  }
+  /* Only TCP handles are supported for sharing. */
+  if (handle && ((handle->type != UV_DEVICE) ||
```
saghul: This check doesn't really apply. Just check if handle is not NULL.

include/uv-win.h
saghul: I see you no longer malloc your own buffer, good! Can you please rename this to read_buffer?

src/win/device.c
```diff
@@ -0,0 +1,365 @@
...
+
+int uv_device_open(uv_loop_t* loop,
+                   uv_device_t* handle,
+                   const char* path,
+                   int flags) {
+  DWORD dwCreationDisposition;
+  int uvflags;
+
+  assert(handle && handle->handle == INVALID_HANDLE_VALUE);
+  uv_connection_init((uv_stream_t*) handle);
```
saghul: Does this need to be called here? What for?

src/win/device.c
```diff
@@ -0,0 +1,365 @@
...
+
+  if (r) 
+    return 0;
+
+  return uv_translate_sys_error(GetLastError());
+}
+
+void uv_device_endgame(uv_loop_t* loop, uv_device_t* handle) {
+  if (handle->buf.base) {
+    free(handle->buf.base);
```
saghul: The buffer needs to be freed by the user in the read callback if she wants to, not by libuv, since it didn't allocate it.

src/win/device.c
```diff
@@ -0,0 +1,365 @@
...
+
+void uv_process_device_read_req(uv_loop_t* loop,
+                                uv_device_t* handle,
+                                uv_req_t* req) {
+  DWORD err;
+
+  assert(handle->type == UV_DEVICE);
+
+  handle->flags &= ~UV_HANDLE_READ_PENDING;
+  assert(handle->flags & UV_HANDLE_READING);
```
saghul: This assert should be before the previous statement, or it's superfluous.

src/win/device.c
```diff
@@ -0,0 +1,365 @@
...
+                      uv_translate_sys_error(err),
+                      &handle->buf);
+    }
+  } else {
+    if (req->overlapped.InternalHigh > 0) {
+      /* Successful read */
+      handle->read_cb((uv_stream_t*) handle,
+                      req->overlapped.InternalHigh,
+                      &handle->buf);
+      /* not sure this, are we need repeat read ? */
```
saghul: If the buffer was filled it's very possible the kernel has more data for us, that's why we read again. Please remove the comment.

src/win/device.c
```diff
@@ -0,0 +1,365 @@
...
+      /* Read again only if bytes == buf.len */
+      if (req->overlapped.InternalHigh < sizeof(handle->buf))
```
saghul: sizeof(handle->buf) is wrong, it should be handle->buf.len.

src/win/device.c
```diff
@@ -0,0 +1,365 @@
...
+  if (handle->shutdown_req != NULL &&
+      handle->write_reqs_pending == 0) {
+      uv_want_endgame(loop, (uv_handle_t*) handle);
+  }
+
+  DECREASE_PENDING_REQ_COUNT(handle);
+}
+
+
+static int uv_device_try_cancel_io(uv_device_t* device) {
```
saghul: Where is this used?

src/win/device.c
zhaozg: Make stream to connected state

src/win/device.c
zhaozg: Ok, I take a mistake

cjdelisle:
Without having really followed this conversation except to periodically sweep the mess of mail out of my inbox, I thought perhaps you would enjoy reading [parkinson's law](http://lib.ru/DPEOPLE/PARKINSON/parkinson.txt) :)

src/win/device.c
saghul: That's only used internally in TCP and Pipe handles. I don't think we need it here, just initialize the fields we need.

saghul:
@cjdelisle :-)

include/uv.h
zhaozg: device-ioctl should define ioargs too, because ioctl has variable use method,eg.return serial boud bit rates.

include/uv-win.h
zhaozg: This used by device_ioctl,must keep

saghul: I don't see that in your code :-O and moreover, it's commented out here :-)

include/uv.h
bnoordhuis: @saghul No strong opinion.

src/unix/stream.c
zhaozg: OK

include/uv.h
zhaozg: changed to uv_ioargs_t\* arg

src/unix/device.c
zhaozg, line 46: Done

src/unix/device.c
zhaozg: Ok

src/unix/stream.c
zhaozg: change to assert(uv__is_stream(stream));

src/win/device.c
zhaozg, line 54: changed

src/win/device.c
zhaozg: ok

src/win/device.c
zhaozg: done

src/win/device.c
zhaozg: I persist to use uv_connection_init, but call only after device open successfully.

src/win/device.c
zhaozg: Done

src/win/device.c
zhaozg: Done

src/win/device.c
zhaozg: not use, removed it

include/uv-win.h
zhaozg: good

src/win/device.c
zhaozg: Done

include/uv-win.h
zhaozg: please check function uv_device_ioctl

saghul: I have checked it, you are currently passing NULL as the last parameter to DeviceIoControl. Maybe you intended to pass the overlapped? If so, please remove the comment and just call it "overlapped".

src/unix/device.c
```diff
@@ -0,0 +1,78 @@
...
+#include <sys/ioctl.h>
+
+int uv_device_init(uv_loop_t* loop,
+                   uv_device_t* device,
+                   const char*path,
+                   int flags) {
+  int fd, err;
+  int stream_flags;
+
+  assert(flags == O_RDONLY || flags == O_WRONLY || flags == O_RDWR);
```
saghul: Return -EINVAL instead of asserting.

src/win/device.c
```diff
@@ -0,0 +1,339 @@
...
+                    uv_write_cb cb) {
+  int result;
+  DWORD err = 0;
+  
+  if (nbufs != 1 && (nbufs != 0 || !handle)) {
+    return ERROR_NOT_SUPPORTED;
+  }
+  /* Only TCP handles are supported for sharing. */
+  if (handle && (!(handle->flags & UV_HANDLE_CONNECTION))) {
+        return ERROR_NOT_SUPPORTED;
```
saghul: style: indent with 2 spaces

src/unix/device.c
```diff
@@ -0,0 +1,78 @@
...
+    return err;
+  }
+  return 0;
+}
+
+int uv_device_ioctl(uv_device_t* device, int cmd, uv_ioargs_t* args) {
+  int err;
+
+  args->result = ioctl(uv__stream_fd((uv_stream_t*) device), cmd, args->arg);
+  if (args->result < 0)
```
saghul: I'm not sure we need the result field, we can just return the value directly.

src/win/device.c
```diff
@@ -0,0 +1,339 @@
...
+
+  dwCreationDisposition = 0;
+  uvflags = 0; 
+
+  uv_stream_init(loop, (uv_stream_t*) handle, UV_DEVICE);
+  handle->handle = INVALID_HANDLE_VALUE;
+  handle->read_buffer.base = NULL;
+  handle->read_buffer.len = 0;
+
+  if (flags & O_RDONLY) {
```
saghul, line 54: Check the flags as we do on Unix.

src/unix/device.c
zhaozg:
``` c
int uv_device_ioctl(uv_device_t* device, int cmd, uv_ioargs_t* args) {
  int err;

  err = ioctl(uv__stream_fd((uv_stream_t*) device), cmd, args->arg);
  if (err < 0)
    return -errno;

  return err;
}
```

src/win/device.c
zhaozg, line 54: Done

src/win/device.c
zhaozg: Done

zhaozg:
@saghul I should make a new RP to libuv/libuv?

saghul:
If you don't mind, that would be great!

zhaozg:
Close this RP, because replaced with https://github.com/libuv/libuv/pull/19.

